PR #661, opened 2015-01-07
Improved microfacet2 tests

est77:
- Improved MDF2 tests.
- Removed refraction case from the V-cavity masking shadowing function (was unused).


src/appleseed/foundation/meta/tests/test_microfacet2.cpp
```diff
@@ -102,49 +102,80 @@ TEST_SUITE(Foundation_Math_Microfacet2)
     //   http://hal.inria.fr/docs/00/96/78/44/PDF/RR-8468.pdf
     //
 
+    struct weak_white_furnace_test_result
```
dictoon: WeakWhiteFurnaceTestResult

PR #671, opened 2015-01-15
More appleseed python improvements

est77:
- Renamed files to remove underscores, like the rest of appleseed's code.
- Cleanup headers.
- Do not expose get_assembly, get_object and similar methods that only work after rendering is started.


src/appleseed.python/bindtexture.cpp
```diff
@@ -27,23 +27,16 @@
...
+#include "bindentitycontainers.h"
 #include "dict2dict.h"
 #include "metadata.h"
-#include "unaligned_transformd44.h"
+#include "unalignedtransformd44.h"
 
 // appleseed.renderer headers.
 #include "renderer/api/texture.h"
-#include "renderer/modeling/scene/textureinstance.h"
+#include "renderer/api/scene.h"
```
dictoon: Order alphabetically.

PR #673, opened 2015-01-19
Added support for object alpha maps

est77:
- Objects can now have an alpha map.
In case both the object and its material have alpha maps, they are combined.
The changes are backward compatible.

- Added an alpha accessor to ShadingPoint that can be reused later to support curves and meshes per vertex alpha.

- Refactored alpha handling code a bit.

- Does not include the changes needed to set object alpha maps from appleseed.studio.


src/appleseed/renderer/kernel/intersection/intersectionfilter.cpp
```diff
@@ -117,13 +117,15 @@ IntersectionFilter::IntersectionFilter(
     Object&                 object,
     const MaterialArray&    materials,
     TextureCache&           texture_cache)
+  : m_obj_alpha_mask(0)
+  , m_obj_alpha_map_signature(0)
 {
     // Initialize the material -> alpha mask mapping.
     m_alpha_map_signatures.assign(materials.size(), 0);
     m_alpha_masks.assign(materials.size(), 0);
```
dictoon: Maybe rename this to ` m_material_alpha_masks`?

src/appleseed/renderer/kernel/intersection/intersectionfilter.cpp
```diff
@@ -227,6 +246,9 @@ size_t IntersectionFilter::get_masks_memory_size() const
 {
     size_t size = 0;
 
+    if (m_obj_alpha_mask)
+        size = m_obj_alpha_mask->get_memory_size();
```
dictoon: `size += ...`

src/appleseed/renderer/kernel/shading/shadingpoint.cpp
```diff
@@ -31,6 +31,7 @@
 #include "shadingpoint.h"
 
 // appleseed.renderer headers.
+#include "renderer/modeling/input/source.h"
```
dictoon: Order alphabetically.

src/appleseed/renderer/kernel/shading/shadingpoint.cpp
```diff
@@ -565,6 +566,43 @@ void ShadingPoint::compute_point_velocity() const
...
+                    a);
+
+                m_alpha *= a;
+            }
+        }
+    }
+    else
+    {
+        assert(m_primitive_type == PrimitiveCurve1 || m_primitive_type == PrimitiveCurve3);
+    }
```
dictoon: Add todo note.

src/appleseed/renderer/kernel/shading/shadingpoint.h
```diff
@@ -625,6 +636,12 @@ inline const Material* ShadingPoint::get_material() const
                 m_material = materials[m_primitive_pa];
         }
 
+        m_shade_alpha_cutouts = m_object->shade_alpha_cutouts();
+
+        if (m_material)
```
dictoon: Simplify:

    if (m_material && m_material->shade_alpha_cutouts())
        m_shade_alpha_cutouts = true;

src/appleseed/renderer/kernel/shading/shadingpoint.h
```diff
@@ -684,6 +701,25 @@ inline size_t ShadingPoint::get_primitive_attribute_index() const
...
+        compute_alpha();
+        m_members |= HasAlpha;
+    }
+
+    return m_alpha;
+}
+
+inline bool ShadingPoint::shade_alpha_cutouts() const
+{
+    if (!(m_members & HasMaterial))
```
dictoon: Remove check, just call `get_material()`.

src/appleseed/renderer/modeling/object/meshobject.cpp
```diff
@@ -121,6 +123,42 @@ const char* MeshObject::get_model() const
...
+}
+
+bool MeshObject::has_alpha_map() const
+{
+    if (!m_params.strings().exist("alpha_map"))
+        return false;
+
+    const char* value = m_params.strings().get("alpha_map");
+
+    return strlen(value) > 0 ? value : 0;
```
dictoon: Won't compile with VS. Use:

    return strlen(value) > 0 ? true : false;

src/appleseed/renderer/modeling/object/meshobject.h
```diff
@@ -69,6 +69,22 @@ class APPLESEED_DLLSYMBOL MeshObject
     // Return a string identifying the model of this object.
     virtual const char* get_model() const APPLESEED_OVERRIDE;
 
+    // This method is called once before rendering each frame.
+    // Returns true on success, false otherwise.
+    virtual bool on_frame_begin(
```
dictoon: Add missing forward declarations (for `Project`, etc.).

src/appleseed/renderer/modeling/surfaceshader/constantsurfaceshader.cpp
```diff
@@ -192,7 +192,7 @@ DictionaryArray ConstantSurfaceShaderFactory::get_input_metadata() const
             .insert("items",
                 Dictionary()
                     .insert("Alpha channel of the color", "color")
-                    .insert("Alpha map of the material", "material"))
+                    .insert("Alpha map of the object and material", "material"))
```
dictoon: "object and/or material"

src/appleseed/renderer/modeling/object/meshobject.cpp
```diff
@@ -121,6 +123,42 @@ const char* MeshObject::get_model() const
...
+}
+
+const Source*MeshObject::get_uncached_alpha_map() const
```
dictoon: Add missing space after `*`.

src/appleseed/renderer/modeling/object/meshobject.cpp
```diff
@@ -37,6 +37,7 @@
```
dictoon: Could object alpha maps be handled entirely in the base `Object` class?

dictoon:
Wow, that was quite a bit of work :) Great job as usual.

src/appleseed/renderer/modeling/object/meshobject.cpp
est77: Alpha maps don't make sense for some kind of objects, for example curves.
I prefer to enable them only for objects that make sense.

dictoon: They could make sense on curves, if you use them for feathers for instance (or grass).

est77: I prefer to keep alpha maps in MeshObject for now. If we find a good reason to support them in other object types, we can do it in a backwards compatible way without problems.

dictoon: The downside is that we already have duplicated handling of alpha maps in `Object` and `MeshObject`. We could have *more* functionality, more orthogonality, and *less* code if everything was just in `Object`. But again, your call.

est77: I don't think there's duplicated handling of alpha maps in the PR.

Currently, alpha maps are used for two different things in appleseed. Removed unwanted parts of surfaces when intersecting rays with geometry and compositing transparency.

I think that it would be good to try to separate the two cases in the future and this PR goes in that direction. For object alpha maps I'm considering only the first case. Compositing transparency is best handled in materials and / or OSL shaders.

I extended the Object interface to support alpha maps. For some object types it makes sense to support them, for some others it's not clear if it's a good idea and for others alpha maps will not work.

Objects that can have alpha maps, need to implement the part of the interface that's related to alpha maps, like MeshObject.

I can see at some point using object alpha maps to clip unwanted parts of geometry, 
renaming the alpha maps inputs from materials to opacity maps or something like that and 
use them just for compositing transparency. 

So I only want to enable them for objects where it makes sense and not for all present and future object types by default.

PR #697, opened 2015-02-19
OSL background support #1

est77:
- Bind scene shader groups (previously, they were ignored).
- Added code to execute OSL background shaders.

### DIFF
--- a/src/appleseed/renderer/kernel/shading/oslshadergroupexec.h
+++ b/src/appleseed/renderer/kernel/shading/oslshadergroupexec.h
@@ -35,6 +35,7 @@
 
 // appleseed.foundation headers.
 #include "foundation/core/concepts/noncopyable.h"
+#include "foundation/image/color.h"
 
 // OSL headers.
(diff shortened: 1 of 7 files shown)


src/appleseed/renderer/kernel/shading/oslshadergroupexec.h
```diff
@@ -89,6 +90,10 @@ class OSLShaderGroupExec
         const ShadingPoint&             shading_point,
         const float                     surface_area) const;
 
+    foundation::Color3f execute_background(
+        const ShaderGroup&          shader_group,
```
dictoon: Align arguments with those of surrounding methods.

src/appleseed/renderer/kernel/shading/oslshadergroupexec.h
```diff
@@ -89,6 +90,10 @@ class OSLShaderGroupExec
...
+        const foundation::Vector3d& outgoing) const;
```
dictoon: You need to include `"foundation/math/vector.h"`.

PR #702, opened 2015-02-20
Assembly factories

est77:
This PR adds assembly factories and the possibility to have more than one type of assembly.
Refactored from some of my experimental branches.

### DIFF
--- a/src/appleseed/renderer/modeling/scene/assembly.cpp
+++ b/src/appleseed/renderer/modeling/scene/assembly.cpp
@@ -56,6 +56,8 @@ namespace renderer
 // Assembly class implementation.
 //
 
+const char* Model = "assembly";
+
 namespace
 {
     const UniqueID g_class_uid = new_guid();
@@ -108,6 +110,11 @@ Assembly::~Assembly()
 void Assembly::release()
 {
     delete this;
+}
+
+const char* Assembly::get_model() const
+{
+    return Model;
 }
 
 BSDFContainer& Assembly::bsdfs() const
@@ -262,6 +269,11 @@ void Assembly::on_frame_end(const Project& project)
 //
 // AssemblyFactory class implementation.
 //
+
+const char* AssemblyFactory::get_model()
+{
+    return Model;
+}
 
 auto_release_ptr<Assembly> AssemblyFactory::create(
     const char*         name,
--- a/src/appleseed/renderer/modeling/scene/assembly.cpp
+++ b/src/appleseed/renderer/modeling/scene/assembly.cpp
@@ -56,7 +56,7 @@ namespace renderer
 // Assembly class implementation.
 //
 
-const char* Model = "assembly";
+const char* Model = "generic_assembly";
 
 namespace
 {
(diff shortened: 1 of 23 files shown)


src/appleseed/renderer/modeling/scene/assemblyfactoryregistrar.cpp
```diff
@@ -0,0 +1,115 @@
...
+using namespace foundation;
+using namespace std;
+
+namespace renderer
+{
+
+namespace
+{
+
+class IAssemblyAssemblyFactoryAdapter
```
dictoon: Indent code in anonymous namespaces.

src/appleseed/renderer/modeling/scene/assembly.h
```diff
@@ -132,9 +135,15 @@ class APPLESEED_DLLSYMBOL Assembly
 // Assembly factory.
 //
 
+// To preserve API compatibility, this class
```
dictoon: What would deriving from `IAssemblyFactory` break?

src/appleseed/renderer/modeling/scene/assembly.cpp
```diff
@@ -56,6 +56,8 @@ namespace renderer
 // Assembly class implementation.
 //
 
+const char* Model = "assembly";
```
dictoon: `generic_assembly`? Like we have `generic_material`, etc.

src/appleseed/renderer/modeling/scene/iassemblyfactory.h
```diff
@@ -0,0 +1,70 @@
...
+// appleseed.foundation headers.
+#include "foundation/core/concepts/noncopyable.h"
+#include "foundation/utility/autoreleaseptr.h"
+
+// appleseed.main headers.
+#include "main/dllsymbol.h"
+
+// Forward declarations.
+namespace renderer      { class Assembly; }
+namespace foundation    { class Dictionary; }
```
dictoon: Remove redundant forward declarations.

src/appleseed/renderer/modeling/scene/assembly.cpp
est77: I was following the convention of using the class name for the model, but I think generic_assembly is ok. I'll change it.

dictoon: And indeed it would make sense to rename `Assembly` to `GenericAssembly` if we start to have many kinds of assemblies.

dictoon: Although, right now I guess your idea is to stick to a single `Assembly` class that can be populated differently at instantiation-time, right?

est77: > And indeed it would make sense to rename Assembly to GenericAssembly if we start to have many kinds of assemblies.

Ideally, we would have an abstract Assembly class and the current Assembly class would be a subclass of it, called GenericAssembly. But I'm not sure what would be the real advantage of doing it and in exchange we break source compatibility. I'd like to be really sure before going in 
this direction. What about renaming the current class to GenericAssembly and having a typedef GenericAssembly Assembly?

dictoon: Or even simpler, we leave everything as is for now, i.e. we keep `Assembly` (of model `generic_assembly`).

PR #711, opened 2015-03-03
More ray differentials work

est77:
- Add setters to Dual class.
- Use duals instead of vectors in places where differentials are relevant.
- Do not use BSDFSample::get_incoming() in BSDFs.


src/appleseed/foundation/math/dual.h
```diff
@@ -102,6 +104,13 @@ inline const T& Dual<T>::get_value() const
 }
 
 template <typename T>
+inline void Dual<T>::set_value(const T& value)
+{
+    m_value = value;
+    m_has_derivatives = false;
```
dictoon: That's dangerous, e.g. if `set_value()` is called after `set_derivatives()`. I would recommend against clearing the flag here.

src/appleseed/foundation/math/dual.h
```diff
@@ -123,6 +132,14 @@ const T& Dual<T>::get_dy() const
     return m_dy;
 }
 
+template <typename T>
+void Dual<T>::set_derivatives(const T& dx, const T& dy)
```
dictoon: Mark as inline.

src/appleseed/renderer/kernel/lighting/pathtracer.h
```diff
@@ -341,14 +348,31 @@ size_t PathTracer<PathVisitor, Adjoint>::trace(
...
+            sample.get_incoming_vector(),
             ray.m_time,
             ray.m_dtime,
             bsdf_mode_to_ray_flags(sample.get_mode()),
             ray.m_depth + 1);
 
+        if (sample.get_incoming().has_derivatives())
+        {
+            scattered_ray.m_rx.m_org =
+                    scattered_ray.m_org + vertex.m_shading_point->get_dpdx();
```
dictoon: Fix indentation.

src/appleseed/renderer/modeling/bsdf/oslashikhminbrdf.cpp
```diff
@@ -148,20 +147,22 @@ namespace
             const double sin_theta = sqrt(1.0 - cos_theta * cos_theta);
 
             // Compute the halfway vector in world space.
-            h = sample.get_shading_basis().transform_to_parent(
+            const Vector3d h = sample.get_shading_basis().transform_to_parent(
                         Vector3d::unit_vector(cos_theta, sin_theta, cos_phi, sin_phi));
```
dictoon: Fix indentation.

dictoon:
Looking pretty good!

src/appleseed/foundation/math/dual.h
est77: It is expected that you call set_value() before set_derivatives().
But checking the code, we don't need setters at all and then there's no 
possibility of doing the wrong thing anymore.

Issue #718, opened 2015-03-12
Keep the OSL shading system alive between progressive renders


est77:
and only optimize the shader groups that need it.

PR #719, opened 2015-03-15
Keep OSL shading system alive between progressive renders

est77:
And optimize only the shaders that have been edited since the last render. Fixes issue #718.
Also fixed an unrelated bug (triangle with no normals crash) that I found while testing this.


src/appleseed/renderer/kernel/rendering/baserenderer.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+// Interface header.
+#include "baserenderer.h"
+
+// appleseed.renderer headers.
+#include "renderer/global/globallogger.h"
+#include "renderer/modeling/project/project.h"
+#include "renderer/modeling/scene/scene.h"
+#ifdef APPLESEED_WITH_OIIO
+#include "renderer/kernel/rendering/oiioerrorhandler.h"
```
dictoon: Order alphabetically (kernel before modeling).

src/appleseed/renderer/kernel/rendering/baserenderer.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+BaseRenderer::BaseRenderer(
+    Project&                project,
+    const ParamArray&       params)
+  : m_project(project)
+  , m_params(params)
+{
+#ifdef APPLESEED_WITH_OIIO
+
+    RENDERER_LOG_DEBUG("%s.", "creating OpenImageIO error handler");
```
dictoon: I don't think this message is useful, I would suggest to remove it.

dictoon: Although, it's in line with the other messages that announce the creation of the various OSL/OIIO components... Why not.

src/appleseed/renderer/kernel/rendering/baserenderer.cpp
```diff
@@ -0,0 +1,225 @@
...
+    RENDERER_LOG_DEBUG("%s.", "destroying OSL renderer services");
+    delete m_renderer_services;
+#endif
+
+#ifdef APPLESEED_WITH_OIIO
+    const string stats = m_texture_system->getstats();
+    const string trimmed_stats = trim_right(stats, "\r\n");
+    RENDERER_LOG_INFO("%s", trimmed_stats.c_str());
+
+    RENDERER_LOG_DEBUG("%s.", "destroying OpenImageIO texture system");
```
dictoon: Simplify: `RENDERER_LOG_DEBUG("destroying OpenImageIO texture system.");`

src/appleseed/renderer/kernel/rendering/baserenderer.cpp
```diff
@@ -0,0 +1,225 @@
...
+        "setting OpenImageIO texture cache size to %s.",
+        pretty_size(texture_cache_size_bytes).c_str());
+
+    const float texture_cache_size_mb =
+        static_cast<float>(texture_cache_size_bytes) / (1024 * 1024);
+
+    m_texture_system->attribute("max_memory_MB", texture_cache_size_mb);
+
+    // search paths
+    {
```
dictoon: Move this block to a separate method?

src/appleseed/renderer/kernel/rendering/baserenderer.cpp
```diff
@@ -0,0 +1,225 @@
...
+            m_texture_system->clear();
+            m_texture_system->attribute("searchpath", new_search_path);
+        }
+    }
+#endif
+
+#ifdef APPLESEED_WITH_OSL
+    m_renderer_services->initialize(texture_store);
+
+    // search paths
```
dictoon: Move this block to a separate method?

src/appleseed/renderer/kernel/rendering/baserenderer.cpp
est77: > I don't think this message is useful, I would suggest to remove it.

There are too many debug messages. I'll leave only the interesting ones.

src/appleseed/renderer/kernel/rendering/baserenderer.cpp
est77: I don't think we get much moving them to separate methods.
But some comments would be nice. I'll add them.

dictoon: What we get is a much shorter function, no need for artificial scopes, and no need for comments (just name the functions based on the comments you added, e.g. `update_oiio_texture_sys_attrs`).

PR #720, opened 2015-03-17
Small improvements

est77:
- Refactor BaseRenderer initialize method.
- Prevent black frames when doing progressive rendering.

### DIFF
--- a/src/appleseed/renderer/kernel/rendering/baserenderer.cpp
+++ b/src/appleseed/renderer/kernel/rendering/baserenderer.cpp
@@ -151,9 +151,20 @@ bool BaseRenderer::initialize_shading_system(
     TextureStore& texture_store,
     IAbortSwitch& abort_switch)
 {
-    // Update OIIO Texture system attributes
-    // if they changed since the last render.
 #ifdef APPLESEED_WITH_OIIO
+    initialize_oiio();
+#endif
+
+#ifdef APPLESEED_WITH_OSL
+    return initialize_osl(texture_store, abort_switch);
+#endif
+
+    return true;
+}
+
+#ifdef APPLESEED_WITH_OIIO
+void BaseRenderer::initialize_oiio()
+{
     const ParamArray& params = m_params.child("texture_store");
 
     const size_t texture_cache_size_bytes =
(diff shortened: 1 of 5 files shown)


src/appleseed/renderer/kernel/rendering/baserenderer.cpp
```diff
@@ -151,9 +151,20 @@ bool BaseRenderer::initialize_shading_system(
...
-    // if they changed since the last render.
 #ifdef APPLESEED_WITH_OIIO
+    initialize_oiio();
+#endif
+
+#ifdef APPLESEED_WITH_OSL
+    return initialize_osl(texture_store, abort_switch);
+#endif
+
+    return true;
```
dictoon: Maybe put the `return true` inside an `#else` block?

src/appleseed/renderer/kernel/rendering/baserenderer.cpp
```diff
@@ -151,9 +151,20 @@ bool BaseRenderer::initialize_shading_system(
...
+}
+
+#ifdef APPLESEED_WITH_OIIO
+void BaseRenderer::initialize_oiio()
```
dictoon: I believe the artificial scopes are no longer needed.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -238,6 +238,13 @@ const AABB2u& Frame::get_crop_window() const
     return impl->m_crop_window;
 }
 
+size_t Frame::get_pixel_count() const
+{
+    return
```
dictoon: `return impl->m_crop_window.volume()` ought to be sufficient (to be verified).

src/appleseed/renderer/modeling/frame/frame.h
```diff
@@ -105,6 +105,9 @@ class APPLESEED_DLLSYMBOL Frame
     void set_crop_window(const foundation::AABB2u& crop_window);
     const foundation::AABB2u& get_crop_window() const;
 
+    // Return the number of pixels in the Frame, taking into account the crop window.
```
dictoon: "Frame" -> "frame"

PR #721, opened 2015-03-21
Misc changes

est77:
- Adjusted min pixel update percentange for progressive rendering.
- Disabled in-source builds.
- Removed unused code from OSL specific BSDFs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,6 +47,11 @@ include (src/cmake/utilities.txt)
 # Build configurations.
 #--------------------------------------------------------------------------------------------------
 
+# Disable in source builds.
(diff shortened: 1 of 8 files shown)


CMakeLists.txt
```diff
@@ -47,6 +47,11 @@ include (src/cmake/utilities.txt)
 # Build configurations.
 #--------------------------------------------------------------------------------------------------
 
+# Disable in source builds.
```
dictoon: Out of curiosity, is there a particular reason to forbid it? Not that it's a good idea to do in-source builds...

est77: It's not really a good idea. My main reason is that it generates a lot of files that we are not ignoring in the git config.
I prefer to disable it by default and if anyone needs to do in-source builds review the decision.

PR #743, opened 2015-05-03
Python script to extract osl shader metadata.

SecretChief:
Python script to parse compiled osl shaders and put the metadata in a JSON file. The JSON file can be used by integration plugins to build shader UIs.

### DIFF
--- a/scripts/oslextractmeta.py
+++ b/scripts/oslextractmeta.py
@@ -1,0 +1,466 @@
+#! /usr/bin/env python
+
+"""
+The MIT License (MIT)
+
+Copyright (c) 2015, Hans Hoogenboom, the appleseedhq Organization
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+"""
+
+# TODO: check the amount of times open is used on a file, 
+#       i bet this could be reduced quite a bit.
+# add option to specify path to oslinfo -> use ini file
+# DONE: add option for rsl shaders and vex shaders, detect file extension
+# add option to query json file? YES, maybe other tool
+# add option to add or overwrite
+# add option to check if shader already exists
+# DONE: add option to parse the/multiple directory/ies (or is default behaviour, check if it is a path or if we have files
+# DONE: add option to read paths and files from a file
+
+
+import json
+import glob
+import os.path
+import sys
+import optparse
+import hashlib
+import datetime
+import getpass
+
+DEBUG = True
+
+
+class Shader():
+    def __init__( self ):
+        self.name = "name"
+        self.path = "path"
+        self.hash = "hash"
+        self.num  = "num"
+    
+
+FileTypes = {'.oso' : "openshadinglanguage",
+             '.sdl' : "3delight rsl",
+             '.slo' : "RSL",
+             '.otl' : "vex"
+            }
+
+# metadata according to the OSL specification
+_shaderTypes = ["surface", "displacement", "light", "volume", "shader"]
+_dataTypes   = ["int", "float", "point", "vector", "normal", "color", "matrix", "string", "void"]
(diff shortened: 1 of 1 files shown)


dictoon:
Can you tell a bit more what this script does and when it is useful? Maybe you could also provide sample input and output files?

scripts/oslextractmeta.py
```diff
@@ -0,0 +1,466 @@
...
+import sys
+import optparse
+import hashlib
+import datetime
+import getpass
+
+DEBUG = True
+
+
+class Shader():
```
Narann: Why not derive from `object`? It offer a lot of interesting stuff.

scripts/oslextractmeta.py
```diff
@@ -0,0 +1,466 @@
...
+             '.sdl' : "3delight rsl",
+             '.slo' : "RSL",
+             '.otl' : "vex"
+            }
+
+# metadata according to the OSL specification
+_shaderTypes = ["surface", "displacement", "light", "volume", "shader"]
+_dataTypes   = ["int", "float", "point", "vector", "normal", "color", "matrix", "string", "void"]
+_shaderKeys  = ["name", "label", "type", "help", "url", "value", "page", "widget", "units"]
+#_parmWidgets = ["number", "string", "boolean", "checkBox", "popup", "mapper", "filename", "null"]
```
Narann: Why comment those lines.

scripts/oslextractmeta.py
```diff
@@ -0,0 +1,466 @@
...
+
+    return hashMethod.digest()[:16]
+
+
+def parseShaderInfo( compiledShader, FileTypes ):
+    (name, extension) = os.path.splitext( compiledShader )
+    shaderUI = None
+    if extension == '.oso':
+        shaderUI = parseOslInfo( compiledShader )
+    #if extension == '.sdl':
```
Narann: Why comments here?

scripts/oslextractmeta.py
```diff
@@ -0,0 +1,466 @@
...
+        _error( "Could not process %s" % compiledShader )
+        return None
+    else:
+        compShader = dict()
+        #initialize id here, set one level up
+        compShader[ 'element' ]    = 0
+        compShader[ 'name' ]  = shaderUI[ 'name' ]
+        compShader[ 'path' ] = compiledShader
+        compShader[ 'hash' ]  = 1
+        #str( calcCheckSum( compiledShader ) )
```
Narann: Why comments here?

scripts/oslextractmeta.py
```diff
@@ -0,0 +1,466 @@
...
+        compShader[ 'mtime' ] = str( os.path.getmtime( compiledShader ) )
+        compShader[ 'ctime' ] = str( datetime.datetime.now() )
+        compShader[ 'language' ] = FileTypes[ extension ]
+        compShader[ 'ui'   ]  = shaderUI
+
+        return compShader
+
+
+def getNumberOfElements( jsonFile ):
+    for i in range ( len( jsonFile ) ):
```
Narann: Not pythonic, please use something like:

```python
for data in jsonFile:
    data['elements']
```

scripts/oslextractmeta.py
```diff
@@ -0,0 +1,466 @@
...
+        if update and existingFile:
+            newHeader = updateJsonHeader( json_temp, numElements )
+            deleteHeader( json_temp )
+            json.dump( json_temp, sui_json )
+            # and update the header
+            json.dump( newHeader, fp_json )
+        else:
+            json.dump( writeJsonHeader( sui_json, numElements ), fp_json )
+
+#if DEBUG:
```
Narann: Why comments here?

dictoon:
To be honest, as soon as I understand what this is for and why it's useful, I'll merge it and we can all work on improving and cleaning up the code. It doesn't have to be perfect from the get go :)

SecretChief:
The idea was to accumulate the ui of (osl) shaders into a JSON file. The JSON file can then be parsed  to build an interface in appleseed/houdini/maya whatever. For sample input you can use the osl shaders from appleseed.

Narann:
I like the idea but I have the impression it's very specific. What if the third party application don't use JSON? What if the third party application use JSON but already has its own JSON hierarchy? You have done 90% of the work (extract info from OSL file). It's a shame to lost this in favor of a specific file format.

Let me give you some suggestion:

You say you want to use python script to generate JSON file. I would suggest to replace the _scripting_ idea by a generic _module_.

A tiny python module `osl_meta`

```python
shader = osl_meta.load('foo.osl') # where data is a class istance with all metadata values
shader.type # "surface", "displacement", "light", "volume", "shader"
shader.name
shader.help
# etc...
```

With such this, every application having python can use it to directly extract info without having to export to a JSON file.

Once you have this, each third party devs can generate their own organized JSON file from a dict. Or an XML, or another one.

But if you _really_ want to push application specific JSON into appleseed master you could do this:

```python
data.to_json(file_path)
```

Providing a _default_ json file.

My 2 cts.

SecretChief:
Yes, it is a very specific application. The parseOslInfo function basically does what you describe in osl_meta, I could split up the file in some generic library and one Esteban wants to use for the appleseed shaders. I will probably rewrite houseed as well or make a separate script to parse the JSON file. 
Initially I wanted to do a version that does parse 3delight and Renderman files as well but that would make it more of a separate project and not appleseed specific...

est77:
This script will be really useful. We can run it as part of appleseed build to extract metadata for OSL shaders and use it later in appleseed.studio, blenderseed, ...

But I think that the tool could be simpler, both the implementation and the usage.
We probably don't need any of the RSL or VEX stuff, we could also restrict 
the output to json or XML, always overwriting the output file and removing the update options and the hashing code.

The way shaders are specified could be simplified. Maybe allowing the user to specify a 
directory, recursively walking the dir and collecting the metadata for all oso files found.

An option to specify a path for oslinfo would be nice in case users don't have it on the PATH.

dictoon:
Alright, sounds good to me. Let's get this merged and let's improve on it. Thanks all!

SecretChief:
Narann: class name() creates a class parenting the pre python 2.2 object model. It misses some functionality like super which you'll get with the new python2.2+ object class, which you will get with class name(object).

Refactoring a lot of code with the suggestions of Narann (remove very guly for loops) and stripping out references to other shading languages.

Issue #747, opened 2015-05-17
appleseed.python: crashes when a MasterRenderer instance is garbage collected


est77:
appleseed sometimes crashes when instances of MasterRenderer are garbage collected.
Since PR #719, MasterRenderers must be destroyed before the project is destroyed. 
This is not taken into account in the current bindings and crashes can happen if Python destroys Project instances before MasterRenderers.

PR #748, opened 2015-05-17
Fix possible appleseed python crashes when MasterRenderer instances are destroyed.

est77:
Fix for issue #747.
Introduced a wrapper class for MasterRenderer in the bindings that keeps a python reference to 
the Project instance and makes sure everything is destroyed in the correct order.

### DIFF
--- a/src/appleseed.python/bindmasterrenderer.cpp
+++ b/src/appleseed.python/bindmasterrenderer.cpp
@@ -45,57 +45,100 @@ using namespace renderer;
 
 namespace
 {
-    std::auto_ptr<MasterRenderer> create_master_renderer(
-        Project*                project,
+    // A class that wraps MasterRenderer and keeps a python
+    // reference to the project object to prevent it being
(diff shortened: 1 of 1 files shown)


src/appleseed.python/bindmasterrenderer.cpp
```diff
@@ -45,57 +45,100 @@ using namespace renderer;
...
                     bpy_dict_to_param_array(params),
                     renderer_controller,
                     tile_callback));
     }
 
-    bpy::dict master_renderer_get_parameters(const MasterRenderer* m)
+    bpy::dict master_renderer_get_parameters(const MasterRendererWrapper* m)
     {
-        return param_array_to_bpy_dict(m->get_parameters());
+        return param_array_to_bpy_dict(m->m_renderer->get_parameters());
```
dictoon: You could even overload `MasterRendererWrapper::operator->` to make it more transparent and minimize the number of changes...

Issue #746, opened 2015-05-13
Metadata for render settings


est77:
Configurations could have metadata describing the different options, just like we already do for Entities.

Could be useful in DCC integration plugins, automatic documentation tools, ...

dictoon:
Absolutely!

PR #749, opened 2015-05-21
Started adding metadata to Configurations

est77:
This PR add some basic metadata to Configurations (render settings). Ticket #746
Metadata can be useful in DCC plugins, to generate automatic documentation, ...

Metadata is accesible from C++ and from appleseed.python as nested dictionaries.

I only added metadata for the settings used in GafferAppleseed for now.
Some help strings are missing and many others can certainly be improved.

But I think we can merge the PR if the implementation is ok and improve things as needed.

### DIFF
--- a/src/appleseed/renderer/kernel/lighting/drt/drtlightingengine.cpp
+++ b/src/appleseed/renderer/kernel/lighting/drt/drtlightingengine.cpp
@@ -53,6 +53,7 @@
 #include "foundation/math/population.h"
 #include "foundation/math/vector.h"
 #include "foundation/platform/types.h"
+#include "foundation/utility/containers/dictionary.h"
 #include "foundation/utility/statistics.h"
 #include "foundation/utility/string.h"
 
@@ -458,6 +459,31 @@ void DRTLightingEngineFactory::release()
 ILightingEngine* DRTLightingEngineFactory::create()
 {
     return new DRTLightingEngine(m_light_sampler, m_params);
+}
+
+Dictionary DRTLightingEngineFactory::get_params_metadata()
+{
+    Dictionary metadata;
+    add_common_params_metadata(metadata, true);
+
+    metadata.dictionaries().insert(
+        "max_path_length",
+        Dictionary()
+            .insert("type", "int")
+            .insert("default", "8")
+            .insert("unlimited", "true")
+            .insert("min", "1")
+            .insert("help", "Maximum ray trace depth"));
+
+    metadata.dictionaries().insert(
+        "rr_min_path_length",
+        Dictionary()
+            .insert("type", "int")
+            .insert("default", "3")
+            .insert("min", "1")
+            .insert("help", "Consider pruning low contribution paths starting with this bounce"));
+
+    return metadata;
 }
 
 }   // namespace renderer
--- a/src/appleseed/renderer/kernel/lighting/ilightingengine.cpp
+++ b/src/appleseed/renderer/kernel/lighting/ilightingengine.cpp
@@ -5,8 +5,7 @@
 //
 // This software is released under the MIT license.
 //
-// Copyright (c) 2010-2013 Francois Beaune, Jupiter Jazz Limited
-// Copyright (c) 2014-2015 Francois Beaune, The appleseedhq Organization
+// Copyright (c) 2015 Esteban Tovagliari, The appleseedhq Organization
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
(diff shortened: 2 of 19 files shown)


src/appleseed/renderer/kernel/lighting/drt/drtlightingengine.cpp
```diff
@@ -460,4 +461,36 @@ ILightingEngine* DRTLightingEngineFactory::create()
     return new DRTLightingEngine(m_light_sampler, m_params);
 }
 
+Dictionary DRTLightingEngineFactory::get_params_metadata()
+{
+    Dictionary metadata;
+    add_common_params_metadata(metadata);
+
+    metadata.dictionaries().insert(
+        "max_path_lenght",
```
dictoon: Typo: `max_path_length`

My advice: search & fix `lenght` everywhere in appleseed...

src/appleseed/renderer/kernel/lighting/ilightingengine.cpp
```diff
@@ -0,0 +1,55 @@
...
+// appleseed.foundation headers.
+#include "foundation/utility/containers/dictionary.h"
+
+using namespace foundation;
+
+namespace renderer
+{
+
+//
+// ILightingEngine class implementation.
```
dictoon: I tend to remove these comment blocks noawadays, if there's a single class definition in the file (I keep them when multiple classes are defined in the file).

In any case, it should be `ILightingEngineFactory` if you decide to keep it.

src/appleseed/renderer/kernel/lighting/ilightingengine.cpp
```diff
@@ -0,0 +1,55 @@
...
+//
+
+void ILightingEngineFactory::add_common_params_metadata(foundation::Dictionary& metadata)
+{
+    metadata.dictionaries().insert(
+        "enable_ibl",
+        Dictionary()
+            .insert("type", "bool")
+            .insert("default", "on")
+            .insert("help", "Enable image based lighting"));
```
dictoon: "image-based"

src/appleseed/renderer/kernel/lighting/ilightingengine.cpp
```diff
@@ -0,0 +1,55 @@
+
+//
+// This source file is part of appleseed.
+// Visit http://appleseedhq.net/ for additional information and resources.
+//
+// This software is released under the MIT license.
+//
+// Copyright (c) 2010-2013 Francois Beaune, Jupiter Jazz Limited
```
dictoon: Update copyright notice.

src/appleseed/renderer/kernel/lighting/drt/drtlightingengine.h
```diff
@@ -62,6 +63,10 @@ class DRTLightingEngineFactory
     // Return a new DRT lighting engine instance.
     virtual ILightingEngine* create() APPLESEED_OVERRIDE;
 
+    // Get the metadata dictionary describing
+    // the drt lighting engine params.
```
dictoon: drt -> DRT

src/appleseed/renderer/kernel/lighting/pt/ptlightingengine.cpp
```diff
@@ -710,4 +710,61 @@ ILightingEngine* PTLightingEngineFactory::create()
...
+
+    metadata.dictionaries().insert(
+        "enable_caustics",
+        Dictionary()
+            .insert("type", "bool")
+            .insert("default", "false")
+            .insert("help", "Enable caustics"));
+
+    metadata.dictionaries().insert(
+        "max_path_lenght",
```
dictoon: `max_path_length`

src/appleseed/renderer/kernel/lighting/pt/ptlightingengine.cpp
```diff
@@ -710,4 +710,61 @@ ILightingEngine* PTLightingEngineFactory::create()
...
+        Dictionary()
+            .insert("type", "int")
+            .insert("help", "Maximum number of light bounces"));
+
+    metadata.dictionaries().insert(
+        "rr_min_path_length",
+        Dictionary()
+            .insert("type", "int")
+            .insert("help", "TODO: Not documented yet"));
```
dictoon: "Consider pruning low contribution paths after that many bounces"

src/appleseed/renderer/kernel/lighting/pt/ptlightingengine.cpp
```diff
@@ -710,4 +710,61 @@ ILightingEngine* PTLightingEngineFactory::create()
...
+            .insert("type", "int")
+            .insert("help", "Number of samples used to estimate direct lighting"));
+
+    metadata.dictionaries().insert(
+        "ibl_env_samples",
+        Dictionary()
+            .insert("type", "int")
+            .insert("help", "Number of samples used to estimate environment lighting"));
+
+    metadata.dictionaries().insert(
```
dictoon: Next event estimation should pretty much always be enabled. It is kept for educational purposes only. I advise against exposing it.

src/appleseed/renderer/kernel/lighting/pt/ptlightingengine.cpp
```diff
@@ -710,4 +710,61 @@ ILightingEngine* PTLightingEngineFactory::create()
...
+        "next_event_estimation",
+        Dictionary()
+            .insert("type", "bool")
+            .insert("help", "TODO: Not documented yet"));
+
+    metadata.dictionaries().insert(
+        "max_ray_intensity",
```
dictoon: "Clamp ray intensity to this value to reduce fireflies"

src/appleseed/renderer/kernel/lighting/pt/ptlightingengine.h
```diff
@@ -62,6 +63,10 @@ class PTLightingEngineFactory
     // Return a new path tracing lighting engine instance.
     virtual ILightingEngine* create() APPLESEED_OVERRIDE;
 
+    // Get the metadata dictionary describing
+    // the pt lighting engine params.
```
dictoon: pt -> PT

src/appleseed/renderer/kernel/lighting/sppm/sppmlightingengine.cpp
```diff
@@ -664,4 +665,89 @@ ILightingEngine* SPPMLightingEngineFactory::create()
...
+        Dictionary()
+            .insert("type", "bool")
+            .insert("default", "false")
+            .insert("help", "Enable caustics"));
+
+    metadata.dictionaries().insert(
+        "photon_type",
+        Dictionary()
+            .insert("type", "enum")
+            .insert("options", "mono|poly")
```
dictoon: Maybe options should be a nested dictionary. That way we could add a description for each value...

src/appleseed/renderer/kernel/lighting/sppm/sppmlightingengine.h
```diff
@@ -64,6 +65,10 @@ class SPPMLightingEngineFactory
     // Return a new SPPM lighting engine instance.
     virtual ILightingEngine* create() APPLESEED_OVERRIDE;
 
+    // Get the metadata dictionary describing
+    // the sppm lighting engine params.
```
dictoon: sppm -> SPPM

src/appleseed/renderer/kernel/texturing/texturestore.cpp
```diff
@@ -77,6 +78,19 @@ StatisticsVector TextureStore::get_statistics() const
     return StatisticsVector::make("texture store statistics", stats);
 }
 
+Dictionary TextureStore::get_params_metadata()
+{
+    Dictionary metadata;
+    metadata.dictionaries().insert(
+        "max_size",
```
dictoon: Let's expose this as MB instead of bytes maybe?

src/appleseed/renderer/kernel/rendering/final/uniformpixelrenderer.cpp
```diff
@@ -275,4 +276,31 @@ IPixelRenderer* UniformPixelRendererFactory::create(
...
+            .insert("type", "int")
+            .insert("default", "64")
+            .insert("help", "Number of anti-aliasing samples"));
+
+    metadata.dictionaries().insert(
+        "force_antialiasing",
+        Dictionary()
+            .insert("type", "bool")
+            .insert("default", "false")
+            .insert("help", "TODO: Not documented yet"));
```
dictoon: When using 1 sample/pixel and this is disabled, samples are placed in the middle of the pixels.

src/appleseed/renderer/kernel/rendering/final/uniformpixelrenderer.cpp
```diff
@@ -275,4 +276,31 @@ IPixelRenderer* UniformPixelRendererFactory::create(
...
+
+    metadata.dictionaries().insert(
+        "decorrelate_pixels",
```
dictoon: Not sure this should be exposed. It should be enabled when the number of passes > 1, and disabled when number of passes == 1.

src/appleseed/renderer/kernel/texturing/texturestore.cpp
est77: I think that metadata should reflect the units and values of the parameters as used in the code, 
and it would be the plugin responsibility to expose some parameters in a different way if that's more convenient.

Or do you mean switching max_size everywhere in the code to MBs?

dictoon: You're right, let's keep as is for now.

Maybe in the future it'll make sense to expose MB instead of bytes everywhere where it makes sense (such as for texture caches), but for now it's fine, and we're consistent.

est77:
default = "unlimited" has some problems.
We loose the default value in case unlimited == off, for example pt.max_path_length defaults to 8 if unlimited is unchecked. The other issue is that code using the metadata becomes more complex.
param["default"] can now return two different types, a string or an int / float.

My idea was adding a ("unlimited", true) field to unlimited params. Plugins can check the tag and decide to expose the parameters in their UI as two controls: a checkbox and a numeric input, just like we do in studio.

est77:
Rebased

PR #750, opened 2015-05-23
Oslextractmeta

SecretChief:
oslextractmeta improvements and code clean up.

Changes:
- switched to argparse instead of optionparser,
- added support for a configuration file as requested,
- cleaned code to be compliant with pep008,
- added code for directory traversal

### DIFF
--- a/scripts/oslextractmeta.py
+++ b/scripts/oslextractmeta.py
@@ -64,6 +64,8 @@ def _error(msg, crash=False):
         sys.exit(1)
     return False
 
+def _fatalError(msg):
+    _error(msg,True)
 
 def _formatVal(st):
     value = st.replace('"','',2)
@@ -150,7 +152,7 @@ def createFileList(filetypes, osl_cfg, recursive=False, args=None, pathfile=None
 
     # if there are no files/paths quit
     if len(filelist) < 1:
-        _error("No files or directories found, exiting.", True)
+        _fatalError("No files or directories found, exiting.", True)
     
     return filelist
 
@@ -170,7 +172,7 @@ def parseOslInfo(compiledShader, osl_cfg):
     try:
         fp = subprocess.check_output(cmd)
     except subprocess.CalledProcessError as fp_ret:
-        _error("Could not run oslinfo, exiting.\nReturncode: %s" % fp_ret.returncode, True)
+        _fatalError("Could not run oslinfo, exiting.\nReturncode: %s" % fp_ret.returncode, True)
     
     # check if output of oslinfo is correct
     # if false skip shader and write error message to console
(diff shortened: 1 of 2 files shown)


scripts/oslextractmeta.py
```diff
@@ -0,0 +1,453 @@
...
+#_parmFloat   = _parmInteger + ["digits"]
+#_parmSlider  = ["slidermin", "slidermax", "slidercenter", "sliderexponent"]
+#_parmKeyword = ["output"]
+
+
+#----------------------------------------------------------
+# Functions to sanitize olsinfo output
+#----------------------------------------------------------
+
+def _error(msg, crash=False):
```
est77: Maybe we can have two error functions, _error() and _fatalError(), instead of using a bool arg?

scripts/oslextractmeta.py
```diff
@@ -0,0 +1,453 @@
...
+        for x in jsondict['shaders'].values():
+            print x['ui']
+        ''')
+    parser.add_argument('-i', '--input', nargs='+', action='store', dest='files', metavar='compiled shaders',help='List of file(s) to parse.')
+    parser.add_argument('-v', '--verbose', action='store_true', dest='verbosity', help='Increase output verbosity.')
+    parser.add_argument('-o', '--output', nargs=1, action='store', dest='output', required=True, metavar='output file', help="Store shader UI in file.")
+    parser.add_argument('-f', '--file', nargs='+', action='store', dest='read_file', metavar='file',  help="Read file paths from file(s).")
+    parser.add_argument('-U', '--update', action='store_true', dest='update', help="Update existing shader file.")
+    parser.add_argument('-O', '--overwrite', action='store_true', dest='overwrite', help="Overwrite existing files.")
+    parser.add_argument('-c', '--clean', action='store_true', dest='clean', help="Clean file, remove non existant shaders.")
```
est77: I'd remove the update option and always overwrite the json file.
Running the script shouldn't be too expensive and it's not something users will do frequently.
Overwriting all the time should be ok and all the code related to update and clean could be removed.

SecretChief:
The error functions would both have 90% of same code, does it make sense to split them up?

Regarding updating/overwriting, default action is adding new shaders to the json file. Overwriting as standard only makes sense if we use the script only for shaders supplied by appleseed in which case we could also hard code the relative path. If you have shaders in other directories you need to supply the paths each time or, when you added a new shader, add a path to the configuration file or an external file. If the functionality proves unwanted or cumbersome with the follow up compile scripts we can remove it.

est77:
> The error functions would both have 90% of same code, does it make sense to split them up?

I prefer having them separate. But internally, _fatalError() can call _error() and then exit.
without repeating the code.

PR #751, opened 2015-05-24
oslextractmeta improvements and code clean up.

SecretChief:
oslextractmeta improvements and code clean up.

Changes:
-switched to argparse instead of optionparser,
-added support for a configuration file as requested,
-cleaned code to be compliant with pep008,
-added code for directory traversal
-_fatalError function added


est77:
Manually merged. Thanks!!

PR #756, opened 2015-05-28
python: use nested dicts for entity params metadata

est77:
Previously lists of dictionaries were used. 
Now, metadata for entity parameters can now be accessed as a dict:

md = appleseed.Light.get_input_metadata()
md['directional_light']['irradiance']['help']


dictoon:
Cool!

PR #767, opened 2015-06-16
fixed BezierCurveBase<T, N>::compute_recursion_depth() about z component.

KojiNakamaru:
This fix introduces a subexpression about z as described in Errata (http://eaflux.com/index.html.en):

> Page 2, bottom right: If control points are placed non-uniformly in ray direction, the resulting v parameter may have a large margin of error because the algorithm computes v in 2D space. To compensate this, we should add a subexpression about z in the maximum depth computation. (Thanks to Emil Kirichev)

Anyway, thank you very much for implementing the algorithm!

### DIFF
--- a/src/appleseed/foundation/math/beziercurve.h
+++ b/src/appleseed/foundation/math/beziercurve.h
@@ -365,7 +365,8 @@ size_t BezierCurveBase<T, N>::compute_recursion_depth(const ValueType epsilon) c
             max(
                 l0,
                 std::abs(m_ctrl_pts[i].x - ValueType(2.0) * m_ctrl_pts[i + 1].x + m_ctrl_pts[i + 2].x),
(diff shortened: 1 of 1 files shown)


dictoon:
Hi Koji, thanks for the pull request!

How one would construct a curve that would exhibit the problem that this PR fixes?

KojiNakamaru:
The following control points for the ray with direction `(0, 0, 1)`

```
( 0.0,  2.0,  4000.0)
( 0.0,  1.0,  2000.0)
( 0.0, -1.0,  1000.0)
( 0.0, -2.0,     0.0)
```

will demonstrate the issue. As in this example, if a curve has non-uniform x or y coordinates and also has non-uniform z coordinates. coarse subdivision by the original 2D equation -- though it works well in many cases -- may lead relatively large numerical errors.

est77:
Thanks!!

PR #776, opened 2015-06-29
SSS work

est77:
- BSSRDF interface improvements.
- Started directional dipole implementation.


src/appleseed/renderer/modeling/bssrdf/directionaldipolebssrdf.h
```diff
@@ -60,14 +60,41 @@ namespace renderer
...
+    double      m_mean_free_path_multiplier;
+    double      m_anisotropy;
+    double      m_from_ior;
+    double      m_to_ior;
+
+    // Precomputed values.
+    Spectrum    m_cdf;
+    double      m_max_mean_free_path;
+
+    // Convenience accessors for use after reflectance and
```
dictoon: Missing end of sentence?

src/appleseed/renderer/modeling/bssrdf/directionaldipolebssrdf.cpp
```diff
@@ -79,10 +103,230 @@ namespace
...
+            const double r2 = square(r);
+
+            // Modified normal.
+            const Vector3d ni_s = cross(normalize(xoxi), normalize(cross(ni, xoxi)));
+
+            // Directions of ray sources.
+            const double eta = values->m_to_ior / values->m_from_ior;
+            const double nnt = 1.0 / eta;
+            const double ddn = -dot(wi, ni);
+            Vector3d wr = normalize(wi * -nnt - ni * (ddn * nnt + std::sqrt(1.0 - nnt * nnt * (1.0 - ddn * ddn))));
```
dictoon: Make const?

src/appleseed/renderer/modeling/bssrdf/directionaldipolebssrdf.h
dictoon: Ah I understand what you mean: At some point `m_reflectance` and `m_mean_free_path` change meaning, and so it's clearer to use different identifiers.

I think it would be just cleaner (and no less efficient, really, since we don't store these values) to just introduce new members `m_sigma_a` and `m_sigma_s_prime`.

src/appleseed/renderer/modeling/bssrdf/directionaldipolebssrdf.cpp
```diff
@@ -79,10 +103,230 @@ namespace
...
+                value);
+
+            value *= 0.5f;
+#endif
+        }
+      private:
+
+        static double compute_rd(double alpha_prime, double two_c1, double three_c2)
+        {
+            double cphi = 0.25 * (1.0 - two_c1);
```
dictoon: Make everything const?

src/appleseed/renderer/modeling/bssrdf/directionaldipolebssrdf.cpp
```diff
@@ -79,10 +103,230 @@ namespace
...
+        }
+
+        static double compute_alpha_prime(double rd, double c1, double c2)
+        {
+            const double c12 = 2.0 * c1;
+            const double c23 = 3.0 * c2;
+
+            double x0 = 0.0, x1 = 1.0, xmid;
+
+            // For now simple bisection.
```
dictoon: We have a solver based on Newton's method in `foundation/math/root.h`, if that can be useful.

PR #808, opened 2015-07-21
Temporary switch dirpole to absorption and scattering inputs

est77:
To simplify testing and debugging. Presets can now be used directly in the directional dipole BSSRDF.

### DIFF
--- a/src/appleseed/renderer/modeling/bssrdf/directionaldipolebssrdf.cpp
+++ b/src/appleseed/renderer/modeling/bssrdf/directionaldipolebssrdf.cpp
@@ -83,10 +83,10 @@ namespace
             const ParamArray&       params)
           : BSSRDF(name, params)
         {
-            m_inputs.declare("reflectance", InputFormatSpectralReflectance);
-            m_inputs.declare("reflectance_multiplier", InputFormatScalar, "1.0");
-            m_inputs.declare("dmfp", InputFormatSpectralReflectance);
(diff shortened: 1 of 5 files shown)


src/appleseed/renderer/modeling/bssrdf/directionaldipolebssrdf.cpp
```diff
@@ -184,19 +186,6 @@ namespace
                 outgoing_point.get_point(),
                 outgoing_point.get_shading_normal(),
                 value);
-
```
dictoon: Keep this. We'll want it eventually, and it doesn't hurt as long as it's commented out.

src/appleseed/renderer/modeling/bssrdf/directionaldipolebssrdf.cpp
```diff
@@ -387,47 +376,53 @@ DictionaryArray DirectionalDipoleBSSRDFFactory::get_input_metadata() const
...
                 Dictionary()
                     .insert("color", "Colors")
                     .insert("texture_instance", "Textures"))
             .insert("use", "required")
-            .insert("default", "0.5"));
+            .insert("default", "2.29"));
 
+    // Since presets are usually specified in mm.,
+    // we set the multipliers so that the values are correct
+    // for scenes modelled in meters.
```
dictoon: modeled

PR #826, opened 2015-07-31
SSS work

est77:
- Precompute sigma_tr for BSSRDFs that use it.
- Renamed dipole sampling functions.
- Added more tests and plots.


src/appleseed/renderer/meta/tests/test_sss.cpp
```diff
@@ -95,6 +96,98 @@ TEST_SUITE(Renderer_Modeling_BSSRDF_SSS)
...
+                Eta,
+                Anisotropy,
+                sigma_a,
+                sigma_s,
+                sigma_tr);
+
+            const float sa = sigma_a1[0] / static_cast<float>(dmfp);
+            const float ss = sigma_s1[0] / static_cast<float>(dmfp);
+
+            EXPECT_FEQ_EPS(sa, sigma_a[0], 0.0001f);
```
dictoon: You can probably use a much smaller epsilon value here, like 1.0e-6.

src/appleseed/renderer/meta/tests/test_sss.cpp
```diff
@@ -501,4 +600,178 @@ TEST_SUITE(Renderer_Modeling_BSSRDF_SSS)
...
+            Spectrum result;
+
+            dp_bssrdf->evaluate(
+                &dp_values,
+                outgoing,
+                normal,
+                incoming,
+                normal,
+                result);
+            dp_points.push_back(Vector2d(r, result[0] /  r));
```
dictoon: I don't think this makes sense, the directional dipole's `evaluate()` method must (and does) return r * R(r), not just R(r).

src/appleseed/renderer/modeling/bssrdf/sss.cpp
```diff
@@ -112,7 +112,38 @@ double reduced_extinction_coefficient(
     const double    dmfp,
     const double    alpha_prime)
 {
-    return 1.0 / (sqrt(3.0 * (1.0 - alpha_prime)) * dmfp);
+    const double sigma_tr = 1.0 / dmfp;
```
dictoon: The original expression was more efficient (one less divide). Maybe simply add a comment saying that 1.0 / dmfp is sigma_tr?

src/appleseed/renderer/modeling/bssrdf/sss.h
```diff
@@ -183,24 +195,23 @@ double normalized_diffusion_max_radius(
...
 //
-// Attenuation sampling.
+// Dipole sampling.
 //
 // Reference:
 //
-//   Volumetric Path Tracing, Steve Marschner, section 3.
-//   http://www.cs.cornell.edu/courses/cs6630/2012sp/notes/09volpath.pdf
+//   A Practical Model for Subsurface Light Transport
+//   https://graphics.stanford.edu/papers/bssrdf/bssrdf.pdf
```
dictoon: Was the Volumetric Path Tracing reference relevant / useful / interesting? If so, it might make sense to keep it.

src/appleseed/renderer/meta/tests/test_sss.cpp
```diff
@@ -31,6 +31,7 @@
 #include "renderer/kernel/shading/shadingpointbuilder.h"
 #include "renderer/modeling/bssrdf/bssrdf.h"
 #include "renderer/modeling/bssrdf/directionaldipolebssrdf.h"
+#include "renderer/modeling/bssrdf/normalizeddiffusionbssrdf.h"
```
dictoon: Normalized diffusion tests must be conditioned by the `APPLESEED_WITH_NORMALIZED_DIFFUSION_BSSRDF` preprocessor symbol.

PR #827, opened 2015-08-02
SSS work

dictoon:


src/appleseed/foundation/math/scalar.h
```diff
@@ -71,8 +71,9 @@ static const double TwoPi           =  6.2831853071795865;      // 2 * Pi
 static const double HalfPi          =  1.5707963267948966;      // Pi / 2
 static const double RcpPi           =  0.3183098861837907;      // 1 / Pi
 static const double RcpTwoPi        =  0.1591549430918953;      // 1 / (2 * Pi)
+static const double RcpFourPi       =  0.0795774715459477;      // 1 / (4 * Pi)
 static const double RcpHalfPi       =  0.6366197723675813;      // 1 / (Pi/2) = 2 / Pi
-static const double RcpPiSq         =  0.1013211836423377;      // 1 / (Pi^2) = (1 / Pi)^2
+static const double RcpPiSq         =  0.1013211836423378;      // 1 / (Pi^2) = (1 / Pi)^2
```
Narann: Is `static const` better than a `#define`? I'm just curious. :)

dictoon: In this particular case, yes, since the whole point of these constants is to force evaluation of the expressions.

dictoon: Woops, sorry, not properly awaken: the expressions are already evaluated :) So no, `static const` and `#define` are essentially the same. It's simply a question of preference.

dictoon: Well now it appears to me that we *could* write the actual expressions and let the compiler evaluate them, if we had the guarantee it would evaluate them at compile-time...

Narann: lol I think all those confusing is my answer. :D

dictoon: Let's see: What do you do when you need sqrt(2)/2 somewhere in the renderer's code?

1. You could just write it:

        const Vector3d u = (sqrt(2.0) / 2, 0.0, -sqrt(2.0) / 2);

    Unfortunately, this *may* compute `sqrt(2.0)` twice, which is expensive. Of course the compiler *may* compute it ony once and reuse the value twice, or it may compute `sqrt(2.0)` at compile-time. There are no guarantees on what it will actually do since this will depend on the compiler and the compilation settings used.

2. Writing

        #define RcpSqrtTwo (sqrt(2.0) / 2)

    and then

        const Vector3d u = (RcpSqrtTwo , 0.0, -RcpSqrtTwo);

    is exactly the same thing as case 1, since preprocessor symbols are just replaced lexically.

3. Another option would be to write

        static const double RcpSqrtTwo = sqrt(2.0) / 2;

    This is definitely better. There are still two options though: either the compiler evaluates it at compile-time and everything is great, or it evaluates it once when the program starts and store the value somewhere in memory.

4. We wrote

        static const double RcpSqrtTwo =  0.7071067811865475;

    which has the advantage that no computation or precomputation will be needed, and we're *almost* sure that value won't be stored in memory and will instead be used directly where it's needed. In other words, it's the safest option, even though we are *still* not guaranteed (as far as I know) that this is equivalent to

        const Vector3d u = (0.7071067811865475 , 0.0, -0.7071067811865475);

5. What we could have done is writing

        #define RcpSqrtTwo 0.7071067811865475

    In this case we would be 100% sure of what is happening.

If we find out that option 4 is not equivalent to option 5, we'll switch to option 5.

PR #833, opened 2015-08-04
SSS OSL work

est77:
- Exposed BSSRDFs as OSL closures.
- BSSRDF interface changes to support OSL.
- Implemented OSL BSSRDF.
- Cleanup stdosl.h OSL header.
- Execute OSL shaders when computing SSS contribution.

### DIFF
--- a/src/appleseed/renderer/kernel/shading/closures.cpp
+++ b/src/appleseed/renderer/kernel/shading/closures.cpp
@@ -83,6 +83,11 @@ namespace
         OSL::ustring    tag;
     };
 
+    struct DiffuseBSDFClosureParams
+    {
+        OSL::Vec3       N;
+    };
+
     struct DisneyBRDFClosureParams
     {
         OSL::Vec3       N;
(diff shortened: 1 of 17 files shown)


src/appleseed/renderer/kernel/shading/closures.cpp
```diff
@@ -100,9 +105,24 @@ namespace
         float           clearcoat_gloss;
     };
 
-    struct DiffuseBSDFClosureParams
+    OSL::ustring better_dipole_profile_name("better_dipole");
+    OSL::ustring dipole_profile_name("standard_dipole");
```
dictoon: Rename string to `standard_dipole_profile_name` and move to first position.

dictoon:
Looks pretty good!

I'm wondering about the introduction of a second `evaluate_inputs()` method in `BSSRDF`... Does it really make things simpler?

dictoon:
Also, if we decide to keep both `evaluate_inputs()` method, I think the `data` argument for the simpler one should be of type `void*` instead of `foundation::uint8*`, for the sake of consistency (data arguments are `void*` everywhere else).

PR #834, opened 2015-08-05
Split inputs evaluation and inputs precomputations for both BSDFs and BSSRDFs.

est77:
This is needed when using BSSRDFs as OSL closures.
In that case, we don't need evaluation. Input values are initialized with the arguments of OSL closures 
and prepare_inputs can now be called to perform precomputations needed on input values.

For consistency, do the same with BSDFs.

### DIFF
--- a/src/appleseed/renderer/modeling/bsdf/bsdf.cpp
+++ b/src/appleseed/renderer/modeling/bsdf/bsdf.cpp
@@ -94,7 +94,7 @@ void BSDF::evaluate_inputs(
     const size_t            offset) const
 {
     input_evaluator.evaluate(get_inputs(), shading_point.get_uv(0), offset);
-    prepare_inputs(reinterpret_cast<uint8*>(input_evaluator.data()) + offset);
+    prepare_inputs(input_evaluator.data() + offset);
 }
 
 void BSDF::prepare_inputs(void* data) const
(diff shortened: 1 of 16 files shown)


src/appleseed/renderer/modeling/bsdf/bsdf.cpp
```diff
@@ -94,6 +94,11 @@ void BSDF::evaluate_inputs(
     const size_t            offset) const
 {
     input_evaluator.evaluate(get_inputs(), shading_point.get_uv(0), offset);
+    prepare_inputs(reinterpret_cast<uint8*>(input_evaluator.data()) + offset);
```
dictoon: `reinterpret_cast<>` redundant, `data()` already returns `uint8*`.

src/appleseed/renderer/modeling/bsdf/osl/oslbsdf.cpp
```diff
@@ -226,6 +226,12 @@ namespace
                 this,
                 shading_point.get_shading_basis(),
                 shading_point.get_osl_shader_globals().Ci);
+
+            for (size_t i = 0, e = c->get_num_closures(); i < e; ++i)
+            {
+               bsdf_from_closure_id(c->get_closure_type(i)).prepare_inputs(
```
dictoon: Fix indentation.

src/appleseed/renderer/modeling/bssrdf/bssrdf.cpp
```diff
@@ -87,7 +87,7 @@ void BSSRDF::evaluate_inputs(
     const size_t            offset) const
 {
     input_evaluator.evaluate(get_inputs(), shading_point.get_uv(0), offset);
-    evaluate_inputs(input_evaluator.data(), offset);
+    prepare_inputs(reinterpret_cast<uint8*>(input_evaluator.data()) + offset);
```
dictoon: Same as above.

dictoon:
Much nicer!

Couple totally minor things to fix and you can merge it. Thanks!

PR #841, opened 2015-08-08
SSS work

est77:
SSS now works with non physical lights & IBL.


src/appleseed/renderer/kernel/lighting/imagebasedlighting.cpp
```diff
@@ -174,6 +229,98 @@ void compute_ibl_bsdf_sampling(
...
+        environment_edf.evaluate(
+            shading_context,
+            input_evaluator,
+            incoming,
+            env_value,
+            env_prob);
+
+        const double mis_weight =
+            mis_power2(
+                sample_count * bssrdf_prob,
```
dictoon: Are we sure `bssrdf_prob` and `env_prob` are expressed with respect to the same measure?

PR #846, opened 2015-08-27
Unbreak the build when WITH_DISNEY_MATERIAL is OFF (default setting)

danfe:
We should not attempt to build or use `renderer::DisneyMaterial` or any other
Disney-related class if `APPLESEED_WITH_DISNEY_MATERIAL` is not defined; put
missing `#ifdef`'s where they're due.


dictoon:
Merged, thanks!

PR #847, opened 2015-08-27
Improve Clang support

danfe:
Fix the build against modern (3.4.1 and above) Clang versions and report correct compiler in the "About" dialog box.

### DIFF
--- a/src/appleseed.cli/continuoussavingtilecallback.cpp
+++ b/src/appleseed.cli/continuoussavingtilecallback.cpp
@@ -85,7 +85,7 @@ namespace
         }
 
       private:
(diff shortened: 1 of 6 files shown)


est77:
Thanks!

dictoon:
Perfect, thank you!

It might be that some of the `using namespace boost` have become unnecessary and could be removed now... I haven't check thoroughly.

danfe:
Right.  I'm quite pleased with general code quality of appleseed (and its portability across matrix of different operating systems and compilers is a good showcase of it), but those "using namespace foo" had irritated me as well. :-)

PR #848, opened 2015-08-28
SSS work

est77:
- Refactored exponential distribution sampling functions.
- Cleanup SSS tests.
- Ignore CMake 3.0 warnings.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,6 +40,10 @@ if (POLICY CMP0043)
     cmake_policy (SET CMP0043 OLD)  # consume the content of the suffixed COMPILE_DEFINITIONS_<CONFIG>
 endif ()
 
+if (POLICY CMP0026)
+    cmake_policy (SET CMP0026 OLD)  # disallow use of the LOCATION target property (cmake 3.0)
+endif ()
+
 include (src/cmake/utilities.txt)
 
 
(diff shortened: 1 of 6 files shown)


est77:
> In fact, I think mappings.h and sampling.h should be merged into one. They have the same purpose: mapping uniform random numbers in [0,1)^N to various domains.

Ok. Maybe renaming mappings to sampling.h and moving the exponential sampling there?

est77:
> I liked these plots... Why did you remove them?

I'm planning to rewrite the Dirpole tests and plots.

dictoon:
> Ok. Maybe renaming mappings to sampling.h and moving the exponential sampling there?

I would advise to keep `mappings.h` for the simple reason that header files with the same name as a module (`sampling` here) are already used to include all files of the module.

For instance, `foundation/math/knn.h` is a "shortcut file" to include all files under `foundation/math/knn/`. We have these "shortcut files" for modules where it makes no sense to include only some of the files. Examples of such modules are `knn`, `bsp`, `bvh`, etc.

dictoon:
> I'm planning to rewrite the Dirpole tests and plots.

Ok.

est77:
Fixed & rebased.

dictoon:
Thanks!

PR #850, opened 2015-09-04
First part of adding FreeBSD support

danfe:
Do not assume that, except Windows and Apple OS X, there's only Linux. ;-)

### DIFF
--- a/src/appleseed.cli/CMakeLists.txt
+++ b/src/appleseed.cli/CMakeLists.txt
@@ -63,7 +63,7 @@ if (UNIX AND NOT WIN32 AND NOT APPLE)
     set_target_properties (appleseed.cli PROPERTIES
         INSTALL_RPATH "\$ORIGIN/../lib"
     )
-endif (UNIX AND NOT WIN32 AND NOT APPLE)
+endif ()
 
 
 #--------------------------------------------------------------------------------------------------
--- a/src/appleseed.cli/CMakeLists.txt
+++ b/src/appleseed.cli/CMakeLists.txt
@@ -59,11 +59,11 @@ add_executable (appleseed.cli
     ${appleseed.cli_sources}
 )
 
-if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
+if (UNIX AND NOT WIN32 AND NOT APPLE)
     set_target_properties (appleseed.cli PROPERTIES
         INSTALL_RPATH "\$ORIGIN/../lib"
     )
-endif ()
+endif (UNIX AND NOT WIN32 AND NOT APPLE)
 
 
 #--------------------------------------------------------------------------------------------------
--- a/src/appleseed.cli/CMakeLists.txt
+++ b/src/appleseed.cli/CMakeLists.txt
@@ -59,7 +59,7 @@ add_executable (appleseed.cli
     ${appleseed.cli_sources}
 )
 
-if (UNIX AND NOT WIN32 AND NOT APPLE)
+if (USE_RPATH_ORIGIN)
     set_target_properties (appleseed.cli PROPERTIES
         INSTALL_RPATH "\$ORIGIN/../lib"
     )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,6 +97,7 @@ option (WITH_NORMALIZED_DIFFUSION_BSSRDF    "Build Normalized diffusion BSSRDF (
 option (USE_STATIC_BOOST                    "Use static Boost libraries"                            ON)
 option (USE_STATIC_OIIO                     "Use static OpenImageIO libraries"                      ON)
 option (USE_STATIC_OSL                      "Use static OpenShadingLanguage libraries"              ON)
+option (USE_RPATH_ORIGIN                    "Use $ORIGIN in rpath to locate shared libraries"       OFF)
 
 option (USE_SSE                             "Use SSE instruction sets up to version 2"              ON)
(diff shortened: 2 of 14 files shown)


src/appleseed.cli/CMakeLists.txt
```diff
@@ -59,11 +59,11 @@ add_executable (appleseed.cli
     ${appleseed.cli_sources}
 )
 
-if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
+if (UNIX AND NOT WIN32 AND NOT APPLE)
```
dictoon: Aren't `UNIX`, `WIN32` and `APPLE` orthogonal such that you could just check if `UNIX` is defined?

src/appleseed.cli/CMakeLists.txt
```diff
@@ -59,11 +59,11 @@ add_executable (appleseed.cli
...
     set_target_properties (appleseed.cli PROPERTIES
         INSTALL_RPATH "\$ORIGIN/../lib"
     )
-endif ()
+endif (UNIX AND NOT WIN32 AND NOT APPLE)
```
dictoon: We actually omit repeating the clause in `endif ()`, in my experience they add nothing except gratuitous duplication, meaning two places to change when you change the condition.

src/appleseed.cli/CMakeLists.txt
danfe: François, with all due respect, but did you read the commit log? :-)  I've explicitly explained why `if (UNIX)` alone it not enough.

src/appleseed.cli/CMakeLists.txt
danfe: Totally agree, I also find this weird.  OK so now I have to change those lines back to `endif ()` and add another commit to the merge request, correct?

src/appleseed.cli/CMakeLists.txt
dictoon: Of course I did. What I mean is that

    if (UNIX)

is sufficient.

What we had before essentially was

    if (LINUX)

which I agree is now wrong.

src/appleseed.cli/CMakeLists.txt
dictoon: Yes. If you make a new commit and push it (to the same branch), it will appear here in the pull request. It's magical :)

src/appleseed.cli/CMakeLists.txt
danfe: Per what I read from [CMake documentation](http://www.cmake.org/Wiki/CMake_Useful_Variables):
```
UNIX 
    is TRUE on all UNIX-like OS's, including Apple OS X and CygWin
WIN32 
    is TRUE on Windows. Prior to 2.8.4 this included CygWin
```
That means that I can indeed remove `NOT WIN32` check, but the `NOT APPLE` check should still be there because `APPLE` is `UNIX` to CMake, while OS X linker does not support `$ORIGIN`.

dictoon: Oh right, `UNIX` is implied by `APPLE`, had forgotten that...

danfe: OK, so what shall I pick: just `if (UNIX AND NOT APPLE)` or `if (UNIX AND NOT APPLE AND NOT CYGWIN)`?  Provided I know little about Cygwin, it can be wrong yet cluttering the code...

dictoon: Then I suggest leaving it as you originally did.

est77: As we are doing this test multiple times, wouldn't it make sense to do the test once, save the result in a cmake variable and test that instead?

something like (in the top level CMakefile)

if (UNIX AND NOT APPLE AND NOT CYGWIN)
   set (APPLESEED_IS_UNIX True)
endif ()

?

danfe: Roger that, this means I'd only have to fix `endif ()` lines.

danfe: Good point Esteban, François what do you think?

dictoon: I'm fine with `APPLESEED_USE_RPATH`; not so with `APPLESEED_IS_UNIX` as it suggested earlier, since this would introduce conflicting definitions of OS X: a Unix in CMake's definition (and in most everyone's definition), but not a Unix in appleseed's definition...

danfe: Actually it looks like Cygwin [does support](https://cygwin.com/ml/crossgcc/2009-02/msg00019.html) `$ORIGIN` in rpath, so the check can be spelled as `UNIX AND NOT APPLE`.

I'll try to cook up a patch for top level to get `APPLESEED_USE_RPATH` support.

danfe: I suggest to name the new option as `USE_RPATH_ORIGIN` for consistency with `USE_STATIC_BOOST`, `USE_SSE`, etc.  Opinions?

dictoon: Sounds good to me!

dictoon:
Thanks!

danfe:
Awesome collaborative work guys, now I start to understand why people love GitHub and pull requests. ;-)

est77:
I added a small PR to remove the cmake option. It's ok on OSX or Windows for USE_RPATH_ORIGIN to be undefined, as undefined vars in cmake compare to false.

Thanks!

PR #851, opened 2015-09-04
Small change for last commit, remove rpath option.

est77:
We don't want it to be an option, because our packaging script depends on it.
On OSX or Windows, were USE_RPATH_ORIGIN is not defined it is still ok 
as undefined cmake variables compare to false.


danfe:
Not that I have a strong opinion (i.e. I don't mind), originally I just had thought it can be nice to potential porters/package builders to explicitly expose it as an option.

est77:
I don't mind exposing it as an option when we have a need for it.
The problem was that it was appearing as an option but then it was ignored.
It was set to true for unix, false for windows and OSX without considering if 
the user turned it on or off.

We use runtime paths and $ORIGIN on linux when making our release packages 
to avoid having to install libraries in system paths or having to deal with LD_LIBRARY_PATH.

PR #857, opened 2015-09-15
Travis containers update

est77:
Issue #788.

Switch to container based builds. This cuts building time by more than 50%,
from 24 mins to 10 mins approx.

The good points:

- Builds start much faster now.
- We can use 2 cores for building.

The bad points:

- It seems it's not possible to update gcc to 4.8 like we did before. Since our deps are build with gcc 4.8, this could be a problem at some point. So far it seems to work.

Also, removed an unused clang format file.


dictoon:
Cool!

Maybe squash the last three commits into a single one before merging?

est77:
Yes. It's still work in progress, not ready to be merged yet. 
I'm not being able to install gcc 4.8 at the moment.

est77:
Merged, with the idea of rolling back if we have problems and don't find a way to install gcc 4.8.

dictoon:
As far as I can tell, you have closed the PR, not merged it.

PR #860, opened 2015-09-16
Add support for optional studio and cli user settings files.

est77:
Suggested by @danfe on the dev mailing list.

On linux, appleseed.cli and appleseed.studio now try to read and save settings files first to $HOME/.appleseed/settings and if not possible, use the default install settings path.

Useful for system installs, where appleseed_install_dir/settings may not be writeable.

### DIFF
--- a/src/appleseed.cli/main.cpp
+++ b/src/appleseed.cli/main.cpp
@@ -90,14 +90,36 @@ namespace
     void load_settings()
     {
         const filesystem::path root_path(Application::get_root_path());
-        const filesystem::path settings_file_path = root_path / "settings" / "appleseed.cli.xml";
         const filesystem::path schema_file_path = root_path / "schemas" / "settings.xsd";
 
         SettingsFileReader reader(g_logger);
-        reader.read(
-            settings_file_path.string().c_str(),
-            schema_file_path.string().c_str(),
-            g_settings);
+
+        // First try to read the settings from the user path.
+        if (const char* p = Application::get_user_settings_path())
+        {
+            const filesystem::path user_settings_path(p);
+            const filesystem::path user_settings_file_path = user_settings_path / "appleseed.cli.xml";
+
+            if (reader.read(
+                    user_settings_file_path.string().c_str(),
+                    schema_file_path.string().c_str(),
+                    g_settings))
+            {
+                RENDERER_LOG_INFO("successfully loaded settings from %s.", user_settings_file_path.string().c_str());
+                return;
+            }
+        }
+
+        // As a fallback, try to read the settings from the appleseed install.
+        const filesystem::path settings_file_path = root_path / "settings" / "appleseed.cli.xml";
+
+        if (reader.read(
+                settings_file_path.string().c_str(),
+                schema_file_path.string().c_str(),
+                g_settings))
+        {
+            RENDERER_LOG_INFO("successfully loaded settings from %s.", settings_file_path.string().c_str());
+        }
     }
 
     void apply_settings()
--- a/src/appleseed.shared/application/application.cpp
+++ b/src/appleseed.shared/application/application.cpp
@@ -137,6 +137,40 @@ const char* Application::get_root_path()
     return root_path_buffer;
 }
 
+const char* Application::get_user_settings_path()
+{
+    static char user_settings_buffer[FOUNDATION_MAX_PATH_LENGTH + 1];
+    static bool user_settings_initialized = false;
+
+    if (!user_settings_initialized)
+    {
+// Windows.
+#if defined _WIN32
+
+        return 0;
+
+// OS X.
+#elif defined __APPLE__
+
+        return 0;
+
+#elif defined __linux__ || defined __FreeBSD__
+
+        filesystem::path p(get_home_directory());
+        p /= ".appleseed/settings";
+        copy_directory_path_to_buffer(p, user_settings_buffer);
+
+#else
+
+        #error Unsupported platform.
+
+#endif
+        user_settings_initialized = true;
+    }
+
+    return user_settings_buffer;
+}
+
 const char* Application::get_tests_root_path()
 {
     static char tests_root_path_buffer[FOUNDATION_MAX_PATH_LENGTH + 1];
--- a/src/appleseed.studio/mainwindow/mainwindow.cpp
+++ b/src/appleseed.studio/mainwindow/mainwindow.cpp
@@ -1255,22 +1255,45 @@ void MainWindow::slot_project_file_changed(const QString& filepath)
 void MainWindow::slot_load_settings()
 {
     const filesystem::path root_path(Application::get_root_path());
-    const string settings_file_path = (root_path / "settings" / "appleseed.studio.xml").string();
     const string schema_file_path = (root_path / "schemas" / "settings.xsd").string();
 
     SettingsFileReader reader(global_logger());
 
     Dictionary settings;
+    bool success = false;
 
-    const bool success =
-        reader.read(
-            settings_file_path.c_str(),
+    // First try to read the settings from the user path.
+    if (const char* p = Application::get_user_settings_path())
+    {
+        const filesystem::path user_settings_path(p);
+        const filesystem::path user_settings_file_path = user_settings_path / "appleseed.studio.xml";
+
+        success = reader.read(
+            user_settings_file_path.string().c_str(),
             schema_file_path.c_str(),
             settings);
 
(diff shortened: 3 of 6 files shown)


src/appleseed.cli/main.cpp
```diff
@@ -90,14 +90,36 @@ namespace
...
-            schema_file_path.string().c_str(),
-            g_settings);
+
+        // First try to read the settings from the user path.
+        if (const char* p = Application::get_user_settings_path())
+        {
+            const filesystem::path user_settings_path(p);
+            const filesystem::path user_settings_file_path = user_settings_path / "appleseed.cli.xml";
+
+            if (reader.read(
```
dictoon: Add one indentation level, like you did below.

src/appleseed.shared/application/application.cpp
```diff
@@ -137,6 +137,44 @@ const char* Application::get_root_path()
...
+#if defined _WIN32
+
+        return 0;
+
+// OS X.
+#elif defined __APPLE__
+
+        return 0;
+
+#elif defined __linux__
```
dictoon: I think it should be the same on FreeBSD.

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -1292,11 +1315,32 @@ void MainWindow::slot_load_settings()
...
+            filesystem::create_directories(user_settings_path);
+            const filesystem::path user_settings_file_path = user_settings_path / "appleseed.studio.xml";
+
+            if (writer.write(user_settings_file_path.c_str(), m_settings))
+            {
+                RENDERER_LOG_INFO("successfully saved settings to %s.", user_settings_file_path.c_str());
+                return;
+            }
+        }
+        catch(filesystem::filesystem_error&)
```
dictoon: Space before parenthesis.
Make exception object const.

src/appleseed/foundation/platform/path.cpp
```diff
@@ -133,6 +135,49 @@ const char* get_executable_directory()
...
+    {
+#if defined _WIN32
+
+        return 0;
+
++#elif defined __APPLE__
+
+        return 0;
+
+#elif defined __linux__
```
dictoon: Same thing, I'm pretty sure we want to have the same behavior on FreeBSD, as discussed with danfe.

src/appleseed/foundation/platform/path.h
```diff
@@ -83,6 +83,8 @@ APPLESEED_DLLSYMBOL const char* get_executable_path();
 // Return the path to the directory containing the application's executable. NOT thread-safe.
 APPLESEED_DLLSYMBOL const char* get_executable_directory();
 
+// Return the path to the user home directory.
```
dictoon: "the user's home directory"

dictoon:
Cool, thanks!

@danfe Please free to chime in and give your opinion on the PR.

src/appleseed.shared/application/application.cpp
danfe: Correct, both GNU/Linux, FreeBSD (all *BSDs, FWIW), and other Unix-like systems behave the same WRT storing user's settings in `$HOME`.

I cannot be 100% sure about OSX, but likely there as well (OSX often behaves like Unix) — or does the user have write permissions to `foobar.app` bundle installation directory?

src/appleseed/foundation/platform/path.cpp
danfe: That is correct, `$HOME` is standard environment variable and semantically 100% equivalent on *BSD and some other Unix-like systems.

src/appleseed/foundation/platform/path.cpp
```diff
@@ -133,6 +135,49 @@ const char* get_executable_directory()
...
+
+        const char *home_dir = getenv("HOME");
+
+        if (home_dir == 0)
+            home_dir = getpwuid(getuid())->pw_dir;
+
+        assert(home_dir);
+        strcpy(path, home_dir);
```
danfe: I don't think `strcpy()`ing to an array from a pointer (even if array is declared `static`) is safe, would you consider using `strncpy()` instead?  Since `path` is an array, you can use `sizeof(path)` as `size_t len` parameter.

src/appleseed/foundation/platform/path.h
danfe: Agreed; @dictoon I appreciate your attention to these little details. ;-)

src/appleseed/foundation/platform/path.cpp
```diff
@@ -69,11 +71,11 @@ const char* get_executable_path()
...
-	    const DWORD result =
+        const DWORD result =
             GetModuleFileName(
-		        GetModuleHandle(NULL),
-		        path,
-		        sizeof(path));
+                GetModuleHandle(NULL),
+                path,
+                sizeof(path));
         assert(result);
```
danfe: Here (and below): technically `assert()`ing raw value is OK, but 1) it might be a good idea to avoid implicit casting to boolean value (i.e. assert pointers against `NULL` or `0x0` if you agree with Bjarne that macros are evil, and integers/opaque types against plain `0`), and 2) explicit comparison already used in some other places (consistency reason).

src/appleseed/foundation/platform/path.cpp
est77: Lets start with this:

const char *home_dir = getenv("HOME");

if (home_dir == 0)
    home_dir = getpwuid(getuid())->pw_dir;

Is it a correct way to get the home dir on FreeBSD?

dictoon: @est77 All the functions exist on FreeBSD, but of course that doesn't mean this is the correct way to retrieve the home directory. If you're not sure, let the FreeBSD case unimplemented and maybe @danfe can extend your implementation to FreeBSD if it turns out to be the correct way to do it?

danfe: @est77 yes, this code works identically on Linux/glibc and FreeBSD; complete version with required headers:
```c++
#include <sys/types.h>  // not really required (we don't declare any uid_t, etc. variables)
#include <stdio.h>
#include <stdlib.h>
#include <pwd.h>
#include <unistd.h>

int main()
{
    const char *home_dir = getenv("HOME2"); // HOME2 to emulate missing variable

    if (home_dir == 0)
        home_dir = getpwuid(getuid())->pw_dir;

    printf("my sweet home: %s\n", home_dir);
    return 0;
}
```

danfe:
Shall we also remove this part of `CMakeLists.txt`:
```
install (
    CODE "FILE (MAKE_DIRECTORY ${CMAKE_INSTALL_PREFIX}/images)"
    CODE "FILE (MAKE_DIRECTORY ${CMAKE_INSTALL_PREFIX}/images/autosave)"
)
```
It does not look right that build system should worry about creating run-time directories (that might not even be needed).

dictoon:
@danfe That's right. These actually *are* needed if you enable autosaving, but they should really be created in the user's home directory when accessed the first time. I would suggest doing this in a separate PR.

est77:
Comments addressed & rebased.

src/appleseed.shared/application/application.cpp
```diff
@@ -137,6 +137,40 @@ const char* Application::get_root_path()
...
+// Windows.
+#if defined _WIN32
+
+        return 0;
+
+// OS X.
+#elif defined __APPLE__
+
+        return 0;
+
```
danfe: For consistency, you could've added `// Unix.` comment here (or something like "Other Unix-like" or whatever), but that's not really that important.

src/appleseed/foundation/platform/path.cpp
```diff
@@ -133,6 +137,44 @@ const char* get_executable_directory()
...
+
+#elif defined __linux__ || defined __FreeBSD__
+
+        const char *home_dir = getenv("HOME");
+
+        if (home_dir == 0)
+            home_dir = getpwuid(getuid())->pw_dir;
+
+        strncpy(path, home_dir, FOUNDATION_MAX_PATH_LENGTH);
+        path[FOUNDATION_MAX_PATH_LENGTH] = 0;
```
danfe: Since `path` is an array of `char`s, NULL-termination reads better as `= '\0';`.  This is quite common practice in C/C++ code that deals with strings.  It also helps compiler to do a better job on type checking.

dictoon:
We can address the other comments in subsequent commits. Thanks!

PR #861, opened 2015-09-17
Miscellaneous developments

dictoon:


src/appleseed/foundation/platform/path.cpp
```diff
@@ -47,13 +47,9 @@
...
 #if defined __APPLE__
 #include <mach-o/dyld.h>
-#elif defined __linux__
-#include <pwd.h>
+#elif defined __linux__ || defined __FreeBSD__
 #include <sys/types.h>
-#include <unistd.h>
-#elif defined __FreeBSD__
 #include <pwd.h>
-#include <sys/sysctl.h>
```
danfe: Yes, I know it's tempting to move all Linux/FreeBSD includes under one hood, but right now Linux code path does not use `sys/sysctl.h` header: its `get_executable_path()` implementation reads from `/proc/self/exe`.

Technically it does not hurt anything, because `sys/sysctl.h` exists on both systems, but it's not very neat.  (Let me think if I can improve Linux implementation while here: reading from `procfs` involves VFS layer, perhaps there is a more direct way to ask kernel for executable path, like we do for FreeBSD.)

src/appleseed/foundation/platform/path.cpp
```diff
@@ -144,23 +140,26 @@ const char* get_home_directory()
...
 
         return 0;
 
-+#elif defined __APPLE__
+// OS X.
+#elif defined __APPLE__
 
         return 0;
 
+// Other Unixes.
```
danfe: More commonly used plural of *Unix* is *Unices* (`google://plural of unix`).

src/appleseed/foundation/platform/path.cpp
dictoon: Oh right, I missed that. Thanks.

src/appleseed/foundation/platform/path.cpp
dictoon: Will fix, thanks.

src/appleseed/foundation/platform/x86timer.cpp
```diff
@@ -102,13 +107,67 @@ uint64 X86Timer::read()
...
+// gcc.
+#elif defined __GNUC__
+
+    uint32 h, l;
+
+    asm volatile (
+        "rdtscp\n\t"
+        "mov %%edx, %0\n\t"
+        "mov %%eax, %1\n\t"
+        "cpuid\n\t"                         // prevent instructions coming afterward from executing before the RDTSCP instruction
```
danfe: CPUID-related asm() code might need some improvements, I'll comment more of this later after second part of FreeBSD support hits the tree.  As a teaser:
- Register use could be denoted more explicitly: for CPUID, %eax and %ecx are input/output and %ebx and %edx are output only
- Good code should always preserve %ebx/%rbx because ABI defines them as callee-saved (on both x64 and x86)
- Typically separate implementations required for x86-64 and i386, because when preserving %rbx, 32-bit operations would set the upper 32 bits to zero

src/appleseed/foundation/platform/x86timer.cpp
```diff
@@ -102,13 +107,67 @@ uint64 X86Timer::read()
...
+        : "=r" (h), "=r" (l)                // outputs
+        :                                   // inputs
+        : "%rax", "%rbx", "%rcx", "%rdx"    // clobbered registers
```
danfe: For clobbered register list, you do not need to use `%`.  However, in correct CPUID implementation, you wouldn't have to declare clobbered list explicitly in the first place.

src/appleseed/foundation/platform/x86timer.cpp
dictoon: We should address these issues in a separate PR. You seem to be pretty knowledgeable on this, would you care fixing the issues you mentioned?

danfe: Yes, these require deeper investigation and certainly a different PR.  I'll try to review all x86-related `asm()` code as soon as my time permits.

dictoon: Awesome, thanks.

danfe: Just sent an email to the list about X86 time-keeping code; I'd like to get some wider exposure to the patch before opening a pull request.

PR #862, opened 2015-09-17
FreeBSD support (part 2)

danfe:
Add missing methods in the `System` class.

### DIFF
--- a/src/appleseed/foundation/platform/system.cpp
+++ b/src/appleseed/foundation/platform/system.cpp
@@ -74,6 +74,14 @@
     #include <sys/types.h>
     #include <unistd.h>
 
+// FreeBSD.
+#elif defined __FreeBSD__
+
+    #include <sys/types.h>
+    #include <sys/resource.h>
+    #include <sys/sysctl.h>
+    #include <unistd.h>
+
 // Other platforms.
 #else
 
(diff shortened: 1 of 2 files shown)


src/appleseed/foundation/platform/system.cpp
```diff
@@ -421,6 +429,388 @@ uint64 System::get_process_virtual_memory_size()
...
+    uint32_t regs[4];
+    int no_higher_level_cache = 0;
+
+    regs[eax] = 2;
+    cpuid(regs);
+    // Doing only one call is technically wrong, but all CPUs up to Core i7
+    // require a single call.  Since this is a fallback code path for really
+    // old CPUs anyways (modern ones will provide Function 4), we should be
+    // safe, but let's add an assert() on the lower 8 bits just in case.
+    assert ((regs[eax] & 0xFF) == 1);
```
danfe: Oops, superfluous space after function name.  Let's wait for more feedback to see if a follow-up commit is required, or it could be fixed in subsequent "style-fixing" commits.

src/appleseed/foundation/platform/system.cpp
```diff
@@ -421,6 +429,388 @@ uint64 System::get_process_virtual_memory_size()
     return static_cast<uint64>(rss) * sysconf(_SC_PAGESIZE);
 }
 
+// ------------------------------------------------------------------------------------------------
+// FreeBSD.
+// ------------------------------------------------------------------------------------------------
```
dictoon: Very nice code!

Regarding the formatting and coding style, for the sake of consistency there are a few minor things that could be improved:

* Wrap utility types and functions into an anonymous namespace, instead of making them `static`.
* Classes/structures are spelled in CamelCase.
* Macros are spelled in UPPERCASE.
* The pointer `*` goes on the type's side rather than the variable's side.
* We don't repeat the `struct` keyword, which is optional in C++.

src/appleseed/foundation/platform/system.cpp
```diff
@@ -421,6 +429,388 @@ uint64 System::get_process_virtual_memory_size()
...
+    uint32_t regs[4];
+
+    // Cycle up to ten possible caches to be extra sure.
+    for (size_t i = 0; i < 10; i++)
+    {
+        regs[eax] = 4;
+        regs[ecx] = i;
+        cpuid(regs);
+
+        unsigned type = bitfield(regs[eax], 4, 0);
```
dictoon: We usually tag with `const` all variables whose value don't change after initialization.

src/appleseed/foundation/platform/system.cpp
dictoon: `assert()` is a "function", so we don't put a space before the opening parenthesis.

dictoon: Ahah, you caught it before me. Tiny tiny details :)

danfe: Yeah, that's why I appreciate your attention to details: because so do I.  I'll address remaining comments you did below (all very valid) tomorrow, need some sleep now.  Thanks!

src/appleseed/foundation/platform/system.cpp
danfe: Quick question: do I have to reindent all code when I move it inside anonymous namespace?
About class names: so correct name is `TrivialX86Cache`, right?

dictoon: Yes, the contents of a namespace is indented (with 4 spaces). The only exception is for top-level namespaces (`foundation`, `renderer`) since that would indent entire files but provide no practical value.

`TrivialX86Cache`, that's correct!

dictoon:
Great, thank you!

PR #863, opened 2015-09-18
Unbreak the build when WITH_OSL is OFF (default setting)

danfe:
Inclusion of `renderer/kernel/shading/closures.h` should be guarded by `APPLESEED_WITH_OSL` macro because that header file includes OSL headers.

Lest us forget to test the builds against default settings. ;-)


dictoon:
Good catch, thanks!

PR #866, opened 2015-09-23
Optimized LRU cache

dictoon:


Narann:
Any performance boost on high cache usage?

dictoon:
On high cache usage, acquiring available tiles from the shared texture cache is significantly faster, thus improving the scalability as well (since the lock protecting the texture cache is held for a shorter duration).

On low cache usage, the thread-local "front-end" cache allows to bypass locking altogether.

PR #868, opened 2015-09-23
SSS Work

est77:


CMakeLists.txt
```diff
@@ -93,6 +93,7 @@ option (WITH_ALEMBIC                        "Build Alembic support"
 option (WITH_OSL                            "Build OSL support"                                     OFF)
 option (WITH_DISNEY_MATERIAL                "Build Disney material"                                 OFF)
 option (WITH_NORMALIZED_DIFFUSION_BSSRDF    "Build Normalized diffusion BSSRDF (patented)"          OFF)
```
dictoon: "Normalized Diffusion"
"Directional Dipole"

CMakeLists.txt
```diff
@@ -93,6 +93,7 @@ option (WITH_ALEMBIC                        "Build Alembic support"
...
+option (WITH_DIRECTIONAL_DIPOLE_BSSRDF      "Build Directional dipole BSSRDF"                       OFF)
```
dictoon: But truth be told, I wouldn't even bother with a dedicated `#define` unless we actually need to make the release and we haven't fixed the directional dipole code. My hope is that we can fix it before the release so this `#define` would never be required.

src/appleseed/renderer/kernel/lighting/subsurfacesampler.h
```diff
@@ -194,18 +194,6 @@ void SubsurfaceSampler::sample(
         if (incoming_point.get_material() == outgoing_material ||
             incoming_point.get_opposite_material() == outgoing_material)
         {
-#ifdef APPLESEED_WITH_OSL
```
dictoon: Why is it not needed here, but needed in `pathtracer.h`? I'm not sure to know the rationale...

src/appleseed/renderer/kernel/shading/closures.cpp
```diff
@@ -107,7 +107,6 @@ namespace
 
     OSL::ustring standard_dipole_profile_name("standard_dipole");
     OSL::ustring better_dipole_profile_name("better_dipole");
-    OSL::ustring directional_dipole_profile_name("directional_dipole");
```
dictoon: Why not use the `#define` you introduced earlier? (Even though I think we should not introduce this define yet, and similarly we shouldn't remove the directional dipole from OSL yet.)

src/appleseed/renderer/modeling/bssrdf/dipolebssrdf.cpp
```diff
@@ -144,8 +121,14 @@ double DipoleBSSRDF::evaluate_pdf(
...
     // PDF of the sampled radius.
-    const double sigma_tr = 1.0 / values->m_dmfp;
-    const double pdf_radius = exponential_distribution_pdf(radius, sigma_tr);
+    double pdf_radius = 0.0;
+    for (size_t i = 0, e = values->m_sigma_tr.size(); i < e; ++i)
+    {
+        const double sigma_tr = values->m_sigma_tr[i];
+        pdf_radius += exponential_distribution_pdf(radius, sigma_tr);
+    }
+
```
dictoon: Remove this blank line to keep the "PDF of the sampled radius" block together.

src/appleseed/renderer/modeling/bssrdf/dipolebssrdf.h
```diff
@@ -101,10 +101,50 @@ class DipoleBSSRDF
...
+        rd_fun,
+        values->m_reflectance,
+        values->m_dmfp,
+        values->m_anisotropy,
+        values->m_sigma_a,
+        values->m_sigma_s);
+
+    // Precompute effective extinction coefficient.
+    values->m_sigma_tr.resize(values->m_sigma_a.size());
+    values->m_sigma_tr.set(static_cast<float>(1.0f / values->m_dmfp));
```
dictoon: Use `1.0 / dmfp` instead of `1.0f / dmfp` if `dmfp` is a `double`.

src/appleseed/renderer/modeling/bssrdf/dipolebssrdf.h
```diff
@@ -101,10 +101,50 @@ class DipoleBSSRDF
...
+
+  protected:
+    template <typename ComputeRdFun>
+    void do_prepare_inputs(
+        ComputeRdFun                rd_fun,
+        DipoleBSSRDFInputValues*    values) const;
 };
 
 
 //
```
dictoon: Move implementation at the end of the header file.

src/appleseed/renderer/modeling/bssrdf/directionaldipolebssrdf.cpp
```diff
@@ -99,12 +108,70 @@ namespace
...
+            return align(sizeof(DirectionalDipoleBSSRDFInputValues), 16);
+        }
+
+        void prepare_inputs(void* data) const APPLESEED_OVERRIDE
+        {
+            DirectionalDipoleBSSRDFInputValues* values =
+                reinterpret_cast<DirectionalDipoleBSSRDFInputValues*>(data);
+
+            // Presets.
+            const size_t preset = floor(values->m_preset);
```
dictoon: That won't compile on Windows (and it certainly shouldn't on Linux!). Use

    const size_t preset = truncate<size_t>(values->m_preset);

src/appleseed/renderer/modeling/bssrdf/directionaldipolebssrdf.cpp
```diff
@@ -146,22 +213,131 @@ namespace
...
+                reinterpret_cast<const DirectionalDipoleBSSRDFInputValues*>(data);
+
+            // PDF of the sampled radius.
+            double pdf_radius = 0.0;
+            for (size_t i = 0, e = values->m_sigma_tr.size(); i < e; ++i)
+            {
+                const double sigma_tr = values->m_sigma_tr[i];
+                pdf_radius += exponential_distribution_pdf(radius, sigma_tr);
+            }
+
```
dictoon: Remove blank line.

src/appleseed/renderer/modeling/bssrdf/directionaldipolebssrdf.cpp
```diff
@@ -146,22 +213,131 @@ namespace
...
+            // Compute and return the final PDF.
+            return pdf_radius * pdf_angle;
         }
 
       private:
+        void apply_preset(const size_t preset, DirectionalDipoleBSSRDFInputValues* values) const
+        {
+            switch(preset)
+            {
+                case 1: // Apple
```
dictoon: Indent `case` statements with two spaces.
Indent code below `case` statements with two more spaces.
Indent `break` with the rest of the code.

Example:

    switch (preset)
    {
      case 1:
        ...
        break;

src/appleseed/renderer/modeling/bssrdf/directionaldipolebssrdf.cpp
```diff
@@ -146,22 +213,131 @@ namespace
...
+            const double pdf_angle = RcpTwoPi;
+
+            // Compute and return the final PDF.
+            return pdf_radius * pdf_angle;
         }
 
       private:
+        void apply_preset(const size_t preset, DirectionalDipoleBSSRDFInputValues* values) const
+        {
+            switch(preset)
```
dictoon: Add space after `switch`.

dictoon:
Do I understand correctly that the directional dipole model now only accepts presets or manually-entered sigma values? But that the other dipole models continue to accept reflectance and dmfp?

CMakeLists.txt
est77: I agree. If we can fix it, we should remove the option. In the meantime, I think it's ok to have the option to disable it.

src/appleseed/renderer/kernel/lighting/subsurfacesampler.h
est77: My idea is running the shader only on the point that was choosen, instead of on all candidate points found by the subsurface sampler (16 in the worst case now), as running the shader could be expensive.
I'm not sure how much of a difference ignoring OSL bump / normal mapping in the subsurface sampler would make. It's something I'd like to check.

dictoon: Ah, makes sense. Thanks.

PR #893, opened 2015-10-24
Started adding appleseed.python tests

est77:
Rescued from an old branch.


est77:
> assertTrue? Shouldn't it be some kind of equality check rather?

Fixed and rebased.

est77:
> In the other libraries (appleseed.foundation, appleseed.renderer, etc.) unit tests are placed inside a meta/tests/ directory. I suggest doing the same here.

I think that the usual practice when adding unit tests to python modules is test/
The tests are now part of the package and users can run them.
I think it's easier if running the tests is similar to other common python modules.

PR #909, opened 2015-11-01
OSL BSSRDF fixes & test suite update

est77:


Narann:
Looking at the images, some new results look weird on the contact plane.

dictoon:
I kind of agree, actually. Esteban, can you remind me again why you modified the non-OSL test scenes?

est77:
I added UV coordinates to the box and the sphere, for the bump tests.

dictoon:
Ah, right. For some reason it seems to lead to noticeably more noise on the floor planes...

PR #922, opened 2015-11-14
Misc python work

est77:


est77:
> But won't extracting a string from a string always succeed?

This function converts from Python objects to appleseed params.
In the string case, extract<const char*> will only succeed if the object is a Python string.
bool, int and floats can be implicitly converted, so we use the Python API to check the types 
and decide which is the correct conversion.

PR #927, opened 2015-11-22
Some minor improvements

danfe:
1. Guard x86-specific SIMD code with `#ifdef APPLESEED_USE_SSE`.  This helps to get PowerPC port going.
2. Unbreak the build in 32-bit mode with GCC: improvements for x86 timer code as discussed earlier; GCC-only for now.
3. Declare Boost includes are "system": on modern compilers it is no-op, but 1) it fixes the build against GCC 4.2.1 which is still the default on FreeBSD 9.X, and 2) guards us from potential warnings is Boost headers in the future.  Strictly speaking, all 3rd-party headers should be marked as `SYSTEM`.


est77:
Thanks!

PR #928, opened 2015-11-26
Metadata improvements

est77:


src/appleseed/renderer/kernel/texturing/texturestore.cpp
```diff
@@ -88,6 +88,7 @@ Dictionary TextureStore::get_params_metadata()
         Dictionary()
             .insert("type", "int")
             .insert("default", DefaultTextureStoreSizeMB * 1024 * 1024)
+            .insert("label", "Texture cache size")
```
dictoon: "Texture Cache Size"

PR #934, opened 2015-12-20
Added exposure parameter to some appleseed lights and latlong environment

est77:
An alternative way to specify intensity multipliers.
Mostly for Gaffer ImageEngine/gaffer#1590, but could be useful for other plugins.

### DIFF
--- a/src/appleseed/renderer/modeling/environmentedf/latlongmapenvironmentedf.cpp
+++ b/src/appleseed/renderer/modeling/environmentedf/latlongmapenvironmentedf.cpp
@@ -102,11 +102,13 @@ namespace
             TextureCache&   texture_cache,
             const Source*   radiance_source,
             const Source*   multiplier_source,
+            const Source*   exposure_source,
             const size_t    width,
             const size_t    height)
           : m_texture_cache(texture_cache)
           , m_radiance_source(radiance_source)
           , m_multiplier_source(multiplier_source)
+          , m_exposure_source(exposure_source)
           , m_rcp_width(1.0 / width)
           , m_rcp_height(1.0 / height)
         {
@@ -131,7 +133,9 @@ namespace
 
             double multiplier;
             m_multiplier_source->evaluate(m_texture_cache, uv, multiplier);
-            payload.m_color *= static_cast<float>(multiplier);
+            double exposure;
+            m_exposure_source->evaluate(m_texture_cache, uv, exposure);
+            payload.m_color *= static_cast<float>(multiplier * pow(2.0, exposure));
 
             importance = static_cast<double>(luminance(payload.m_color));
         }
@@ -140,6 +144,7 @@ namespace
         TextureCache&   m_texture_cache;
         const Source*   m_radiance_source;
         const Source*   m_multiplier_source;
+        const Source*   m_exposure_source;
         const double    m_rcp_width;
         const double    m_rcp_height;
     };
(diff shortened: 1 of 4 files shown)


src/appleseed/renderer/modeling/environmentedf/latlongmapenvironmentedf.cpp
```diff
@@ -102,11 +102,13 @@ namespace
             TextureCache&   texture_cache,
             const Source*   radiance_source,
             const Source*   multiplier_source,
+            const double    exposure,
             const size_t    width,
             const size_t    height)
           : m_texture_cache(texture_cache)
           , m_radiance_source(radiance_source)
           , m_multiplier_source(multiplier_source)
+          , m_multiplier(pow(2.0, exposure))
```
dictoon: Maybe rename this one to `m_exposure_multiplier`?

src/appleseed/renderer/modeling/environmentedf/latlongmapenvironmentedf.cpp
```diff
@@ -131,7 +133,7 @@ namespace
 
             double multiplier;
             m_multiplier_source->evaluate(m_texture_cache, uv, multiplier);
-            payload.m_color *= static_cast<float>(multiplier);
+            payload.m_color *= static_cast<float>(multiplier * m_multiplier);
```
dictoon: Actually, why not allow the exposure to be textured?

est77:
Addressed comments and rebased.

Narann:
I'm not sure the "exposure" word is the good one. "exposure has a real meaning in photography and can be disturbing as a light parameter. I think Guerilla name this "exponent" (search for "exponent" in [this page](http://guerillarender.com/doc/1.4/User%20Guide_Lighting_Light%20attributes.html)).

dictoon:
Good point @Narann, I kind of agree. Exponent would probably be clearer.

est77:
I prefer exposure. It's a common term in other renderers, image viewers, ... and in my opinion 
easier for users. I find the name exponent confusing. Exponent of what?

Narann:
I've read your commit content more and realize we speak about many things. I will try to be more specific:

"Exposure" is good for:
* Image viewer (Nuke, Rv, RenderView). It's often combined with "Gamma" parameter and "color space" choice. Finality is the visual feedback for screen display (user eyes).
* Camera "exposition" (time or fstop) but it's more on "light simulators" like Maxwell.
* Light's map (and this is the confusing part).

When you use a HDR map as an environment light (or even as an area square light) the word "exposure" actually refere to the map itself: You "expose" the image.

This mean have "exposure" for a directional or a point light without image doesn't make much sense as we don't expose anything and talk more about "intensity".

I try to resume:

Light:
* Intensity
* Exponent

Image:
* Exposure (of the image)

This mean you can totally have:

EnvLight:
* Itensity
* Exponent
* Exposure (of the image)

Directly from the Guerilla doc (because it's well explained):

`Color * Intensity * pow(2, Exponent)`

In the case of an environment light, `Color` is the "HDR image + its exposure". This would mean:

`"Exposured" Image * Intensity * pow(2, Exponent)`

Hope it's clearer. :)

PS: Guerilla choose to have a `Exponent` parameter to avoid to the user typing intensity with many zeroes (like "1000000") which doesn't mean anything when you talk to your mates. It's more user friendly to manipulate/talk with tiny numbers (try to have a conversation like "hey you should put your intensity to one hundred thousand, if it's not enough push to one hundred and fifty thousand. Still doesn't work, use hundred and fifty six thousand, just to try").

est77:
I understand what you mean. I agree, everyone does, that exposure is not a correct term for light emission. But it's a common term that it's easy to understand and users are familiar with.

If you check this https://github.com/ImageEngine/gaffer/pull/1590 (the origin of this PR) the param is called exposure. It has the same name for Arnold lights and I think it's also used in some Renderman area lights and probably more renderers.

I don't like exponent. It is a very generic name that does not make sense unless you know the formula used to compute final light intensities.

dictoon:
I personally prefer exponent over exposure because it's meaning is more clear to me: it's just an exponent (of the form 10^e), while I'm never sure about exposure functions, which could be way more complicated and custom. I'm not asking to change it in this context, either is fine really, but it's interesting to think about it.

PR #939, opened 2015-12-27
Add APPLESEED prefix to macros in foundation/compiler.h

est77:


src/appleseed/foundation/platform/compiler.h
```diff
@@ -96,20 +96,20 @@ namespace foundation
 // Named APPLESEED_ALIGN instead of simply ALIGN to prevent a collision with the ALIGN
```
dictoon: I think it's now safe to remove the last two lines of this comment ("Named APPLESEED_ALIGN...").

dictoon:
Looks good! You can merge when the build passes.

PR #945, opened 2016-01-10
misc work

est77:
- Fix orientation of tangents in ShadingPoint.
- Fix orientation of tangents in OSL.
- Added APPLESEED_LIKELY/UNLIKELY branch prediction macros for gcc.
- Added Basis3d to python bindings.
- Fallback to the original shading basis if the normal from OSL is zero.
- Small microfacet related code refactors.


src/appleseed.python/bindbasis.cpp
```diff
@@ -0,0 +1,68 @@
...
+using namespace std;
+
+namespace
+{
+    template <typename T>
+    void do_bind_basis(const char* class_name)
+    {
+        bpy::class_<Basis3<T> >(class_name)
+            .def(bpy::init<>())
+            .def(bpy::init<Vector<T,3> >())
```
dictoon: `<T, 3>`

src/appleseed/renderer/kernel/shading/shadingpoint.cpp
```diff
@@ -410,6 +410,10 @@ void ShadingPoint::compute_world_space_partial_derivatives() const
             m_dpdu = (dv1 * dp0 - dv0 * dp1) * rcp_det;
             m_dpdv = (du0 * dp1 - du1 * dp0) * rcp_det;
 
+            // Make sure dPdv x dPdu == N
```
dictoon: Why should dPdv x dPdu == N?

est77: The comment is not 100% correct, I'll fix it.
The idea is that dPdv x dPdu and dPdy x dPdx point in the same direction as N.
If not, then we flip dPdu or dPdx.

dictoon:
Couple minor things to fix and then it's good to merge!

PR #963, opened 2016-02-16
Enable more warnings when using gcc.

est77:
- Added APPLESEED_UNUSED to mark unused variables.
- Removed unused variables, unused local typedefs, ...
- Added missing virtual destructor.

### DIFF
--- a/src/appleseed.studio/utility/widgetzoomhandler.cpp
+++ b/src/appleseed.studio/utility/widgetzoomhandler.cpp
@@ -227,9 +227,6 @@ void WidgetZoomHandler::apply_scale_factor()
     if (m_scroll_area->widget()->layout())
         m_scroll_area->widget()->layout()->activate();
 
-    const int new_widget_width = m_scroll_area->widget()->width();
-    const int new_widget_height = m_scroll_area->widget()->height();
-
     const double actual_h_multiplier = static_cast<double>(m_scroll_area->widget()->width()) / old_widget_width;
     const double actual_v_multiplier = static_cast<double>(m_scroll_area->widget()->height()) / old_widget_height;
 
--- a/src/appleseed/foundation/meta/tests/test_dictionary.cpp
+++ b/src/appleseed/foundation/meta/tests/test_dictionary.cpp
@@ -110,7 +110,7 @@ TEST_SUITE(Foundation_Utility_StringDictionary)
 
         EXPECT_EXCEPTION(ExceptionDictionaryItemNotFound,
         {
-            const char* item = sd.get("key");
+            APPLESEED_UNUSED const char* item = sd.get("key");
         });
     }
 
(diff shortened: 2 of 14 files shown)


src/appleseed/renderer/modeling/project/projectfilewriter.cpp
```diff
@@ -110,7 +110,6 @@ namespace renderer
 namespace
 {
     // Floating-point formatting settings.
-    const char* VectorFormat     = "%.15f";
```
dictoon: Surprised about this one. How do we control precision when writing vectors then?

src/appleseed/foundation/platform/compiler.h
```diff
@@ -191,6 +191,17 @@ namespace foundation
 
 
 //
+//  A macro to mark a variable as unused. Useful in unit tests.
```
dictoon: Cool!

src/appleseed.studio/utility/widgetzoomhandler.cpp
```diff
@@ -227,9 +227,6 @@ void WidgetZoomHandler::apply_scale_factor()
     if (m_scroll_area->widget()->layout())
         m_scroll_area->widget()->layout()->activate();
 
-    const int new_widget_width = m_scroll_area->widget()->width();
-    const int new_widget_height = m_scroll_area->widget()->height();
-
     const double actual_h_multiplier = static_cast<double>(m_scroll_area->widget()->width()) / old_widget_width;
```
dictoon: I think the idea here was to use theses variables to compute `actual_h_multiplier`, etc.

src/appleseed/foundation/meta/tests/test_dictionary.cpp
```diff
@@ -110,7 +110,7 @@ TEST_SUITE(Foundation_Utility_StringDictionary)
 
         EXPECT_EXCEPTION(ExceptionDictionaryItemNotFound,
         {
-            const char* item = sd.get("key");
+            APPLESEED_UNUSED const char* item = sd.get("key") ;
```
dictoon: Remove redundant space before semicolon.

src/cmake/config/linux-gcc.txt
```diff
@@ -60,8 +60,13 @@ set (git_command "git")
 
 # Flags common to all configurations.
 set (c_compiler_flags_common
-    -Werror                                             # Treat Warnings As Errors
+    -Wall
```
dictoon: Thanks for that, very useful.

src/appleseed/renderer/modeling/project/projectfilewriter.cpp
est77: Could it be that we are not writting any vector?

dictoon: You're right, we don't. The vectors I had in mind were rotation axes for `<rotation>` elements and scaling factors for `<scaling>` elements, but these are collapsed to matrices when writing project files.

PR #965, opened 2016-02-20
linux -Werror fixes for release builds

est77:


src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -1132,7 +1132,7 @@ void MainWindow::slot_open_cornellbox_builtin_project()
     if (!can_close_project())
         return;
 
-    const bool successful = m_project_manager.load_builtin_project("cornell_box");
+    APPLESEED_UNUSED const bool successful = m_project_manager.load_builtin_project("cornell_box");
```
dictoon: I find this quite surprising. Doesn't the compiler complain in debug? Are we sure this is going to work with other compilers?

dictoon: Wouldn't some kind of `APPLESEED_DEBUG_ONLY` macro be more appropriate?

src/appleseed/renderer/modeling/bsdf/ashikhminbrdf.cpp
```diff
@@ -393,7 +393,7 @@ namespace
         struct SVal
         {
             double      m_kg;               // constant factor of glossy component
-            double      m_k;                // constant factor needed during hemisphere (isotropic case only)
+            double      m_k;                // constant factor needed during hemisphere (anisotropic case only)
```
dictoon: Looks like another word is missing: "during hemisphere sampling"?

src/cmake/config/linux-gcc.txt
```diff
@@ -66,6 +66,7 @@ set (c_compiler_flags_common
     -Wno-sign-compare
     -Wno-unused-function
     -Wno-reorder
+    -Wno-strict-aliasing
```
dictoon: I don't think we want to disable warnings about strict aliasing violations.

src/appleseed.studio/mainwindow/mainwindow.cpp
est77: if you have something like:
const bool ok = ...
assert(ok);

in debug builds the ok variable is used in the assert check.
In release builds, assert expands to nothing and the variable is unused.

dictoon: Yes, I understand. What I mean is that, in debug mode, the variable *is* actually used.

src/cmake/config/linux-gcc.txt
est77: Enabling it produces a number of  "dereferencing type-punned pointer will break strict-aliasing rules" warnings that we'd need to fix in a future PR.
For now, this fixes the release build and we didn't have this warning enabled before in any case.

est77: Enabling it produces a number of  "dereferencing type-punned pointer will break strict-aliasing rules" warnings that we'd need to fix in a future PR.
For now, this fixes the release build and we didn't have this warning enabled before in any case.

dictoon: Interesting; we really shouldn't have those I think. Can you show me one example where we have this problem in our own code?

src/appleseed.studio/mainwindow/mainwindow.cpp
est77: in gcc attribute unused means that the variable could be unused. It only supress the warning.

est77:
In file included from /home/est/Devel/appleseedhq/appleseed/src/appleseed/foundation/math/bvh.h:38:0,
                 from /home/est/Devel/appleseedhq/appleseed/src/appleseed/renderer/kernel/intersection/curvetree.h:42,
                 from /home/est/Devel/appleseedhq/appleseed/src/appleseed/renderer/kernel/intersection/assemblytree.h:34,
                 from /home/est/Devel/appleseedhq/appleseed/src/appleseed/renderer/kernel/intersection/assemblytree.cpp:31:
/home/est/Devel/appleseedhq/appleseed/src/appleseed/foundation/math/bvh/bvh_node.h: In instantiation of ‘const U& foundation::bvh::Node<AABB>::get_user_data() const [with U = renderer::CurveTree::LeafUserData; AABB = foundation::AABB<float, 3ul>]’:
/home/est/Devel/appleseedhq/appleseed/src/appleseed/renderer/kernel/intersection/curvetree.h:277:92:   required from here
/home/est/Devel/appleseedhq/appleseed/src/appleseed/foundation/math/bvh/bvh_node.h:261:51: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
     return *reinterpret_cast<const U*>(m_bbox_data);
                                                   ^
/home/est/Devel/appleseedhq/appleseed/src/appleseed/foundation/math/bvh/bvh_node.h: In instantiation of ‘U& foundation::bvh::Node<AABB>::get_user_data() [with U = renderer::AssemblyTree::Item; AABB = foundation::AABB<double, 3ul>]’:
/home/est/Devel/appleseedhq/appleseed/src/appleseed/renderer/kernel/intersection/assemblytree.cpp:237:61:   required from here
/home/est/Devel/appleseedhq/appleseed/src/appleseed/foundation/math/bvh/bvh_node.h:269:45: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
     return *reinterpret_cast<U*>(m_bbox_data);
                                             ^
/home/est/Devel/appleseedhq/appleseed/src/appleseed/foundation/math/bvh/bvh_node.h: In instantiation of ‘const U& foundation::bvh::Node<AABB>::get_user_data() const [with U = renderer::AssemblyTree::Item; AABB = foundation::AABB<double, 3ul>]’:
/home/est/Devel/appleseedhq/appleseed/src/appleseed/renderer/kernel/intersection/assemblytree.cpp:638:55:   required from here
/home/est/Devel/appleseedhq/appleseed/src/appleseed/foundation/math/bvh/bvh_node.h:261:51: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
     return *reinterpret_cast<const U*>(m_bbox_data);
                                                   ^
cc1plus: all warnings being treated as errors
src/appleseed/CMakeFiles/appleseed.dir/build.make:5078: recipe for target 'src/appleseed/CMakeFiles/appleseed.dir/renderer/kernel/intersection/assemblytree.cpp.o' failed
make[2]: *** [src/appleseed/CMakeFiles/appleseed.dir/renderer/kernel/intersection/assemblytree.cpp.o] Error 1
CMakeFiles/Makefile2:117: recipe for target 'src/appleseed/CMakeFiles/appleseed.dir/all' failed
make[1]: *** [src/appleseed/CMakeFiles/appleseed.dir/all] Error 2
Makefile:149: recipe for target 'all' failed

dictoon:
Interesting, thanks.

The solution is to use `foundation::binary_cast<>` (from `foundation/utility/casts.h`) instead of `reinterpret_cast<>`.

We shouldn't do the change without carefully monitoring ray tracing performance.

PR #975, opened 2016-03-20
Shading work

est77:
- Fixed some crashes when using OSL.
- Added smooth tangents to OSL.
- Better argument names for closures.
- Updated shader lib to use the new closures.
- Start using shader lib shaders in test scenes.
- Absorption bug fixes.


src/appleseed/renderer/kernel/shading/shadingpoint.cpp
```diff
@@ -411,10 +411,6 @@ void ShadingPoint::compute_world_space_partial_derivatives() const
             m_dpdu = (dv1 * dp0 - dv0 * dp1) * rcp_det;
             m_dpdv = (du0 * dp1 - du1 * dp0) * rcp_det;
 
-            // Make sure dPdv x dPdu points in the same direction as N.
```
dictoon: What's the rationale behind the removing of this code? Do these changes affect rendering of non-OSL bump mapping scenes?

dictoon:
Thanks!

PR #723, opened 2015-03-29
OS X 10.7 Build Fixes

johnhaddon:
These are just a couple of small fixes to get the latest Appleseed building on my Mac.

PR #991, opened 2016-04-01
OSX fixes, gcc 4.2

est77:
- Fixed the new BRDFs. (apply the same changes as PR #723 to glossy and metal).

- Temporary workaround for python's EntityVector and EntityMap remove method:
Since remove is not very useful in python as it is right now, we can have a proper fix when 
the code is reviewed / improved.
 
Reported by John.


johnhaddon:
Thanks Est!

PR #992, opened 2016-04-02
Improvements to python entity containers bindings

est77:
- Removed methods exposed in base classes that were always overriden in subclasses.
- remove now returns the entity being removed.
- Added basic unit tests for entity containers.


dictoon:
Excellent!

PR #1000, opened 2016-04-21
Small ShaderQuery improvements.

est77:
PR 1000! :trophy: 
- ShaderQuery::get_metadata now returns a Dictionary.
- Added constructor with one optional search paths for convinience.

### DIFF
--- a/src/appleseed.python/bindshadergroup.cpp
+++ b/src/appleseed.python/bindshadergroup.cpp
@@ -71,6 +71,16 @@ namespace
       : public boost::noncopyable
     {
       public:
+        ShaderQueryWrapper()
+        {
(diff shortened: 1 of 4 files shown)


src/appleseed.python/bindshadergroup.cpp
```diff
@@ -116,8 +126,8 @@ namespace
 
         bpy::dict get_metadata() const
         {
-            const DictionaryArray metadata = m_shader_query->get_metadata();
-            return dictionary_array_to_bpy_dict(metadata, "name");
+            const Dictionary metadata = m_shader_query->get_metadata();
```
dictoon: Is the copy here on purpose?

dictoon:
Congrats! And thanks for the amazing work over the years!

PR #1002, opened 2016-04-23
ShaderQuery fixes.

est77:
- Reset the query when opening a new shader. Workaround for an OSL bug?
- Do not return dictionaries by value, important if ShaderQuery is used across shared library boundaries.
- Do no use default parameters in the public API.

### DIFF
--- a/src/appleseed/renderer/modeling/shadergroup/shaderquery.cpp
+++ b/src/appleseed/renderer/modeling/shadergroup/shaderquery.cpp
@@ -44,8 +44,12 @@ BEGIN_OSL_INCLUDES
 #include "OSL/oslquery.h"
 END_OSL_INCLUDES
 
+// Boost headers.
+#include "boost/optional.hpp"
+
 // Standard headers.
 #include <string>
+#include <vector>
 
 using namespace foundation;
 using namespace std;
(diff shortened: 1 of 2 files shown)


src/appleseed/renderer/modeling/shadergroup/shaderquery.cpp
```diff
@@ -44,8 +44,12 @@ BEGIN_OSL_INCLUDES
 #include "OSL/oslquery.h"
 END_OSL_INCLUDES
 
+// Boost headers,
```
dictoon: Replace comma by dot.

src/appleseed/renderer/modeling/shadergroup/shaderquery.cpp
```diff
@@ -164,13 +172,33 @@ struct ShaderQuery::Impl
 
         return dictionary;
     }
+
+    bool open(const char *shader_name)
```
dictoon: Move star to the type's side.

src/appleseed/renderer/modeling/shadergroup/shaderquery.h
```diff
@@ -99,11 +100,13 @@ class APPLESEED_DLLSYMBOL ShaderQuery
...
-    // If the search path is not passed, shaders can only be queried
-    // using absolute paths.
+    // Create a new shader query without a search path.
+    // Shaders can only be queried using absolute paths.
+    static foundation::auto_release_ptr<ShaderQuery> create();
+
+    // Create a new shader query with a single search path.
     static foundation::auto_release_ptr<ShaderQuery> create(
-        const char* search_path = 0);
+        const char* search_path);
```
dictoon: Write the declaration on a single line?

PR #1013, opened 2016-05-21
Improved UV diagnostic surface shader.

est77:
Map U to R and V to G. B is set to 0.
Fix discontinuity across triangles.
Works correctly for uv coordinates greater than 1.
old: 
uvdiagold
new: 
uvdiagnew

### DIFF
--- a/src/appleseed/renderer/modeling/surfaceshader/diagnosticsurfaceshader.cpp
+++ b/src/appleseed/renderer/modeling/surfaceshader/diagnosticsurfaceshader.cpp
@@ -165,6 +165,14 @@ namespace
         return Color3f(u, v, w);
     }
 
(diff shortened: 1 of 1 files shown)


Narann:
Interesting, would this would have fixed the uv problem with alembic import?

est77:
> Interesting, would this would have fixed the uv problem with alembic import?

No. This only affects the diagnostic mode visualization. There are no changes to UVs.
I didn't know uvs in Alembic were wrong. The Alembic support has to be rewritten 
in any case. At this point, I'm wondering if moving directly to USD would be better.

Narann:
Thanks,

IIRC, USD and abc can be used complementary.

dictoon:
Thanks!

PR #1014, opened 2016-05-24
Small shader lib improvements

est77:
- Updates to subsurface shader. 
- Added Mari UDIMs support to texture shaders.


dictoon:
Nice!

PR #1026, opened 2016-07-14
Revamp asset files handling code

dictoon:


src/appleseed/renderer/modeling/project/assethandler.cpp
```diff
@@ -0,0 +1,256 @@
...
+        scene->collect_asset_paths(paths);
+
+    Frame* frame = m_project.get_frame();
+    if (frame)
+        frame->collect_asset_paths(paths);
+
+    do_collect_asset_paths(paths, m_project.render_layer_rules());
+    do_collect_asset_paths(paths, m_project.configurations());
+
+    vector<string> unique_paths = array_vector<vector<string> >(paths);
```
Narann: space in `<vector<string> >`?

Narann:
Why not use the word "resource" instead of "asset"? From the code I've read here, we try to gather external resources no? And "asset" has a real meaning in production which is not limited to files.

src/appleseed/renderer/modeling/project/assethandler.cpp
dictoon: On purpose.

In C++03 and earlier, `<<` and `>>` are the left-shift and right-shift tokens. If I had written

    array_vector<vector<string>>

the parser would have been confused by the presence of a right-shift token.

This situation is properly handled starting with C++11, to which we're finally about to switch!

dictoon:
Your remark is interesting, because from my experience, the term *resource* is more general than the term *asset*. For instance, memory or disk space are resources but are obviously not assets.

In this code, we call *asset* any external file referenced by a project. Textures, meshes, simulation grids, etc. are all considered assets.

In any case, this term is not exposed to the user, it's only used in the code. As long as we know what we call an asset, we should be fine.

Narann:
> memory or disk space are resources

You're right, in computing, _resource_ are hardware related and it's certainly better like this.

Let's keep _asset_ but let me explain what, from my perspective, an asset is:

An asset is an abstract representation of _something_ related to the project identity: Characters are assets, props are assets, environments (aggregation of props) are assets, etc.

And from each asset come various version type: For asset character Toto there is a modeling version, a lookdev version etc. Versions are often a hierarchy of version (lookdev version contain a desription of shading parameter for each object and a list of texture version for example). Each version is often a single file but it can be a sequence (UDIM, render, etc.) or multiple different files (eg. I remember publish particles in Maya was actually publish a whole folder). This is also why a version or a _publication_ can not be considered as one single file.

Of course there is many variations, each pipeline is different but hope you get the idea.

Just that when I read the code I saw _asset_ and tried to understand what  you where trying to achieved based on my asset interpretation. :D

src/appleseed/renderer/modeling/project/assethandler.cpp
Narann: > the parser would have been confused by the presence of a right-shift token.

I'm very surprised about this. I have the impression I never had this problem. Maybe because `array_vector` and `vector` you use are `typedef` or macro or things like that? Anyway, I lean something.

dictoon: > I'm very surprised about this. I have the impression I never had this problem. Maybe because array_vector and vector you use are typedef or macro or things like that? Anyway, I lean something.

It's unrelated to `array_vector` or `vector` being typedefs. You probably used C++11 or later without knowing it, being the standard for most compilers nowadays.

Example: https://godbolt.org/g/NL7d1B

Narann: You're right, Maybe I'm not used to templates and std enough. :D

src/appleseed/renderer/meta/tests/test_projectfilewriter.cpp
```diff
@@ -60,301 +62,230 @@ TEST_SUITE(Renderer_Modeling_Project_ProjectFileWriter)
...
-            filesystem::remove_all(m_base_output);
-            filesystem::remove_all(m_alternate_output);
+            remove_all(m_output_directory);
 
-            filesystem::create_directory(m_base_output);
-            filesystem::create_directory(m_alternate_output);
+            // On Windows, the create_directory() call below will fail with an Access Denied error
+            // if a File Explorer window was opened in the output directory that we just deleted.
+            // A small pause solves the problem.
+            sleep(50);
```
est77: I have a compile error here: call of overloaded ‘sleep(int)’ is ambiguous.
APPLESEED_DLLSYMBOL void sleep(const uint32 ms)
/usr/include/unistd.h:444:21: note: unsigned int sleep(unsigned int)

dictoon: Right, it probably needs namespace qualification. I'll push a fix right away.

Interestingly, the Travis build has succeeded.

src/appleseed/renderer/meta/tests/test_projectfilewriter.cpp
```diff
@@ -60,301 +62,230 @@ TEST_SUITE(Renderer_Modeling_Project_ProjectFileWriter)
...
-            filesystem::copy_file("unit tests/inputs/test_projectfilewriter_object.obj", fullpath);
+            get_assembly()->objects().insert(
+                auto_release_ptr<Object>(
+                    MeshObjectFactory::create(
+                        object_name,
+                        ParamArray().insert("filename", filename))));
         }
 
-        string get_texture_entity_filepath() const
+        template <>
```
est77: error: explicit specialization in non-namespace scope ‘struct  TestSuiteRenderer_Modeling_Project_ProjectFileWriter::Fixture’

I think it doesn't need to be a specialization. Simple overloading should work fine here.

dictoon: Interesting. I started with an overload but then thought a specialization would be clearer. VS does not complain, obviously. Will fix now.

dictoon:
Thanks for details @Narann! I fully agree with your definition of an asset.

I tried in most places to have the code speaks of *asset files* and *asset paths* instead of just *assets*, but there are still a few places where there could be confusion.

src/appleseed/renderer/meta/tests/test_projectfilewriter.cpp
est77: Travis was reporting the build as ok, but if you check the logs, it failed.

dictoon: That's pretty bad! Any idea on how to fix this?

est77: I'm going to look into it now.

PR #1027, opened 2016-07-16
compute smooth normals mesh operation

est77:
- Added a compute smooth normals mesh operation, similar to compute smooth tangents.
- Exported mesh operations from the appleseed library.
- Added python bindings for mesh ops.

### DIFF
--- a/src/appleseed.python/bindmeshobject.cpp
+++ b/src/appleseed.python/bindmeshobject.cpp
@@ -56,6 +56,16 @@ namespace
         return MeshObjectFactory::create(name.c_str(), bpy_dict_to_param_array(params));
     }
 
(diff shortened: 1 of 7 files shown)


src/appleseed.python/bindmeshobject.cpp
```diff
@@ -56,6 +56,16 @@ namespace
         return MeshObjectFactory::create(name.c_str(), bpy_dict_to_param_array(params));
     }
 
+    const Triangle& get_triangle(const MeshObject *object, const size_t index)
```
dictoon: Move star to the type's side.

dictoon:
Looks good!

est77:
Addressed comments and rebased.
The travis error is not related to this PR.

PR #1055, opened 2016-08-04
Color pipeline optimizations

dictoon:


src/appleseed/foundation/image/regularspectrum.h
```diff
@@ -589,6 +593,19 @@ inline RegularSpectrum<T, N> saturate(const RegularSpectrum<T, N>& s)
 }
 
 template <typename T, size_t N>
+inline void saturate_in_place(RegularSpectrum<T, N>& s)
```
est77: Could saturate_in_place be implemented as clamp_in_place(s, 0, 1)?

dictoon: Absolutely, will do.

PR #1057, opened 2016-08-07
Microfacet code simplifications and improvements.

est77:
- Removed non-virtual interface layer and simplified code.
- Switched to a more accurate masking-shadowing function for Beckmann and GGX.
- Removed side checks for distributions that work correctly on both sides.
- Added faster approximations to erf and erf_inv. Removed boost.

### DIFF
--- a/src/appleseed/foundation/math/microfacet.h
+++ b/src/appleseed/foundation/math/microfacet.h
@@ -35,11 +35,6 @@
 #include "foundation/math/vector.h"
 #include "foundation/platform/compiler.h"
 
-// Boost headers.
-#include "boost/math/special_functions/erf.hpp"
-#include "boost/math/special_functions/sign.hpp"
-#include "boost/mpl/bool.hpp"
-
 // Standard headers.
 #include <algorithm>
 #include <cassert>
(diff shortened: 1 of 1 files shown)


src/appleseed/foundation/math/microfacet.h
```diff
@@ -660,22 +491,85 @@ class BeckmannMDF
...
+        const T a4 = T(-1.453152027);
+        const T a5 = T(1.061405429);
+        const T p = T(0.3275911);
+
+        // A&S formula 7.1.26
+        const T abs_x = std::abs(x);
+        const T t = T(1.0) / (T(1.0) + p * abs_x);
+        const T y = T(1.0) - (((((a5 * t + a4) * t) + a3) * t + a2) * t + a1) * t * std::exp(-square(abs_x * abs_x));
+        const T sign_x = x < T(0.0) ? T(-1.0) : T(1.0);
+        return sign_x * y;
```
dictoon: return x < T(0.0) ? -y : +y;

is probably going to be at least as fast, if not faster, in all cases.

src/appleseed/foundation/math/microfacet.h
```diff
@@ -660,22 +491,85 @@ class BeckmannMDF
...
+    //
+    //    Handbook of Mathematical Functions.
+    //    Abramowitz and Stegun.
+    //    http://people.math.sfu.ca/~cbm/aands/toc.htm
+    //
+    //  Copied from from pbrt-v3.
+
+    inline T erf(const T x) const
+    {
+        // constants
```
dictoon: Full line comments must start with a capital and end with a period.

src/appleseed/foundation/math/microfacet.h
```diff
@@ -331,59 +212,30 @@ class MDF
...
             return T(0.0);
 
         return
-            mdf.do_eval_G1(v, h, alpha_x, alpha_y) *
+            mdf.G1(v, h, alpha_x, alpha_y) *
             std::abs(dot(v, h)) *
-            mdf.do_eval_D(h, alpha_x, alpha_y) /
-            std::abs(cos_theta(v));
+            mdf.D(h, alpha_x, alpha_y) /
+            std::abs(cos_theta_v);
```
dictoon: I would suggest moving the std::abs() statements to the same line as the mdf calls to increase readability:

    return
        mdf.G1(v, h, alpha_x, alpha_y) * std::abs(dot(v, h)) *
        mdf.D(h, alpha_x, alpha_y) / std::abs(cos_theta_v);

dictoon:
Very nice, the simplifications and removal of Boost from this code are welcome.

est77:
Fixed and rebased

PR #1062, opened 2016-08-17
Added support for OSL array parameters.

### LINKED ISSUE #1061: Add support for OSL array parameters.

est77:
Added support for OSL float, vector, point, normal and color array params in projects.
Syntax is "float[] 1.0 2.0 3.0 4.0 ...", "vector[] 1.0 2.0 3.0 1.0 2.0 3.0 ..."
For vectors, colors, points and normals, the number of floats after the type[], must be a multiple of 3.

Closes #1061

### DIFF
--- a/src/appleseed/renderer/meta/tests/test_shaderparamparser.cpp
+++ b/src/appleseed/renderer/meta/tests/test_shaderparamparser.cpp
@@ -72,6 +72,46 @@ TEST_SUITE(Renderer_Modeling_ShaderParamParser)
         EXPECT_EQ("test_string", parser.parse_string_value());
     }
 
+    TEST_CASE(ShaderParamParserFloatArray)
+    {
+        ShaderParamParser parser("float[] 1.0 2.0 3.0 4.0 5.0 6.0 7.0");
+        EXPECT_EQ(OSLParamTypeFloatArray, parser.param_type());
+        std::vector<float> values;
+        parser.parse_float_array(values);
+        EXPECT_EQ(7, values.size());
+    }
+
+    TEST_CASE(ShaderParamParserEmptyFloatArray)
+    {
+        ShaderParamParser parser("float[] ");
+        std::vector<float> values;
+
+        EXPECT_EXCEPTION(ExceptionOSLParamParseError,
+        {
+            parser.parse_float_array(values);
+        });
+    }
+
+    TEST_CASE(ShaderParamParserColorArray)
+    {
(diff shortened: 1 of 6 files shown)


src/appleseed/renderer/meta/tests/test_shaderparamparser.cpp
```diff
@@ -72,6 +72,24 @@ TEST_SUITE(Renderer_Modeling_ShaderParamParser)
         EXPECT_EQ("test_string", parser.parse_string_value());
     }
 
+    TEST_CASE(ShaderParamParserFloatArray)
+    {
+        ShaderParamParser parser("float[] 1.0 2.0 3.0 4.0 5.0 6.0 7.0");
+        EXPECT_EQ(OSLParamTypeFloatArray, parser.param_type());
+        std::vector<float> values;
+        parser.parse_float_array(values);
+        EXPECT_EQ(values.size(), 7);
```
dictoon: Make expected value the first argument, this will lead to better messages if the assertion fails.

src/appleseed/renderer/meta/tests/test_shaderparamparser.cpp
```diff
@@ -72,6 +72,24 @@ TEST_SUITE(Renderer_Modeling_ShaderParamParser)
...
+    }
+
+    TEST_CASE(ShaderParamParserColorArray)
+    {
+        ShaderParamParser parser("color[] 1.0 2.0 3.0 4.0 5.0 6.0 7.0 8.0 9.0");
```
dictoon: What happens if, say, the number of values in the array is not a multiple of 3? I think the edge cases deserve test cases, even more so than the main cases...

src/appleseed/renderer/modeling/shadergroup/shaderparam.h
```diff
@@ -100,51 +105,76 @@ class APPLESEED_DLLSYMBOL ShaderParam
...
-        const float     vz);
+        const char*         name,
+        const float         vx,
+        const float         vy,
+        const float         vz);
+
+    // Create a vector array param.
+    static foundation::auto_release_ptr<ShaderParam> create_vector_array_param(
+        const char*         name,
+        std::vector<float>&         value);
```
dictoon: Fix indentation.

src/appleseed/renderer/meta/tests/test_shaderparamparser.cpp
est77: In that case, an exception will be thrown.

PR #1071, opened 2016-08-24
OS X build fixes

johnhaddon:
I found these tweaks necessary to get Appleseed to build on my (aging) Mac setup. They're not particularly pretty, so if you have suggestions for better ways of achieving the same, I'm all ears.

I wasn't sure how to run the test suite, but I ran `src/appleseed.python/test/runtests.py` manually after installing, and that passed.

### DIFF
--- a/src/appleseed.python/bindentitycontainers.h
+++ b/src/appleseed.python/bindentitycontainers.h
@@ -64,21 +64,30 @@ namespace detail
     }
 
     template <typename T>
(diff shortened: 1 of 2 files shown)


dictoon:
Hi John, nice, thanks!

You can run the full unit test suite from appleseed.cli:

    ./appleseed.cli --run-unit-tests

or simply

    ./appleseed.cli -ut

You can also run them from appleseed.studio (Debug menu).

dictoon:
Looking good, merging! Thanks John.

PR #1074, opened 2016-08-24
Update glass BSDF reference images

dictoon:


Narann:
There is more grain now. :smiley:

dictoon:
Here there is, there's the explanation from the commit:

The previous reference images were wrong: a bug introduced in 672b0d7 ("Add SSE implementations of min/max_value() on spectra") essentially disabled russian roulette, leading to longer paths and less noise. The bug was subsequently fixed in e1069e1 ("Fix min_value() and max_value() for renderer::DynamicSpectrum when size=3").

Narann:
Thanks! :)

PR #1089, opened 2016-09-17
SSS refactors

est77:


src/appleseed/renderer/modeling/bssrdf/normalizeddiffusionbssrdf.cpp
```diff
@@ -133,15 +133,12 @@ namespace
             float cumulated_pdf = 0.0f;
             for (size_t i = 0, e = values->m_channel_pdf.size(); i < e; ++i)
             {
-                const double a = static_cast<double>(values->m_reflectance[i]);
+                const float a = values->m_reflectance[i];
                 const double s = normalized_diffusion_s_dmfp(a);
```
dictoon: Would make sense for `normalized_diffusion_s_dmfp()` to work natively with floats. We might want to templatize them.

est77: Maybe it would be better to move all to floats when we move to float shading?

dictoon: Indeed, you're right.

PR #1099, opened 2016-10-02
Build improvements

est77:
- Added options for missing SIMD instruction sets.
- Renamed APPLESEED_SSE_ALIGN to APPLESEED_SIMD4_ALIGN.
- Added APPLESEED_SIMD8_ALIGN (currently unused).
- Use compiler intrinsics wrapper header instead of including individual SIMD headers.
- Added DynamicSpectrum multiply-add function (for now using SSE).
- Added experimental GNU visibility map for libappleseed.so (disabled by default).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -434,7 +434,7 @@ if (is_x86)
             APPLESEED_USE_SSE
        )
        if (CMAKE_SYSTEM_NAME STREQUAL "Linux" AND CMAKE_SIZEOF_VOID_P MATCHES 4)
-           message (WARNING "Building appleseed with SSE/SSE2 instruction sets on 32-bit Linux is not supported; continuing without SSE/SSE2.")
+           message (WARNING "Building appleseed with SSE/SSE2 instruction sets on 32-bit Linux is not recommended.")
        endif ()
     endif ()
 elseif (TARGET_ARCH MATCHES "ARM")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -121,10 +121,6 @@ option (USE_STATIC_OSL                      "Use static OpenShadingLanguage libr
 option (WARNINGS_AS_ERRORS                  "Treat compiler warnings as errors"                     ON)
 option (HIDE_SYMBOLS                        "When using gcc, hide symbols not on the public API"    ON)
 
-if (CMAKE_SYSTEM_NAME MATCHES "Linux|FreeBSD")
-    option (USE_VISIBILITY_MAP              "When using gcc, hide symbols not on the public API"    OFF)
-endif ()
-
 # SIMD options.
 if (is_x86)
     option (USE_SSE                         "Use SSE instruction sets up to version 2"              ON)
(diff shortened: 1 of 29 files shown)


CMakeLists.txt
```diff
@@ -121,6 +121,10 @@ option (USE_STATIC_OSL                      "Use static OpenShadingLanguage libr
 option (WARNINGS_AS_ERRORS                  "Treat compiler warnings as errors"                     ON)
 option (HIDE_SYMBOLS                        "When using gcc, hide symbols not on the public API"    ON)
 
+if (CMAKE_SYSTEM_NAME MATCHES "Linux|FreeBSD")
+    option (USE_VISIBILITY_MAP              "When using gcc, hide symbols not on the public API"    OFF)
```
dictoon: How does that interact with the `HIDE_SYMBOLS` option that has the exact same description?
Edit: I see that you changed the description in a subsequent commit, but my question remains.

CMakeLists.txt
```diff
@@ -391,29 +390,64 @@ if (UNIX AND NOT APPLE)
...
        set (preprocessor_definitions_common
            ${preprocessor_definitions_common}
-           APPLESEED_USE_SSE
+           APPLESEED_USE_SSE42
        )
-   endif ()
+    endif ()
+    if (USE_SSE)
+       if (CMAKE_SYSTEM_NAME STREQUAL "Linux" AND CMAKE_SIZEOF_VOID_P MATCHES 4)
+           message (WARNING "Building appleseed with SSE/SSE2 instruction sets on 32-bit Linux is not supported; continuing without SSE/SSE2.")
```
dictoon: The problem is that, at this point, APPLESEED_USE_SSE42 has already been defined...

src/cmake/config/linux-gcc.txt
```diff
@@ -76,24 +76,64 @@ if (WARNINGS_AS_ERRORS)
...
+    if (USE_AVX2)
+        set (c_compiler_flags_common
+             ${c_compiler_flags_common}
+            -mavx2                                          # enable AVX2 instruction sets
+        )
+    endif ()
+    if (USE_FMA)
+        set (c_compiler_flags_common
+             ${c_compiler_flags_common}
+            -mfma                                          # enable FMA instruction sets
```
dictoon: Adjust vertical alignment of the comment.

src/cmake/config/mac-clang.txt
```diff
@@ -87,6 +87,25 @@ if (USE_AVX)
...
+if (USE_AVX2)
+    set (c_compiler_flags_common
+         ${c_compiler_flags_common}
+        -mavx2                                          # enable AVX2 instruction sets
+    )
+endif ()
+if (USE_FMA)
+    set (c_compiler_flags_common
+         ${c_compiler_flags_common}
+        -mfma                                          # enable FMA instruction sets
```
dictoon: Adjust vertical alignment of the comment.

CMakeLists.txt
est77: It happens after. It can be used to hide more symbols.
The visibility map included hides OSL and boost symbols that are still on the library for example.
I'm still experimenting with this, that's why by default it is off.

CMakeLists.txt
est77: I'm not sure I understand what is the problem. I didn't do any change here.
I think that if you are on a 32 bits intel system, the only option you have is SSE.
The others only show on 64 bits systems.

PR #1100, opened 2016-10-04
Small fixes and cleanups

dictoon:


src/cmake/config/win-vs.txt
```diff
@@ -154,18 +154,6 @@ if (USE_AVX2)
         /arch:AVX2                          # Advanced Vector Extensions 2
     )
 endif ()
-if (USE_FMA)
```
est77: Maybe we could hide this options for windows?
Or hide them everywhere and decide that AVX2 = AVX2 + FMA + F16C on unix?

dictoon: It doesn't look like F16C is supported by any CPU yet. In any case, AVX2 support certainly doesn't imply F16C support. However I checked and it *does* look like AVX2 support implies FMA support, so there's at least that.

dictoon: Correction, it looks like F16C is supported at least by my CPU (Haswell) :) Not sure to which instruction set it belongs, but it looks like you were right: the first CPU with AVX2 support is Haswell, and it also supports FMA and F16C, so let's go with that.

dictoon: > Maybe we could hide this options for windows?
It's not because the compiler doesn't explicitely support this architecture that we can't write specialized code using instrinsics, so I would keep exposing this option, even on Windows.

PR #1101, opened 2016-10-05
Simplify architecture tuning options

dictoon:


dictoon:
Here is a proposal.

est77:
Yes, I think this is simpler. Thanks!

PR #1141, opened 2016-11-19
EDF OSL closures.

est77:
- Remove special case for OSL emission closures.
- Allow adding different light emitting OSL closures in the future.
- Added special cases for the common case of num closures == 1.
- Small tweaks and improvements.

### DIFF
--- a/src/appleseed/renderer/kernel/shading/closures.cpp
+++ b/src/appleseed/renderer/kernel/shading/closures.cpp
@@ -31,6 +31,7 @@
 
 // appleseed.renderer headers.
 #include "renderer/global/globallogger.h"
+#include "renderer/global/globaltypes.h"
 #include "renderer/modeling/input/inputevaluator.h"
 
 // appleseed.foundation headers.
@@ -365,6 +366,8 @@ namespace
 
     struct EmissionClosure
     {
+        typedef DiffuseEDFInputValues InputValues;
+
         struct Params
         {
         };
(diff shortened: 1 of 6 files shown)


src/appleseed/renderer/kernel/shading/closures.cpp
```diff
@@ -388,6 +391,23 @@ namespace
...
+            const float                 max_weight_component)
+        {
+            InputValues* values =
+                composite_closure.add_closure<InputValues>(
+                    id(),
+                    weight,
+                    max_weight_component);
+
+            const float rcp_max_weight_component = 1.0f / max_weight_component;
+            values->m_radiance = weight * rcp_max_weight_component;
```
dictoon: Looks like you can just divide by `max_weight_component` and get rid of `rcp_max_weight_component`.

src/appleseed/renderer/kernel/shading/closures.cpp
```diff
@@ -1522,29 +1574,42 @@ void CompositeEmissionClosure::process_closure_tree(
...
+            const float max_weight_component = max_value(w);
+
+            if (max_weight_component > 0.0f)
             {
-                // For now, we just recurse.
-                const OSL::ClosureColor* nested = get_nested_closure_color(c->id, c->data());
-                process_closure_tree(nested, weight * Color3f(c->w), total_weight);
+                if (c->id == EmissionID)
+                {
+
```
dictoon: Remove blank line.

src/appleseed/renderer/kernel/shading/closures.cpp
```diff
@@ -1167,6 +1167,13 @@ void CompositeClosure::compute_cdf()
     const size_t closure_count = get_num_closures();
     if (closure_count > 0)
     {
+        if (closure_count == 1)
```
dictoon: What's the most common case? I suggest rearranging the tests to minimize the number of tests in the common case.

dictoon: Regarding the various CDFs built in this file: can we switch to using `foundation::CDF` instead, which handles subtle cases not handled here?

dictoon:
Looks good, thanks! I've made a few remarks.

src/appleseed/renderer/kernel/shading/closures.cpp
est77: No, because foundation::CDF allocates memory.
Maybe we could refactor foundation::CDF to remove the container part out of it?

dictoon: That sounds like a good idea. Just you like for CDF sampling, I would suggest extracting the construction procedure into a free function.

PR #1145, opened 2016-11-30
Initial set of appleseed-maya2 OSL nodes, appleseed library headers

luisbarrancos:


dictoon:
Brilliant man, thanks a lot!

src/appleseed.shaders/include/appleseed/math/as_math_helpers.h
```diff
@@ -0,0 +1,71 @@
...
+// THE SOFTWARE.
+//
+
+#ifndef AS_MATH_HELPERS_H
+#define AS_MATH_HELPERS_H
+
+///////////////////////////////////////////////////////////////////////////////
+
+#ifndef SQR
+#define SQR(X) ( (X) * (X) )
```
dictoon: Wouldn't that be better as a function instead, forcing a single evaluation of the argument?

src/appleseed.shaders/CMakeLists.txt
```diff
@@ -36,7 +37,6 @@ configure_file (
     ${PROJECT_SOURCE_DIR}/src/appleseed.shaders/as_osl_extensions.h
 )
 
-
```
luisbarrancos: Got it, thanks.

src/appleseed.shaders/include/appleseed/math/as_math_helpers.h
```diff
@@ -0,0 +1,71 @@
...
+
+// Perlin bias
+float bias(float value, float b)
+{
+    return pow(value, -log2(b));
+}
+
+float gain(float value, float g)
+{
+    return (value < 0.5)
```
dictoon: A nice way to format expressions with ternary operators is this:

    return value < 0.5
        ? bias(value * 2, 1 - g) * 0.5
        : 1 - bias(2 - (value * 2), 1 - g) * 0.5;

If the condition is long, it goes on the next line:

    return
        very_long_condition() < another_very_long_function_name()
            ? bias(value * 2, 1 - g) * 0.5
            : 1 - bias(2 - (value * 2), 1 - g) * 0.5;

We're using this style rather consistently.

src/appleseed.shaders/include/appleseed/math/as_math_helpers.h
```diff
@@ -0,0 +1,71 @@
...
+
+// Schlick's bias function: http://dept-info.labri.fr/~schlick/DOC/gem2.ps.gz
+float fast_bias(float value, float b)
+{
+    return value / ((1 / b - 2) * (1 - value) + 1);
+}
+
+float fast_gain(float value, float g)
+{
+    return ( value < 0.5 )
```
dictoon: We usually try to avoid redundant parentheses, although this is a really a personal taste and I won't object to them.

However, for the sake of consistency, we normally don't add spaces before/after parentheses, i.e. we prefer

    (value < 0.5)

to

    ( value < 0.5 )

src/appleseed.shaders/include/appleseed/math/as_math_helpers.h
luisbarrancos: @dictoon ok. For a simple single line function, how would you format it? Everything in a single line? i.e:
```
float sqr(float x) { return x * x; }
```

src/appleseed.shaders/include/appleseed/math/as_math_helpers.h
luisbarrancos: @dictoon thanks, i'll update the code and will use this.

src/appleseed.shaders/src/maya/as_maya_vectorProduct.osl
```diff
@@ -0,0 +1,113 @@
...
+    output vector out_output = vector(0)
+    [[
+        string maya_attribute_name = "output",
+        string maya_attribute_type = "vector",
+        string label = "Output",
+        string widget = "null"
+    ]]
+)
+{
+    if ( in_operation == 0 )
```
dictoon: Our formatting convention doesn't put spaces before/after parentheses:

    if (in_operation == 0)

You can fix this in subsequent commits, as you go. No hurry.

luisbarrancos: will do, thanks.

est77:
Amazing!

Thanks!

PR #1149, opened 2016-12-03
OSL Maya utilities header, formatting fixes and cleanups

luisbarrancos:


src/appleseed.shaders/src/maya/as_maya_condition.osl
```diff
@@ -118,7 +118,7 @@ shader as_maya_condition
         string shadername = "";
         getattribute("shader:shadername", shadername);
 
-        warning( "[WARNING]: invalid operation in shader %s, %s:%i\n",
+        warning("[WARNING]: invalid operation in shader %s, %s:%i\n",
```
dictoon: Do we need the [WARNING] prefix here, given that we're already calling the `warning()` function?

src/appleseed.shaders/include/appleseed/maya/as_maya_helpers.h
```diff
@@ -0,0 +1,169 @@
...
+    color in_colorGain,
+    color in_colorOffset,
+    color in_defaultColor,
+    float in_alphaGain,
+    float in_alphaOffset,
+    int in_invert,
+    int in_alphaIsLuminance,
+    output color out_outColor,
+    output float out_outAlpha
+    )
```
dictoon: Not a big deal, but we usually place that closing parenthesis at the end of the previous line:

    void maya_colorBalance(
        color in_colorGain,
        color in_colorOffset,
        color in_defaultColor,
        float in_alphaGain,
        float in_alphaOffset,
        int in_invert,
        int in_alphaIsLuminance,
        output color out_outColor,
        output float out_outAlpha)

luisbarrancos: @dictoon ok, i'll change this and update functions in other headers to follow the same convention.

PR #1151, opened 2016-12-03
Initial set of color spaces, illuminants, RGB<>XYZ data

luisbarrancos:


src/appleseed.shaders/CMakeLists.txt
```diff
@@ -64,6 +64,8 @@ source_group ("include\\gaffer" FILES
 
 set (appleseed_library_osl_headers
     include/appleseed/math/as_math_helpers.h
+    include/appleseed/color/as_color_data.h
```
dictoon: Whenever possible, we try to keep all file lists ordered alphabetically. It makes it easier to find files and to prevent duplicates.

PR #1152, opened 2016-12-03
Add placement2dTexture node

luisbarrancos:


src/appleseed.shaders/CMakeLists.txt
```diff
@@ -163,6 +163,7 @@ set (appleseed_maya2_osl_sources
      src/maya/as_maya_matrixInverse.osl
      src/maya/as_maya_multDoubleLinear.osl
      src/maya/as_maya_multMatrix.osl
+     src/maya/as_maya_place2dTexture.osl
```
dictoon: Order files alphabetically.

luisbarrancos: @dictoon it's fixed in my WIP branch that will have a PR in the next day or two if that's ok with you.

dictoon: Totally fine, thank you!

PR #1156, opened 2016-12-09
maya shaders small tweaks.

est77:
place2dTexture:
---------------------

- made repeat a float[2]
- default uv coords to shading globals uvs.
- remove isconnected (does not seem to work correctly here anyway).


src/appleseed.shaders/src/maya/as_maya_place2dTexture.osl
```diff
@@ -157,12 +157,7 @@ shader as_maya_place2dTexture
     ]]
 )
 {
-    // The user might override input UV coords for the placement node.
```
luisbarrancos: @est77 my only concern was piping outUv from a node to the input UV of the placement node (it's possible), but if a connection is set, it should automagically be picked up with fallback as global primvars u and v, so everything ok.

PR #1169, opened 2016-12-22
Add more Maya nodes, OSL wip2

luisbarrancos:


dictoon:
Lots of great work, thank you!

PR #1173, opened 2016-12-23
Remove EOF extraneous lines, minor tweaks

luisbarrancos:


Narann:
Just FYI, `newline` is what define a line in [POSIX standard](http://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline).

 That's why many text editor put them automatically.

luisbarrancos:
@Narann those are very good points, excellent points in fact.

> ... If a source file that is not empty does not end in a new-line character, or ends in a new-line character immediately preceded by a backslash character before any such splicing takes place, the behavior is undefined.

From C++03, though this changes in C++11
https://stackoverflow.com/questions/72271/no-newline-at-end-of-file-compiler-warning

Issue #1176, opened 2016-12-24
as_maya_areaLight.oso contains garbage that results in syntax errors at loading


dictoon:
On both Esteban's machine and mine, that file contains uninitialized matrix parameters.

Esteban's version of the file (compiled with oslc 1.7.4):

    oparam	matrix	out_matrix	0 0 -1.3590819e-11 4.5856091e-41 -9.7107791e-27 4.5913544e-41 -9.7113831e-27 4.5913544e-41 1.3118252e-37 0 -7.762992e-09 4.5856091e-41 1.4413581e-37 0 -7.4269604e-09 4.5856091e-41		%meta{string,maya_attribute_name,"matrix"} %meta{string,maya_attribute_type,"matrix"} %meta{string,label,"Matrix"} %meta{string,widget,"null"}  %read{2147483647,-1} %write{2147483647,-1}
    oparam	matrix	out_inverseMatrix	0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0		%meta{string,maya_attribute_name,"inverseMatrix"} %meta{string,maya_attribute_type,"matrix"} %meta{string,label,"Inverse Matrix"} %meta{string,widget,"null"}  %read{2147483647,-1} %write{3,3} %initexpr
    oparam	matrix	out_worldMatrix	0 0 -1.2545481e-11 4.5856091e-41 -9.7107791e-27 4.5913544e-41 -9.7113831e-27 4.5913544e-41 1.3118252e-37 0 -7.762992e-09 4.5856091e-41 1.4419088e-37 0 -7.4269604e-09 4.5856091e-41		%meta{string,maya_attribute_name,"worldMatrix"} %meta{string,maya_attribute_type,"matrix"} %meta{string,label,"World Matrix"} %meta{string,widget,"null"}  %read{2147483647,-1} %write{2147483647,-1}

My version of the file (compiled with oslc 1.6.8):

    oparam	matrix	out_matrix	0 0 4.7087743e-38 4.5905136e-41 0 0 7.2719816e+27 3.7835059e-44 7.2719816e+27 3.7835059e-44 0 0 -nan -nan 7.2719816e+27 3.7835059e-44		%meta{string,maya_attribute_name,"matrix"} %meta{string,maya_attribute_type,"matrix"} %meta{string,label,"Matrix"} %meta{string,widget,"null"}  %read{2147483647,-1} %write{2147483647,-1}
[...]

PR #1177, opened 2016-12-25
Yet Another OSL Fix

luisbarrancos:
This should fix ticket #1176.

### DIFF
--- a/src/appleseed.shaders/src/maya/as_maya_areaLight.osl
+++ b/src/appleseed.shaders/src/maya/as_maya_areaLight.osl
@@ -106,7 +106,7 @@ surface as_maya_areaLight
     ]],
 )
 {
-    out_center = point("center", 0, 0, 0);
+    out_center = point("object", 0, 0, 0);
 
     out_matrix = matrix("object", "camera");
     out_inverseMatrix = inverse(out_matrix);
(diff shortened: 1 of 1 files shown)


src/appleseed.shaders/src/maya/as_maya_areaLight.osl
```diff
@@ -106,6 +106,19 @@ surface as_maya_areaLight
     ]],
 )
 {
+    out_center = point("center", 0, 0, 0);
```
dictoon: Shouldn't that be `point("shader", 0, 0, 0)`?

est77:
I'm wondering if we can remove all the output parameters.
We are not using them and we are not connecting them to other shaders.
It's almost sure that they will be optimized away by OSL.

src/appleseed.shaders/src/maya/as_maya_areaLight.osl
luisbarrancos: @dictoon it should actually be *shader* space, but there is no *shader* space anymore, so the best we can do is to use *object* space, and center will be (0,0,0) of that space. Addressed with 2cfe7f799ec1a8c9fc785af6c52bfc1b252bb62e

luisbarrancos:
@est77 if they will be optimized away unless connected, why would you want to remove Maya features? You can use the center position, to connect to a say, distance node to drive a color or other quantity.
There are much more parameters in the node(s), but they are hidden, and most are pretty useless for us.

screenshot_2016-12-26_08-49-34

PR #1181, opened 2016-12-26
Fix rendering with textured area lights

dictoon:


est77:
Can you update the ref image of the osl test scene 07 - emission.appleseed test scene?

PR #1211, opened 2017-01-10
Many fixes and improvements

dictoon:
Best reviewed commit-per-commit...


dictoon:
The PR fails to build because some unit tests fail to call `on_frame_begin()` on BSSRDFs before using them. I will fix that shortly.

PR #1220, opened 2017-01-20
Plastic BRDF fixes and improvements.

est77:
- Handle roughness == 0 as specular.
- Bug fixes.


src/appleseed/renderer/modeling/bsdf/plasticbrdf.cpp
```diff
@@ -410,9 +449,8 @@ namespace
             value.resize(diffuse_reflectance.size());
             for (size_t i = 0, e = diffuse_reflectance.size(); i < e; ++i)
             {
-                const float pd = diffuse_reflectance[i];
-                const float non_linear_term =
-                    1.0f - (internal_scattering * pd) * fdr;
+                float pd = diffuse_reflectance[i];
```
dictoon: Make const.

PR #1224, opened 2017-01-27
More Maya nodes

luisbarrancos:
Besides more nodes, value noise, dealing with upstream UV filters.


src/appleseed.shaders/include/appleseed/fractal/as_noise_helpers.h
```diff
@@ -59,4 +67,115 @@
...
+float noise_derivative(vector control_p, float x)
+{
+    float tmp = control_p[0] - control_p[1];
+
+    return (tmp - control_p[1] + control_p[2]) * x - tmp;
+}
+
+vector noise_lookup(int y, int vx[3])
+{
+    float rng_table[65536] = { RNG_TABLE };
```
dictoon: Does OSL have the `static` keyword? Any downside in including 256 KB of data in a function?

src/appleseed.shaders/src/appleseed/as_plastic.osl
```diff
@@ -0,0 +1,240 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+#include "appleseed/material/as_material_helpers.h"
+#include "appleseed/maya/as_maya_helpers.h"
+
+shader as_plastic
```
dictoon: So, if I understand correctly, this isn't the port of a Maya node. In which case, we now have a plastic BRDF. Should we use it in this shader?

src/appleseed.shaders/src/maya/as_maya_file.osl
```diff
@@ -205,7 +205,7 @@ shader as_maya_file
             out_outColor,
             out_outAlpha);
 
-        exit();
+        return;
```
dictoon: Ahah, I was wondering about this one when debugging the `missingcolor` issue!

src/appleseed.shaders/src/maya/as_maya_fractal.osl
```diff
@@ -0,0 +1,294 @@
...
+        float time_value = in_time, amplitude = in_amplitude, c = 0.0, val;
+        
+        for (int i = 0; i <= recursion_level; ++i)
+        {
+            if (i == recursion_level)
+            {
+                amplitude *= (iterate - (float) recursion_level);
+            }
+
+            if (!amplitude)
```
dictoon: We do allow omitting curly braces for single-line blocks, if you'd prefer that. But we're totally cool with curly braces in that case too. Whatever you prefer, just know that you have the choice :)

src/appleseed.shaders/include/appleseed/fractal/as_noise_helpers.h
luisbarrancos: @dictoon i wished. What i can tell you from the last week(s) is that in terms of (runtime) performance i don't notice anything abnormal, but the actual shader compilation times increase tremendously.

dictoon: Maybe a question worth asking to the OSL devs?

src/appleseed.shaders/src/appleseed/as_plastic.osl
luisbarrancos: @dictoon This was a generic plastic shader. Dielectric Fresnel specular over a subsurface/diffuse term, something to kickstart a wide library of AS shaders, not particularly tied to any DCC app. I noticed there was some work going on with layered BxDF terms and a plastic shader, so i'm not sure to which extent these issues would overlap. Perhaps @est77 can give some extra info on this?

dictoon: @est77 can definitely shed some light here. I would tend to think that the new plastic BRDF is more advanced than anything we could do with OSL alone, so if this shader is not tied to any DCC app with look constraints, then we probably want to take advantage of the plastic BRDF. Let's hear @est77.

luisbarrancos: @dictoon sure. We can use the new closure(s) and simplify this if needed be or just remove it.

src/appleseed.shaders/src/maya/as_maya_psdFileTex.osl
```diff
@@ -0,0 +1,326 @@
...
+                texture(
+                    in_fileTextureName,
+                    st[0], st[1],
+                    dsdt[0], dsdt[1], dsdt[2], dsdt[3],
+                    "blur", in_filterOffset,
+                    "width", in_filter,
+                    "wrap", "periodic",
+                    "subimage", 0,
+                    "missingcolor", in_defaultColor,
+                    "missingalpha", 1,
```
est77: @luisbarrancos Can you change it by "missingalpha", 1.0?
It seems that for keyword arguments, you need to use the exact type for the values.

PR #1226, opened 2017-01-28
Improve entity editor in .studio, add alternative volume parameterization to glass BSDF

dictoon:


est77:
Do we need to set the parameterization here: https://github.com/appleseedhq/appleseed/blob/master/src/appleseed/renderer/modeling/bsdf/oslbsdf.cpp#L365

or it will pick the default?

est77:
I think that closure parameters can be renamed as long as they are not keyword parameters.
They are matched by position and not by name.

dictoon:
> Do we need to set the parameterization here

Good point, yes we need to. I'll add it.

dictoon:
> I think that closure parameters can be renamed as long as they are not keyword parameters.

Ah, right, good point. I will rename them then, and test that everything still works.

PR #1236, opened 2017-02-06
Small code simplifications and cosmetics changes

dictoon:
Cherry-picked a bunch of small commits from my monster changelist, in preparation to more serious changes.

### DIFF
--- a/src/appleseed.cli/houdinitilecallbacks.h
+++ b/src/appleseed.cli/houdinitilecallbacks.h
@@ -50,7 +50,7 @@ class MPlayTileCallbackFactory
   public:
     MPlayTileCallbackFactory(
         const char*         scene_name,
-        bool                progressive_mode,
+        const bool          progressive_mode,
         foundation::Logger& logger);
 
     virtual void release() APPLESEED_OVERRIDE;
@@ -66,8 +66,8 @@ class HRmanPipeTileCallbackFactory
 {
   public:
     HRmanPipeTileCallbackFactory(
-        int                 socket_number,
-        bool                progressive_mode,
+        const int           socket_number,
+        const bool          progressive_mode,
         foundation::Logger& logger);
 
     virtual void release() APPLESEED_OVERRIDE;
--- a/src/appleseed.python/bindshadergroup.cpp
+++ b/src/appleseed.python/bindshadergroup.cpp
@@ -121,9 +121,9 @@ namespace
             return m_shader_query->get_shader_type();
         }
 
-        size_t get_num_params() const
+        size_t get_param_count() const
         {
-            return m_shader_query->get_num_params();
+            return m_shader_query->get_param_count();
         }
 
         bpy::dict get_param_info(const size_t param_index) const
@@ -159,7 +159,7 @@ void bind_shader_group()
         .def("open", &ShaderQueryWrapper::open)
         .def("get_shader_name", &ShaderQueryWrapper::get_shader_name)
         .def("get_shader_type", &ShaderQueryWrapper::get_shader_type)
-        .def("get_num_params", &ShaderQueryWrapper::get_num_params)
+        .def("get_num_params", &ShaderQueryWrapper::get_param_count)
         .def("get_param_info", &ShaderQueryWrapper::get_param_info)
         .def("get_metadata", &ShaderQueryWrapper::get_metadata)
         ;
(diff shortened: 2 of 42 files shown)


src/appleseed.python/bindshadergroup.cpp
```diff
@@ -121,9 +121,9 @@ namespace
             return m_shader_query->get_shader_type();
         }
 
-        size_t get_num_params() const
+        size_t get_param_count() const
         {
-            return m_shader_query->get_num_params();
+            return m_shader_query->get_param_count();
```
Narann: Indirection in .cpp file? Why not move this into the header?

src/appleseed/renderer/modeling/bssrdf/dipolebssrdf.h
```diff
@@ -241,7 +241,11 @@ void DipoleBSSRDF::do_prepare_inputs(
         values->m_precomputed.m_channel_pdf);
 
     // Precompute the (square of the) max radius.
-    values->m_precomputed.m_rmax2 = foundation::square(dipole_max_radius(foundation::min_value(values->m_precomputed.m_sigma_tr)));
+    values->m_precomputed.m_rmax2 =
+        foundation::square(
+            dipole_max_radius(
+                foundation::min_value(
+                    values->m_precomputed.m_sigma_tr)));
```
Narann: Is precomputation a two step process? You seems to use one precomputated value to precompute another one. Why not precompute everything at the same time (it could have some benefit for future optimization).

src/appleseed/renderer/modeling/surfaceshader/diagnosticsurfaceshader.cpp
```diff
@@ -550,12 +550,9 @@ void DiagnosticSurfaceShader::evaluate(
       case FacingRatio:
         {
             const Vector3d& normal = shading_point.get_shading_normal();
```
Narann: Why not use floats here?

src/appleseed.cli/houdinitilecallbacks.h
```diff
@@ -66,8 +66,8 @@ class HRmanPipeTileCallbackFactory
 {
   public:
     HRmanPipeTileCallbackFactory(
-        int                 socket_number,
-        bool                progressive_mode,
+        const int           socket_number,
```
dictoon: I only added the missing `const` qualifier, I didn't want to introduce too many changes. But your question is very relevant.

src/appleseed.python/bindshadergroup.cpp
dictoon: You're right. That being said, I just renamed a method to conform to our coding conventions, and tried to keep the changes minimal.

src/appleseed/renderer/modeling/bssrdf/dipolebssrdf.h
dictoon: Very good point.

src/appleseed/renderer/modeling/surfaceshader/diagnosticsurfaceshader.cpp
dictoon: Geometric computations are done in double precision on purpose.

src/appleseed/renderer/modeling/bssrdf/dipolebssrdf.h
dictoon: Ah, `values->m_precomputed.m_sigma_tr` is actually computed a few lines before, so there's nothing we can optimize here.

src/appleseed.cli/houdinitilecallbacks.h
dictoon: Fixed, will be part of a later PR.

src/appleseed/renderer/modeling/bssrdf/dipolebssrdf.h
Narann: Mmmh, from my perspective, precomputation optimisation came from the possibility to vectorize.

Where this:

```
for(value in values)
{
    values->m_precomputed.m_rmax2 =
foundation::square(
            dipole_max_radius(
                foundation::min_value(
                    values->m_precomputed.m_sigma_tr)));
}
```

Become this:
```
sigmas; // array of m_sigma_tr
rmax2; // array of m_rmax2
// both having the same size (if values.size() is known, you can allocate all those arrays at once)
for(const auto i : sigmas.size())
{
    rmax2[i] =
foundation::square(
            dipole_max_radius(
                foundation::min_value(sigmas[i])));
}
```

Note I have no idea what `values`, `m_precomputed` and `m_sigma_tr` are. I assume simple types here.

I have also no idea if it worth the cost (I have no numbers) but it looks like the precomputation process can be vectorization friendly without all those indirection.

dictoon: You're right, but that would require shading multiple points in parallel. It's how things should be done in the future, but it's a pretty important change of the architecture.

PR #1237, opened 2017-02-07
Fixes

dictoon:


dictoon:
Woops, just noticed that the first commit isn't quite correct. Will make a new PR.

PR #1238, opened 2017-02-07
Fixes

dictoon:


src/appleseed/renderer/modeling/project/projectfileupdater.cpp
```diff
@@ -1238,13 +1238,13 @@ namespace
...
             else if (strcmp(bsdf.get_model(), GlossyBRDFFactory().get_model()) == 0 ||
-                     strcmp(bsdf.get_model(), MetalBRDFFactory().get_model()) == 0 ||
-                     strcmp(bsdf.get_model(), SpecularBTDFFactory().get_model()) == 0)
+                     strcmp(bsdf.get_model(), MetalBRDFFactory().get_model()) == 0)
             {
                 move_if_exist(bsdf, "anisotropy", "anisotropic");
             }
             else if (strcmp(bsdf.get_model(), SpecularBTDFFactory().get_model()) == 0)
             {
+                move_if_exist(bsdf, "anisotropy", "anisotropic");
```
est77: I think this can be removed. We don't have anisotropy in the specular btdf

dictoon: Ah, good point.

PR #1239, opened 2017-02-07
appleseed.studio improvements

dictoon:


src/appleseed/renderer/modeling/material/imaterialfactory.cpp
```diff
@@ -59,4 +59,71 @@ void IMaterialFactory::add_common_input_metadata(DictionaryArray& metadata)
...
+            .insert("min_value", "0.0")
+            .insert("max_value", "1.0")
+            .insert("use", "optional")
+            .insert("default", "1.0")
+            .insert("visible_if",
+                Dictionary()
+                    .insert("displacement_method", "bump")));
+
+    metadata.push_back(
+        Dictionary()
```
Narann: I wonder if there is a way in C++1x to initialize those kind of "dynamic recursive" parameters. All those `insert()` make me nervous thinking to the allocation calls.

Something similar to:

`Dictionary({"name", "normal_map_up"}{"label", "Normal Map Up Vector"} ... )`

Could be nicer.

dictoon: There's possibly better ways, yes. But keep in mind we're still bound to C++03 for the time being.

dictoon: Also, a better notation wouldn't change much, the allocations would still need to be done (items would still need to be inserted one by one into the `std::map`).

PR #1245, opened 2017-02-09
OSL Maya Noise

luisbarrancos:


src/appleseed.shaders/include/appleseed/fractal/as_fractal_helpers.h
```diff
@@ -34,25 +34,33 @@
...
-    point pp = surface_point;
-    float amp = amplitude, fw = filter_width, sum = 0.0;
+    point xyz = surface_point;
+    float amp = amplitude, filter_size = filter_width;
+    float current_time = initial_time, sum = 0.0;
 
     for (int i = 0; i < octaves; ++i)
     {
-        sum += amp * filtered_snoise(pp, fw);
+        if (!amp)
```
dictoon: This test should probably be moved outside the loop.

src/appleseed.shaders/include/appleseed/maya/as_maya_fractal_helpers.h
```diff
@@ -0,0 +1,380 @@
+
+//
+// This source file is part of appleseed.
+// Visit http://appleseedhq.net/ for additional information and resources.
+//
+// This software is released under the MIT license.
+//
+// Copyright (c) 2016-2017 Luis Barrancos, The appleseedhq Organization
```
dictoon: Copyright year should be simply 2017 if you created the file this year.

src/appleseed.shaders/include/appleseed/maya/as_maya_fractal_helpers.h
```diff
@@ -0,0 +1,380 @@
...
+    float lacunarity,
+    float gain)
+{
+    point xyz = surface_point;
+    float amp = amplitude, filter_size = filter_width;
+    float current_time = initial_time, sum = 0.0;
+
+    for (int i = 0; i < octaves; ++i)
+    {
+        if (!amp)
```
dictoon: Same remark, move out of the loop (change everywhere).

src/appleseed.shaders/include/appleseed/maya/as_maya_fractal_helpers.h
```diff
@@ -0,0 +1,380 @@
...
+
+    for (int i = 0; i < waves; ++i)
+    {
+        do
+        {
+            x = random_noise(seed++);
+            y = random_noise(seed++);
+            h = hypot(x, y);
+        }
+        while (h <= 0.0);
```
dictoon: This looks strange: `h` is almost never going to be <= 0.

Rejection sampling (since this is what this loop implements) is usually rejecting samples outside the unit disk in order to implement uniform disk sampling.

Just checking that this was intentional and not a buggy disk sampling algorithm :)

src/appleseed.shaders/include/appleseed/maya/as_maya_fractal_helpers.h
```diff
@@ -0,0 +1,380 @@
...
+            break;
+        }
+
+        sum += amp * maya_cos_waves_2d(xyz, current_time, i, num_waves);
+        amp *= ratio;
+        xyz *= frequency_ratio;
+
+        current_time *= time_ratio;
+    }
+    return (inflection) ? abs(sum) : sum * 0.5 + 0.5;
```
dictoon: Remove redundant parentheses.

src/appleseed.shaders/src/maya/as_maya_noise.osl
```diff
@@ -0,0 +1,353 @@
+
+//
+// This source file is part of appleseed.
+// Visit http://appleseedhq.net/ for additional information and resources.
+//
+// This software is released under the MIT license.
+//
+// Copyright (c) 2016 Luis Barrancos, The appleseedhq Organization
```
dictoon: 2017 if a new file created this year.

src/appleseed.shaders/include/appleseed/fractal/as_fractal_helpers.h
luisbarrancos: @FrancoisBeaune not really, because *amp* is a copy of *amplitude* that is modified within the loop by *ratio* after accumulating the iteration's noise. It's entirely possible to have a ratio of 0, which would result on the fractal noise returning only the first frequency result.

ratio

As far as potential optimization goes we can either test *amp* within the loop. 
Or test for *ratio* outside the loop, if 0, run the sequence/call for the base frequency only, if > 0, proceed to accumulate all the frequencies. I'm not sure we would gain anything with this 2nd approach but i haven't profiled this.
 
As a side note, the global parameter *amplitude* should be tested as well (we're not testing it yet) but it's way before we reach any of these calls. I'm leaving this global amplitude test for later since i want to find common ground with the remaining noise/fractal nodes.

src/appleseed.shaders/include/appleseed/maya/as_maya_fractal_helpers.h
luisbarrancos: @dictoon not a disk sampling algorithm, but it should be removed since h > 0 (debug code for some random_noise() wtf-ery). Will do it later.

src/appleseed.shaders/include/appleseed/fractal/as_fractal_helpers.h
dictoon: Ah ok, I hadn't paid enough attention.

May I suggest `amp == 0.0` though? I think this would be clearer...

luisbarrancos: @dictoon done.

PR #1264, opened 2017-02-25
MDF refactors

est77:
- Removed template params for MDFs.
- Moved most code to cpp files, cleanup microfacet header.
- Refactored erf and erf_inv out of BackmannMDF.
- Added gamma param to MDFs in preparation for the STD MDF.

### DIFF
--- a/src/appleseed/foundation/math/microfacet.cpp
+++ b/src/appleseed/foundation/math/microfacet.cpp
@@ -311,7 +311,7 @@ Vector3f BlinnMDF::sample(
     const float         alpha_y) const
 {
     const float cos_theta = std::pow(1.0f - s[0], 1.0f / (alpha_x + 2.0f));
-    const float sin_theta = std::sqrt(1.0f - cos_theta * cos_theta);
+    const float sin_theta = std::sqrt(std::max(0.0f, 1.0f - square(cos_theta)));
 
     float cos_phi, sin_phi;
     sample_phi(s[1], cos_phi, sin_phi);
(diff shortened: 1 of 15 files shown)


src/appleseed/foundation/math/microfacet.cpp
```diff
@@ -141,7 +142,7 @@ Vector3f sample_visible_normals(
         stretched[1] < 0.99999f ? std::atan2(stretched[2], stretched[0]) : 0.0f;
 
     // Sample slope.
-    Vector2f slope = mdf.sample11(cos_theta, s);
+    Vector2f slope = mdf.sample11(cos_theta, s, gamma);
```
dictoon: Make const.

src/appleseed/foundation/math/microfacet.cpp
```diff
@@ -29,15 +29,775 @@
...
+
+void sample_phi(
+    const float         s,
+    const float         alpha_x,
+    const float         alpha_y,
+    float&              cos_phi,
+    float&              sin_phi)
+{
+    Vector2f sin_cos_phi(
+        std::cos(TwoPi<float>() * s) * alpha_x,
```
dictoon: Factor out `s * TwoPi<float>()`.

est77:
> What's the rationale for de-templatizing foundation::MDF?

We were not really using the double versions of MDFs, except in some unit tests.
I think the code now is more readable. There's a better separation between interface and implementation.
And now that everything is float only, I could replace cmath by OIIO's fast math routines.

PR #1277, opened 2017-03-01
Minor changes

dictoon:


src/appleseed/renderer/modeling/bsdf/disneybrdf.cpp
```diff
@@ -338,12 +338,12 @@ namespace
             m_inputs.declare("base_color", InputFormatSpectralReflectance);
             m_inputs.declare("subsurface", InputFormatFloat, "0.0");
             m_inputs.declare("metallic", InputFormatFloat, "0.0");
-            m_inputs.declare("specular", InputFormatFloat, "0.5");
+            m_inputs.declare("specular", InputFormatFloat, "0.0");
             m_inputs.declare("specular_tint", InputFormatFloat, "0.0");
             m_inputs.declare("anisotropic", InputFormatFloat, "0.0");
             m_inputs.declare("roughness", InputFormatFloat, "0.5");
```
est77: Could we change this one too, now that we are modifying the defaults? I always thought that 0.5 is too much.

dictoon: You mean roughness? We could set it to 0.0 by default.

est77: Yes. Maybe something like 0.1?

dictoon: Ok.

PR #1281, opened 2017-03-02
Travis: use new linux dependency package.

est77:


dictoon:
Does this provide the exact same functionality as before?

est77:
> Does this provide the exact same functionality as before?

Yes. It's just better organized, I think.

PR #1282, opened 2017-03-04
Add envChrome, Anisotropic, layeredTexture

luisbarrancos:
Add missing EDF, BTDF to surface nodes.
Expose Blinn BRDF to OSL.


src/appleseed.shaders/src/maya/as_maya_lambert.osl
```diff
@@ -172,41 +204,57 @@ shader as_maya_lambert
...
+        }
+        else
+        {
+            transparency_term = in_transparency * transparent();
+        }
 
         if (in_matteOpacityMode == 0)
         {
-            matte_opacity = 0.0;
+            ; // 0
```
dictoon: Maybe remove the `;`?

luisbarrancos: ok, i'll remove that condition in the resp. nodes

PR #1283, opened 2017-03-04
Explicitly set C default library locale

glebmish:
QApplication sets application locale to user's locale when created.
It can lead to erros with decimal point when parsing .obj files.
To prevent this C locale must be explicitly set right after QApplication
is built.

### DIFF
--- a/src/appleseed.studio/main/main.cpp
+++ b/src/appleseed.studio/main/main.cpp
@@ -60,6 +60,7 @@
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
+#include <clocale>
 #include <fstream>
 #include <sstream>
 #include <string>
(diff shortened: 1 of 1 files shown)


src/appleseed.studio/main/main.cpp
```diff
@@ -60,6 +60,7 @@
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
+#include <clocale>
```
dictoon: We require that `#include` clauses are always sorted alphabetically. This helps preventing duplicate includes.

src/appleseed.studio/main/main.cpp
```diff
@@ -280,15 +281,15 @@ int main(int argc, char* argv[])
     g_previous_message_handler = qInstallMsgHandler(message_handler);
 
     QApplication application(argc, argv);
+    // QApplication sets C locale to user's locale, we need to fix this
```
dictoon: May I suggest to move this after the setup of `QApplication`.

src/appleseed.studio/main/main.cpp
```diff
@@ -280,15 +281,15 @@ int main(int argc, char* argv[])
...
+    std::setlocale(LC_ALL, "C");
+
     QApplication::setOrganizationName("appleseedhq");
     QApplication::setOrganizationDomain("appleseedhq.net");
     QApplication::setApplicationName("appleseed.studio");
     QApplication::setApplicationVersion(Appleseed::get_lib_version());
     QApplication::setWindowIcon(QIcon(make_app_path("icons/appleseed.png")));
-
-    // The locale must be set after the construction of QApplication.
-    QLocale::setDefault(QLocale::C);
```
dictoon: Oops, you removed this, but we do need it: the Qt locale must be also set. Please keep this and just add your call after it.

Issue #1258, opened 2017-02-22
Update attribute editor when clearing crop window


dictoon:
When the attribute editor is showing the attributes of a frame, clearing the crop window with the toolbar button should also clear the crop window in the attribute editor.

Narann:
Not always. I guess you should have added: "Only when a relevant entity (beauty, render pass, I don't remember the name) is selected".

dictoon:
Updated the description, I think it's pretty accurate now.

aiivashchenko:
I'll take this one to fix, if you don't mind

aiivashchenko:
I've found another related bug, I guess. If the "Crop window" attribute will be updated through "Project" > "Output" > "beauty" > right click > "Edit" dedicated dialog window while "beauty" attribute editor is active it will also take no effect to field into attribute editor.

aiivashchenko:
Hello, @dictoon! I've achieved some results on that task, however I am not quite sure that this solution is the one you've been expecting, since there are a lot of other ways to reach the goal.

For example, we can
1. directly connect signals in MainWindow
2. emit signal in ClearRenderRegionAction
3. connect signal directly to QLineEdit clear() slot on its creation
4. pass the call down through hierarchy without signals
etc. 

All the points are having some pros and cons, so if you'll have some recommendations please let me know and I'll improve the proposed solution. Here is a [PR](https://github.com/appleseedhq/appleseed/pull/1291).

dictoon:
Done.

PR #1291, opened 2017-03-08
Update attribute editor when clearing crop window #1258

aiivashchenko:


dictoon:
Pretty much perfect, thanks!

I only have two minor suggestions regarding naming:

- I would suggest `signal_crop_window_cleared` for the signal emitted by `MainWindow` because, as far as `MainWindow` is concerned, all it knows is that the crop window was cleared; it doesn't necessarily know that in consequence to that signal, a widget will be cleared...

- I would suggest `slot_clear_crop_window_widget` for the slot, because that's the terminology we used: the user interface is made of widgets...

These are really details. The code is simple, clear and expressive. Excellent work!

dictoon:
Note that you can simply push a new commit, it will automatically get added to this PR.

aiivashchenko:
@dictoon, thank you for the review, 
Signal and slot are named now as requested.

dictoon:
Perfect, merging! Thank you

PR #1295, opened 2017-03-09
Reset EntityEditor auto_ptr properly while clearing the layout #1258

aiivashchenko:
Reset EntityEditor auto_ptr properly while clearing the layout
Update all EntityEditor form widgets with dictionary
Remove unnecessary create_form_layout call

### DIFF
--- a/src/appleseed.studio/mainwindow/mainwindow.cpp
+++ b/src/appleseed.studio/mainwindow/mainwindow.cpp
@@ -1475,8 +1475,7 @@ namespace
             Project&        project) APPLESEED_OVERRIDE
         {
             project.get_frame()->reset_crop_window();
-            m_attribute_editor->refresh(
-                project.get_frame()->get_parameters());
+            m_attribute_editor->refresh();
         }
 
       private:
@@ -1487,9 +1486,9 @@ namespace
       : public RenderingManager::IScheduledAction
     {
       public:
-        explicit SetRenderRegionAction(
-            const QRect& rect,
-            const AttributeEditor* attribute_editor)
+        SetRenderRegionAction(
+            const QRect&            rect,
+            const AttributeEditor*  attribute_editor)
           : m_rect(rect),
             m_attribute_editor(attribute_editor)
         {
@@ -1514,8 +1513,7 @@ namespace
                 AABB2i(
                     Vector2i(x0, y0),
                     Vector2i(x1, y1)));
-            m_attribute_editor->refresh(
-                project.get_frame()->get_parameters());
+            m_attribute_editor->refresh();
         }
 
       private:
@@ -1539,17 +1537,21 @@ void MainWindow::slot_clear_render_region()
 
 void MainWindow::slot_set_render_region(const QRect& rect)
 {
-    m_rendering_manager.schedule(
-        auto_ptr<RenderingManager::IScheduledAction>(
-            new SetRenderRegionAction(rect, m_attribute_editor)));
+    auto_ptr<RenderingManager::IScheduledAction> set_render_region_action(
+        new SetRenderRegionAction(rect, m_attribute_editor));
 
-    if (m_settings.get_path_optional<bool>(SETTINGS_RENDER_REGION_TRIGGERS_RENDERING))
+    if (m_settings.get_path_optional<bool>(SETTINGS_RENDER_REGION_TRIGGERS_RENDERING)
+        && !m_rendering_manager.is_rendering())
     {
-        if (m_rendering_manager.is_rendering())
-            m_rendering_manager.reinitialize_rendering();
-        else start_rendering(InteractiveRendering);
+        m_rendering_manager.schedule(set_render_region_action);
+        start_rendering(true);
     }
-    else m_rendering_manager.reinitialize_rendering();
+    else
+    {
+        set_render_region_action.get()->operator()(
+            *m_project_manager.get_project());
+        m_rendering_manager.reinitialize_rendering();
+    }
 }
 
 void MainWindow::slot_render_widget_context_menu(const QPoint& point)
--- a/src/appleseed.studio/mainwindow/mainwindow.cpp
+++ b/src/appleseed.studio/mainwindow/mainwindow.cpp
@@ -1475,7 +1475,7 @@ namespace
             Project&        project) APPLESEED_OVERRIDE
         {
             project.get_frame()->reset_crop_window();
-            m_main_window->emit_signal_crop_window_cleared();
+            m_main_window->emit_refresh_attribute_editor_signal();
         }
 
       private:
@@ -1562,9 +1562,10 @@ void MainWindow::slot_render_widget_context_menu(const QPoint& point)
     menu->exec(point);
 }
 
-void MainWindow::emit_signal_crop_window_cleared() const {
+void MainWindow::emit_refresh_attribute_editor_signal() const
+{
     emit signal_refresh_attribute_editor(
-            m_project_manager.get_project()->get_frame()->get_parameters());
+        m_project_manager.get_project()->get_frame()->get_parameters());
 }
 
 namespace
--- a/src/appleseed.studio/mainwindow/mainwindow.cpp
+++ b/src/appleseed.studio/mainwindow/mainwindow.cpp
@@ -1466,8 +1466,8 @@ namespace
       : public RenderingManager::IScheduledAction
     {
       public:
-        explicit ClearRenderRegionAction(const MainWindow* main_window)
-          : m_main_window(main_window)
+        explicit ClearRenderRegionAction(const AttributeEditor* attribute_editor)
+          : m_attribute_editor(attribute_editor)
         {
         }
 
@@ -1475,19 +1475,23 @@ namespace
             Project&        project) APPLESEED_OVERRIDE
         {
             project.get_frame()->reset_crop_window();
-            m_main_window->emit_refresh_attribute_editor_signal();
+            m_attribute_editor->refresh(
+                project.get_frame()->get_parameters());
         }
 
       private:
-        const MainWindow* m_main_window;
+        const AttributeEditor* m_attribute_editor;
     };
 
     class SetRenderRegionAction
       : public RenderingManager::IScheduledAction
     {
       public:
-        explicit SetRenderRegionAction(const QRect& rect)
-          : m_rect(rect)
+        explicit SetRenderRegionAction(
+            const QRect& rect,
+            const AttributeEditor* attribute_editor)
+          : m_rect(rect),
+            m_attribute_editor(attribute_editor)
         {
         }
 
@@ -1510,17 +1514,20 @@ namespace
                 AABB2i(
                     Vector2i(x0, y0),
                     Vector2i(x1, y1)));
+            m_attribute_editor->refresh(
+                project.get_frame()->get_parameters());
         }
 
       private:
         const QRect m_rect;
+        const AttributeEditor* m_attribute_editor;
     };
 }
 
 void MainWindow::slot_clear_render_region()
 {
     auto_ptr<RenderingManager::IScheduledAction> clear_render_region_action(
-        new ClearRenderRegionAction(this));
+        new ClearRenderRegionAction(m_attribute_editor));
 
     if (m_rendering_manager.is_rendering())
         m_rendering_manager.schedule(clear_render_region_action);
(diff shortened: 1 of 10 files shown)


src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -1522,6 +1533,11 @@ void MainWindow::slot_render_widget_context_menu(const QPoint& point)
     menu->exec(point);
 }
 
+void MainWindow::emit_signal_crop_window_cleared() const {
```
dictoon: I believe this method should be renamed to something like `emit_refresh_attribute_editor_signal()`.

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -1481,13 +1490,15 @@ namespace
...
 void MainWindow::slot_clear_render_region()
 {
-    m_rendering_manager.schedule(
-        auto_ptr<RenderingManager::IScheduledAction>(
-            new ClearRenderRegionAction()));
+    auto_ptr<RenderingManager::IScheduledAction> clear_render_region_action(
+        new ClearRenderRegionAction(this));
 
-    m_rendering_manager.reinitialize_rendering();
+    if (m_rendering_manager.is_rendering())
```
dictoon: Not sure why this piece of logic has changed?

src/appleseed.studio/mainwindow/project/attributeeditor.cpp
```diff
@@ -83,9 +85,12 @@ void AttributeEditor::edit(
         m_entity_editor.get(), SIGNAL(signal_applied(foundation::Dictionary)),
         receiver, slot_apply);
 
-    QObject::connect(
-        m_parent->window(), SIGNAL(signal_crop_window_cleared()),
-        m_entity_editor.get(), SLOT(slot_clear_crop_window_widget()));
+    FrameItem* frame_item = dynamic_cast<FrameItem*>(receiver);
```
dictoon: This is the problematic part of this PR (and of the previous one, a fact that I missed in my first review):

The attribute editor should not have logic for specific types of items. It's a generic facility that should work in a generic manner.

src/appleseed.studio/mainwindow/mainwindow.cpp
aiivashchenko: Actually, we have two possible cases right here. 
First, the button is clicked while rendering is on the go. Then, `ClearRenderRegionAction` should be spawned and put to the `RenderingManager` task queue. This is the one that was implemented. 
However, if we are not rendering, it won't take the effect until the rendering process will be started. Also, if the button will be clicked like thousand times, you'll get thousand tasks in queue at the beginning of rendering process.
So, the second case is to properly handle button click when rendering process is stopped.

src/appleseed.studio/mainwindow/project/attributeeditor.cpp
aiivashchenko: We can start from the top of the stack then to find the proper place for item determination.
So, the `AttributeEditor` change begins at [ProjectExplorer::slot_item_selection_changed()](https://github.com/appleseedhq/appleseed/blob/master/src/appleseed.studio/mainwindow/project/projectexplorer.cpp#L230).
Then, [ItemBase::slot_edit(AttributeEditor* attribute_editor = 0)](https://github.com/appleseedhq/appleseed/blob/master/src/appleseed.studio/mainwindow/project/itembase.h#L90) is called, the [FrameItem::slot_edit(AttributeEditor* attribute_editor)](https://github.com/appleseedhq/appleseed/blob/master/src/appleseed.studio/mainwindow/project/frameitem.cpp#L77) for our case.
[AttributeEditor::edit(...)](https://github.com/appleseedhq/appleseed/blob/master/src/appleseed.studio/mainwindow/project/attributeeditor.cpp#L65) is the last point of journey, since it spawns `EntityEditor`
And here's ~~Johnny~~ dilemma.
We can pass signals all the way down through hierarchy, until we will reach the 'AttributeEditor'. 
Since 'AttributeEditor' is not a `QObject`, it can only connect signals and slots for derivatives of `QObject` available at the scope, but can't have its own.
We either need to make `AttributeEditor` a `QObject` derivative, either add a getter method for `EntityEditor` instance.
As for the signal/slot transmission, we can try to determine Item type at [ProjectExplorer::slot_item_selection_changed()](https://github.com/appleseedhq/appleseed/blob/master/src/appleseed.studio/mainwindow/project/projectexplorer.cpp#L230). 
Other way is to create a virtual slot at `ItemBase`, keep it always connected, and implement all we need at `FrameItem`. Con - all items will know about event.

src/appleseed.studio/mainwindow/project/attributeeditor.cpp
```diff
@@ -31,6 +31,7 @@
 #include "attributeeditor.h"
 
 // appleseed.studio headers.
+#include "frameitem.h"
```
dictoon: You can remove this `#include` now.

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -1437,19 +1437,32 @@ namespace
...
+        const AttributeEditor* m_attribute_editor;
     };
 
     class SetRenderRegionAction
       : public RenderingManager::IScheduledAction
     {
       public:
-        explicit SetRenderRegionAction(const QRect& rect)
-          : m_rect(rect)
+        explicit SetRenderRegionAction(
```
dictoon: The `explicit` keyword is no longer required here.

I wish C++ had made constructors explicit by default...

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -1437,19 +1437,32 @@ namespace
...
+            const QRect& rect,
+            const AttributeEditor* attribute_editor)
```
dictoon: Please vertically align arguments:

    SetRenderRegionAction(
        const QRect&            rect,
        const AttributeEditor*  attribute_editor)

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -1472,29 +1485,34 @@ namespace
...
 void MainWindow::slot_clear_render_region()
 {
-    m_rendering_manager.schedule(
-        auto_ptr<RenderingManager::IScheduledAction>(
-            new ClearRenderRegionAction()));
+    auto_ptr<RenderingManager::IScheduledAction> clear_render_region_action(
+        new ClearRenderRegionAction(m_attribute_editor));
 
-    m_rendering_manager.reinitialize_rendering();
+    if (m_rendering_manager.is_rendering())
```
dictoon: By the way, you can probably use `RenderingManager::schedule_or_execute()` here.

aiivashchenko: Thank you, I'll try that.

aiivashchenko: @dictoon, I've tried it and it produces a segfault even if I'm removing the refresh code.

aiivashchenko:
Here is an explanation why it is done with a new interface.
Mostly all project-related entities are derived from `EntityItem` class, except the frame. Frame is treated in an absolutely different way right now. 
The join points for both, `FrameItem` and `EntityItem` derivables, are `ItemBase` and `EntityCreatorBase` classes. 
We can't put a fully virtual method to `ItemBase`, since it should have a cabability to be constructed on its own, and `EntityCreatorBase` drives slightly another functions.
So, as you wanted to keep it simple, with dummy logic for `AttributeEditor` and without signals/slots, I've decided to make a new interface for values retrieval as a compromise between desired solution and desired functionality.

src/appleseed.studio/mainwindow/mainwindow.cpp
aiivashchenko: @dictoon, found a reason for this. `RenderingManager` just don't know anything about project if rendering process hasn't been launched at least once before `RenderingManager::schedule_or_execute()` call. As a result, a nullptr is passed to the action. Thus, if we want to use that function, we need to pass `Project` instance to rendering manager once project is loaded.
I'll try to fix it tomorrow.

dictoon: Hey, thanks for the update. That's an interesting problem in fact. I wonder if there are other places where this crash could already happen?

I would suggest leaving the code as you wrote it for now (assuming that it works) and addressing the problem you just reported in a subsequent PR.

dictoon:
About `IEntityValueProvider` : couldn't we get rid of it if we modified `AttributeEditor::edit()` to take a `renderer::Entity*` rather than just its values?

src/appleseed.studio/mainwindow/mainwindow.cpp
aiivashchenko: @dictoon, as far as I understood action itself triggers the `Frame` instance update. If we will just schedule it, frame won't be updated if we are not rendering. Particulary in this case crop value will remain the same until rendering is started.

dictoon: You're right.

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -1472,37 +1484,45 @@ namespace
...
-        auto_ptr<RenderingManager::IScheduledAction>(
-            new SetRenderRegionAction(rect)));
+    auto_ptr<RenderingManager::IScheduledAction> set_render_region_action(
+        new SetRenderRegionAction(rect, m_attribute_editor));
 
-    if (m_settings.get_path_optional<bool>(SETTINGS_RENDER_REGION_TRIGGERS_RENDERING))
+    if (m_settings.get_path_optional<bool>(SETTINGS_RENDER_REGION_TRIGGERS_RENDERING)
+        && !m_rendering_manager.is_rendering())
+    {
+        m_rendering_manager.schedule(set_render_region_action);
```
aiivashchenko: Scheduling an action and starting a rendering process straight after was a part of original logic. It sounds legit, since renderer starts the batch job from iteration over scheduled actions. 
But I can change it, no worries.
And you are right, there should be 3 cases:
1. Execute an action if we are not rendering and we are not going to by the settings (the one missed in original code)
2. Schedule an action if we are rendering (the one missed by me)
3. Schedule or force an action and start rendering if corresponding value is set in options

aiivashchenko:
@dictoon, I've missed that way, sorry, my bad. I'll need just to add an extra argument to `AttributeEditor::edit()` than and, probably, a getter method to `EntityItemBase`.

aiivashchenko:
@dictoon, checked out that approach, and here is a problem. [This](https://github.com/appleseedhq/appleseed/blob/master/src/appleseed.studio/mainwindow/project/projectbuilder.h#L248) method is used for the entity update after changes has been made through some widget. In fact, it just deletes the old `Entity` instance and creates a new one, with new values. So, the `Entity` pointer we are passing to `AttributeEditor::edit()` became useless straight after the first edit has been applied at all cases, except `Frame` one, since it is treated in a different manner. Thus, it looks like that we need to go all the way down to [EntityVector](https://github.com/appleseedhq/appleseed/blob/master/src/appleseed/renderer/modeling/entity/entityvector.cpp#L275) and handle entity update here somehow, or pass a pointer to renderer::Entity pointer to `AttributeEditor::edit()`

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -1472,37 +1484,47 @@ namespace
...
+
+    if (!m_rendering_manager.is_rendering())
+    {
+        set_render_region_action.get()->operator()(
+            *m_project_manager.get_project());
 
-    if (m_settings.get_path_optional<bool>(SETTINGS_RENDER_REGION_TRIGGERS_RENDERING))
+        if (m_settings.get_path_optional<bool>(SETTINGS_RENDER_REGION_TRIGGERS_RENDERING))
+            start_rendering(true);
+        else m_rendering_manager.reinitialize_rendering();
```
dictoon: Not sure we need to call `reinitialize_rendering()` since we know we're not currently rendering.

src/appleseed.studio/mainwindow/project/entityvalueprovider.h
```diff
@@ -0,0 +1,47 @@
+
+//
+// This source file is part of appleseed.
+// Visit http://appleseedhq.net/ for additional information and resources.
+//
+// This software is released under the MIT license.
+//
+// Copyright (c) 2010-2013 Francois Beaune, Jupiter Jazz Limited
+// Copyright (c) 2014-2017 Francois Beaune, The appleseedhq Organization
```
dictoon: This is a new file, so you should replace these two copyright lines by a single one:

    // Copyright (c) 2017 Andrei Ivashchenko, The appleseedhq Organization

dictoon:
Merged, thank you!

PR #1298, opened 2017-03-10
Addition of ToolTip feature to rendering settings

nasehim7:
1. Resolved the styling issues by placing the tooltip just after the creation of qt components so that tooltip creation gets standardized.
2. Addition of a config_param_path to create_bounce_settings() and create_bounce_settings_group() for easy retrieval of setting description.


dictoon:
Looking pretty good, thank you!

PR #1300, opened 2017-03-10
Open packed project 

glebmish:


src/appleseed/CMakeLists.txt
```diff
@@ -701,6 +701,20 @@ source_group ("foundation\\utility\\test" FILES
...
+    foundation/utility/minizip/unzip.c
+    foundation/utility/minizip/unzip.h
+)
+list (APPEND appleseed_sources
+    ${foundation_utility_minizip_sources}
+)
+source_group ("foundation\\utility\\minizip" FILES
+    ${foundation_utility_minizip_sources}
+)
+
```
dictoon: Remove one redundant blank line.

src/appleseed/foundation/utility/unzipper.cpp
```diff
@@ -0,0 +1,217 @@
+
+//
+// This source file is part of appleseed.
+// Visit http://appleseedhq.net/ for additional information and resources.
+//
+// This software is released under the MIT license.
+//
+// Copyright (c) 2010-2013 Francois Beaune, Jupiter Jazz Limited
```
dictoon: Replace these two lines of copyright notice by a single line:

    // Copyright (c) 2017 Gleb Mishchenko, The appleseedhq Organization

Use this copyright notice for all new files you create.

src/appleseed/foundation/utility/unzipper.cpp
```diff
@@ -0,0 +1,217 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
```
dictoon: Remove one blank line.

src/appleseed/foundation/utility/unzipper.cpp
```diff
@@ -0,0 +1,217 @@
...
+#include <string>
+#include <sstream>
+
+using namespace std;
+namespace bf = boost::filesystem;
+
+namespace foundation
+{
+
+const int BUFFER_SIZE = 4096;
```
dictoon: Size types are usually better expressed with `size_t` instead of `int`.

src/appleseed/foundation/utility/unzipper.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+const char* UnzipException::what() const throw() 
+{
+    stringstream exc_message;
+    exc_message << message;
+
+    if (err != 0)
+        exc_message << err;
+
+    return exc_message.str().c_str();
```
dictoon: That's not going to work. You're returning (via `c_str()`) a pointer to the temporary string (returned by `str()`).

src/appleseed/foundation/utility/unzipper.cpp
```diff
@@ -0,0 +1,217 @@
...
+}
+
+bool exists(const string& dirname) {
```
dictoon: I don't think these helper methods are really useful. I would recommend using Boost.Filesystem directly as this will expose the actual Boost methods you're using, and hence will make the exact semantics of your code clearer to readers.

src/appleseed/foundation/utility/unzipper.cpp
```diff
@@ -0,0 +1,217 @@
...
+{
+    bf::path dir(dirname);
+    return bf::remove_all(dir);
+}
+
+bool is_directory(const string& dirname) 
+{
+    // used own implementation of is_directory instead of boost implementation
+    // because this directory is not in filesystem, but in zipfile
+    return dirname[dirname.size() - 1] == '/';
```
dictoon: The comment is clear, but you should somehow reflect that in the name of the function then. `is_zip_entry_directory()`?

src/appleseed/foundation/utility/unzipper.cpp
```diff
@@ -0,0 +1,217 @@
...
+}
+
+void open_current_file(unzFile& zip_file) 
+{
+    const int err = unzOpenCurrentFile(zip_file);
+    if (err != UNZ_OK) 
+        throw UnzipException("Can't open file inside zip: ", err);
+}
+
+int read_chunk(unzFile& zip_file, char* buffer, const int chunk_size) {
```
dictoon: Place braces on their own lines (everywhere).

src/appleseed/foundation/utility/unzipper.cpp
dictoon: The right way to do this is to assemble the exception message when constructing the exception. That way you also don't need to overload `Exception::what()`.

src/appleseed/foundation/utility/unzipper.cpp
```diff
@@ -0,0 +1,217 @@
...
+    const int err = unzReadCurrentFile(zip_file, buffer, chunk_size);
+
+    if (err == UNZ_ERRNO)
+        throw UnzipException("IO error while reading from zip");
+    if (err < 0)
+        throw UnzipException("zLib error with decompessing file: ", err);
```
dictoon: *while*
*decompressing*

src/appleseed/foundation/utility/unzipper.cpp
```diff
@@ -0,0 +1,217 @@
...
+void close_current_file(unzFile& zip_file) {
+    const int err = unzCloseCurrentFile(zip_file);
+
+    if (err == UNZ_CRCERROR)
+        throw UnzipException("CRC32 is not good");
+}
+
+string read_filename(unzFile& zip_file) 
+{
+    char filename_inzip[256];
```
dictoon: How do you know that 256 is enough? Is it enough in all cases?

src/appleseed/foundation/utility/unzipper.cpp
```diff
@@ -0,0 +1,217 @@
...
+    if (is_directory(filepath)) 
+    {
+        make_dir(filepath);
+        return;
+    } else 
+        open_current_file(zip_file);
+
+    fstream out(filepath.c_str(), ios_base::out | ios_base::binary);
+
+    char* buffer = new char[BUFFER_SIZE];
```
dictoon: This code will leak in case of exceptions (there might not be exceptions, but it's a good practice to make code exception-safe).

If `BUFFER_SIZE` is small enough, allocate `buffer` on the stack:

    char buffer[BUFFER_SIZE];

If it's too large to be allocated on the stack, use an `std::vector`.

src/appleseed/foundation/utility/unzipper.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+        has_next = unzGoToNextFile(zip_file);
+    }
+
+    unzClose(zip_file);
+
+    return filenames;
+}
+
+} // namespace foundation
```
dictoon: Add an empty line at the end of the file.

src/appleseed/foundation/utility/unzipper.h
```diff
@@ -0,0 +1,86 @@
+
+
```
dictoon: Remove one blank line. All source files should begin and end with a blank line.

src/appleseed/foundation/utility/unzipper.h
```diff
@@ -0,0 +1,86 @@
...
+//
+// This source file is part of appleseed.
+// Visit http://appleseedhq.net/ for additional information and resources.
+//
+// This software is released under the MIT license.
+//
+// Copyright (c) 2010-2013 Francois Beaune, Jupiter Jazz Limited
```
dictoon: Update copyright notice.

src/appleseed/foundation/utility/unzipper.h
```diff
@@ -0,0 +1,86 @@
...
+class UnzipException
+  : public foundation::Exception
+{
+  public:
+    UnzipException(const char* message, const int err): message(message), err(err) {}
+    UnzipException(const char* message): message(message), err(0) {}
+
+    const char* what() const throw();
+
+    ~UnzipException() throw() {};
```
dictoon: If you don't need the destructor, just leave it out.

src/appleseed/foundation/utility/unzipper.h
```diff
@@ -0,0 +1,86 @@
...
+// throws UnzipException if zip_filename can't be opened or is not a zip file 
+//
+
+std::vector<std::string> get_filenames_with_extension_from_zip(
+    const std::string& zip_filename, 
+    const std::string& extension);
+
+}       // namespace foundation
+
+#endif  // !APPLESEED_FOUNDATION_UTILITY_UNZIPPER_H
```
dictoon: Add blank line at end of file.

src/appleseed/renderer/modeling/project/projectfilereader.cpp
```diff
@@ -3036,15 +3038,49 @@ namespace
     }
 }
 
+bool is_packed_project(const string& project_filepath, string& project_name) {
+    try 
+    {
+        vector<string> appleseed_files = get_filenames_with_extension_from_zip(project_filepath, ".appleseed");
+        
+        if (appleseed_files.size() != 1)
+            throw ExceptionUnsupportedFileFormat("Should be exactly one .appleseed file in project");
```
dictoon: This is a bit weird. You're throwing an exception but immediately catching it below.

Since this is in the renderer, and not generic code:

Instead of using exceptions, I suggest just writing an error message to the log and returning false.

src/appleseed/renderer/modeling/project/projectfilereader.cpp
```diff
@@ -3036,15 +3038,49 @@ namespace
...
+    string project_name;
+
+    // Handle packed projects
+    string actual_project_filepath;
+    if (is_packed_project(project_filepath, project_name))
+    {
+        string unpacked_project_directory(project_filepath);
+        unpacked_project_directory += ".unpacked";
+
+        RENDERER_LOG_INFO("%s appears to be a packed project", project_filepath);
```
dictoon: End messages with a period (`.`).

src/appleseed/renderer/modeling/project/projectfilereader.cpp
```diff
@@ -3036,15 +3038,49 @@ namespace
...
+        RENDERER_LOG_INFO("Unpacking to %s...", unpacked_project_directory.c_str());
```
dictoon: For some reason all our log messages are in lower case... This is highly debatable but for the sake of consistency, please do the same :)

src/appleseed/renderer/modeling/project/projectfilereader.cpp
```diff
@@ -3036,15 +3038,49 @@ namespace
...
+
+        unzip(project_filepath, unpacked_project_directory);
+        
+        actual_project_filepath = (bf::path(unpacked_project_directory) / bf::path(project_name)).string().c_str();
+        project_filepath = actual_project_filepath.data();
```
dictoon: I think it would be clearer to do it like I suggested on Slack, i.e. to use `actual_project_filepath` throughout:

    string actual_project_filepath = project_filepath;

    if (packed)
    {
        ...
        actual_project_filepath = something else;
    }

    // Use actual_project_filepath or actual_project_filepath.c_str() in the following code.

src/appleseed/CMakeLists.txt
glebmish: done (I'll add comments everywhere so I can track what's already done)

dictoon: Hm, that's going to be a lot of mail notifications! :)

src/appleseed/foundation/utility/unzipper.cpp
glebmish: done

src/appleseed/foundation/utility/unzipper.cpp
glebmish: what's wrong with while?

src/appleseed/foundation/utility/unzipper.h
glebmish: I need it to specify throws(). If not specified, it fails to compile

src/appleseed/renderer/modeling/project/projectfilereader.cpp
glebmish: These two are different exceptions.
UnzipException comes from `get_filenames_with_extension_from_zip` function in case of IO errors or if file is not a zip
ExceptionUnsupportedFileFormat throws to outside of the function

src/appleseed/renderer/modeling/project/projectfilereader.cpp
glebmish: I thought that it can be unclear to use `actual_project_filepath` only in this function because `project_filepath` is used everywhere else. What do you think? If you think it will be better, I'll do it this way.
Also I tried your suggestion. There is a possibility that somebody will use `project_filepath` when later in this function. In that case I think good solution will be to make it NULL after initialization of `actual_project_filepath`

src/appleseed/foundation/utility/unzipper.cpp
dictoon: My comment applies to the second error message: I meant to replace *with* by *while*.

src/appleseed/foundation/utility/unzipper.h
dictoon: You should be able to remove the destructor entirely since it does nothing.

src/appleseed/renderer/modeling/project/projectfilereader.cpp
dictoon: You're right, I had overlooked that.

Still, a `is_something()` method shouldn't throw an exception. Maybe inline this whole function where it's used instead, and get rid of the function?

src/appleseed/foundation/utility/unzipper.cpp
glebmish: I don't know how to deal with int
```C++
inline UnzipException::UnzipException(const char* what, const int err)
{
  char *string_error = *int_to_string(err);*
  StringException(what, string_error);
}
```

src/appleseed/foundation/utility/unzipper.cpp
glebmish: understood, will do

src/appleseed/foundation/utility/unzipper.cpp
dictoon: You can use `foundation::to_string()` (make sure you include "foundation/utility/string.h" if needed).

Note that the constructor of the base class must be invoked in an initializer list:

    UnzipException::UnzipException()
      : StringException(...)
    {
    }

Alternatively, you can use `Exception::set_what()` to store the message in the exception.

src/appleseed/foundation/utility/unzipper.cpp
glebmish: http://serverfault.com/questions/9546/filename-length-limits-on-linux
Here is a link and summary in top answer. It says that 255 bytes is a standard file length maximum

dictoon: On Linux maybe, but what about macOS, Windows and the other OS appleseed builds and runs on? :)

glebmish:
There are 2 issues left
* length of filename buffer
* actual_project_filepath

dictoon:
Thanks!

Regarding `actual_project_filepath`, maybe your solution is the best one after all. Let's leave it aside for now.

src/appleseed/foundation/utility/unzipper.cpp
glebmish: Hmm, is there any standard way of resolving thin issue?

glebmish:
One more issue is with exception in `is_packed_project`
I will think how to do it better. I would prefer not to inline this function

src/appleseed/foundation/utility/unzipper.cpp
dictoon: Seems like you should be able to know in advance the length of the filename:
http://stackoverflow.com/questions/16386335/how-does-one-read-files-in-subfolders-with-minizip

src/appleseed/foundation/utility/unzipper.cpp
```diff
@@ -0,0 +1,197 @@
...
+    if (err == UNZ_CRCERROR)
+        throw UnzipException("CRC32 is not good");
+}
+
+string read_filename(unzFile& zip_file) 
+{
+    unz_file_info zip_file_info;
+    unzGetCurrentFileInfo(zip_file, &zip_file_info, NULL, 0, NULL, 0, NULL, 0);
+
+    char* filename = new char[zip_file_info.size_filename + 1];
```
dictoon: We can do even better I think, and in particular we can make this code exception-safe (which it isn't currently: if `unzGetCurrentFileInfo()` throws, memory is leaked).

With C++11 and beyond we probably could do

    string inzip_filename;
    inzip_filename.resize(zip_file_info.size_filename);
    unzGetCurrentFileInfo(zip_file, &zip_file_info, &inzip_filename[0], zip_file_info.size_filename, NULL, 0, NULL, 0);

because `std::string` is guaranteed to use contiguous storage.

Unfortunately, until the next release we are still stuck to C++03, where this would be better:

    vector filename(zip_file_info.size_filename + 1);
    unzGetCurrentFileInfo(zip_file, &zip_file_info, &filename[0], filename.size(), NULL, 0, NULL, 0);
    const string inzip_filename(&filename[0]);

To be tested (especially regarding the +1 for `\0` etc.)

src/appleseed/foundation/utility/unzipper.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+    fstream out(filepath.c_str(), ios_base::out | ios_base::binary);
+
+    char buffer[BUFFER_SIZE];
+    int read;
+
+    do 
+    {
+        read = read_chunk(zip_file, (char*) &buffer, BUFFER_SIZE);         
+        out.write((char*) &buffer, read);
```
dictoon: You can remove the cast here.

src/appleseed/foundation/utility/unzipper.cpp
```diff
@@ -0,0 +1,197 @@
...
+    else open_current_file(zip_file);
+
+    fstream out(filepath.c_str(), ios_base::out | ios_base::binary);
+
+    char buffer[BUFFER_SIZE];
+    int read;
+
+    do 
+    {
+        read = read_chunk(zip_file, (char*) &buffer, BUFFER_SIZE);
```
dictoon: Here too.

src/appleseed/foundation/utility/unzipper.cpp
glebmish: I can't
```
/home/gleb/source/appleseed/appleseed/src/appleseed/foundation/utility/unzipper.cpp: In function ‘void foundation::extract_current_file(void*&, const string&)’:
/home/gleb/source/appleseed/appleseed/src/appleseed/foundation/utility/unzipper.cpp:126:57: error: cannot convert ‘char (*)[4096]’ to ‘char*’ for argument ‘2’ to ‘int foundation::read_chunk(void*&, char*, int)’
         read = read_chunk(zip_file, &buffer, BUFFER_SIZE);         
                                                         ^
/home/gleb/source/appleseed/appleseed/src/appleseed/foundation/utility/unzipper.cpp:127:32: error: no matching function for call to ‘std::basic_fstream<char>::write(char (*)[4096], int&)’
         out.write(&buffer, read);
```

src/appleseed/foundation/utility/unzipper.cpp
glebmish: OK, I'll do it later

src/appleseed/renderer/modeling/project/projectfilereader.cpp
```diff
@@ -3036,15 +3038,54 @@ namespace
...
+
+    // Handle packed projects
+    string actual_project_filepath;
+    if (is_packed_project(project_filepath, project_name))
+    {
+        string unpacked_project_directory(project_filepath);
+        unpacked_project_directory += ".unpacked";
+
+        RENDERER_LOG_INFO("%s appears to be a packed project.", project_filepath);
+        RENDERER_LOG_INFO("unpacking to %s...", unpacked_project_directory.c_str());
```
dictoon: I suggest combining these two messages:

    RENDERER_LOG_INFO("%s appears to be a packed project; unpacking to %s...", project_filepath, unpacked_project_directory.c_str());

src/appleseed/foundation/utility/unzipper.cpp
dictoon: Oh interesting, ok, nevermind then.

dictoon:
Perfect, thanks! Will merge when Travis has done building it.

Issue #1262, opened 2017-02-23
Implement the new STD microfacet distribution function


est77:
https://bv.univ-poitiers.fr/access/content/user/mribar03/website/projects/eg_2017/index.html

est77:
Isotropic variant implemented in #1303

est77:
Rest added in https://github.com/appleseedhq/appleseed/pull/1399

PR #1303, opened 2017-03-11
STD MDF implementation

gospodnetic:
Issue #1262 
Implementation of equations, no unit tests yet.
Notes:
1) since equation for normal sampling uses arctan, I used the second version of function make_unit_vector which uses directly theta and phi.

2) Gamma function in disney code was implemented directly, as cmath already has the function, I used that one.

### DIFF
--- a/src/appleseed/foundation/math/microfacet.cpp
+++ b/src/appleseed/foundation/math/microfacet.cpp
@@ -773,4 +773,50 @@ float GTR1MDF::pdf(
     return D(h, alpha_x, alpha_y, gamma) * h.y;
 }
 
+// 
+// StdMDF class implementation
+//
+
+StdMDF::StdMDF() {}
+
+float StdMDF::D(
+    const Vector3f&     h,
+    const float         alpha_x,
+    const float         alpha_y,
+    const float         gamma) const
+{
+    const float cos_theta = h.y;
+
+    if (cos_theta == 0.0f)
+        return 0.0;
+
+    const float cos_theta_2 = square(cos_theta);
+    const float cos_theta_4 = square(cos_theta_2);
+    const float alpha_x2 = square(alpha_x);
+    const float tan_theta_2 = (1.0f - cos_theta_2) / cos_theta_2;
+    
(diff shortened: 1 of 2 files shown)


src/appleseed/foundation/math/microfacet.cpp
```diff
@@ -773,4 +773,115 @@ float GTR1MDF::pdf(
...
+    const Vector3f&     v,
+    const Vector3f&     s,
+    const float         alpha_x,
+    const float         alpha_y,
+    const float         gamma) const
+{
+    const float phi = TwoPi<float>() * s[0];
+    const float a = gamma - 1.0f;
+    const float b = std::pow(1.0f - s[1], 1.0f / (1.0f - gamma)) - 1.0f;
+    const float theta = atan(alpha_x * std::sqrt(a * b));
```
est77: I think we need std:: here, std::atan(...)

src/appleseed/foundation/math/microfacet.h
```diff
@@ -407,6 +407,61 @@ class GTR1MDF
...
+        const float         alpha_x,
+        const float         alpha_y,
+        const float         gamma) const APPLESEED_OVERRIDE;
+
+    virtual Vector3f sample(
+        const Vector3f&     v,
+        const Vector3f&     s,
+        const float         alpha_x,
+        const float         alpha_y,
+        const float         gamma = 0.0f) const APPLESEED_OVERRIDE;
```
est77: We can remove the default value for the gamma argument here.
Same thing in the following method declaration.

src/appleseed/foundation/math/microfacet.cpp
gospodnetic: atan is a part of cmath which was already included. I'll change it to std:: - is it better to use std:: for trig. functs?

est77: This code used to be in a header (it was a template before). That's why all cmath functions are qualified. we could remove the std:: in all the file and add a using namespace std on top.
But that can be done in a different PR. For now we can just be consistent with the rest of the code in the file and use std:: qualifiers for cmath functions.

Issue #1276, opened 2017-02-28
Use geometric normal "as is" (unflipped) when no vertex normals are present?


dictoon:

est77:
I think we can assume that input meshes are consistently oriented, so I'd agree with this change.

andr2xea:
I will take this issue to fix.

dictoon:
Done.

PR #1304, opened 2017-03-11
Fix issue #1276

andr2xea:
With the help from https://github.com/Aakash1312.

### DIFF
--- a/src/appleseed/renderer/kernel/shading/shadingpoint.cpp
+++ b/src/appleseed/renderer/kernel/shading/shadingpoint.cpp
@@ -669,7 +669,7 @@ void ShadingPoint::compute_geometric_normal() const
                     : ObjectInstance::FrontSide;
 
             // Make the geometric normal face the direction of the incoming ray.
-            if (m_side == ObjectInstance::FrontSide)
+            if (m_side == ObjectInstance::BackSide)
                 m_geometric_normal = -m_geometric_normal;
         }
     }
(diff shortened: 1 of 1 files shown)


src/appleseed/renderer/kernel/shading/shadingpoint.cpp
```diff
@@ -660,17 +660,17 @@ void ShadingPoint::compute_geometric_normal() const
...
-            // In the absence of per-vertex normals, we have no way to know if we are
-            // hitting the front face or the back face of the surface. Assume we are
-            // always hitting the front face...
-            m_side = ObjectInstance::FrontSide;
-
-            // ...and if the geometric normal is not facing the ray, flip it.
-            if (dot(m_ray.m_dir, m_geometric_normal) > 0.0)
-                m_geometric_normal = -m_geometric_normal;
-        }
+        // else
```
dictoon: That cannot be correct, you're leaving `m_side` uninitialized, as far as I can see. It should be initialized based on the direction the geometric normal is pointing to.

Did you try to render with the Sides diagnostic mode? Try to check that the outside is indeed blue, and the inside is indeed red.

Also, it's better to delete code than comment it out. If we ever need it again, it's in Git.

Aakash1312: I think we only need to remove lines 670, 671 and 672

PR #1306, opened 2017-03-12
Fix bug when valid packed project doesn't open

glebmish:
The reason was in has_extension(). It didn't check rfind() result on npos
When filename.size() == extension.size() - 1 substraction of this two values is -1
npos == -1 so in this case function returned true although file don't have such extension

### DIFF
--- a/src/appleseed/foundation/utility/unzipper.cpp
+++ b/src/appleseed/foundation/utility/unzipper.cpp
@@ -31,6 +31,7 @@
 
 // appleseed.foundation headers.
 #include "foundation/core/exceptions/exception.h"
+#include "foundation/utility/string.h"
 #include "foundation/utility/minizip/unzip.h"
 
 // Boost headers.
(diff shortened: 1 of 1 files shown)


src/appleseed/foundation/utility/unzipper.cpp
```diff
@@ -163,7 +162,8 @@ void unzip(const string& zip_filename, const string& unzipped_dir)
 
 bool has_extension(const string& filename, const string& extension) 
 {
-    return filename.rfind(extension) == filename.size() - extension.size();
+    size_t extension_start = filename.rfind(extension);
+    return  extension_start != string::npos && extension_start == filename.size() - extension.size();
```
dictoon: Actually, now that I think of it, I suggest getting rid of the whole function and using `foundation::ends_with()` instead. Sorry, I didn't think of that before :)

PR #1310, opened 2017-03-13
StdMDF option under PlasticBRDF

gospodnetic:
Added possibility to use StdMDF with gamma parameter as Highligh Falloff.
I've also added gamma with default value of 0.0f as argument for functions

- specular_pdf()
- evaluate_specular()

### DIFF
--- a/src/appleseed/renderer/modeling/bsdf/plasticbrdf.cpp
+++ b/src/appleseed/renderer/modeling/bsdf/plasticbrdf.cpp
@@ -219,14 +219,14 @@ namespace
                         values->m_specular_reflectance,
                         *m_mdf,
                         alpha,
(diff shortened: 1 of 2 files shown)


src/appleseed/renderer/modeling/bsdf/plasticbrdf.cpp
```diff
@@ -398,7 +406,8 @@ namespace
             const Vector3f&         wo,
             const Vector3f&         m,
             const float             F,
-            Spectrum&               value)
+            Spectrum&               value,
+            const float             gamma)
```
gospodnetic: Makes complete sense, this was a leftover after removing the default values which required the variable to come at the end. Fixed!

PR #1311, opened 2017-03-14
Add tests for unzipper and packed project

glebmish:


src/appleseed/foundation/meta/tests/test_unzipper.cpp
```diff
@@ -0,0 +1,124 @@
...
+using namespace foundation;
+using namespace std;
+namespace bf = boost::filesystem;
+
+TEST_SUITE(Foundation_Utility_Unzipper)
+{
+    string valid_project = "unit tests/inputs/test_packed_project_valid.appleseedz";
+    string invalid_project = "unit tests/inputs/test_packed_project_invalid.appleseedz";
+
+    set<string> rec_ls(bf::path dir)
```
dictoon: I get what the `ls` part stands for, but what does `rec` mean? Consider using the whole word if it's an abbreviation.

dictoon: Ah, recursive I guess :) It makes sense to call use the full word.

src/appleseed/foundation/meta/tests/test_unzipper.cpp
```diff
@@ -0,0 +1,124 @@
...
+                files.insert(files_in_subdir.begin(), files_in_subdir.end());
+            }
+            else
+                files.insert(current_path.filename().string());
+        }
+
+        return files;
+    }
+
+    TEST_CASE(UnzipTest) {
```
dictoon: Move brace to its own line.

src/appleseed/foundation/meta/tests/test_unzipper.cpp
```diff
@@ -0,0 +1,124 @@
...
+                "plane.obj",
+                "Sphere002.binarymesh",
+                "Plane002.binarymesh",
+                "cube.obj",
+                "Plane001.binarymesh"
+            };
+
+            set<string> actual_files = rec_ls(bf::path(unpacked_dir));
+
+            for (int i = 0; i < 11; ++i)
```
dictoon: Please use size_t unless you really need a signed type. Also, you can use the `countof()` macro:

    for (size_t i = 0; i < countof(expected_files); ++i)

src/appleseed/foundation/meta/tests/test_unzipper.cpp
```diff
@@ -0,0 +1,124 @@
...
+                "Plane002.binarymesh",
+                "cube.obj",
+                "Plane001.binarymesh"
+            };
+
+            set<string> actual_files = rec_ls(bf::path(unpacked_dir));
+
+            for (int i = 0; i < 11; ++i) 
+            {
+                EXPECT_TRUE(actual_files.count(expected_files[i]));
```
dictoon: Are you sure `EXPECT_TRUE()` is what you want? If you are counting things, you should probably check that the count is equal to 1.

src/appleseed/foundation/meta/tests/test_unzipper.cpp
```diff
@@ -0,0 +1,124 @@
...
+        string unpacked_dir = valid_project + ".unpacked";
+
+        try 
+        {
+            unzip(valid_project, unpacked_dir);
+
+            EXPECT_EQ(bf::exists(bf::path(unpacked_dir)), true);
+            EXPECT_EQ(bf::is_empty(bf::path(unpacked_dir)), false);
+
+            string expected_files[] =
```
dictoon: Make const.

src/appleseed/foundation/meta/tests/test_unzipper.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+        // A default constructed directory_iterator acts as the end iterator
+        bf::directory_iterator end_iter;
+        for (bf::directory_iterator dir_itr(dir); dir_itr != end_iter; ++dir_itr) 
+        {
+            bf::path current_path = dir_itr->path();
+
+            if (bf::is_directory(current_path))
+            {
+                set<string> files_in_subdir = rec_ls(current_path);
```
dictoon: Make const.

src/appleseed/foundation/meta/tests/test_unzipper.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+    set<string> rec_ls(bf::path dir) 
+    {
+        set<string> files;
+
+        // A default constructed directory_iterator acts as the end iterator
+        bf::directory_iterator end_iter;
+        for (bf::directory_iterator dir_itr(dir); dir_itr != end_iter; ++dir_itr) 
+        {
+            bf::path current_path = dir_itr->path();
```
dictoon: Make const.

src/appleseed/foundation/meta/tests/test_unzipper.cpp
```diff
@@ -0,0 +1,124 @@
...
+            bf::remove_all(bf::path(unpacked_dir));
+            throw e;
+        }
+    }
+
+    TEST_CASE(FilesnamesWithExtensionTest) 
+    {
+        string extension = ".appleseed";
+
+        vector<string> appleseed_files_4 = get_filenames_with_extension_from_zip(invalid_project, extension);
```
dictoon: Split test case into two or more.

Basically, one test case should check one thing, or one aspect. The name of the test should reflect what it is testing, such that when a test fails, we can get a pretty good idea of what's wrong just by looking at the name of the test case.

src/appleseed/foundation/meta/tests/test_unzipper.cpp
```diff
@@ -0,0 +1,124 @@
...
+        vector<string> appleseed_files_1 = get_filenames_with_extension_from_zip(valid_project, extension);
+
+        EXPECT_EQ(appleseed_files_4.size(), 4);
```
dictoon: In all `EXPECT_*()` and `ASSERT_*()` macros, the expected value should come first, followed by the value to test, e.g.

    EXPECT_EQ(4, appleseed_files_4.size());

src/appleseed/foundation/meta/tests/test_unzipper.cpp
glebmish: As it is set, there will always be 0 or 1. But ok, I'll change it. I wrote EXPECT_EQ(*, 1); originally but changed it later

dictoon: Even if both tests ultimately are the same, the semantic is important. As a reader, I'm left wondering, why is count() returning a bool?

src/appleseed/foundation/meta/tests/test_unzipper.cpp
```diff
@@ -0,0 +1,124 @@
...
+TEST_SUITE(Foundation_Utility_Unzipper)
+{
+    string valid_project = "unit tests/inputs/test_packed_project_valid.appleseedz";
+    string invalid_project = "unit tests/inputs/test_packed_project_invalid.appleseedz";
+
+    set<string> rec_ls(bf::path dir) 
+    {
+        set<string> files;
+
+        // A default constructed directory_iterator acts as the end iterator
```
glebmish: Thanks! Right from boost docs :)

src/appleseed/foundation/meta/tests/test_unzipper.cpp
glebmish: Yes, recursive. I'll rename it

glebmish:
I also changed recursive_ls a bit

src/appleseed/renderer/meta/tests/test_projectfilereader.cpp
```diff
@@ -67,4 +74,41 @@ TEST_SUITE(Renderer_Modeling_Project_ProjectFileReader)
 
         EXPECT_TRUE(identical);
     }
+
+    TEST_CASE(ReadPackedProject)
```
dictoon: I would argue that this test case should also be split in two.

src/appleseed/foundation/meta/tests/test_unzipper.cpp
```diff
@@ -36,30 +36,34 @@
 // Standard headers
 #include <set>
 #include <vector>
+#include <iostream>
```
glebmish: sorry, forgot to delete after some testing

glebmish:
When test for invalid packed project executes, there are some logs printed
```
2017-03-14T20:26:39.109737Z <001>    37 MB error   | while reading unit tests/inputs/test_packed_project_invalid.appleseedz, at line 1, column 1: invalid multi-byte sequence.
2017-03-14T20:26:39.109913Z <001>    37 MB error   | failed to load project file unit tests/inputs/test_packed_project_invalid.appleseedz (1 error, 0 warning).
```
It doesn't effect on test result, but it would be better without it, and I don't how to suppress these logs

dictoon:
> When test for invalid packed project executes, there are some logs printed

I think these log messages actually reveal a bug: we should not attempt to parse a ZIP file as an appleseed (XML) file. Why is this happening?

glebmish:
is_packed_project() checks if it's valid packed project. If it is, project_filepath variable changes on .appleseed file in .unpacked dir. If it's not, nothing happens and ProjectFileReader::read tries to read file as usual

glebmish:
It could be easy to handle this situation by catching exception from is_packed_project() that you asked to remove. Now it could be done for example with error code returned from is_packed_project()

dictoon:
Ok, I see. I think the algorithm should be roughly as follow:

    if file extension is appleseedz:
        unpack
        if a single appleseed file:
            read it
        otherwise
            invalid packed project file error
    else if extension is appleseed:
        read it

At no point do we try to read a ZIP file as an XML file.

If `is_packed_project()` is a liability in the sense that it doesn't allow you to express the above algorithm (or an equivalent one) cleanly, I suggest getting rid of the function.

glebmish:
I'll fix this issue soon and send fix with separate PR, ok?

dictoon:
Ok! Merging now then.

Great work, thanks!

glebmish:
I'm happy to do it
It was interesting to write tests as I don't have much experience here

Issue #1229, opened 2017-02-02
Add support for SSS groups


dictoon:
Suggested by Adrian here:
https://forum.appleseedhq.net/t/does-appleseed-support-a-shader-with-sss-grouping/218?u=franz

The idea is to be able to treat multiple intersecting/overlapping closed objects with the same SSS material as a single object. That means ignoring interior intersections when doing subsurface sampling (a stack would be needed to keep track of enter/leave events and only keep the last event).

Narann:
There is also the invert: Use multiple groups (generically named *trace set*) to separate SSS (or reflection) from two objects.

We relied on this a lot for almost everything on Ballerina (sss, reflection, indirect diffuse, etc.).

http://guerillarender.com/doc/1.4/User%20Guide_Scene%20Graph_Set.html
http://guerillarender.com/doc/1.4/User%20Guide_Rendering_Trace%20Linking.html
https://support.solidangle.com/display/ARP/Using+Trace+Sets

PR #1313, opened 2017-03-14
SSS sets support

Biart95:
Add support for SSS sets.

By default, all object instances are treated as individual SSS sets. User can specify SSS set identifier for each object instance. If two object instances with the same BSSRDF function belong to the same SSS set, they are treated as one object instance during BSSRDF sampling.

This pull request is related to Issue #1229 .


dictoon:
I am very, very impressed by this work. A large, non-trivial task solved elegantly; clean code that respects our coding conventions; test scenes that show that the feature works as expected. Wow!

As I noted in earlier comments, I would suggest to simplify things a little bit, and make them faster:

There is a string class in OpenImageIO (OIIO) called `OIIO::ustring`. From `OpenImageIO/ustring.h`:

    A ustring is an alternative to char* or std::string for storing
    strings, in which the character sequence is unique (allowing many
    speed advantages for assignment, equality testing, and inequality
    testing).

    The implementation is that behind the scenes there is a hash set of
    allocated strings, so the characters of each string are unique.  A
    ustring itself is a pointer to the characters of one of these canonical
    strings.  Therefore, assignment and equality testing is just a single
    32- or 64-bit int operation, the only mutex is when a ustring is
    created from raw characters, and the only malloc is the first time
    each canonical ustring is created.

    The internal table also contains a std::string version and the length
    of the string, so converting a ustring to a std::string (via
    ustring::string()) or querying the number of characters (via
    ustring::size() or ustring::length()) is extremely inexpensive, and does
    not involve creation/allocation of a new std::string or a call to
    strlen.

This class would allow to compare SSS sets in a much faster way than `std::string`.

I would also treat object instances without an explicit SSS set parameter as being in their own SSS set. In that case their SSS set ID would be an empty ustring, and you would treat that case separately in the SSS set ID comparison code (like you're doing now basically).

@est77 also mentioned that you could completely hide the SSS set ID inside the implementation of `ObjectInstance`, and only provide one public method in `ObjectInstance` that returns whether two instances are in the same SSS set or not.

Let me know if something is not clear.

Again, thanks for the hard work!

dictoon:
Also, GitHub says that this PR has conflicts. If you're familiar with Git, you should do an interactive rebase of your branch on top of upstream's master, and skip the commit "Resolve conflicts and fix minor warnings". That should take care of the conflicts, hopefully.

est77:
Another possibility for the conflicts would to create a new branch from master and cherry pick the commits with your changes, since the conflicts are in other merged commits.

dictoon:
Hi, thanks for the fixes!

Unfortunately, it looks like the rebasing didn't work out as expected because now there are even more foreign commits (and conflicts) and this makes the review really hard.

What about following @est77's suggestion to create a new branch off the upstream's master, cherry-picking your commits into that new branch, and opening a new PR from that branch?

Biart95:
Sure! This branch is now flooded with commits and definitely cannot be merged with appleseed. I will put all changes into the new branch and open a new PR. Thanks for the suggestion.

dictoon:
Awesome, thanks!

PR #1315, opened 2017-03-15
Fix stdf mdf overflow

gospodnetic:
Source of the problem: exponential explosion for higher gamma values:

- refactored G1
- refactored D
- introduced type double in D in order to expand numerical range - need comments on that, how to proceed, what is acceptable, what not since it is not consistent with the rest of the code

- sampling warning occurs while moving highlight falloff slider in applseed studio

### DIFF
--- a/src/appleseed/foundation/meta/tests/test_microfacet.cpp
+++ b/src/appleseed/foundation/meta/tests/test_microfacet.cpp
@@ -444,17 +444,40 @@ TEST_SUITE(Foundation_Math_Microfacet)
         const StdMDF std;
 
         float gamma_max = 100.0f;
-        float gamma_step = 1.0f;
+        float gamma_step = 2.0f;
         size_t num_sample = 128;
-        for (float i = 2.0f; i < gamma_max; i += gamma_step)
+        for (float i = 20.0f; i < gamma_max; i += gamma_step)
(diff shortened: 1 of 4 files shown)


src/appleseed/foundation/meta/tests/test_microfacet.cpp
```diff
@@ -419,5 +418,71 @@ TEST_SUITE(Foundation_Math_Microfacet)
...
+        size_t num_sample = 128;
+        for (float i = 10.0f; i < gamma_max; i += gamma_step)
+        {
+            for (size_t j = 0; j < num_sample; ++j)            
+            {         
+                static const size_t Bases[] = { 2 };
+                const Vector2f s = hammersley_sequence<float, 2>(Bases, num_sample, i);
+                const Vector3f v = sample_hemisphere_uniform(s);
+                const float std_G1 = std.G1(v, Vector3f(0.0f, 1.0, 0.0f), 0.5, 0.5, i);
+                EXPECT_TRUE(std::isfinite(std_G1)); // check that G1 doesn't produce NaN
```
dictoon: `std::isfinite()` is available starting with C++11. For now, you'll need to use `foundation::FP<float>::is_finite()`.

PR #1318, opened 2017-03-15
SSS sets support (revisited)

Biart95:
Add support for SSS sets.

By default, all object instances are treated as individual SSS sets. User can specify SSS set identifier for each object instance. If two object instances with the same BSSRDF function belong to the same SSS set, they are treated as one object instance during BSSRDF sampling.

This pull request is related to Issue #1229 .

This is revisited version of Pull Request #1313 with simplified user interface and with usage of `OIIO::ustring` instead of `std::string`.


sandbox/tests/test scenes/sss sets/spheres_common_sss_set.appleseed
```diff
@@ -0,0 +1,188 @@
...
+            <material name="red_material" model="generic_material">
+                <parameter name="bsdf" value="red_material_brdf" />
+                <parameter name="surface_shader" value="physical_shader" />
+            </material>
+            <material name="white_material" model="generic_material">
+                <parameter name="bsdf" value="white_material_brdf" />
+                <parameter name="surface_shader" value="physical_shader" />
+            </material>
+            <object name="spheres" model="mesh_object">
+                <parameter name="filename" value="F:\Appleseed\scenes\spheres.obj" />
```
dictoon: Still some absolute paths. You'll need to replace them by relative paths.

sandbox/tests/test scenes/sss sets/sss_sets_comparison.appleseed
```diff
@@ -0,0 +1,243 @@
...
+                <parameter name="ray_bias_method" value="none" />
+                <parameter name="sss_set_id" value="sss_set" />
+                <assign_material slot="None" side="front" material="marble_sss_material" />
+            </object_instance>
+            <object_instance name="sphere_right_side" object="sss_sets_test.Sphere.004">
+                <parameter name="medium_priority" value="0" />
+                <parameter name="ray_bias_distance" value="0.0" />
+                <parameter name="ray_bias_method" value="none" />
+                <parameter name="sss_set_id" value="sss_set" />
+                <parameter name="use_individual_sss_set" value="false" />
```
dictoon: As far as I understand, this parameter no longer exist and needs to be removed from all test scenes.

src/appleseed/renderer/modeling/bssrdf/separablebssrdf.cpp
```diff
@@ -246,11 +247,23 @@ namespace
...
-                (incoming_point.get_material() == outgoing_material ||
-                 incoming_point.get_opposite_material() == outgoing_material))
+            const Material* incoming_material = incoming_point.get_material();
+            const BSSRDF* incoming_bssrdf =
+                incoming_material == NULL ? NULL :
+                incoming_material->get_uncached_bssrdf();
+            const Material* incoming_opposite_material = incoming_point.get_opposite_material();
+            const BSSRDF* incoming_opposite_bssrdf =
+                incoming_opposite_material == NULL ? NULL :
+                incoming_opposite_material->get_uncached_bssrdf();
```
dictoon: As explained in the review of your previous PR, you must not call the uncached variants of these methods.

src/appleseed/renderer/modeling/scene/objectinstance.h
```diff
@@ -170,6 +176,26 @@ class APPLESEED_DLLSYMBOL ObjectInstance
     // Return the object bound to this instance.
     Object& get_object() const;
 
+    // SSS set
+    // By default, each object instance belongs to its own SSS set
+    struct SubsurfaceScatteringSet
```
dictoon: I suggest moving this entire class to the .cpp file. It doesn't need to be exposed.

Even better, I think you can completely get rid of this class and just use bare `OIIO::ustring` objects stored in object instances, with the convention that an empty ustring means that the object instance is in its own SSS set.

dictoon:
Thanks, looks much better!

There are still a number of things that need to be addressed before we can merge this, but it's getting closer!

Thanks for the high quality work, appreciated.

Biart95:
Thank you for the detailed review! I made all requested changes:

- SubsurfaceScatteringSet removed, OIIO::ustring hide in .cpp. (it really makes the things simpler!)
- All my scenes are fixed
- Only cached accesses are now used

I am still not sure if using cached access is safe enough, but all unit tests are passed (except for LocalSampleAccumulationBuffer) and test scenes are working. Please, inform me if something can potentially cause bugs in the future, and I will try to sort it out.

src/appleseed/renderer/modeling/bssrdf/separablebssrdf.cpp
```diff
@@ -246,11 +247,23 @@ namespace
...
             probe_ray.m_tmax = norm(exit_point - probe_ray.m_org);
 
-            // Only consider incoming points on the same object instance,
-            // and with the same material as the outgoing point.
-            if (incoming_point.get_object_instance().get_uid() == outgoing_obj_inst_uid &&
-                (incoming_point.get_material() == outgoing_material ||
-                 incoming_point.get_opposite_material() == outgoing_material))
+            const Material* incoming_material = incoming_point.get_material();
+            const BSSRDF* incoming_bssrdf =
+                incoming_material == NULL ? NULL :
```
dictoon: You should use `0` instead of `NULL` in C++03 code.

src/appleseed/renderer/modeling/bssrdf/separablebssrdf.cpp
```diff
@@ -246,11 +247,23 @@ namespace
...
+            const BSSRDF* incoming_opposite_bssrdf =
+                incoming_opposite_material == NULL ? NULL :
+                incoming_opposite_material->get_render_data().m_bssrdf;
+
+            bool same_bssrdf =
+                incoming_bssrdf == outgoing_bssrdf || incoming_opposite_bssrdf == outgoing_bssrdf;
+            bool same_sss_set =
+                incoming_point.get_object_instance().is_in_same_sss_set(outgoing_object_instance);
+
+            // Only consider hit points with the same bssrdf as the outgoing point
```
dictoon: Capitalize *BSSRDF* in comments since it's an acronym.

PR #1333, opened 2017-03-20
Low light threshold

aytekaman:
-Calculate approximate contribution to skip shadow rays (very naive at
the moment)
-Exposed "Low Light Threshold" parameter in the UI


src/appleseed/renderer/modeling/edf/osledf.cpp
```diff
@@ -213,6 +214,24 @@ namespace
             return probability;
         }
 
+		float get_uncached_max_radiance_value() const
```
dictoon: Add `APPLESEED_OVERRIDE` at the end of the method to tell the compiler and the reader that this method is supposed to override another.

src/appleseed/renderer/modeling/edf/coneedf.cpp
```diff
@@ -195,6 +196,24 @@ namespace
...
 
+		virtual float get_uncached_max_radiance_value() const
+		{
+			Spectrum radiance;
+
+			Source *source = m_inputs.source("radiance");
+
+			assert(source);
+
+			source->evaluate_uniform(radiance);
```
dictoon: If the source is not uniform (constant), `evaluate_uniform()` will not evaluate it. This will happen for instance is the source is a texture, or a procedural function.

In that case, I think we should return `numeric_limits<float>::max()` to make sure we don't consider this light source a low contributing one.

aytekaman:
Thanks for the comprehensive feedback! I did necessary changes.

src/appleseed/renderer/kernel/lighting/ilightingengine.cpp
```diff
@@ -59,6 +59,14 @@ void ILightingEngineFactory::add_common_params_metadata(
                 .insert("label", "Light Samples")
                 .insert("help", "Number of samples used to estimate direct lighting"));
 
+		metadata.dictionaries().insert(
+			"dl_low_light_threshold",
```
dictoon: There are still tabs here.

Make sure your editor is configured to use 4 spaces for indenting.

src/appleseed/renderer/modeling/edf/osledf.cpp
```diff
@@ -45,7 +46,11 @@
 #include "foundation/platform/compiler.h"
 #include "foundation/utility/arena.h"
 
+// Standard headers.
+#include <cmath>
```
dictoon: `std::numeric_limits<>` is in `<limits>`, not `<cmath>`.

src/appleseed/renderer/modeling/edf/coneedf.cpp
```diff
@@ -195,6 +196,30 @@ namespace
             return sample_cone_uniform_pdf(m_cos_half_angle);
         }
 
+        virtual float get_uncached_max_radiance_value() const APPLESEED_OVERRIDE
+        {
+            Source* radiance_source = m_inputs.source("radiance");
+
+            // If source is not available, simply return max float so that 
+            // light is not considered as a low contributing one.
+            if (!radiance_source)
```
dictoon: This is still not quite correct.

There may well be a source bound to the `radiance` input, but it may not be a uniform source. For instance, it could be a `TextureSource`.

What you need to do is check (1) if there is a source, and (2) if that source is uniform with `Source::is_uniform()`. If both conditions are met, you can use `evaluate_uniform()`. Otherwise, you should return `numeric_limits<float>::max()`.

You'll need to do that for both the `radiance` and `radiance_multiplier` inputs. Only when both have uniform sources bound to them can you return a sensible max radiance value.

Since this code is the same for a few light models, you could factorize it into a helper method in the `EDF` class. This is something that we sometimes do to reduce code duplication. See `ConnectableEntity::check_non_zero_emission()` for an example of how that would look.

dictoon:
My third comment (the important one) is hidden behind a "Show outdated" button, make sure to check it out!

src/appleseed/renderer/modeling/edf/coneedf.cpp
```diff
@@ -33,6 +33,7 @@
 // appleseed.renderer headers.
 #include "renderer/modeling/edf/edf.h"
 #include "renderer/modeling/input/inputarray.h"
+#include "renderer/modeling/input/source.h"
```
dictoon: You can now remove this header file.

src/appleseed/renderer/modeling/edf/diffuseedf.cpp
```diff
@@ -33,6 +33,7 @@
 // appleseed.renderer headers.
 #include "renderer/modeling/edf/edf.h"
 #include "renderer/modeling/input/inputarray.h"
+#include "renderer/modeling/input/source.h"
```
dictoon: You can now remove this header file.

src/appleseed/renderer/modeling/edf/diffuseedf.cpp
```diff
@@ -51,6 +52,7 @@ namespace renderer      { class Assembly; }
 namespace renderer      { class Project; }
 
 using namespace foundation;
+using namespace std;
```
dictoon: You can now remove this `using` clause.

src/appleseed/renderer/modeling/edf/edf.cpp
```diff
@@ -128,4 +132,62 @@ void* EDF::evaluate_inputs(
...
+        return numeric_limits<float>::max();
+
+    Spectrum spectrum;
+
+    source->evaluate_uniform(spectrum);
+
+    return max_value(spectrum);
+}
+
+float EDF::get_max_contribution(const Source* input) const
```
dictoon: This method is a bit weird since it only works for inputs of type spectrum. In fact I don't think you're using it. I'd suggest to remove it.

src/appleseed/renderer/modeling/edf/edf.cpp
```diff
@@ -128,4 +132,62 @@ void* EDF::evaluate_inputs(
...
+{
+    return get_max_contribution_spectrum(input);
+}
+
+float EDF::get_max_contribution(const char* input_name) const
```
dictoon: I'd suggest to remove this one as well.

src/appleseed/renderer/modeling/edf/edf.h
```diff
@@ -99,6 +99,9 @@ class APPLESEED_DLLSYMBOL EDF
     // Retrieve the light near start value.
     double get_uncached_light_near_start() const;
 
+    // Get the cached approximate contribution.
```
dictoon: *approximate maximum contribution*

src/appleseed/renderer/modeling/edf/edf.h
```diff
@@ -146,9 +149,23 @@ class APPLESEED_DLLSYMBOL EDF
         const foundation::Basis3f&  shading_basis,              // world space orthonormal basis around shading normal
         const foundation::Vector3f& outgoing) const = 0;        // world space emission direction, unit-length
 
+  protected:
+        float get_max_contribution_scalar(const Source* source) const;
```
dictoon: Indent these methods like the others.

src/appleseed/renderer/modeling/edf/osledf.cpp
```diff
@@ -37,6 +37,7 @@
 #include "renderer/kernel/shading/shadingpoint.h"
 #include "renderer/modeling/edf/diffuseedf.h"
 #include "renderer/modeling/edf/edf.h"
+#include "renderer/modeling/input/source.h"
```
dictoon: Remove this header file and the `using` clause since they're no longer needed.

src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
```diff
@@ -589,31 +591,47 @@ void DirectLightingIntegrator::add_emitting_triangle_sample_contribution(
...
+    const double rcp_sample_distance = sqrt(rcp_sample_square_distance);
+    
+    // Normalize the incoming direction. 
     cos_on *= rcp_sample_distance;
+    incoming *= rcp_sample_distance;
+    
+    // Decide whether the contributon from this light is significant or not.
+    const double approximate_contribution = 
+        cos_on * 
+        dot(incoming, m_geometric_normal) *
```
dictoon: I'm really not sure it's a good idea to take into account `incoming` at all... But if you do, you should use the shading normal instead of the geometric normal.

src/appleseed/renderer/kernel/lighting/drt/drtlightingengine.cpp
```diff
@@ -125,15 +127,17 @@ namespace
             {
                 RENDERER_LOG_INFO(
                     "distribution ray tracing settings:\n"
-                    "  ibl              %s\n"
-                    "  max path length  %s\n"
-                    "  rr min path len. %s\n"
-                    "  dl light samples %s\n"
-                    "  ibl env samples  %s",
+                    "  ibl                     %s\n"
```
dictoon: These messages are all indented similarly such that they line up nicely in the log output.

I completely agree that the space for the label is way too small, and we'll fix that as some point. In the meantime, I suggest leaving the indentation as it was and using a shorter label for the light threshold, maybe `dl light thresh.`?

src/appleseed/renderer/modeling/edf/edf.cpp
aytekaman: I added it because in the `constantenvironmentedf.cpp` file, I saw that non zero emission is checked using only the spectrum which is like this: `check_non_zero_emission("radiance");`. I will delete it if you think we will never have such case.

src/appleseed/renderer/kernel/lighting/drt/drtlightingengine.cpp
aytekaman: ok, long parameter name was bugging me as well:)

src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
```diff
@@ -606,31 +606,28 @@ void DirectLightingIntegrator::add_emitting_triangle_sample_contribution(
...
     
-    // Decide whether the contributon from this light is significant or not.
+    // Compute the approximate contribution.
     const double approximate_contribution = 
-        cos_on * 
+        (cos_on * 
         rcp_sample_square_distance * 
-        edf->get_max_contribution();
+        edf->get_max_contribution()) 
+        / sample.m_triangle->m_rcp_area;
```
dictoon: `m_rcp_area` is `1.0 / area`, so dividing by `m_rcp_area` is equivalent to multiplying by the area.

src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
```diff
@@ -606,31 +606,28 @@ void DirectLightingIntegrator::add_emitting_triangle_sample_contribution(
...
+
+    // Don't use this sample if the approxmiate contribution is low.
```
dictoon: Typo: *approximate*

src/appleseed/renderer/kernel/lighting/sppm/sppmparameters.cpp
```diff
@@ -109,6 +109,7 @@ SPPMParameters::SPPMParameters(const ParamArray& params)
   , m_alpha(params.get_optional<float>("alpha", 0.7f))
   , m_max_photons_per_estimate(params.get_optional<size_t>("max_photons_per_estimate", 100))
   , m_dl_light_sample_count(params.get_optional<float>("dl_light_samples", 1.0))
+  , m_dl_low_light_threshold(params.get_optional<float>("dl_low_light_threshold", 0.0))
```
dictoon: Should be `0.0f` since it's a float (in fact, the line just above has the same problem).

src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
aytekaman: Yes, i used rcp because there there was not an area property in the triangle and i didn't want to add one. Should i add it?

dictoon: I think it would make sense to add it yes.

dictoon:
There are a few conflicts that will need to be fixed before we can merge this...

PR #1336, opened 2017-03-20
Implemented ability to open packed projects in studio

glebmish:


src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -1122,7 +1122,7 @@ void MainWindow::slot_open_project()
         get_open_filename(
             this,
             "Open...",
-            "Project Files (*.appleseed);;All Files (*.*)",
+            "Project Files (*.appleseed);;Project Packed Files (*.appleseedz);;All Files (*.*)",
```
dictoon: *Packed Project Files* would probably be better.

glebmish: I'll change it later. Not at home already

dictoon: No worries, I can fix it. Thanks man!

glebmish: Thanks!

PR #1340, opened 2017-03-20
Suppress warning

aytekaman:
Apply a minor fix in order to compile project on VC14 with WX flag.


dictoon:
Squashed and merged, thanks!

aytekaman:
Thanks!

PR #1343, opened 2017-03-21
Usability changes for opening packed projects

glebmish:


src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -1122,7 +1122,7 @@ void MainWindow::slot_open_project()
         get_open_filename(
             this,
             "Open...",
-            "Project Files (*.appleseed);;Packed Project Files (*.appleseedz);;All Files (*.*)",
+            "Project Files (*.appleseed *.appleseedz);;Plain Project Files (*.appleseed);;Packed Project Files (*.appleseedz);;All Files (*.*)",
```
dictoon: You're missing a ; between *.appleseed and *.appleseedz (check how other filters are defined).

src/appleseed/renderer/modeling/project/projectfilereader.cpp
```diff
@@ -3065,6 +3065,9 @@ auto_release_ptr<Project> ProjectFileReader::read(
         }
 
         string unpacked_project_directory(project_filepath);
+
+        size_t ext_start = unpacked_project_directory.rfind(".");
```
dictoon: To be robust, you would need to check if the file has an extension. It's not guaranteed since you allow to open any zip file regardless of its extension.

I would suggest to use Boost with something like

    const string unpacked_project_directory =
        bf::path(project_filepath).stem().string() + ".unpacked";

Issue #1348, opened 2017-03-23
AttributeEditor is not receiving changes to Entity made through EntityEditorWindow


aiivashchenko:
E.g. of steps to reproduce: 

1. Open `EntityEditorWindow` with "Project" > "Output" > "beauty" > right click > "Edit" 
2. Make some changes (e.g. crop window value)
3. Submit with "OK" button

aiivashchenko:
In process

PR #1350, opened 2017-03-23
Update AttributeEditor on changes made with EntityEditorWindow #1348

aiivashchenko:
These are kind of basic changes which should be made to update `AttributeEditor` when changes to entity are made through `EntityEditorWindow`.

Any changes made to entity with `EntityEditor` widgets are actually addressed to `EntityItemBaseSlots::slot_edit_accepted()` method, no matter it they are done with `AttributeEditor` and  `EntityEditorWindow`. This slot can schedule the `EntityEditionAction`, which calles`EntityItem::edit()`, if we are rendering, or execute thismethod on its own. So, `EntityItem::edit()` should be the right place to place the `AttributeEditor::refresh()` call. However `ItemBase` don't know anything about `AttributeEditor` untill we will call `ItemBase::slot_edit(AttributeEditor*)`. Thus, we can save the ptr here (it is done now like that) or pass it on construction of project tree items (i.e. add as arg to `ItemBase` constrictor; not done now since it leads to massive changes).

Other possible improvements which could be made:

1. Add some kind of a bool flag to `ItemBase`, or special slots like `EntityItemBaseSlots::slot_edit_accepted()`, which would help to indicate taht we are editing with `EntityEditorWindow`. As a result we will avoid extra `AttributeEditor::refresh()` calls.
2. Make a special action for `EntityEditorWindow`. Same as previous.
3. Remember item in `ProjectExplorer` on select. Thus, we would be able to `ItemBase::finish_edit()` on next selection to clear `AttributeEditor` ptr.

### DIFF
--- a/src/appleseed.studio/mainwindow/project/entityitem.h
+++ b/src/appleseed.studio/mainwindow/project/entityitem.h
@@ -186,6 +186,9 @@ void EntityItem<Entity, ParentEntity, CollectionItem>::edit(const foundation::Di
     // Move the item to its sorted position.
     if (!m_fixed_position && old_entity_name != new_entity_name)
         move_to_sorted_position(this);
+
+    if (Base::m_attrubute_editor)
+        Base::m_attrubute_editor->refresh();
 }
 
 template <typename Entity, typename ParentEntity, typename CollectionItem>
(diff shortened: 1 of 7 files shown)


aiivashchenko:
@dictoon, thank you for review. I'll investigate on that. Not quite sure that there will be a way to do this without any changes to items at all, since management of both `AttributeEditor` and `EntityEditorWindow` is handled here right now in `slot_edit(...)`, but yeah, signals and slots could probably help us not to store a ptr to `AttributeEditor`.

dictoon:
Closing this until we figure out the right way to link and synchronize multiple entity editors.

Issue #1324, opened 2017-03-16
Invalid samples in sss_sets_comparison.appleseed test scene 


est77:
Does not seem related to SSS sets. Disabling them still generates invalid samples.

dictoon:
Do you mean that rendering them lead to NaN/INF pixel samples?

est77:
There's something in this scene that produces NaN / INF pixel samples, yes.
I played a bit with it and it seem the origin is subsurface scattering.
Disabling the SSS sets made no difference.

Biart95:
The bug is located in the following fragment of code in separablebssrdf.cpp:
```c++
    // Compute the PDF of this incoming point.
    const float dot_nn = static_cast<float>(
        abs(dot(projection_basis.get_normal(), incoming_point.get_shading_normal())));
    incoming_point_prob = projection_axis_prob * disk_point_prob * dot_nn;

    // Weight the sample contribution with multiple importance sampling.
    const float mis_weight =
        compute_mis_weight(
            bssrdf,
            bssrdf_data,
            projection_basis,
            projection_axis,
            incoming_point_prob,
            outgoing_point.get_point(),
            incoming_point.get_point(),
            incoming_point.get_shading_normal());

    // Multiplying the contribution by mis_weight is equivalent to dividing the probability by it.
    incoming_point_prob /= mis_weight;
```
The bug occurs when the sampled axis in `outgoing_point` and the normal of the surface in `incoming_point` are perpendicular. In this case both `mis_weight` and `dot_nn` are equal to zero.

Biart95:
I suggest to add a threshold that would discard such a "bad" incoming points during the tracing of a probe ray. Since the bug essentially comes out of the fact that the probability of the intersection must be 0, it is rather safe merely to remove this kind of intersections.

I will test this approach on many different scenes and see the result.

dictoon:
I don't remember if we discussed that on Slack, but yes, ignoring these incoming points is the right approach.

There used to be a test like the one you describe but it appears to have been lost in the refactorings.

Here it was: https://github.com/appleseedhq/appleseed/blob/4e672d5b98c472a7d247794dc159f8fe79ac916f/src/appleseed/renderer/kernel/lighting/subsurfacesampler.h#L219

dictoon:
@Biart95 @est77: Can we close this issue?

Biart95:
[...]

PR #1351, opened 2017-03-23
Fix NaN values during BSSRDF sampling

Biart95:
This is the fix related to Issue #1324 .

### DIFF
--- a/src/appleseed/renderer/modeling/bssrdf/separablebssrdf.cpp
+++ b/src/appleseed/renderer/modeling/bssrdf/separablebssrdf.cpp
@@ -255,9 +255,17 @@ namespace
             const bool same_sss_set =
                 incoming_point.get_object_instance().is_in_same_sss_set(outgoing_object_instance);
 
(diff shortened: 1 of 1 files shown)


src/appleseed/renderer/modeling/bssrdf/separablebssrdf.cpp
```diff
@@ -255,9 +255,17 @@ namespace
             const bool same_sss_set =
                 incoming_point.get_object_instance().is_in_same_sss_set(outgoing_object_instance);
 
+            const float dot_nn = static_cast<float>(
+                abs(dot(projection_basis.get_normal(), incoming_point.get_shading_normal())));
+            const float dot_nn_threshold = 1e-06f;
```
dictoon: We have a different naming convention for constants: `DotNnThreshold`

dictoon:
Looking good, thanks! You can address the minor comment I made in a later PR, it's a detail.

Issue #1347, opened 2017-03-23
Non-robust BeckmannMDF sampling leads to NaN results on some machines


Biart95:
## Description

Appleseed build on my machine (Windows 7, Visual Studio 2013, x64) performs badly on **all** scenes with Beckmann, giving a lot of NaN results.

The problem is in microfacet.cpp, in `sample_visible_normals` and in `BeckmannMDF::sample11`.

When `alpha_x` and `alpha_y` are small, `cos_theta` can become 1.0 due to the rounding errors in the following line:
[https://github.com/appleseedhq/appleseed/blob/master/src/appleseed/foundation/math/microfacet.cpp#L140](url) 
The code clearly checks this case while calculating `phi` but for some reason do not check it for `cos_theta`, which in its turn leads to misbehaviour in `BeckmannMDF::sample11`, in the following line:
[https://github.com/appleseedhq/appleseed/blob/master/src/appleseed/foundation/math/microfacet.cpp#L379](url) 
Here `cot_theta` is divided by zero and the bug happens.

## Corresponding chunks of code with comments

In `sample_visible_normals`:
``` c++
    Vector3f stretched(
        sign_cos_vn * v[0] * alpha_x,   //
        sign_cos_vn * v[1],             // <--- alpha_x and alpha_y are small
        sign_cos_vn * v[2] * alpha_y);  //

    stretched = normalize(stretched); // <--- stretched[1] becomes 1.0 due to the rounding errors

    const float cos_theta = stretched[1]; // <--- cos_theta also becomes 1.0, checks provided only for phi
    const float phi =
        stretched[1] < 0.99999f ? atan2(stretched[2], stretched[0]) : 0.0f; 

    // Sample slope.
    Vector2f slope = mdf.sample11(cos_theta, s, gamma); // <--- returns #IND
```

in `BeckmannMDF::sample11`:
``` c++
    const float ct = max(cos_theta, 1.0e-6f);             // <--- 1.0
    const float tan_theta = sqrt(1.0f - square(ct)) / ct; // <--- 0.0
    const float cot_theta = 1.0f / tan_theta;             // <--- INF

    const float c = erf(cot_theta);
```

Biart95:
Clamping cos_theta between `1e-06` and `1.0 - 1e-06` resolves the Issue.

dictoon:
Thanks for the detailed investigation!

@est77 What do you think of the fix proposed by @Biart95?

dictoon:
[...]

PR #1352, opened 2017-03-23
Fix NaNs during Beckmann MDF sampling

Biart95:
This is the fix related to Issue #1347 .

When cos_theta = 1 (special case):
1. tan_theta = 0
2. cot_theta = INF
3. c = Erf(INF) = 1
4. K = 0
5. y_approx = y_exact = 2*s[0]
6. b = y_approx - 1 = 2*s[0] - 1
7. slope[0] = Erf_inv(b) = Erf_inv(2*s[0] - 1)

### DIFF
--- a/src/appleseed/foundation/math/microfacet.cpp
+++ b/src/appleseed/foundation/math/microfacet.cpp
@@ -368,28 +368,42 @@ Vector3f BeckmannMDF::sample(
             gamma);
 }
 
+
 // This code comes from OpenShadingLanguage test render.
 Vector2f BeckmannMDF::sample11(
     const float         cos_theta,
     const Vector3f&     s,
     const float         gamma) const
 {
-    const float ct = max(cos_theta, 1.0e-6f);
+    const float threshold = 1e-06f;
+    Vector2f slope;
+
+    // Sample slope Y.
+    slope[1] = erf_inv(2.0f * s[1] - 1.0f);
+
+    // Sample slope X:
+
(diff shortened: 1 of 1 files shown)


src/appleseed/foundation/math/microfacet.cpp
```diff
@@ -368,38 +368,50 @@ Vector3f BeckmannMDF::sample(
             gamma);
 }
 
+
```
dictoon: Remove superfluous blank line.

src/appleseed/foundation/math/microfacet.cpp
```diff
@@ -368,38 +368,50 @@ Vector3f BeckmannMDF::sample(
...
 // This code comes from OpenShadingLanguage test render.
 Vector2f BeckmannMDF::sample11(
     const float         cos_theta,
     const Vector3f&     s,
     const float         gamma) const
 {
-    const float ct = max(cos_theta, 1.0e-6f);
+    const float threshold = 1e-06f;
```
dictoon: It's a constant so it should be spelled `Threshold`.

Sometimes, it doesn't make sense to declare a constant for the threshold/epsilon value. A good hint is in the name of the constant: does `Threshold` tell anything to the reader as to how this value should be tuned and what effect it will have?

I suspect that here you would be better off with using `1.0e-6f` directly where you need it since it's not really something that needs adjusting and it would be hard to give it a more descriptive name.

PR #1354, opened 2017-03-24
Improve handling of geometric & shading normals

dictoon:
- Rely on the geometric normal for side determination when vertex normals are absent (we used to default to front side in that case)
- Allow to flip normals of a given object instance: <parameter name="flip_normals" value="true" /> (default is false obviously)
- Simplify computation of geometric and shading normals
- Formalize our CCW=front convention with a new renderer::compute_triangle_normal() free function
- The osl/04 test scene was actually not rendering correctly; it now does. Improve the scene to make it more obvious that things work properly now.


est77:
In this area, we might want to handle the case when object or assembly transforms flip the handedness of the coordinate system. We already compute the flag for OSL. 

We can check it in a different PR.

dictoon:
Sure, but I'm not 100% clear what needs to be done.

PR #1361, opened 2017-03-27
Fix potential wide gamut negative values

luisbarrancos:


luisbarrancos:
As explained earlier, adding this reference: https://github.com/colour-science/colour/issues/257

Issue #1359, opened 2017-03-25
Allow filtering unit tests by a regex in appleseed.studio


dictoon:

htmlboss:
I've made some progress here but it's a little more complicated than I originally thought. The QTreeWidget that displays all the unit tests cannot be filtered it seems so I've changed it to a QTreeView to allow a custom model. Right now it is repopulated correctly with the parent-children relationship.

Currently I am working on hooking up the checkboxes again and THEN implementing the filtering (which I'm designing to be real-time so the list updates as you change the filter text).

htmlboss:
Just submitted a PR implementing this.

dictoon:
Done.

PR #1364, opened 2017-03-30
Added filtering to appleseed.studio unit tests (Issue #1359)

htmlboss:
Primary changes:
- Unit tests in appleseed.studio are filtered by regex in realtime (new feature).
- const-corrected TestWindow class.
- Removed unused header files in TestWindow.h/.cpp.
- Various minor fixes to code layout to keep things consistent.

Screenshot of relevant UI changes (I tried to follow the same layout from the main window):

ui


src/appleseed.studio/debug/tests/testwindow.cpp
```diff
@@ -365,19 +374,56 @@ namespace
...
+{
+    bool do_filter_items(QTreeWidgetItem* item, const QRegExp& regexp)
+    {
+        bool any_children_visible = false;
+
+        for (int i = 0; i < item->childCount(); ++i)
+        {
+            if (do_filter_items(item->child(i), regexp))
+            {
+                any_children_visible = true;
```
dictoon: You can add a `break` here to exit the loop as soon as one visible child is found.

src/appleseed.studio/debug/tests/testwindow.cpp
```diff
@@ -307,7 +314,9 @@ namespace
         {
             QTreeWidgetItem* item = parent->child(i);
             if (item->checkState(column) == state)
+            {
```
dictoon: We're actually fine with omitting the curly braces for single line blocks.

dictoon:
Thanks for the super nice PR! The code is extremely clear and I have literally nothing to complain about in this area.

A few remarks on the UI/UX front:

* For the sake of consistency with the project explorer and the color/texture picker (two places where filtering is available), I would suggest placing the textbox and button _above_ the unit test list rather than below.

* Another thing to consider is whether we should add a way to check/uncheck the tests that are shown in the list _after filtering_. What do you think?

htmlboss:
> Another thing to consider is whether we should add a way to check/uncheck the tests that are shown in the list after filtering. What do you think?

In terms of convenience I really like that. Are you suggesting adding buttons to select/de-select all items presently shown in the list (as opposed to toggling all loaded tests)? It would be very easy to implement, but wouldn't that clutter the UI though?

dictoon:
Maybe we should actually modify the existing buttons to operate on the shown tests instead of the whole test suite?

dictoon:
Merged, thank you!

PR #1371, opened 2017-04-02
Branchless ortho-basis construction algorithm

aytekaman:
Reference: http://jcgt.org/published/0006/01/01/

### DIFF
--- a/src/appleseed/foundation/math/basis.h
+++ b/src/appleseed/foundation/math/basis.h
@@ -178,57 +178,23 @@ inline void Basis3<T>::build(const VectorType& normal)
     //
     // Reference:
     //
-    //   Hughes, J. F., and Moller, T. Building an Orthonormal Basis from a Unit Vector.
-    //   Journal of Graphics Tools 4, 4 (1999), 33-35.
-    //   http://www.cs.brown.edu/research/pubs/pdfs/1999/Hughes-1999-BAO.pdf
+    //   Tom Duff, James Burgess, Per Christensen, Christophe Hery, Andrew Kensler, 
+    //   Max Liani, and Ryusuke Villemin, Building an Orthonormal Basis, Revisited, 
+    //   Journal of Computer Graphics Techniques (JCGT), vol. 6, no. 1, 1-8, 2017
+    //   http://jcgt.org/published/0006/01/01/paper-lowres.pdf
     //
 
     assert(is_normalized(normal));
 
-    // n is simply the input vector.
     m_n = normal;
 
-    // Compute u so that it is orthogonal to n.
-    if (std::abs(m_n[0]) < std::abs(m_n[1]))
-    {
-        if (std::abs(m_n[0]) < std::abs(m_n[2]))
-        {
-            // m_n[0] is the smallest component.
-            m_u[0] =  T(0.0);
-            m_u[1] = -m_n[2];
-            m_u[2] =  m_n[1];
-        }
-        else
-        {
-            // m_n[2] is the smallest component.
-            m_u[0] = -m_n[1];
-            m_u[1] =  m_n[0];
-            m_u[2] =  T(0.0);
-        }
-    }
-    else
-    {
-        if (std::abs(m_n[1]) < std::abs(m_n[2]))
-        {
-            // m_n[1] is the smallest component.
-            m_u[0] = -m_n[2];
-            m_u[1] =  T(0.0);
-            m_u[2] =  m_n[0];
-        }
-        else
-        {
-            // m_n[2] is the smallest component.
-            m_u[0] = -m_n[1];
-            m_u[1] =  m_n[0];
-            m_u[2] =  T(0.0);
-        }
-    }
+    T sign = copysignf(1.0f, m_n[2]);
 
-    // u is orthogonal to n, but not unit-length. Normalize it.
-    m_u = normalize(m_u);
+    const T a = T(-1.0) / (sign + m_n[2]);
+    const T b = m_n[0] * m_n[1] * a;
 
-    // Compute v.
-    m_v = cross(m_u, m_n);
+    m_u = VectorType(T(1.0) + sign * m_n[0] * m_n[0] * a, sign * b, -sign * m_n[0]);
+    m_v = VectorType(b, sign + m_n[1] * m_n[1] * a, -m_n[1]);
 
 #ifndef NDEBUG
     checks();
--- a/src/appleseed/foundation/math/basis.h
+++ b/src/appleseed/foundation/math/basis.h
@@ -188,13 +188,13 @@ inline void Basis3<T>::build(const VectorType& normal)
 
     m_n = normal;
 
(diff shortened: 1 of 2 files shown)


src/appleseed/foundation/math/basis.h
```diff
@@ -178,57 +178,23 @@ inline void Basis3<T>::build(const VectorType& normal)
...
-            m_u[2] =  T(0.0);
-        }
-    }
-
-    // u is orthogonal to n, but not unit-length. Normalize it.
-    m_u = normalize(m_u);
-
-    // Compute v.
-    m_v = cross(m_u, m_n);
+    T sign = copysignf(1.0f, m_n[2]);
```
dictoon: Should be `T(1.0)`.

dictoon:
Could you add unit benchmarks for this, in single and double precision, comparing with the old method (which you can reproduce in the unit benchmark file)?

src/appleseed/foundation/math/basis.h
est77: and maybe copysign?

dictoon:
It looks like some unit tests fail because some of the assertions checking the orthonormal basis fail. It would be interesting to know how precise or imprecise the new method is.

src/appleseed/foundation/math/basis.h
dictoon: Right!

aytekaman: Thanks! will correct them.

est77: I guess we'd can wrap copysign and copysignf into a overloaded function and then the compiler will do the right thing.

dictoon: We don't need to, `std::copysign()` is already a bunch of overloads.

The main problem is that it's C++11 only.

dictoon: Let's try without copysign for now?

aytekaman:
I just saw the unit test results. I will investigate this further.

aytekaman:
İ just swapped n_u, n_v. Now unit tests don't fail. Maybe their coordinate system assumptions are different?

dictoon:
Interesting. Our coordinate system is right-handed, with U pointing right, N pointing up and V pointing away from the screen, toward the viewer.

src/appleseed/foundation/math/basis.h
aytekaman: you mean i should use `copysignf` or the branchless version in the paper which doesn't have `copysign` at all?

dictoon: I would suggest doing without `copysign` at all for now.

dictoon: So,

    const T sign = m_n[2] < 0.0f ? T(-1.0) : T(1.0);

Also, could you please add const qualifiers?

aytekaman:
Now I will add unit tests.

dictoon:
Awesome!

aytekaman:
Here are the results for convenience:
-------------------------------------------------------------------------------------------------
 BranchlessBuild_SinglePrecision:   2,148,606 clock ticks (1.070K calls/s at 2299.415 MHz)
 OriginalBuild_SinglePrecision:       6,410,783 clock ticks (377.4 calls/s at 2419.280 MHz)
-------------------------------------------------------------------------------------------------
 BranchlessBuild_DoublePrecision: 3,963,067 clock ticks (602.1 calls/s at 2386.097 MHz)
 OriginalBuild_DoublePrecision:     6,577,000 clock ticks (380.7 calls/s at 2503.997 MHz)
-------------------------------------------------------------------------------------------------

aytekaman:
Here are the results from the _release_ build for convenience:

```
BranchlessBuild SinglePrecision: 1,942,049 clock ticks (1.194K calls/s at 2318.736 MHz)
OriginalBuild   SinglePrecision: 6,370,353 clock ticks (386.4 calls/s at 2461.768 MHz)

BranchlessBuild DoublePrecision: 4,019,368 clock ticks (581.1 calls/s at 2335.743 MHz)
OriginalBuild   DoublePrecision: 6,548,406 clock ticks (378.6 calls/s at 2479.023 MHz)
```

src/appleseed/foundation/meta/benchmarks/benchmark_basis.cpp
```diff
@@ -0,0 +1,133 @@
+
+//
+// This source file is part of appleseed.
+// Visit http://appleseedhq.net/ for additional information and resources.
+//
+// This software is released under the MIT license.
+//
+// Copyright (c) 2010-2013 Francois Beaune, Jupiter Jazz Limited
```
dictoon: If this is a new file (it appears so), you should replace these two lines by a single one with your name:

    // Copyright (c) 2017 Aytek Aman, The appleseedhq Organization

The rules are explained here: https://github.com/appleseedhq/appleseed/wiki/Copyright-Notices

src/appleseed/foundation/meta/benchmarks/benchmark_basis.cpp
```diff
@@ -0,0 +1,133 @@
...
+        Vector<T, 3>  m_normal[N];
+        Basis3<T>     m_result[N];
+
+        BuildFixture()
+        {
+            MersenneTwister rng;
+
+            for (size_t i = 0; i < N; ++i)
+            {
+                m_normal[i] = rand_vector1<Vector<T, 3>>(rng);
```
dictoon: `rand_vector1<Vector<T, 3>>` is only allowed in C++11 and later. In C++03 and earlier, you need to write

    rand_vector1<Vector<T, 3> >

src/appleseed/foundation/meta/benchmarks/benchmark_basis.cpp
```diff
@@ -0,0 +1,133 @@
...
+                m_normal[i] = normalize(m_normal[i]);
+            }
+        }
+
+        Basis3<T> build_Original(const Vector<T, 3> normal) const
```
dictoon: `build_original` (method names are lower case).

dictoon: Can this method be made static instead of merely const?

dictoon:
Very nice work, great attention to details, thanks!

Did the performance change after your last commit ("Make comparison more fair")?

dictoon:
(Make sure to show the outdated comments.)

aytekaman:
I forgot to mention that I had updated the results with the newest commit. Before that, branchless code was faster (ex: Single precision branchless version was around 2,2 M clock ticks)

Maybe we can keep the old method in the Basis implementation to make the comparison more reliable. What do you think?

dictoon:
> Maybe we can keep the old method in the Basis implementation to make the comparison more reliable. What do you think?

What you can do is try move it back there in your local repo, and check whether it makes a noticeable performance difference. I suspect it won't, in which case it's fine to keep the old method in the unit benchmark file.

PR #1373, opened 2017-04-02
Change delimiter ';' on ' ' in QT file dialog filters

glebmish:
With ';' wildcards of extensions are not recognized on some systems.
For example with filter equals (*.png;*.exr) not png not exr files are shown.
On the other hand, ' ' delimiter works well on any system and also advised in documentation.


dictoon:
Thanks!

PR #1378, opened 2017-04-06
Bug fixes and tests

glebmish:


src/appleseed.studio/utility/miscellaneous.cpp
```diff
@@ -187,13 +187,18 @@ QString get_extension(ParamArray& settings, const QString& target_dialog)
...
+    if (ext_start == -1 || ext_end == -1 || ext_start > ext_end)
+        return "";
+
+    QStringList extensions = filter.mid(ext_start, ext_end - ext_start).split(' ');
+
+    if (extensions.length() == 1 && extensions[0] != "*.*")
+        return extensions[0].mid(1);
     else
-        return QString("");
+        return ".appleseed";
```
dictoon: `get_extension()` appears to be a fairly generic function so this "default" extension should be passed in argument.

dictoon: I think we might have discussed this some days or weeks ago, but I'm actually a bit surprised that we need to do this at all.

Do you confirm that the file dialog does not append a file extension automatically?

What happens in other situations (such as saving a render as an image file on disk) where we don't do anything special, is an extension appended if it's missing?

src/appleseed/foundation/meta/tests/test_zip.cpp
```diff
@@ -81,6 +81,40 @@ TEST_SUITE(Foundation_Utility_Zip)
...
+        try
+        {
+            ASSERT_TRUE(bf::exists(InitialDirectory));
+            ASSERT_FALSE(bf::exists(TargetZip));
+            ASSERT_FALSE(bf::exists(TargetDirectory));
+
+            zip(TargetZip, InitialDirectory);
+            unzip(TargetZip, TargetDirectory);
+
+            const set<string> ExpectedFiles = recursive_ls(InitialDirectory);
```
dictoon: Use lower case for variables: `expected_files`, `actual_files`.

src/appleseed/foundation/meta/tests/test_zip.cpp
```diff
@@ -81,6 +81,40 @@ TEST_SUITE(Foundation_Utility_Zip)
         }
     }
 
+    TEST_CASE(ZipUnzip_UnzipsCorrectDirectory)
```
dictoon: Maybe `ZipUnzipRoundtrip`?

The code looks really good in any case!

src/appleseed/renderer/meta/tests/test_projectfilewriter.cpp
```diff
@@ -287,4 +287,22 @@ TEST_SUITE(Renderer_Modeling_Project_ProjectFileWriter)
...
+        const bool success =
+            ProjectFileWriter::write(
+                m_project.ref(),
+                (m_output_directory / PackedProjectName).string().c_str(),
+                ProjectFileWriter::Defaults
+            );
+
+        ASSERT_TRUE(success);
+        ASSERT_TRUE(exists(m_output_directory / PackedProjectName));
+        ASSERT_FALSE(exists((m_output_directory / PackedProjectName).replace_extension(".unpacked")));
```
dictoon: Use the actual file/directory name in

    exists((m_output_directory / PackedProjectName).replace_extension(".unpacked"))

instead of using `replace_extension()`:

    exists(m_output_directory / "test_pack_valid_project.unpacked")

It's best to avoid as much logic as possible in unit tests, especially in assertions.

src/appleseed.studio/utility/miscellaneous.cpp
glebmish: `Do you confirm that the file dialog does not append a file extension automatically?`
Yes, I'm totally sure about it. It was previously added in slot_save_project_as() inside mainwindow.cpp
As it was only one extension, there was no need to know which extension is chosen in filters. When this need appeared, I had to wrote function to manually parse filter string to get extension
As this get_extension() function used only in the code I changed/wrote, in other case everything works like before

src/appleseed/foundation/meta/tests/test_zip.cpp
glebmish: Previously you said that constants should be LikeThat. Why it's not the case here?

dictoon: Because they're not constants? :)

dictoon: A constant is really just that, a plain value that has a name. Anything else is a variable, even if it's prefixed by `const` (which just means the variable is immutable).

glebmish: still not clear enough. I declare it as const set and it's not constant? Why?

glebmish: ah, ok!

src/appleseed/foundation/meta/tests/test_zip.cpp
```diff
@@ -96,13 +96,13 @@ TEST_SUITE(Foundation_Utility_Zip)
             zip(TargetZip, InitialDirectory);
             unzip(TargetZip, TargetDirectory);
 
-            const set<string> ExpectedFiles = recursive_ls(InitialDirectory);
-            const set<string> ActualFiles = recursive_ls(TargetDirectory);
+            const set<string> expected_files = recursive_ls(InitialDirectory);
+            const set<string> actual_files = recursive_ls(TargetDirectory);
 
-            ASSERT_EQ(ActualFiles.size(), ExpectedFiles.size());
+            ASSERT_EQ(actual_files.size(), expected_files.size());
```
dictoon: The arguments should be swapped (the expected value must come first). It's important for error messages, if/when the assertion fails.

dictoon:
Looks good, thanks!

PR #1380, opened 2017-04-08
Alpha map optimization

aytekaman:
`has_uniform_opacity()` returns true when `alpha_map = 1.0`. Not sure about the naming of this method. Also, it can be broken into two parts such as `has_uniform_alpha()` and `get_uniform_alpha_value()`.


src/appleseed/renderer/modeling/material/material.cpp
```diff
@@ -94,6 +94,20 @@ bool Material::has_alpha_map() const
     return get_non_empty(m_params, "alpha_map") != 0;
 }
 
+bool Material::has_uniform_opacity() const
```
est77: I think we can rename this method. It's actually returning true if the material has an uniform alpha map with value 1.
I'd also not use the word opacity.

src/appleseed/renderer/modeling/scene/objectinstance.cpp
```diff
@@ -83,7 +83,7 @@ bool uses_alpha_mapping(const MaterialArray& materials)
     {
         if (materials[i])
         {
-            if (materials[i]->has_alpha_map())
+            if (materials[i]->has_alpha_map() && !materials[i]->has_uniform_opacity())
                 return true;
 
             if (const ShaderGroup* sg = materials[i]->get_uncached_osl_surface())
```
est77: Can we move this test just before if (materials[i]->has_alpha_map() ...) to avoid any issue with OSL materials?

src/appleseed/renderer/modeling/material/material.cpp
aytekaman: I thought a lot about naming this correctly. How about these:
`is_uniformly_opaque()` or a more expressive one `has_uniform_alpha_map_value_of_one()` ?
I am open to suggestions here:)

est77: I like the last suggestion. It's very clear what it does.

dictoon: I liked the first suggestion, `is_uniformly_opaque()`: shorter, consistent with other methods, and IMHO descriptive enough... Oh well :)

PR #1389, opened 2017-04-16
Misc work

est77:
- alSurface improvements.
- added alFlowNoise shader.
- removed unused headers in closures.h.
- Fixed wrong warning when using object alpha maps.


src/appleseed/renderer/modeling/scene/objectinstance.cpp
```diff
@@ -459,7 +459,11 @@ bool ObjectInstance::on_frame_begin(
 
     const EntityDefMessageContext context("object instance", this);
 
-    if (uses_alpha_mapping())
+    const bool uses_materials_alpha_mapping =
```
dictoon: What was the problem exactly?

est77: When you use an object alpha map but you have different front and back materials you get this warning. I think that in this case, the result is correct (the object alpha map is used in both sides) 
and the warning is wrong.

src/appleseed.shaders/include/alshaders/noise.h
```diff
@@ -0,0 +1,367 @@
+
```
luisbarrancos: @est77 i'll be moving this to the AS noise helpers headers since it'll be used by the asNoise custom node as well (for the crocodile noise, advection noise, etc).

est77: Ok. I didn't know if we had more uses of it or not. That's why I put it in its own header,

PR #1391, opened 2017-04-16
misc changes

est77:
- Fix OSL EDF get_max_contribution method.
- alSurface transparency related fixes.
- Remove if (color) {...} from shaders. Fixes wrong transparency closure use detection.

### DIFF
--- a/src/appleseed.shaders/src/alshaders/alSurface.osl
+++ b/src/appleseed.shaders/src/alshaders/alSurface.osl
@@ -293,11 +293,14 @@ shader alSurface
     int do_glossy2 = 1;
     int do_glossy = 1;
 
-    if (raytype("transparency"))
+    if (raytype("transparency") || raytype("shadow"))
     {
-        out_result = (color(1) - in_opacity) * transparent();
-        return;
+        if (min(in_opacity) < 1.0)
+            out_result = (color(1) - in_opacity) * transparent();
     }
+
+    if (raytype("transparency"))
+        return;
 
     if (raytype("shadow"))
     {
--- a/src/appleseed.shaders/src/maya/as_maya_anisotropic.osl
+++ b/src/appleseed.shaders/src/maya/as_maya_anisotropic.osl
@@ -269,7 +269,7 @@ shader as_maya_anisotropic
                 );
         }
 
-        if (transparency)
+        if (max(transparency) > 0.0)
         {
             out_outColor = transparency * transparent();
         }
@@ -314,7 +314,7 @@ shader as_maya_anisotropic
 
             opacity = 1.0;
         }
-        else if (in_transparency)
+        else if (max(in_transparency) > 0.0)
         {
             transparency_term = in_transparency * transparent();
         }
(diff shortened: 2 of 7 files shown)


src/appleseed.shaders/src/alshaders/alSurface.osl
```diff
@@ -293,12 +293,15 @@ shader alSurface
     int do_glossy2 = 1;
     int do_glossy = 1;
 
-    if (raytype("transparency"))
+    if (raytype("transparency") || raytype("shadow"))
     {
-        out_result = (color(1) - in_opacity) * transparent();
-        return;
+        if (min(in_opacity) < 1.0)
```
luisbarrancos: @est77 why the usage of min()? OSL inequality test will test component by component, so a != 1 would suffice, since there is no opacity value above 1.0, any != 1.0 value would be < 1.0 necessarily.

src/appleseed.shaders/src/maya/as_maya_anisotropic.osl
```diff
@@ -269,7 +269,7 @@ shader as_maya_anisotropic
                 );
         }
 
-        if (transparency)
+        if (max(transparency) > 0.0)
```
luisbarrancos: @est77 same applies, why max() ? The inequality test will suffice since transparency is in [0,1] range, any component != 0 would trigger the clause.

est77: There are two reasons. The most important is that maybe this is triggering a bug in OSL.
I was having false positives regarding the transparent() closure usage when using if (color) {}.
The second one is a style issue. if (color) {} implies a conversion from a color to a boolean and 
I think that's not a natural conversion. I find it strange.

dictoon: While technically correct, `max()` is (in theory) more computationally intensive than necessary; `any(color) > 0.0` would have been enough.

This is purely a theoretical remark since of course the max of three values can be computed in 1 cycle or less using SSE, and it's likely that the OSL compiler does just that...

est77: All this is constant folded by OSL in most cases. There's usually no runtime cost.

dictoon: Yes, _if_ transparency is a constant.

luisbarrancos: @est77 i agree with if (color), we talked about this last week. i just had no time yet to replace all those tests, but to recap, i'll leave if (foo) only and only for boolean types, which in OSL means ints that can only take 0 or 1 really since we have no bool.
We cannot test if (color > 0), OSL fails with this expression, but we can test if (color != 0), that was my suggestion.

est77: Yes, in the case we want to optimize transparency is constant, equal to 0. If OSL can remove the test and the closure assignment, the closure does not appear in the used closures list and 
we know we can skip shader exec. for shadows and transparency.

I don't mind if (color != 0) if OSL can constant fold it correctly. That's something we need to test.

luisbarrancos: @est77 ok, i'll test with isconstant() to verify this later, no hurries.

PR #1392, opened 2017-04-17
Remove file shader sequence options.

est77:
This is done now at export time by the Maya plugin using Maya's API.
https://github.com/appleseedhq/appleseed-maya/commit/0755ab2227500e12dc1ef293165702694be794f9

### DIFF
--- a/src/appleseed.shaders/src/maya/as_maya_file.osl
+++ b/src/appleseed.shaders/src/maya/as_maya_file.osl
@@ -49,26 +49,6 @@ shader as_maya_file
         string label = "Disable File Load",
         string widget = "checkBox"
     ]],
-    int in_useFrameExtension = 0
-    [[
-        string as_maya_attribute_name = "useFrameExtension",
-        string label = "Use Frame Extension",
-        string widget = "checkBox"
-    ]],
-    int in_frameExtension = 1
-    [[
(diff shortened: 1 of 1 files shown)


src/appleseed.shaders/src/maya/as_maya_file.osl
```diff
@@ -49,26 +49,6 @@ shader as_maya_file
...
-        string widget = "number",
-        int slider = 0
-    ]],
-    int in_frameOffset = 0
-    [[
-        string as_maya_attribute_name = "frameOffset",
-        string label = "Frame Offset",
-        string widget = "number",
-        int slider = 0
-    ]],
```
luisbarrancos: @est77 only for the file node? Because there are 2 more file related nodes. The "movie" node needs the frame extension and offset, since that is used by OIIO to search a frame in a movie sequence (via FFMPEG, and subimage token), so this should not apply to the "movie" node.
The PSD texture node probably is going to require something similar to access multichannel images, though in practice not many people would (hopefully) use it.

est77: This is only for the file node. Does not apply to movie or psd.
I have to check in the Maya API if there's something for the psd and movie files.

PR #1396, opened 2017-04-22
Implement using of packed projects as archive projects

glebmish:


src/appleseed/renderer/modeling/project/assethandler.cpp
```diff
@@ -208,6 +208,13 @@ bool AssetHandler::handle_asset(string& asset_path) const
     APIString search_path;
     m_project.search_paths().qualify(asset_path, &qualified_asset_path, &search_path);
 
+    // Relative asset path, that does not exist. Handling is skipped.
+    // Means that this file is in unprovided search path (therefore should be provided by parent project)
+    if (search_path.empty() && !exists(qualified_asset_path.c_str()))
```
dictoon: Is the `search_path.empty()` condition really necessary here?

src/appleseed/renderer/modeling/project/projectfilereader.cpp
```diff
@@ -3153,6 +3153,28 @@ namespace
...
+    // returns filename of .appleseed file inside archive
+    // returns "" in case there are more than 1 .appleseed files
+    // "" means it's not a valid packed project
+    string get_project_name(const char* project_filepath)
+    {
+        const vector<string> appleseed_files = get_filenames_with_extension_from_zip(project_filepath, ".appleseed");
+        return appleseed_files.size() == 1 ? appleseed_files[0] : "";
+    }
+
+    string get_unpacked_project(
```
dictoon: I suggest calling this method simply `unpack_project()` since it's what it does really.

I believe this code can also be factored into `unpack_project()`:

    const string unpacked_project_directory =
        bf::path(project_filepath).replace_extension(".unpacked").string();

src/appleseed/renderer/modeling/project/projectfilereader.cpp
```diff
@@ -3171,9 +3193,8 @@ auto_release_ptr<Project> ProjectFileReader::read(
     string actual_project_filepath;
     if (is_zip_file(project_filepath))
     {
-        const vector<string> appleseed_files = get_filenames_with_extension_from_zip(project_filepath, ".appleseed");
-
-        if (appleseed_files.size() != 1)
+        string project_name = get_project_name(project_filepath);
```
dictoon: `project_filename` seems more appropriate?

dictoon: Can this variable be made const?

src/appleseed/renderer/modeling/project/projectfilereader.cpp
```diff
@@ -3153,6 +3153,28 @@ namespace
 
         return false;
     }
+
+    // returns filename of .appleseed file inside archive
+    // returns "" in case there are more than 1 .appleseed files
+    // "" means it's not a valid packed project
+    string get_project_name(const char* project_filepath)
```
dictoon: `get_project_filename_from_archive()` maybe? `get_project_name()` is a bit too generic...

src/appleseed/renderer/modeling/project/assethandler.cpp
glebmish: I think yes.  
Comment says: Relative asset path, that does not exist.
search_path.empty() checks relativness, !exists(qualified_asset_path.c_str()) checks existence

glebmish: if asset with global path does not exist, it's still an error

dictoon: Not really: you know that the asset path is relative because the case where it is absolute is handled at the beginning of the method.

Checking `search_path.empty()` means checking that there are no search paths at all for this project.

src/appleseed/renderer/modeling/project/projectfilereader.cpp
glebmish: `I believe this code can also be factored into unpack_project():`

unpacked_project_directory is used before unpack_project is called in logs
Also since it's used outside of unpack_project it should be seen there

dictoon: You're right, but maybe the log message itself could be moved to the method, I think it would make sense. In that case it should probably be changed from

    "%s appears to be a packed project; unpacking to %s..."

to something like

    "Unpacking %s to %s..."

src/appleseed/renderer/modeling/project/assethandler.cpp
glebmish: I agree. I was puzzled by `else` at the end of function, but I read comment there thoroughly and now I think you're right

glebmish:
Oh, I finally understood the convention! Thanks!

PR #1409, opened 2017-04-30
Add shader/surface, drawdb node classification

luisbarrancos:
For issue appleseedhq/appleseed-maya#41. This adds back the maya shader/surface classification where needed. Perhaps we can find a workaround later to restrict AS nodes to the AS menus/entries only.


est77:
I have a fix for this too. It's a bit different. Let me push mine instead.

est77:
Replaced by https://github.com/appleseedhq/appleseed/pull/1410

PR #1411, opened 2017-05-01
Bool tests for bool types only

luisbarrancos:
As discussed, bool types with bool tests, int and floats test for values explicitly.


luisbarrancos:
Closing. It was a minor change.

PR #1394, opened 2017-04-21
Parser for for IESNA LM-63 Photometric Data Files

Biart95:

PR #1414, opened 2017-05-05
Parser for for IESNA LM-63 Photometric Data Files (Fixed)

Biart95:


Biart95:
Previous pull request: https://github.com/appleseedhq/appleseed/pull/1394

src/appleseed/foundation/utility/iesparser.cpp
```diff
@@ -415,466 +403,470 @@ void IESParser::parse_tilt_data(std::istream& input_stream)
...
-    return boost::regex_match(line, regex);
+    static const boost::regex Regex(TiltLineRegex);
+    return boost::regex_match(line, Regex);
 }
 
-
 void IESParser::parse_keyword_line(const std::string& line)
 {
-    boost::regex regex(KEYWORD_LINE_REGEX);
+    static const boost::regex Regex(KeywordLineRegex);
```
dictoon: This is subtly dangerous (I think): by marking this variable static, the compiler will ensure that there is a single instance of it shared across all calls, and that it is constructed "some time" before the method is first invoked. The trouble is, there isn't any guarantee that this "lazy construction" will work properly if the method is invoked by multiple threads at the same time. The chance of this failing is rather tiny, but as far as I know, it exists. I would suggest to simply mark the variable as `const`, and remove `static`.

On a sidenote, even though the variable is const, it's not a true constant fully defined at compile-time (e.g. `const int Something = 12`) but rather just a normal variable that's made immutable. As a normal variable, we would spell it in lower case: `regex` (or `re` if you need to avoid conflicts with `boost::regex`).

src/appleseed/foundation/utility/iesparser.h
```diff
@@ -189,128 +188,127 @@ class IESParser
...
-        return keywords_dictionary.at(keyword);
+        return m_keywords_dictionary.at(keyword);
     }
 
     // Options:
-    bool ignore_allowed_keywords;  // if true, all keywors are allowed not depending on format version
-    bool ignore_required_keywords; // if true, some required keywors can be missing
-    bool ignore_empty_lines;       // if true, than the file can contain whitespace lines
-    bool ignore_tilt;              // if true, TILT section is not parsed
+    bool m_ignore_allowed_keywords;  // if true, all keywors are allowed not depending on format version
```
dictoon: Typo: *keywords*

PR #1422, opened 2017-05-19
Removed normalized diffusion cmake build option.

est77:


dictoon:
Thanks! Don't forget to update the build instructions in the wiki.

PR #1429, opened 2017-05-26
Fix relative luminance, LCHab|uv, add XYZ<>UCS

luisbarrancos:


luisbarrancos:
Relatively trivial fix. Though we're preparing for release i would prefer to include this for 1.7.

dictoon:
Thanks! It'll be part of 1.7.

PR #1431, opened 2017-05-27
1.7 release work

est77:


Narann:
Why are the reasons to remove partio? I guess it's because it's not often use while alembic support points (or even OpenVDB points)? I just ask. :)

Keep the good work! :+1:

est77:
> Why are the reasons to remove partio? I guess it's because it's not often use while alembic support points (or even OpenVDB points)? I just ask. :)

We were only using it in an unit test which didn't have much value. We cannot render points or particles at the moment in any case. 
When we add them, Alembic, OpenVDB points or even USD would be better choices.

Note that you can still use PARTIO inside OSL, for point cloud lookups, and similar stuff.

Narann:
Thanks! :)

PR #1442, opened 2017-06-09
More 3D tex nodes, custom AS nodes

luisbarrancos:
Custom 100% colorspace aware Luminance node, ColorTransform node, converts from **any** input space to **any** output space, using CIE XYZ as an intermediary space.
Almost all Maya 3d texture nodes added. Add correlated color temperature and test RNG functions.
Requires https://github.com/appleseedhq/appleseed-maya/pull/49


src/appleseed.shaders/src/maya/as_maya_granite.osl
```diff
@@ -0,0 +1,322 @@
...
+        string label = "Randomness",
+        string page = "Granite Attributes"
+    ]],
+    float in_spottyness = 0.3
+    [[
+        string as_maya_attribute_name = "spottyness",
+        string as_maya_attribute_name_short = "s",
+        float min = 0.0,
+        float softmax = 1.0,
+        string label = "Spottyness",
```
dictoon: Shouldn't it be *Spottiness*?

luisbarrancos: I think so, that the right term is *spottiness*, but i was following Maya's nomenclature, bugs quirks and all :)

dictoon: Oh ok :)

PR #1447, opened 2017-06-11
True branchless orthonormal basis construction

aytekaman:
Now that we are switching to C++ 11, we should be able use copysign safely. However, In OIIO's `missing_math` header, copysign is directed to  the native _copysign function via a macro, which is not properly overloaded. How should we handle this?

### DIFF
--- a/src/appleseed/foundation/math/basis.h
+++ b/src/appleseed/foundation/math/basis.h
@@ -188,7 +188,7 @@ inline void Basis3<T>::build(const VectorType& normal)
 
     m_n = normal;
 
-    const T sign = m_n[2] < T(0.0) ? T(-1.0) : T(1.0);
+    const T sign = copysign(T(1.0), m_n[2]);
 
     const T a = T(-1.0) / (sign + m_n[2]);
     const T b = m_n[0] * m_n[1] * a;


dictoon:
Hey, thanks so much for remembering to update the code!

We have a mechanism in place to cleanup after "unhygienic" libraries: pairs of macros that surround the libraries' header files, defined in [`foundation/platform/oiioheaderguards.h`](https://github.com/appleseedhq/appleseed/blob/master/src/appleseed/foundation/platform/oiioheaderguards.h) in the case of OIIO.

Unfortunately this mechanism won't work in this case because a C/C++ macro cannot `#undef` a symbol.

It seems to me that the simplest option is to simply implement `copysign()` ourselves. It's very simple to implement, and to my surprise it seems that the standard implementation is never inlined, so our implementation would also have a performance advantage.

I've just done it: https://github.com/appleseedhq/appleseed/pull/1448. Review appreciated!

dictoon:
Actually, a closer study of the assembly code generated by gcc 7.1 and clang 4.0.0 indicates that their implementation are probably superior to mine since they use SSE registers only, instead of ping-ponging between integer and floating point registers like my implementation does.

Maybe a better alternative to my PR is to replace OIIO's header guards by a pair of include files that would surround OIIO's header files. It's something I've done in our Maya plugin:
https://github.com/appleseedhq/appleseed-maya/blob/master/src/appleseedmaya/alphamapnode.cpp#L36

dictoon:
Here is a better mechanism that takes care of undefining the `copysign()` macro: https://github.com/appleseedhq/appleseed/pull/1450

Review appreciated!

aytekaman:
Maybe we can merge this now and have a true _branchless_ method:)

dictoon:
Yes! I'm not in front of my computer right now but feel free to merge it!
If you don't have the rights I'll give them to you tomorrow.

PR #1450, opened 2017-06-11
Better handling of unhygienic header files

dictoon:


aytekaman:
Ok, now instead of using macro definitions from one header file per dependency, we need to create begin/end header pairs and include them. This brings an additional flexibility i.e. we can #undef symbols which is what we want. Things look good to me:) Maybe @est77 should take a look too?

dictoon:
Exactly! Thanks for the review. If it looks good to you then it's fine :)

PR #1461, opened 2017-06-17
Fix CIE Lab,Luv,LCh, remapping, glass, see notes

luisbarrancos:
CIE L*a*b* and CIE L*u*v* return values with L in [0,100] range, and a,b
and u,v in [-100,100] range. Color samples however were discarded for
being negative, and it's not really easy to adjust this in a normal
workflow, so CIELAB, CIELUV are mapped to [0,1] overall.
The same applies to CIE LCh, and a minor bug.


luisbarrancos:
@dictoon oops, wrong version of color xform header committed. Fixed in commit dd011f8dc9405770b0515e53563ccc2538a39b8d

PR #1463, opened 2017-06-17
API headers dependencies removal

est77:
Wrap OIIO and OSL classes used in API headers, replace OSL and OIIO includes in public headers by forward declarations.

With these changes, OIIO and OSL headers are not required anymore to build appleseed.maya.

### DIFF
--- a/src/appleseed/renderer/api/rendering.h
+++ b/src/appleseed/renderer/api/rendering.h
@@ -45,7 +45,6 @@
 #include "renderer/kernel/rendering/masterrenderer.h"
 #include "renderer/kernel/rendering/nulltilecallback.h"
 #include "renderer/kernel/rendering/progressive/progressiveframerenderer.h"
-#include "renderer/kernel/rendering/scenepicker.h"
 #include "renderer/kernel/rendering/tilecallbackbase.h"
 #include "renderer/kernel/rendering/timedrenderercontroller.h"
 
--- a/src/appleseed/renderer/meta/tests/test_environmentedf.cpp
+++ b/src/appleseed/renderer/meta/tests/test_environmentedf.cpp
@@ -73,9 +73,6 @@
 #include "OpenImageIO/texture.h"
 #include "foundation/platform/_endoiioheaders.h"
 
-// Boost headers.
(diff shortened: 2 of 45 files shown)


src/appleseed/renderer/api/rendering.h
```diff
@@ -45,7 +45,6 @@
 #include "renderer/kernel/rendering/masterrenderer.h"
 #include "renderer/kernel/rendering/nulltilecallback.h"
 #include "renderer/kernel/rendering/progressive/progressiveframerenderer.h"
-#include "renderer/kernel/rendering/scenepicker.h"
```
dictoon: This one should remain I think.

src/appleseed/renderer/meta/tests/test_environmentedf.cpp
```diff
@@ -191,16 +189,17 @@ TEST_SUITE(Renderer_Modeling_EnvironmentEDF)
             TextureStore texture_store(m_scene);
             TextureCache texture_cache(texture_store);
 
-            boost::shared_ptr<OIIO::TextureSystem> texture_system(
-                OIIO::TextureSystem::create(),
-                boost::bind(&OIIO::TextureSystem::destroy, _1));
+            std::shared_ptr<OIIOTextureSystem> texture_system(
+                OIIOTextureSystemFactory::create(),
+                [](OIIOTextureSystem* object) { object->release(); });
```
dictoon: Lovely!

src/appleseed/renderer/api/rendering.h
est77: scenepicker.h brings ShadingPoint and that brings OSL back again.
We cannot remove the OSL includes from shadingpoint. The class holds an instance of OSL::ShaderGlobals.

ScenePicker is not being used in any of our plugins as far as I know. It's only used in studio.

dictoon: The fact that it's used in .studio makes removing this line problematic though. How can it compile?

est77: I added the include in studio where it is used.

PR #1464, opened 2017-06-17
Studio python module

glebmish:


src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -181,6 +183,16 @@ void MainWindow::open_project(const QString& filepath)
     m_project_manager.load_project(filepath.toAscii().constData());
 }
 
+void MainWindow::new_project()
+{
+    if (!can_close_project())
```
est77: I think that in scripting, we always want to close existing project, even if it has unsaved changes.
We should move the can_close_project() test back to the slot.
What do you think?

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -514,6 +526,8 @@ void MainWindow::build_log_panel()
 
 void MainWindow::build_python_console_panel()
 {
+    PythonInterpreter::instance().set_mainwindow(this);
```
est77: Can we move this to the place where the mainwindow is created?
It's not really related to the python console panel.

src/appleseed.studio/mainwindow/mainwindow.cpp
glebmish: agree
but maybe we don't want to close project if it's unsaved? Maybe it would be a good idea to introduce force flag here? And if force==false then throw exception

src/appleseed.studio/mainwindow/mainwindow.cpp
glebmish: sure

src/appleseed.studio/python/pythoninterpreter.h
```diff
@@ -32,6 +32,8 @@
 // appleseed.foundation headers.
 #include "foundation/core/concepts/noncopyable.h"
 
+#include "mainwindow/mainwindow.h"
```
est77: Use a forward declaration.

glebmish: that way I need to include mainwindow.h in module.cpp and in every file where I use PythonInterpreter and get MainWindow from it. Is it ok?

src/appleseed.studio/mainwindow/mainwindow.cpp
est77: yes, good idea.

src/appleseed.studio/python/pythoninterpreter.h
est77: Yes, it is ok. It's more work, but it's better and we already do it everywhere in the rest of the code.

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -98,6 +98,8 @@
 #include <algorithm>
 #include <cassert>
 
+#include "python/pythoninterpreter.h"
```
dictoon: Move to the `appleseed.studio headers` section.

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -181,6 +183,16 @@ void MainWindow::open_project(const QString& filepath)
     m_project_manager.load_project(filepath.toAscii().constData());
 }
 
+void MainWindow::new_project()
```
dictoon: Maybe move before `open_project()` to follow the natural progression: new, open, save, save as, etc.

src/appleseed.studio/python/pythoninterpreter.cpp
```diff
@@ -85,7 +98,9 @@ PythonInterpreter::PythonInterpreter()
     // Imports appleseed module and creates asd alias for it
```
dictoon: Update comment now that there are two modules.

src/appleseed.studio/python/pythoninterpreter.cpp
```diff
@@ -85,7 +98,9 @@ PythonInterpreter::PythonInterpreter()
...
     // so we can refer to it by both names.
     PyRun_SimpleString("import appleseed\n"
```
dictoon: All imports first, all namespace aliases next.

src/appleseed.studio/python/module.cpp
```diff
@@ -0,0 +1,51 @@
...
+namespace studio = appleseed::studio;
+
+namespace bpy = boost::python;
+
+void open_project(const char* project_path)
+{
+    studio::PythonInterpreter::instance().get_mainwindow()->open_project(project_path);
+}
+
+void new_project()
```
dictoon: Reorder here too: `new_project()` first, `open_project()` second

src/appleseed.studio/python/pythoninterpreter.h
```diff
@@ -46,10 +48,14 @@ class PythonInterpreter
     void execute_command(const char* command);
 
     void redirect_output(OutputRedirector redirector);
+    void set_mainwindow(MainWindow* mainWindow);
```
dictoon: I'd suggest reordering methods of this class to follow a more logical progression:

    instance

    set_mainwindow
    get_mainwindow

    redirect_output

    execute_command

src/appleseed.studio/python/pythoninterpreter.h
```diff
@@ -46,10 +48,14 @@ class PythonInterpreter
...
+    MainWindow* get_mainwindow();
```
dictoon: `get_mainwindow()` should be made const.

src/appleseed.studio/python/module.cpp
```diff
@@ -0,0 +1,51 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+
+#include "foundation/platform/python.h"
+
+#include "pythoninterpreter.h"
```
dictoon: Headers should be ordered from the most specific first (appleseed.studio headers first in this case) to the most general (typically standard headers). This order was chosen because it helps catching missing header files. Several explanations and details can be found here: https://stackoverflow.com/questions/2762568/c-c-include-file-order-best-practices

Also, add the section comments: "appleseed.studio headers", "appleseed.foundation headers", etc.

src/appleseed.studio/python/pythoninterpreter.cpp
```diff
@@ -42,6 +42,8 @@
 namespace bpy = boost::python;
 namespace bf = boost::filesystem;
 
+extern "C" void init_appleseedstudio();
```
dictoon: Where is this function defined?

src/appleseed.studio/python/pythoninterpreter.h
glebmish: I can't because then I can't use its non-const functions
const MainWindow* PythonInterpreter::get_mainwindow()
{
    return mainWindow;
}

PythonInterpreter::instance().get_mainwindow().->open_project();

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -209,6 +190,31 @@ namespace
...
+    remove_render_widgets();
+
+    set_file_widgets_enabled(false, NotRendering);
+    set_project_explorer_enabled(false);
+    set_rendering_widgets_enabled(false, NotRendering);
+
+    m_project_manager.load_project(filepath.toAscii().constData());
+}
+
 void MainWindow::open_and_render_project(const QString& filepath, const QString& configuration)
```
glebmish: idk if I should expose this

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -902,6 +936,13 @@ bool MainWindow::can_close_project()
     return false;
 }
 
+bool MainWindow::can_close_project()
```
glebmish: I separated it to use in python module, but don't use it now. Should I revert it?

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -224,6 +230,39 @@ void MainWindow::open_and_render_project(const QString& filepath, const QString&
...
+        update_workspace();
+    }
+}
+
+Project* MainWindow::current_project()
+{
+    return m_project_manager.get_project();
+}
+
+bool MainWindow::is_project_dirty()
```
glebmish: is it ok to add this function here?

src/appleseed.studio/python/module.cpp
```diff
@@ -0,0 +1,93 @@
...
+#include "renderer/api/project.h"
+
+namespace bpy = boost::python;
+using namespace appleseed::studio;
+using namespace renderer;
+using namespace foundation;
+
+MainWindow* mainwindow()
+{
+    return PythonInterpreter::instance().get_mainwindow();
```
glebmish: Does it have any significant overhead to call for interpreter instance every time I need MainWindow?

src/appleseed.studio/python/pythoninterpreter.cpp
glebmish: it is generated by BOOST_PYTHON_MODULE(_appleseedstudio) in module.cpp

src/appleseed.studio/python/pythoninterpreter.h
dictoon: I'm not talking about returning a `const MainWindow*`, I'm just saying you should make `get_mainwindow()` itself const, which isn't the same thing:

    MainWindow* get_mainwindow() const;

src/appleseed.studio/mainwindow/mainwindow.cpp
est77: for now lets not expose it

src/appleseed.studio/mainwindow/mainwindow.cpp
est77: Yes, if it is unused, lets remove it.

src/appleseed.studio/mainwindow/mainwindow.cpp
est77: yes, I think so.

src/appleseed.studio/python/module.cpp
est77: Not really. Scripting is not really about performance and on top of that Python is super slow.

src/appleseed.studio/mainwindow/mainwindow.cpp
glebmish: maybe change all appearances of m_project_manager.is_project_dirty in MainWindow on this then?

src/appleseed.studio/mainwindow/mainwindow.cpp
dictoon: I'm really not enthusiastic about it. I see no value in duplicating what the project manager already offers. I'm in favor of removing this method.

src/appleseed.studio/mainwindow/mainwindow.cpp
dictoon: I'm really not enthusiastic about it. I see no value in duplicating what the project manager already offers. I'm in favor of removing this method.

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -870,19 +909,14 @@ namespace
     }
 }
 
-bool MainWindow::can_close_project()
+bool MainWindow::attempt_close_project()
```
dictoon: The name of this method is misleading: it isn't closing the project, it's just checking if the project can be closed. Maybe `ask_can_close_project()`? Not ideal but better. Maybe you have a better suggestion? :)

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -902,6 +936,13 @@ bool MainWindow::can_close_project()
     return false;
 }
 
+bool MainWindow::can_close_project()
+{
+    return !m_project_manager.is_project_loading() && // Project is not being loaded: no problem
```
dictoon: Rewrite as a series of `if` with early exits as it used to be done. This makes the logic much clearer.

src/appleseed.studio/python/module.cpp
```diff
@@ -0,0 +1,93 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+// appleseed.studio headers.
+#include "mainwindow/mainwindow.h"
+#include "pythoninterpreter.h"
```
dictoon: Should be `python/pythoninterpreter.h`.

src/appleseed.studio/mainwindow/mainwindow.cpp
glebmish: I will revert it back to old can_close_project function as I don't use separated part anyway

src/appleseed.studio/mainwindow/mainwindow.cpp
glebmish: same as above

src/appleseed.studio/python/module.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+// appleseed.renderer headers.
+#include "renderer/api/project.h"
+
+namespace bpy = boost::python;
+using namespace appleseed::studio;
+using namespace renderer;
+using namespace foundation;
+
+MainWindow* mainwindow()
```
dictoon: Should be `main_window()`.

src/appleseed.studio/python/module.cpp
dictoon: Should be `get_main_window()`.

src/appleseed.studio/python/pythoninterpreter.cpp
```diff
@@ -55,9 +59,14 @@ PythonInterpreter& PythonInterpreter::instance()
     return interpreter;
 }
 
-void PythonInterpreter::execute_command(const char* command)
+void PythonInterpreter::set_mainwindow(MainWindow* window)
```
dictoon: Should be `set_main_window()`.

src/appleseed.studio/python/pythoninterpreter.h
```diff
@@ -36,20 +36,27 @@ namespace appleseed {
...
 
     void redirect_output(OutputRedirector redirector);
 
+    void execute_command(const char* command);
+
   private:
     PythonInterpreter();
     ~PythonInterpreter();
+
+    MainWindow* mainWindow;
```
dictoon: Should be `m_main_window`.

dictoon:
Starts to look pretty awesome!

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -895,15 +935,15 @@ bool MainWindow::can_close_project()
     // The current project has been modified, ask the user what to do.
     switch (show_modified_project_message_box(this))
     {
-      case QMessageBox::Save:
-        slot_save_project();
-        return true;
+        case QMessageBox::Save:
```
dictoon: Revert indentation for the `switch` block.

src/appleseed.studio/mainwindow/mainwindow.cpp
dictoon: I thought you wanted to remove `MainWindow::open_and_render_project()`?

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -1268,26 +1295,8 @@ void MainWindow::slot_save_project_as()
...
-
-        m_project_manager.save_project_as(filepath.toAscii().constData());
-
-        if (m_project_file_watcher)
-            start_monitoring_project_file();
-
-        update_recent_files_menu(filepath);
-        update_workspace();
-    }
+    save_project(filepath);
```
dictoon: There is a regression here: if the user cancelled the *Save As* dialog, `filepath` is empty. It appears like this case is no longer handled correctly.

src/appleseed.studio/mainwindow/mainwindow.h
```diff
@@ -88,8 +89,13 @@ class MainWindow
     // Destructor.
     ~MainWindow();
 
+    ProjectManager* get_project_manager();
```
dictoon: Make the method const.

src/appleseed.studio/python/module.cpp
```diff
@@ -0,0 +1,103 @@
...
+using namespace appleseed::studio;
+using namespace renderer;
+using namespace foundation;
+
+MainWindow* main_window()
+{
+    return PythonInterpreter::instance().get_main_window();
+}
+
+ProjectManager* project_manager() {
```
dictoon: Move brace to its own line.

src/appleseed.studio/python/module.cpp
```diff
@@ -0,0 +1,103 @@
...
+    return project_manager()->get_project();
+}
+
+Project* open_project(const char* project_path)
+{
+    main_window()->open_project(project_path);
+    return project_manager()->get_project();
+}
+
+void save_project(const char* project_path = 0)
```
dictoon: Prefer `nullptr` now that we're using C++11.

src/appleseed.studio/python/module.cpp
```diff
@@ -0,0 +1,103 @@
...
+{
+    if (project_path == 0)
```
dictoon: `nullptr`

src/appleseed.studio/python/pythoninterpreter.cpp
```diff
@@ -42,6 +43,9 @@
 namespace bpy = boost::python;
 namespace bf = boost::filesystem;
 
+// Function prototypes.
+extern "C" void init_appleseedstudio();
```
dictoon: Where is this function defined?

dictoon:
Almost there! I think you introduced a regression that needs to be fixed, apart from that I've made a handful of minor cosmetics/formatting remarks. Once this is fixed we can merge!

src/appleseed.studio/mainwindow/mainwindow.cpp
glebmish: no, it was there before. Should I?

src/appleseed.studio/mainwindow/mainwindow.cpp
glebmish: I think there isn't
This check is in save_prject(): `if (!filepath.isEmpty())`

src/appleseed.studio/python/pythoninterpreter.cpp
glebmish: it is generated by BOOST_PYTHON_MODULE(_appleseedstudio) in module.cpp

src/appleseed.studio/mainwindow/mainwindow.cpp
dictoon: Ah ok, my bad :) It's fine then.

src/appleseed.studio/mainwindow/mainwindow.cpp
dictoon: But it isn't in `update_recent_files_menu()` if I'm not mistaken.

src/appleseed.studio/python/pythoninterpreter.cpp
dictoon: Got it. Maybe worth adding a comment to explain this.

Issue #1451, opened 2017-06-12
Camera jumps the first time it's moved after opening a project


dictoon:

aytekaman:
I did some debugging on this. I might be wrong but it seems like camera matrix and controller target are not properly set in the project till first camera update via mouse. I managed to fix this. But now, we don't get same results in BSDF test scenes i.e. camera is pointing to the empty space. But I think this is consistent with the camera parameters in the project file.

dictoon:
Hey, thanks for looking into this!!

> It seems like camera matrix and controller target are not properly set in the project till first camera update via mouse.

That's well possible, yes.

> But now, we don't get same results in BSDF test scenes i.e. camera is pointing to the empty space

That sounds like a bug. If the render shows something, that means that the camera is set properly, and the camera controller should be set accordingly.

Did you try other scenes from the test suite?

aytekaman:
This is getting interesting:)

Just rotated the camera a little bit and than saved the scene. Newly saved scene works fine. I checked the project files and the only difference is the camera matrices.

> Did you try other scenes from the test suite?

Yes, same thing happens in the _autofocus_ scene too.

dictoon:
> the only difference is the camera matrices

How different are the matrices? Can you copy/paste them here?

Isn't there also a difference in that the initial scene has no controller pivot point, while the newly saved scene has one?

aytekaman:
Original matrix for the _autofocus_ scene:

`0.000000000000000 0.000000000000000 -1.000000000000000 0.000000000000000`
`0.000000000000000 1.000000000000000 0.000000000000000 0.000000000000000`
`1.000000000000000 0.000000000000000 0.000000000000000 0.000000000000000`
`0.000000000000000 0.000000000000000 0.000000000000000 1.000000000000000`

Matrix after moving the camera a little bit:

`0.020990486104180 0.136938778876766 -0.990357092332281 0.001843854808597`
`-0.000000000000000 0.990575340373520 0.136968956497031 0.011070143145411`
`0.999779675475107 -0.002875044978055 0.020792657917254 0.033553467009628`
`0.000000000000000 0.000000000000000 0.000000000000000 1.000000000000000`

> Isn't there also a difference in that the initial scene has no controller pivot point, while the newly saved scene has one?

No. I diffed the files. They both have the same `controller_target` values. Only difference is the matrices.

I will investigate this further. If you have any suggestions, I appreciate it=)

dictoon:
Fixed.

PR #1468, opened 2017-06-20
Fix camera movement bug 

aytekaman:
First version of the fix for #1451.


aytekaman:
There is a problem with the file format i think. I will fix that.

dictoon:
Thanks!

Looks like line endings have changed, yes.

dictoon:
Woops, my review comments are shown as outdated (because you pushed your line endings fix just before my review).

dictoon:
Thanks a lot!

aytekaman:
You are welcome!

PR #1472, opened 2017-06-22
Many improvements around Python in appleseed.studio

dictoon:


src/appleseed.studio/python/pythoninterpreter.cpp
```diff
@@ -98,8 +97,7 @@ PythonInterpreter::PythonInterpreter()
     // Add path to appleseed module to sys.path so python can find it.
     bpy::import("sys").attr("path").attr("append")(lib.string());
 
-    // Imports appleseed module with 'asd' alias for it
-    // and _appleseedstudio module with 'studio' alias for it.
+    // Import appleseed module with 'asd' alias and _appleseedstudio module with 'studio' alias.
```
est77: asr alias

dictoon: Good catch, thanks.

Issue #1467, opened 2017-06-20
Fractional direct light samples with spot light leads to incorrect render


dictoon:
Repro case:

[test_scene_dl_samples.zip](https://github.com/appleseedhq/appleseed/files/1088077/test_scene_dl_samples.zip)

(Rename to *.appleseedz then load directly in appleseed.studio.)

Change the number of Direct Lighting samples to a value lower than 1 (e.g. 0.2) and observe that the render becomes much brighter.

dictoon:
Fixed.

PR #1489, opened 2017-06-28
Sun soft shadows

usakhelo:
This PR adds size_multiplier to sun light and soft shadows to sun's sample function.
It also fixes over-brightness when light sample count is less than 1.0
I'll comment code lines for the sake of clarity..

### DIFF
--- a/src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
+++ b/src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
@@ -234,19 +234,16 @@ void DirectLightingIntegrator::compute_outgoing_radiance_light_sampling_low_vari
     }
 
     // Add contributions from non-physical light sources only.
-    if (m_light_sample_count > 0)
+    for (size_t i = 0, e = m_light_sampler.get_non_physical_light_count(); i < e; ++i)
     {
-        for (size_t i = 0, e = m_light_sampler.get_non_physical_light_count(); i < e; ++i)
-        {
-            LightSample sample;
-            m_light_sampler.sample_non_physical_light(m_time, i, sample);
+        LightSample sample;
+        m_light_sampler.sample_non_physical_light(m_time, i, sample);
 
-            add_non_physical_light_sample_contribution(
-                sampling_context,
-                sample,
-                outgoing,
-                radiance);
-        }
+        add_non_physical_light_sample_contribution(
+            sampling_context,
+            sample,
+            outgoing,
+            radiance);
     }
 }
 
@@ -693,16 +690,6 @@ void DirectLightingIntegrator::add_non_physical_light_sample_contribution(
         probability
     );
 
-    //sample.m_probability *= probability;
-
-    //light->evaluate(
-    //    m_shading_context,
-    //    sample.m_light_transform,
-    //    m_material_sampler.get_point(),
-    //    emission_position,
-    //    emission_direction,
-    //    light_value);
-
     // Compute the incoming direction in world space.
     const Vector3d incoming = -emission_direction;
 
@@ -733,7 +720,7 @@ void DirectLightingIntegrator::add_non_physical_light_sample_contribution(
     // Add the contribution of this sample to the illumination.
     const float attenuation = light->compute_distance_attenuation(
         m_material_sampler.get_point(), emission_position);
-    const float weight = transmission * attenuation / sample.m_probability;
+    const float weight = transmission * attenuation / sample.m_probability / probability;
     light_value *= weight;
     light_value *= material_value;
     radiance += light_value;
--- a/src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
+++ b/src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
@@ -173,6 +173,7 @@ void DirectLightingIntegrator::compute_outgoing_radiance_light_sampling(
         else
         {
             add_non_physical_light_sample_contribution(
+                sampling_context,
                 sample,
                 outgoing,
                 radiance);
@@ -661,7 +662,8 @@ void DirectLightingIntegrator::add_emitting_triangle_sample_contribution(
 }
 
 void DirectLightingIntegrator::add_non_physical_light_sample_contribution(
-    const LightSample&          sample,
+    SamplingContext&            sampling_context,
+    LightSample&                sample,
     const Dual3d&               outgoing,
     Spectrum&                   radiance) const
 {
@@ -674,13 +676,31 @@ void DirectLightingIntegrator::add_non_physical_light_sample_contribution(
     // Evaluate the light.
     Vector3d emission_position, emission_direction;
     Spectrum light_value(Spectrum::Illuminance);
-    light->evaluate(
+
+    // Generate a uniform sample in [0,1).
+    sampling_context.split_in_place(2, 1);
+    const Vector2d s = sampling_context.next2<Vector2d>();
+    float probability;
+
+    light->sample(
         m_shading_context,
         sample.m_light_transform,
-        m_material_sampler.get_point(),
+        s,
         emission_position,
         emission_direction,
-        light_value);
+        light_value,
+        probability
+    );
+
+    sample.m_probability *= probability;
+
+    //light->evaluate(
+    //    m_shading_context,
+    //    sample.m_light_transform,
+    //    m_material_sampler.get_point(),
+    //    emission_position,
+    //    emission_direction,
+    //    light_value);
 
     // Compute the incoming direction in world space.
     const Vector3d incoming = -emission_direction;
--- a/src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
+++ b/src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
@@ -233,15 +233,19 @@ void DirectLightingIntegrator::compute_outgoing_radiance_light_sampling_low_vari
     }
 
     // Add contributions from non-physical light sources only.
-    for (size_t i = 0, e = m_light_sampler.get_non_physical_light_count(); i < e; ++i)
+    if (m_light_sample_count > 0)
     {
-        LightSample sample;
-        m_light_sampler.sample_non_physical_light(m_time, i, sample);
+        for (size_t i = 0, e = m_light_sampler.get_non_physical_light_count(); i < e; ++i)
+        {
+            LightSample sample;
+            m_light_sampler.sample_non_physical_light(m_time, i, sample);
 
-        add_non_physical_light_sample_contribution(
-            sample,
-            outgoing,
-            radiance);
+            add_non_physical_light_sample_contribution(
+                sampling_context,
+                sample,
+                outgoing,
+                radiance);
+        }
     }
 }
 
--- a/src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
+++ b/src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
@@ -685,6 +685,7 @@ void DirectLightingIntegrator::add_non_physical_light_sample_contribution(
     light->sample(
         m_shading_context,
         sample.m_light_transform,
+        m_material_sampler.get_point(),
         s,
         emission_position,
         emission_direction,
@@ -692,7 +693,7 @@ void DirectLightingIntegrator::add_non_physical_light_sample_contribution(
         probability
     );
 
-    sample.m_probability *= probability;
+    //sample.m_probability *= probability;
 
     //light->evaluate(
     //    m_shading_context,
--- a/src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
+++ b/src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
@@ -660,7 +660,7 @@ void DirectLightingIntegrator::add_emitting_triangle_sample_contribution(
 
 void DirectLightingIntegrator::add_non_physical_light_sample_contribution(
     SamplingContext&            sampling_context,
-    LightSample&                sample,
+    const LightSample&          sample,
     const Dual3d&               outgoing,
     Spectrum&                   radiance) const
 {
@@ -689,6 +689,8 @@ void DirectLightingIntegrator::add_non_physical_light_sample_contribution(
         light_value,
         probability
     );
+
+    light_value *= probability;
 
     // Compute the incoming direction in world space.
     const Vector3d incoming = -emission_direction;
@@ -720,7 +722,7 @@ void DirectLightingIntegrator::add_non_physical_light_sample_contribution(
     // Add the contribution of this sample to the illumination.
     const float attenuation = light->compute_distance_attenuation(
         m_material_sampler.get_point(), emission_position);
-    const float weight = transmission * attenuation / sample.m_probability / probability;
+    const float weight = transmission * attenuation / (sample.m_probability * probability);
     light_value *= weight;
     light_value *= material_value;
     radiance += light_value;
--- a/src/appleseed/renderer/kernel/lighting/directlightingintegrator.h
+++ b/src/appleseed/renderer/kernel/lighting/directlightingintegrator.h
@@ -145,7 +145,7 @@ class DirectLightingIntegrator
 
     void add_non_physical_light_sample_contribution(
         SamplingContext&                sampling_context,
-        LightSample&                    sample,
+        const LightSample&              sample,
         const foundation::Dual3d&       outgoing,
         Spectrum&                       radiance) const;
 };
--- a/src/appleseed/renderer/kernel/lighting/directlightingintegrator.h
+++ b/src/appleseed/renderer/kernel/lighting/directlightingintegrator.h
@@ -144,7 +144,8 @@ class DirectLightingIntegrator
         Spectrum&                       radiance) const;
 
     void add_non_physical_light_sample_contribution(
-        const LightSample&              sample,
+        SamplingContext&                sampling_context,
+        LightSample&                    sample,
         const foundation::Dual3d&       outgoing,
         Spectrum&                       radiance) const;
 };
--- a/src/appleseed/renderer/kernel/lighting/lighttracing/lighttracingsamplegenerator.cpp
+++ b/src/appleseed/renderer/kernel/lighting/lighttracing/lighttracingsamplegenerator.cpp
@@ -706,7 +706,6 @@ namespace
             light_sample.m_light->sample(
                 m_shading_context,
                 light_sample.m_light_transform,
-                Vector3d(0.0),
                 sampling_context.next2<Vector2d>(),
                 emission_position,
                 emission_direction,
--- a/src/appleseed/renderer/kernel/lighting/lighttracing/lighttracingsamplegenerator.cpp
+++ b/src/appleseed/renderer/kernel/lighting/lighttracing/lighttracingsamplegenerator.cpp
@@ -706,6 +706,7 @@ namespace
             light_sample.m_light->sample(
                 m_shading_context,
                 light_sample.m_light_transform,
+                Vector3d(0.0),
                 sampling_context.next2<Vector2d>(),
                 emission_position,
                 emission_direction,
--- a/src/appleseed/renderer/kernel/lighting/sppm/sppmphotontracer.cpp
+++ b/src/appleseed/renderer/kernel/lighting/sppm/sppmphotontracer.cpp
@@ -450,6 +450,7 @@ namespace
             light_sample.m_light->sample(
                 shading_context,
                 light_sample.m_light_transform,
+                Vector3d(0.0),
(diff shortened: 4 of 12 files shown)


src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
```diff
@@ -233,15 +233,19 @@ void DirectLightingIntegrator::compute_outgoing_radiance_light_sampling_low_vari
     }
 
     // Add contributions from non-physical light sources only.
-    for (size_t i = 0, e = m_light_sampler.get_non_physical_light_count(); i < e; ++i)
+    if (m_light_sample_count > 0)
```
usakhelo: I've found that when light sample count < 0 in studio settings direct lights are strangely brighter than usually. In such cases this variable (m_light_sample_count) gets value 0 sometimes and I thought direct light should not be added when it is 0. It fixed the brightness but I'm not sure if fix is correct.

src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
```diff
@@ -657,7 +662,8 @@ void DirectLightingIntegrator::add_emitting_triangle_sample_contribution(
 }
 
 void DirectLightingIntegrator::add_non_physical_light_sample_contribution(
-    const LightSample&          sample,
+    SamplingContext&            sampling_context,
```
usakhelo: sampling_context added to be passed to sample() function

src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
```diff
@@ -670,13 +676,32 @@ void DirectLightingIntegrator::add_non_physical_light_sample_contribution(
     // Evaluate the light.
     Vector3d emission_position, emission_direction;
     Spectrum light_value(Spectrum::Illuminance);
-    light->evaluate(
+
+    // Generate a uniform sample in [0,1).
+    sampling_context.split_in_place(2, 1);
+    const Vector2d s = sampling_context.next2<Vector2d>();
```
usakhelo: I'm not sure about these two lines. I figured since it is 2d vector dimension should be 2.

src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
```diff
@@ -670,13 +676,32 @@ void DirectLightingIntegrator::add_non_physical_light_sample_contribution(
...
+    float probability;
```
usakhelo: this variable is passed to sample to calculate, but I don't know what to do with it later.

src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
```diff
@@ -670,13 +676,32 @@ void DirectLightingIntegrator::add_non_physical_light_sample_contribution(
...
         m_material_sampler.get_point(),
+        s,
         emission_position,
         emission_direction,
-        light_value);
+        light_value,
+        probability
+    );
+
+    //sample.m_probability *= probability;
```
usakhelo: I've tried to apply calculated probability to light sample but it made light many times brighter.

src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
```diff
@@ -670,13 +676,32 @@ void DirectLightingIntegrator::add_non_physical_light_sample_contribution(
...
+
+    // Generate a uniform sample in [0,1).
+    sampling_context.split_in_place(2, 1);
+    const Vector2d s = sampling_context.next2<Vector2d>();
+    float probability;
+
+    light->sample(
         m_shading_context,
         sample.m_light_transform,
         m_material_sampler.get_point(),
```
usakhelo: target point parameter added to sample()

src/appleseed/renderer/kernel/lighting/lighttracing/lighttracingsamplegenerator.cpp
```diff
@@ -706,6 +706,7 @@ namespace
             light_sample.m_light->sample(
                 m_shading_context,
                 light_sample.m_light_transform,
+                Vector3d(0.0),
```
usakhelo: This and next change calls light's sample() function. I didn't know where to take surface point at this stage so I passed zero vector to make compiler happy.

src/appleseed/renderer/kernel/lighting/sppm/sppmphotontracer.cpp
```diff
@@ -450,6 +450,7 @@ namespace
             light_sample.m_light->sample(
                 shading_context,
                 light_sample.m_light_transform,
+                Vector3d(0.0),
```
usakhelo: same as above

src/appleseed/renderer/modeling/light/directionallight.cpp
```diff
@@ -125,6 +125,7 @@ namespace
         virtual void sample(
             const ShadingContext&   shading_context,
             const Transformd&       light_transform,
+            const Vector3d&         target_point,
```
usakhelo: Dummy target_point parameters, never used in any light except sun light. Third version of sample should added to light.h probably?

src/appleseed/renderer/modeling/light/sunlight.cpp
```diff
@@ -96,6 +96,7 @@ namespace
             m_inputs.declare("environment_edf", InputFormatEntity, "");
             m_inputs.declare("turbidity", InputFormatFloat);
             m_inputs.declare("radiance_multiplier", InputFormatFloat, "1.0");
+            m_inputs.declare("size_multiplier", InputFormatFloat, "1.0");
```
usakhelo: size multiplier parameter added

src/appleseed/renderer/modeling/light/sunlight.cpp
```diff
@@ -141,17 +148,24 @@ namespace
...
             const Vector2d&         s,
             Vector3d&               position,
             Vector3d&               outgoing,
             Spectrum&               value,
             float&                  probability) const override
         {
+            //angular_diameter = one_radian * sun_diameter / distance
+            //sun_diameter = angular_diameter * distance / one_radian = 0.5331 degrees * 1.496 * 10^8 km / 57.29578 degrees
+            double sun_diameter = 0.5331 * m_safe_scene_diameter / 57.29578;
+            sun_diameter *= m_values.m_size_multiplier;
```
usakhelo: calculating sun's diameter based on distance from scene center

src/appleseed/renderer/modeling/light/sunlight.cpp
```diff
@@ -418,6 +437,9 @@ namespace
                 + disk_radius * p[0] * basis.get_tangent_u()
                 + disk_radius * p[1] * basis.get_tangent_v();
 
+            if (target_point != Vector3d(0.0))
+                outgoing = normalize(target_point - position);
```
usakhelo: this checks for dummy target_point passed from sppmphotontracer.cpp and lighttracingsamplegenerator.cpp

src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
dictoon: > I've found that when light sample count < 0

I assume you mean `light sample count < 1`? Because `light sample count< 0` doesn't make sense and shouldn't be allowed from .studio.

The `light sample count < 1` case is a confirmed bug: https://github.com/appleseedhq/appleseed/issues/1467

I suggest you don't worry about it for now, we'll fix it at some point.

src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
dictoon: Makes sense.

src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
dictoon: That's correct, here is the details of how this works:

    sampling_context.split_in_place(2, 1);

This creates a new "trajectory" or "sampling sequence": the 2 means "2D samples" and the 1 means that we want a single sample in that sequence.

    sampling_context.next2<Vector2d>();

This returns the next (and in this case, the only) 2D sample (of type Vector2d, i.e. as two double-precision values). The 2 in `next2()` means that we want a semi-open intervals for samples: [0, 1) (closed on the left, open on the right, i.e. 0 is part of the interval but 1 isn't). That's the only variant available with sampling contexts since it's the only one that we need so far.

With pure random numbers, we have two additional variants: Functions suffixed with 1 such as `rand_double1()` return values in the interval [0, 1] (closed on both ends) and functions suffixed with 3 such as `rand_double3()` return values in the interval (0, 1) (open on both ends).

src/appleseed/renderer/modeling/light/sunlight.cpp
dictoon: Cool, I'll trust your calculations for now :)

src/appleseed/renderer/modeling/light/sunlight.cpp
dictoon: Looks good!

src/appleseed/renderer/modeling/light/sunlight.cpp
dictoon: That won't work because (0,0,0) is otherwise a perfectrly valid target point. We probably need to find a better mechanism here.

src/appleseed/renderer/modeling/light/directionallight.cpp
dictoon: Yes probably.

src/appleseed/renderer/modeling/light/sunlight.cpp
```diff
@@ -128,6 +129,12 @@ namespace
             // Apply turbidity bias.
             m_values.m_turbidity += BaseTurbidity;
 
+            const Source* size_multiplier_src = get_inputs().source("size_multiplier");
+            if (size_multiplier_src && size_multiplier_src->is_uniform())
```
dictoon: We should probably even emit a warning if the size multiplier is not uniform (uniform = constant) since it wouldn't make much sense.

src/appleseed/renderer/modeling/light/sunlight.cpp
```diff
@@ -141,17 +148,24 @@ namespace
...
+            double sun_diameter = 0.5331 * m_safe_scene_diameter / 57.29578;
+            sun_diameter *= m_values.m_size_multiplier;
+
             sample_disk(
                 light_transform,
+                target_point,
                 s,
                 m_scene_center,
-                m_scene_radius,
+                sun_diameter/2,
```
dictoon: Would make sense to compute the Sun's _radius_ directly, thus avoiding a division.

src/appleseed/renderer/modeling/light/sunlight.cpp
```diff
@@ -486,6 +508,17 @@ DictionaryArray SunLightFactory::get_input_metadata() const
             .insert("default", "1.0")
             .insert("help", "Light intensity multiplier"));
 
+    metadata.push_back(
+        Dictionary()
+            .insert("name", "size_multiplier")
+            .insert("label", "Sun Size Multiplier")
```
dictoon: Remove *Sun*.

src/appleseed/renderer/modeling/light/sunlight.cpp
```diff
@@ -486,6 +508,17 @@ DictionaryArray SunLightFactory::get_input_metadata() const
...
+            .insert("type", "numeric")
+            .insert("min_value", "0.0")
+            .insert("max_value", "100.0")
+            .insert("use", "optional")
+            .insert("default", "1.0")
+            .insert("help", "Shadow softness multiplier"));
```
dictoon: Maybe more explicative: "The size multiplier allows to make the sun bigger or smaller, hence making it cast softer or harder shadows."

src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
dictoon: The principle is the same as for mesh lights:

* Given a light emitting mesh, we first choose a triangle of the mesh with probability `Ptri`. Then, on this triangle, we choose a point with uniform probability `Ppoint` which is simply `1/triangleArea`. The total probability for that point is then `Ptri * Ppoint`.

* Given a set of non-physical lights, we first choose one of the light with probability `Plight` (in this method, `Plight` is `sample.m_probability`). Then, for this light, we choose a direction with probability `Pdir`. For most lights, given a target point to be illuminated, `Pdir` is just 1 since there isn't any choice to be made (given a point light and a target point, there is only one ray joining the two, hence its probability is 1). Similarly as for the mesh light case, the total probability for that light direction is then `Plight * Pdir`.

src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
dictoon: According to my explanation above, that's essentially the correct thing to do. There must be another mistake then, I need to investigate more carefully.

dictoon: Ah, I think I know the problem. The sun's "intensity" we calculate assume that the sun is a pure directional light, i.e. it's the full sun's energy, not the energy per unit of area. So in a sense, _not_ dividing the total contribution by the probability of the sampled direction is correct. To be correct in the general case however, we must do that division (i.e. uncomment the light above, or better: divide by `probability` when computing the overall weight on line 736) and at the same time, we need to divide the sun's intensity by the disk's surface area. These two computations will cancel out, but we need to do it for other light models where they won't necessarily cancel out.

usakhelo:
Thanks for review!
I'll be fixing this pr according to your comments. Should I close it for now or is it ok if it will be updated from time to time? I'll do rebase when it'll be final.

P.S. I have to say that lighting is very different when `sample()` takes `target_point` into account. Without it lighting is flat. I don't have good example but here's what I have (upper is with target_point):
image

dictoon:
Hey, thanks! I'll check your changes when I have a moment.

> Should I close it for now or is it ok if it will be updated from time to time? I'll do rebase when it'll be final.

It's actually best to keep it open and amend it until it reaches a point where we can merge it, and then rebase it.

> I have to say that lighting is very different when sample() takes target_point into account

That's kind of surprising, since the sun is so far away (in theory) the lighting should be rather flat, like in your second picture. What's the sun size in this case?

src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
dictoon: This bug was fixed by @Biart95.

src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
```diff
@@ -670,13 +676,32 @@ void DirectLightingIntegrator::add_non_physical_light_sample_contribution(
...
         emission_position,
         emission_direction,
-        light_value);
+        light_value,
+        probability
+    );
+
+    //sample.m_probability *= probability;
+
+    //light->evaluate(
```
dictoon: Remove dead code.

src/appleseed/renderer/modeling/light/light.h
```diff
@@ -107,6 +107,15 @@ class APPLESEED_DLLSYMBOL Light
     virtual void sample(
         const ShadingContext&           shading_context,
         const foundation::Transformd&   light_transform,            // light space to world space transform
+        const foundation::Vector3d&     target_point,               // world space target point
+        const foundation::Vector2d&     s,                          // sample in [0,1)^2
+        foundation::Vector3d&           position,                   // world space emission position
+        foundation::Vector3d&           outgoing,                   // world space emission direction, unit-length
+        Spectrum&                       value,                      // light value
+        float&                          probability) const;     // PDF value
```
dictoon: Align comments.

usakhelo:
Hi Franz,
I've fixed some of the things:

* Probability applied to all the lights. In case of sunlight probability = 1/sun_area, so it is applied to light value and then `weight` is divided by it to compensate. For directional light I've made `probability =1.0` and `disk_radius=0`
* overloaded `sample()` function with `target_point` argument added to all the lights, so they look as expected. 
* added separate `sample_disk_parallel()` function to sunlight to be called from `sample()` without `target_point` argument
* removed `DirectLightingIntegrator` fixes

Also you asked me about sun size on those renderings - sun diameter is 0.74mm and distance is 88mm (`m_scene_radius`)

src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
```diff
@@ -708,7 +720,7 @@ void DirectLightingIntegrator::add_non_physical_light_sample_contribution(
     // Add the contribution of this sample to the illumination.
     const float attenuation = light->compute_distance_attenuation(
         m_material_sampler.get_point(), emission_position);
-    const float weight = transmission * attenuation / sample.m_probability;
+    const float weight = transmission * attenuation / sample.m_probability / probability;
```
dictoon: You can avoid one division by dividing by the product of probabilities:

    const float weight = transmission * attenuation / (sample.m_probability * probability);

This is the kind of optimization the compiler probably cannot do because it may slightly change the result (unless "fast math" is enabled, which it isn't in appleseed).

src/appleseed/renderer/kernel/lighting/directlightingintegrator.h
```diff
@@ -144,7 +144,8 @@ class DirectLightingIntegrator
         Spectrum&                       radiance) const;
 
     void add_non_physical_light_sample_contribution(
-        const LightSample&              sample,
+        SamplingContext&                sampling_context,
+        LightSample&                    sample,
```
dictoon: Why is `sample` passed as a non-const reference?

src/appleseed/renderer/modeling/light/directionallight.cpp
```diff
@@ -242,7 +242,7 @@ namespace
 
             value = m_values.m_irradiance;
 
-            probability = 1.0f / (Pi<float>() * square(static_cast<float>(disk_radius)));
+            probability = 1.0f;
```
dictoon: That's only correct when `disk_radius` is 0.

You shouldn't change the way the probability is computed, it was correct before. What you need to do instead is no longer call `sample_disk()` when you have a target point. The code will also be much simpler in that case.

src/appleseed/renderer/modeling/light/directionallight.cpp
```diff
@@ -135,7 +135,7 @@ namespace
                 light_transform,
                 s,
                 m_scene_center,
-                m_scene_radius,
+                0,
```
dictoon: Here, since you have a target point, you can compute `position`, `outgoing` and `probability` (and `value` but this one doesn't change) in a much simpler manner.

src/appleseed/renderer/modeling/light/maxomnilight.cpp
```diff
@@ -125,6 +142,7 @@ namespace
             outgoing = sample_sphere_uniform(s);
             value = m_values.m_intensity;
             probability = RcpFourPi<float>();
+            value *= probability;
```
dictoon: `value` is not supposed to be multiplied by probability here, it's the reason why we return both of them separately.

src/appleseed/renderer/modeling/light/maxomnilight.cpp
```diff
@@ -115,6 +115,23 @@ namespace
...
+            const Vector2d&         s,
+            Vector3d&               position,
+            Vector3d&               outgoing,
+            Spectrum&               value,
+            float&                  probability) const override
+        {
+            position = light_transform.get_parent_origin();
+            outgoing = normalize(target_point - position);
+            value = m_values.m_intensity;
+            probability = RcpFourPi<float>();
```
dictoon: Since there is no sampling, I believe that in this case `probability` should be 1.

src/appleseed/renderer/modeling/light/maxspotlight.cpp
```diff
@@ -139,6 +140,31 @@ namespace
             const Vector3d axis = -normalize(light_transform.get_parent_z());
 
             compute_radiance(shading_context, light_transform, axis, outgoing, value);
+
+            outgoing = normalize(target_point - position);
+            value *= probability;
```
dictoon: Don't multiply value by probability.

usakhelo:
Hi,
Pushed another attempt to fix comments.
I moved `light_value *= probability` to DirectLightingIntegrator, is that right?

src/appleseed/renderer/modeling/light/sunlight.cpp
usakhelo: added warning if size_multiplier is not uniform

dictoon:
Superseded by #1519.

PR #1492, opened 2017-06-28
Fixes of path tracer and lighting engines

Biart95:


src/appleseed/renderer/kernel/lighting/pathtracer.h
```diff
@@ -588,8 +588,10 @@ bool PathTracer<PathVisitor, VolumeVisitor, Adjoint>::process_bounce(
     SamplingContext&    sampling_context,
     PathVertex&         vertex,
     BSDFSample&         sample,
-    ShadingRay&         ray)
+    ShadingRay&         next_ray)
 {
+    const ShadingRay& ray = vertex.get_ray();
```
dictoon: I suggest moving this right before it is needed.

src/appleseed/renderer/kernel/lighting/pathtracer.h
```diff
@@ -634,22 +636,22 @@ bool PathTracer<PathVisitor, VolumeVisitor, Adjoint>::process_bounce(
 
     // Construct the scattered ray.
     const foundation::Vector3d incoming(sample.m_incoming.get_value());
-    ray = ShadingRay(
+    next_ray = ShadingRay(
```
dictoon: That's a ray copy we should considering eliminating if it isn't too complex.

Biart95:
I tried to tackle #1467 as well. I am aware that my solution can increase the variance a little, but it touches the existing code minimally and improves performance.

PR #1504, opened 2017-07-05
Fix raydepth in shaders, UDIM

luisbarrancos:
Minor fixes for raydepth on the shader side, fixes for UDIM lookups, and ongoing work for Substance support. It seems OIIO/OSL replace the lookups by UDIM tiles if the explicit literal **\<UDIM\>** (for Mari's case here) is supported. This mean getting the filename with the pattern. Trying to access the default filename accesses only the explicit tile/file loaded (i.e, foobar_1001, etc.).


luisbarrancos:
@dictoon Maya can refer to full attribute names, or 2, 3 character tops attribute names (short name). For our custom nodes we provide the short names, maya provides them automagically for its own nodes.

luisbarrancos:
@dictoon merging now, i'll submit a UDIM shaderball with substance ongoing work later in the week.

PR #1510, opened 2017-07-06
Update CMakeLists.txt

aytekaman:
Just noticed that we haven't added the `benchmark_basis.cpp` to the CMakeLists.txt.


dictoon:
Great, thank you!

PR #1512, opened 2017-07-07
Fix default OIIO texture cache size.

est77:


src/appleseed/renderer/kernel/texturing/texturestore.h
```diff
@@ -119,6 +119,9 @@ class TextureStore
     // Retrieve performance statistics.
     foundation::StatisticsVector get_statistics() const;
 
+    // Return the default texture store size in MB.
+    static size_t get_default_texture_store_size_MB();
```
dictoon: We're in the texture store already so we can make the name shorter. I'd also prefer to return bytes since that's what's needed in both cases. In conclusion I propose `get_default_size()`.

PR #1524, opened 2017-07-10
Cosmetics

gospodnetic:
Added space between the keyword and its bracket.


gospodnetic:
Those files are not written by us so we don't touch them :)

PR #1527, opened 2017-07-10
Remove render layer rules

est77:


Narann:
I was the man who pushed for this before I understood appleseed scenes are not lighting software. ^^'

dictoon:
You were the man who push to _remove_ render layer rules, do I get that right?

Narann:
Ooops, I wasn't clear. I mean I originally pushed to put render layer rules in the appleseed scene at the first place and this was a bad idea as appleseed scenes is supposed to expressed a flat scene, not put a mechanism in the renderer to interpret rules.

src/appleseed/renderer/modeling/project/project.xsd
```diff
@@ -294,22 +294,6 @@
                         </xsd:complexContent>
                     </xsd:complexType>
                 </xsd:element>
-                <xsd:element name="rules" minOccurs="0">
```
dictoon: I'm not sure we want to remove them from the schema just yet because that prevents loading scenes with render layer rules.

src/appleseed/renderer/modeling/project/projectfileupdater.cpp
```diff
@@ -647,62 +646,8 @@ namespace
...
-                        rule_name.c_str(),
-                        ParamArray()
-                            .insert("render_layer", render_layer_name)
-                            .insert("order", "1")
-                            .insert("pattern", "^" + entity_path + "$")));
-
-                string_params.remove("render_layer");
-            }
+            // Here, we used to update render layer rules
+            // but render layers were removed in appleseed 1.7.
```
dictoon: *1.7.0-beta*

Narann:
Does the removal of render layer rules also mean we will have to bump to a backward incompatible version? appleseed 2?

dictoon:
@Narann That's a very good question and I thought about it, but ultimately I don't think it's such a big deal to break compatibility as render layers weren't used that much (Fetch was probably the biggest user as every shot used them).

src/appleseed/renderer/modeling/project/project.xsd
est77: I don't think we can load them in any case, no? I removed the code that loads them.
I'm not very familiar with Xerces. What happens if we don't handle an element?

est77:
Fixed and rebased.

src/appleseed/renderer/modeling/project/project.xsd
dictoon: Not loading them is one thing, but failing to load a project file altogether because it now contains unknown elements is another...

PR #1531, opened 2017-07-11
Get rid of templates in Tracer and fix bug in distance sampling

Biart95:
- Get rid of templates in Tracer (Simple overloading instead: now can pass `const ShadingPoint*&` as last argument for the full tracing)
- Fix bug with bad sampling when `vertex.throughput` has more components than `extinction` of phase function.

### DIFF
--- a/src/appleseed/renderer/kernel/lighting/materialsamplers.cpp
+++ b/src/appleseed/renderer/kernel/lighting/materialsamplers.cpp
@@ -131,14 +131,15 @@ const ShadingPoint& BSDFSampler::trace(
         m_shading_point.get_ray().m_depth + 1);
     ray.copy_media_from(m_shading_point.get_ray());
 
-    const ShadingPoint& shading_point =
-        shading_context.get_tracer().trace<const ShadingPoint&>(
-            shading_context,
-            m_shading_point,
-            ray,
-            transmission);
-    if (shading_point.hit()) transmission.set(0.0f);
-    return shading_point;
+    const ShadingPoint* shading_point;
+    shading_context.get_tracer().trace(
+        shading_context,
+        m_shading_point,
+        ray,
+        transmission,
+        shading_point);
+    if (shading_point->hit()) transmission.set(0.0f);
+    return *shading_point;
 }
 
 void BSDFSampler::trace_between(
@@ -258,13 +259,14 @@ const ShadingPoint& PhaseFunctionSampler::trace(
         m_volume_ray.m_depth + 1);
     ray.copy_media_from(m_volume_ray);
 
-    const ShadingPoint& shading_point =
-        shading_context.get_tracer().trace<const ShadingPoint&>(
-            shading_context,
-            ray,
-            transmission);
-    if (shading_point.hit()) transmission.set(0.0f);
-    return shading_point;
+    const ShadingPoint* shading_point;
+    shading_context.get_tracer().trace(
+        shading_context,
+        ray,
+        transmission,
+        shading_point);
+    if (shading_point->hit()) transmission.set(0.0f);
+    return *shading_point;
 }
 
 void PhaseFunctionSampler::trace_between(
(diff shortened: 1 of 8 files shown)


src/appleseed/renderer/kernel/lighting/materialsamplers.cpp
```diff
@@ -131,14 +131,15 @@ const ShadingPoint& BSDFSampler::trace(
...
-    const ShadingPoint& shading_point =
-        shading_context.get_tracer().trace<const ShadingPoint&>(
-            shading_context,
-            m_shading_point,
-            ray,
-            transmission);
-    if (shading_point.hit()) transmission.set(0.0f);
-    return shading_point;
+    const ShadingPoint* shading_point;
+    shading_context.get_tracer().trace(
```
dictoon: I preferred the version that returns a const ref to a `ShadingPoint`; can we switch back to that?

src/appleseed/renderer/kernel/lighting/pt/ptlightingengine.cpp
```diff
@@ -908,16 +908,17 @@ namespace
                     precomputed_mis_weights = scattering_coef;
                 else
                     Spectrum::upgrade(scattering_coef, precomputed_mis_weights);
-                precomputed_mis_weights *= vertex.m_throughput;
+                if (vertex.m_throughput.size() <= channel_count)
```
dictoon: I'm not sure to understand this test, in fact it looks wrong.

There are really only two cases when working with spectra: either they are the same size (both RGB or both spectral), or they're not in which case the RGB one must be upgraded.

src/appleseed/renderer/kernel/lighting/tracer.h
```diff
@@ -79,34 +79,52 @@ class Tracer
...
         const ShadingRay&           ray,
         Spectrum&                   transmission);
 
-    template <typename ReturnType = void>
-    ReturnType trace(
+    void trace(
         const ShadingContext&       shading_context,
         const ShadingRay&           ray,
+        Spectrum&                   transmission,
+        const ShadingPoint*&        shading_point);
```
dictoon: What about changing `trace()` to always return a const ref to the shading point?

Narann: Const ref to a pointer? I'm also curious about the rationals.

dictoon: Just a const ref to a `ShadingPoint`?

Narann: I was not talking about your comment but the code itself. I also agree a const ref to `ShadingPoint` make more sense but because the current code wants a const ref to a pointer of `ShadingPoint`, I ask "why would you do that?".

dictoon: The reference itself is not const, it's the pointer that is. This is one way to return a pointer from a function. Another way would be to pass a pointer to a pointer to a ShadingPoint: `const ShadingPoint**`

src/appleseed/renderer/kernel/lighting/materialsamplers.cpp
Biart95: Yes, I also like that version more. But we need to change the name of this function (because they differ only in return values). `trace_full` and `trace__between_full`?

Biart95: Of course, more complex methods must have more complex names, in this case programmers cannot mistakenly use more complex method by calling `trace()`

src/appleseed/renderer/kernel/lighting/pt/ptlightingengine.cpp
Biart95: It is wrong, you are right. It is the result of my laziness: the fact is if we upgrade everything to the max size, we need to tell `PhaseFunction::sample_distance` what spectrum size we are using. Changing `PhaseFunction::sample_distance` is starting to get annoying... Possibly, we should remove this functionality from `PhaseFunction` forever and ever.

src/appleseed/renderer/kernel/lighting/tracer.h
Biart95: It is possible, but, as I already mentioned, we need to come up with a new name.

The rationals in my solution are my conservative strive to preserve the name of `trace`.

PR #1533, opened 2017-07-12
Reparametrization for volume rendering (Do not merge)

Biart95:
Allows to compare path tracing with single scattering and SSS:

volume
sss


Biart95:
Maintained in #1545

PR #1536, opened 2017-07-14
Add stdSurface, use max(color C) where possible

luisbarrancos:
Since OSL seems not be constant folding color equatlity, inequality tests, but max(C) are folded.
Initial standardSurface for feedback. The interface should build fine on Gaffer as well.


src/appleseed.shaders/src/appleseed/as_disney_material.osl
```diff
@@ -154,36 +154,42 @@ shader as_disney_material
...
-        string as_maya_attribute_short_name = "mo",
-        int as_maya_attribute_hidden = 1,
+        string as_maya_attribute_short_name = "mao",
         float min = 0.0,
         float max = 1.0,
         string label = "Matte Opacity",
         string page = "Matte Opacity Parameters"
     ]],
-    int in_maximumRayDepth = 4
+    color in_matte_opacity_color = color(1,0.5,0)
```
dictoon: Always insert a space after a comma: `color(1, 0.5, 0)`

PR #1538, opened 2017-07-14
Fix critical bug in volume distance sampling

Biart95:


src/appleseed/renderer/modeling/phasefunction/henyeyphasefunction.cpp
```diff
@@ -195,16 +196,25 @@ class HenyeyPhaseFunction
     {
         extinction_coefficient(data, volume_ray, distance, spectrum);
         for (size_t i = 0, e = spectrum.size(); i < e; ++i)
-            spectrum[i] = std::exp(-distance * spectrum[i]);
+        {
+            const float x = -distance * spectrum[i];
+            assert(!isinf(x) && !isnan(x));
```
dictoon: Missing `std::` qualifiers.

Issue #1540, opened 2017-07-16
Adding a specular coating to glass BSDF darkens output abnormally


luisbarrancos:
Rendering the glass BSDF is fine, rendering a coating layer separately is also fine. Rendering a coating (ior=1.35), and adding it to a glass BSDF (ior=1.49) however, without any modulating factor and glass BSDF surface transmittance set to 1.0, results in abnormally dark values (one would expect that without any scaling, adding these two closures would break energy conservation).

Glass BSDF only:
![glass1](https://user-images.githubusercontent.com/387352/28246263-78b3a60a-6a48-11e7-9d45-61e6c8ba5212.png)

Dielectric coating over glass BSDF, surface transmittance set to 1.0:
![glass2](https://user-images.githubusercontent.com/387352/28246271-7fc98a9a-6a48-11e7-8194-7ab5f7169662.png)

est77:
Couldn't reproduce the issue.I re-enabled coatings over glass in PR #1893.
We can open a new ticket if this issue appears again.

PR #1541, opened 2017-07-16
Minor fixes for glass, DEBUG case for stdSurface

luisbarrancos:
Glass would not evaluated, fixed it. Build with DEBUG to test in relation to issue #1540

### DIFF
--- a/src/appleseed.shaders/src/appleseed/as_standard_surface.osl
+++ b/src/appleseed.shaders/src/appleseed/as_standard_surface.osl
@@ -657,6 +657,13 @@ shader as_standard_surface
         {
             color refraction_color = in_refraction_amount * in_refraction_tint;
 
+            // NOTE: Without the layered closure this returns an unusually
+            //       dark result when coating is on.
+            if (coating)
+            {
+                warning("[WARNING]: Coated glass BSDF is unsupported!\n");
+            }
+
             if (max(refraction_color) > 0.0)
             {
                 out_outColor += opacity * as_glass(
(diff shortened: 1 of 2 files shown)


src/appleseed.shaders/src/appleseed/as_standard_surface.osl
```diff
@@ -657,13 +657,29 @@ shader as_standard_surface
         {
             color refraction_color = in_refraction_amount * in_refraction_tint;
 
+#ifdef DEBUG
+            if (compute_coating || in_coating_depth > 0.0)
+            {
+                warning("[WARNING]: Coated glass BSDF is unsupported!!\n");
```
dictoon: One exclamation mark ought to be enough :)

luisbarrancos:
Hi @Narann. Rhe path:ray_length attribute should always be available, always. The reasoning for that test was to have immediate feedback when for some reason that test is not available, in which case the ray depth user choice becomes meaningless since it won´t be respected. 

Not bailing out in that case could be hard to spot depending on the number of bounces that were being ran in the scene. This way we have an red flag. If we use the and, we will bailout according to the user depth of course, but any invalid exit status from getattribute will make the shader continuing execution as if nothing had happened.

Generally, i add DEBUG tests and check status, values, and in case of error, output via OSL´ s warning() together with OSL shader:shadername, and built-in preprocessor macros __FILE__ and __LINE__ which should suffice to pinpoint anything. We can use error() to immediate stop execution altogether, but i found OSL´ s warning() suffices.
Perhaps i should split the status test into a DEBUG clause.

Narann:
> Rhe path:ray_length attribute should always be available, always.

Thanks for the answer. It was what I guessed.

About the "red flag", I have no idea what is the OSL idiom in such situation (mandatory attribute is missing) so do what you want. Maybe a comment can be enough because this is not an _obvious_ logic (a simple "// "path:ray_length" must exists so we stop tracing as a red flag" could help).

Do it if you think it make sense.

PR #1542, opened 2017-07-17
Empty bdpt lighting engine

aytekaman:


src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
```diff
@@ -618,6 +618,7 @@ namespace
         {
             QComboBox* combobox = create_combobox("engine");
             combobox->setToolTip(m_params_metadata.get_path("lighting_engine.help"));
+            combobox->addItem("Bidirectional Path Tracer", "bdpt");
```
dictoon: I suggest making it the second item so that engines get sorted from the simplest (unidirectional PT) to the most advanced (probably VCM, eventually).

src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
```diff
@@ -632,6 +633,7 @@ namespace
           : LightingPanel(config, parent)
         {
             QComboBox* combobox = create_combobox("engine");
+            combobox->addItem("Bidirectional Path Tracer", "bdpt");
```
dictoon: Same remark.

src/appleseed/renderer/kernel/rendering/renderercomponents.cpp
```diff
@@ -143,6 +144,13 @@ bool RendererComponents::create_lighting_engine_factory()
     {
         return true;
     }
+    else if (name == "bdpt")
```
dictoon: Move in second position, like in the dropdown.

PR #1545, opened 2017-07-19
Volumetric entities redesign

Biart95:
- Replace Phase Function entity by Volume entity
- Move maths behind volume sampling into a new `PhaseFunction` class
- Include reparametrization from #1533 [ Excluded =( ]

Pros:
- Avoid code duplication between different phase function models
- Intuitive name that will not scare users
- Names are more logical now, i.e. phase function is a pure mathematical object and does not contain volume properties.

Cons:
- Limited capabilities for adding new phase function models (but we hardly need more)
- `volume.h` already exists in appleseed codebase, which may lead to confusion


src/appleseed/foundation/math/phasefunction.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+using namespace std;
+
+namespace foundation
+{
+
+namespace
+{
+
+inline float henyey_pdf(const float g, const float sqr_g, const float cosine)
```
dictoon: Indent methods in anonymous namespaces.

The only reason we don't indent anything in top-level namespaces is that this would mean basically indenting the whole file, which doesn't help anything. But we indent everything else.

src/appleseed/foundation/math/phasefunction.h
```diff
@@ -0,0 +1,78 @@
...
+// appleseed.renderer headers.
+#include "foundation/math/vector.h"
+
+namespace foundation
+{
+
+class PhaseFunction
+{
+  public:
+    virtual float evaluate(const Vector3f& outgoing, const Vector3f& incoming) const = 0;
```
dictoon: It's a good practice to always add empty virtual destructor to virtual classes:

    virtual ~PhaseFunction() {}

(One of the rare place where it's ok to define the function in the header file.)

src/appleseed/renderer/kernel/lighting/materialsamplers.h
```diff
@@ -45,7 +45,7 @@ namespace renderer  { class ShadingContext; }
 namespace renderer  { class ShadingPoint; }
 
 //
-// This file contains wrappers over samplable functions, such as BSDF and PhaseFunction.
+// This file contains wrappers over samplable functions, such as BSDF and Volume.
```
dictoon: *such as BSDFs and volumes*

src/appleseed/renderer/kernel/lighting/pathtracer.h
```diff
@@ -556,7 +556,7 @@ size_t PathTracer<PathVisitor, VolumeVisitor, Adjoint>::trace(
         shading_points[shading_point_index].clear();
         const ShadingRay::Medium* current_medium = next_ray.get_current_medium();
         if (current_medium != 0 &&
```
dictoon: Write as a single line? Also use `nullptr`.

src/appleseed/renderer/meta/tests/test_volume.cpp
```diff
@@ -148,23 +162,23 @@ TEST_SUITE(Renderer_Modeling_PhaseFunction)
...
+                integral += volume.evaluate(data, shading_ray, 0.5f, incoming);
             }
 
             return integral * foundation::FourPi<float>() / NumberOfSamples;
         }
 
         // Check if probabilistic sampling is consistent with the returned PDF values.
         static bool check_sampling_consistency(
-            PhaseFunction& phase_function,
+            Volume& volume,
```
dictoon: Align arguments vertically.

src/appleseed/renderer/meta/tests/test_volume.cpp
```diff
@@ -71,22 +71,28 @@ using namespace foundation;
...
-TEST_SUITE(Renderer_Modeling_PhaseFunction)
+TEST_SUITE(Renderer_Modeling_Volume)
 {
     struct Fixture
       : public TestFixtureBase
     {
+        Fixture()
+          : TestFixtureBase()
+        {
+            m_base_parameters = ParamArray().insert("phase_function_model", "henyey");
```
dictoon: I think the tests would be clearer if we would get rid of `m_base_parameters` and just repeat the phase function model.

src/appleseed/renderer/kernel/lighting/materialsamplers.h
est77: sampleable?

dictoon: Both appear to be correct: https://en.wiktionary.org/wiki/samplable#English

PR #1553, opened 2017-07-23
Miscellaneous fixes and cleanups

dictoon:


src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
```diff
@@ -155,42 +155,47 @@ void DirectLightingIntegrator::compute_outgoing_radiance_light_sampling_low_vari
...
+            // Add the contribution of the chosen light.
             if (sample.m_triangle)
             {
                 add_emitting_triangle_sample_contribution(
                     sampling_context,
                     sample,
                     mis_heuristic,
                     outgoing,
-                    radiance);
+                    ls_radiance);
```
gospodnetic: I suggest using the name `sample_radiance` or `light_sample_radiance`.

dictoon: That wouldn't be accurate, it's not the radiance of one particular light sample, but of _all_ light samples for the light set (hence the `ls_` prefix). Maybe `lightset_radiance`?

PR #1554, opened 2017-07-23
Studio python work

glebmish:


src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -201,7 +201,7 @@ void MainWindow::new_project()
     on_project_change();
 }
 
-void MainWindow::open_project(const QString& filepath)
+void MainWindow::open_project_sync(const QString& filepath)
```
est77: Lets remove the _sync suffix here and in load_project_sync too.
Everything is sync by default, the suffix does not add any extra information.

src/appleseed.studio/python/module.cpp
```diff
@@ -68,7 +68,7 @@ Project* new_project()
 
 Project* open_project(const char* project_path)
```
est77: Checking the prototype of MainWindow::open_project it returs a bool if opening fails.
The way we are exposing the method now, we loose the result of MainWindow::open_project and 
we don't have a way to know if opening was successful or if it failed.

Can we return a bool here?
Users can still retrieve the project from python using current_project().

glebmish: hm
`void MainWindow::open_project`

est77: Sorry, it's MainWindow::do_load_project the method that returns a bool.
Only in the new open_project sync method, can we return this bool to python?

glebmish: I don't know if it's a good idea. Maybe add function like `open_successfull` or make open_project return NULL (None) in case of error? I'm trying to imagine how script would look like and I don't like it very much
```Python
project = None
if studio.open_project(project_path):
    project = studio.current_project()
else:
    throw Exception(error_msg)
project.some_func()
```
what I don't like here is that semantic is wrong. You can expect that open_project returns opened project but it isn't. Also two different behaviors - performing action and querying information - are mixed here in `studio.open_project`

other two approaches:
1. open_successfull
```Python
project = studio.open_project(project_path)
if not studio.open_successfull():
    throw Exception(error_msg)
project.some_func()
```
2. return None
```Python
project = studio.open_project(project_path)
if not project:
    throw Exception(error_msg)
project.some_func()
```

Last one feels like "Python way" for me

dictoon: I think I agree, I also prefer the second option. Moreover, the user is free to ignore the return value and use `current_project()` to check if the project has been successfully loaded:

    studio.open_project(project_path)
    if studio.current_project() is None:
       throw Exception(error_msg)

est77: I prefer the boolean. load_project if it fails it returns false and leaves the previously opened project (if any) open. Users can't test studio.current_project() because that might point to another project.

Also, I don't think there's a requirement for open_project to return a Project.
The real purpouse of the function is to replace the currently opened project in studio by a new one. We should not promote users storing references to studio Projects in variables. 
They will become invalid the next time you do new or open project.

Lets return a boolean from open project. If it's possible, lets try to print the open error (if any) to the python console to give some feedback to users, and for consistency, lets not return anything from new_project.

glebmish: I added exception in case of failed open. I think fail to open project is a situation that should be explicitly resolved by script writer because if `open_project` will just return false it can be ignored and all following actions will be performed on wrong project. With exception approach I think bool status is useless here

PR #1557, opened 2017-07-27
Refactorings and cleanups

dictoon:


src/appleseed/renderer/kernel/lighting/lighttree.h
```diff
@@ -101,10 +101,10 @@ class LightTree
         // and emitting_triangles vectors within the BackwardLightSampler.
         Item(
             const foundation::AABB3d&       bbox,
-            const size_t                    source_index,
+            const size_t                    light_source_index,
             const size_t                    external_source_index)
```
gospodnetic: Maybe to rename this one just to stay consistent?

dictoon: In fact, if possible, I'd like to completely get rid of this. Let's leave it as is for now.

PR #1558, opened 2017-07-27
Studio python bugfixes

glebmish:


src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -271,12 +274,14 @@ void MainWindow::save_project(QString filepath)
     if (m_project_file_watcher)
         stop_monitoring_project_file();
 
-    m_project_manager.save_project_as(filepath.toAscii().constData());
+    bool successful = m_project_manager.save_project_as(filepath.toAscii().constData());
```

src/appleseed.studio/python/module.cpp
```diff
@@ -74,15 +74,13 @@ bool save_project(const char* project_path = nullptr)
 {
     if (project_path == nullptr)
     {
-        if (project_manager()->get_project()->has_path())
-            main_window()->save_project(project_manager()->get_project()->get_path());
+        if (project_manager()->get_project() && project_manager()->get_project()->has_path())
```
dictoon: Let's move `project_manager()->get_project()` to a local variable :)

PR #1566, opened 2017-07-30
Add OSL object and obj instance names

luisbarrancos:


src/appleseed/renderer/modeling/object/object.h
```diff
@@ -92,9 +92,18 @@ class APPLESEED_DLLSYMBOL Object
     // Return whether surface shaders should be invoked for fully transparent shading points.
     bool shade_alpha_cutouts() const;
 
+    void* get_oiio_name() const;
+
+    // Destructor.
+    ~Object();
```
est77: We could try making the destructor protected.

dictoon:
A few remarks:

- Right now there's a bug if an entity's name is changed with `set_name()`
- We need to change the name, in particular we shouldn't mention oiio as it's confusing
- We should consider moving that to the base `Entity` class (and it already has a private impl)

src/appleseed/renderer/modeling/entity/entity.h
```diff
@@ -102,6 +102,9 @@ class APPLESEED_DLLSYMBOL Entity
     void set_name(const char* name);
     const char* get_name() const;
 
+    // Return the name of the object as a pointer to a ustring for OSL.
```
dictoon: *object* -> *entity*

dictoon: *ustring* -> *OIIO::ustring*

PR #1569, opened 2017-08-01
Implemented studio.add_dock function

glebmish:


src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -313,6 +313,28 @@ ParamArray& MainWindow::get_settings()
...
 }
 
+QDockWidget* MainWindow::create_dock(const char* dock_name)
+{
+    QDockWidget* dock_widget = new QDockWidget(this);
+
+    QString object_name = QString(dock_name).toLower().split(' ').join("_");
+    dock_widget->setObjectName(object_name);
+    dock_widget->setWindowTitle(dock_name);
+    dock_widget->setAllowedAreas(Qt::BottomDockWidgetArea);
```
dictoon: Is this restriction intended?

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -313,6 +313,28 @@ ParamArray& MainWindow::get_settings()
     return m_settings;
 }
 
+QDockWidget* MainWindow::create_dock(const char* dock_name)
+{
+    QDockWidget* dock_widget = new QDockWidget(this);
+
+    QString object_name = QString(dock_name).toLower().split(' ').join("_");
```

src/appleseed.studio/mainwindow/mainwindow.cpp
glebmish: yes to simulate behavior of python console and log. It can be changed in script. I can delete it if you think it shouldn't be default.

dictoon: The Python console, Log panel and all other docks can be docked anywhere. It should be the same with user panels.

est77: If we remove the line, we can let the users call setAllowedAreas from python.

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -313,6 +313,28 @@ ParamArray& MainWindow::get_settings()
...
+    dock_widget->setObjectName(object_name);
+    dock_widget->setWindowTitle(dock_name);
+    dock_widget->setAllowedAreas(Qt::BottomDockWidgetArea);
+
+    addDockWidget(Qt::BottomDockWidgetArea, dock_widget);
```
est77: We should probably remove this line too. Users can call addDockWidget and place the dock in the most convenient dock area for what they want to do.

src/appleseed.studio/python/module.cpp
```diff
@@ -124,4 +133,5 @@ BOOST_PYTHON_MODULE(_appleseedstudio)
     bpy::def("is_project_dirty", is_project_dirty);
 
     bpy::def("main_window", main_window_as_pylong);
+    bpy::def("create_dock", create_dock, bpy::args("dock_name"));
```
est77: I suggest renaming the method and python function name to create_dock_widget

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -313,6 +313,35 @@ ParamArray& MainWindow::get_settings()
...
 }
 
+QDockWidget* MainWindow::create_dock_widget(const char* dock_name)
+{
+    QDockWidget* dock_widget = new QDockWidget(this);
+
+    const QString object_name = QString(dock_name).toLower().split(' ').join("_");
+    dock_widget->setObjectName(object_name);
+    dock_widget->setWindowTitle(dock_name);
+
```
dictoon: Remove redundant blank line.

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -313,6 +313,35 @@ ParamArray& MainWindow::get_settings()
...
+
+    const QList<QAction*>& actions = m_ui->menu_view->actions();
+    QAction* menu_separator = actions.last();
+    for (size_t i = (size_t) actions.size() - 2; i != 0; --i)
```
dictoon: Use `static_cast<>()`.
Add curly braces around single-statement but multi-line `for` loop body.

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -313,6 +313,35 @@ ParamArray& MainWindow::get_settings()
...
+QDockWidget* MainWindow::create_dock_widget(const char* dock_name)
+{
+    QDockWidget* dock_widget = new QDockWidget(this);
+
+    const QString object_name = QString(dock_name).toLower().split(' ').join("_");
+    dock_widget->setObjectName(object_name);
+    dock_widget->setWindowTitle(dock_name);
+
+
+    const QList<QAction*>& actions = m_ui->menu_view->actions();
```
dictoon: Probably a good place to use `const auto& actions` :)

src/appleseed.studio/python/module.cpp
```diff
@@ -110,6 +111,14 @@ bpy::long_ main_window_as_pylong()
     return bpy::long_(ptr);
 }
 
+bpy::long_ create_dock_widget(const char* dock_name)
+{
+    QDockWidget* dock_widget = main_window()->create_dock_widget(dock_name);
+
+    const uintptr_t ptr = binary_cast<uintptr_t>(dock_widget);
```
dictoon: Include `foundation/utility/casts.h`.

glebmish: thanks, I'll fix this remarks in next pr

PR #1570, opened 2017-08-02
Fix building & packaging on macOS

dictoon:


dictoon:
It's expected that Travis cannot build 1.7 anymore since the dependency packages was updated to no longer bundle Python.

PR #1575, opened 2017-08-03
LightSamplerBase lambdas

gospodnetic:
We created a `LightSamplerBase` to contain overlapping functions used by both `ForwardLightSampler` and `BackwardLightSampler`. However, some of those functions overlap _almost_ completely. It is this _almost_ part that bugged us. In order to avoid duplication, we moved those functions completely to the `LightSamplerBase`, which implied that the `LightSamplerBase` must contain some functions which should logically be only in the `BackwardLightSampler` (i.e. `m_light_tree_lights`, and `m_use_light_tree`).

We wanted to avoid the useless members using lambdas.

If we introduce lambdas to the `LightSamplerBase` there is no duplication whatsoever and useless members are gone...however, I am not so happy with this approach...
I find it hard to follow and way too much complicated for something that we could have avoided just by leaving the things as they are now and using the `m_use_light_tree` flag. In that case, however, we have a useless member m_light_tree_lights within ForwardLightSampler
If I extract the code outside the LightSamplerBase, we will have duplication. To be honest, I don't think we want it even temporarily.

Also, I can remove the emitting triangle lambda and leave only the non physical light lambda.
In that case we would at least avoid having useless m_light_tree_lights but the m_use_light_tree would still stay. But, the code would still be hard to follow. The fact that I dislike the most is the need to write code within the initializer list - it is something I don't find logical when reading the code... Also, within that code, we refer to a piece of code which is waaay lower in the file and the reader must scroll up and down to understand what the heck is going on here.

Anyway, I would really appreciate your thoughts on this.

### DIFF
--- a/src/appleseed/renderer/kernel/lighting/lightsamplerbase.h
+++ b/src/appleseed/renderer/kernel/lighting/lightsamplerbase.h
@@ -45,6 +45,7 @@
 // Forward declarations.
 namespace renderer  { class Assembly; }
 namespace renderer  { class AssemblyInstance; }
+namespace renderer  { class Material; }
 namespace renderer  { class MaterialArray; }
 
 namespace renderer
@@ -91,7 +92,9 @@ class LightSamplerBase
         const Light&)>                          LightHandlingLambda;
     
     typedef std::function<void (
-        const EmittingTriangle&)>               TriangleHandlingLambda;
+        const Material*,
+        const float,
+        const size_t)>                          TriangleHandlingLambda;
  
     const Parameters                        m_params;
  
@@ -110,6 +113,7 @@ class LightSamplerBase
     bool                                    m_use_light_tree;
 
     LightHandlingLambda                     m_cdf_light_handling;
+    TriangleHandlingLambda                  m_cdf_triangle_handling;
 
     // Build a hash table that allows to find the emitting triangle at a given shading point.
     void build_emitting_triangle_hash_table();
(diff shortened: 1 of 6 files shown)


src/appleseed/renderer/kernel/lighting/lightsamplerbase.h
```diff
@@ -83,8 +86,15 @@ class LightSamplerBase
     typedef std::vector<EmittingTriangle>       EmittingTriangleVector;
     typedef foundation::CDF<size_t, float>      EmitterCDF;
 
+    typedef std::function<void (
```
dictoon: I think we can completely get rid of the type and just pass `std::function<>` directly. We can do that because we can use `auto` at the site of declaration of the lambda.

Edit: I see that we need to pass the lambda to multiple sub-functions, so maybe it makes sense to have a dedicated type.

src/appleseed/renderer/kernel/lighting/lightsamplerbase.h
```diff
@@ -93,12 +103,14 @@ class LightSamplerBase
...
     EmitterCDF                              m_emitting_triangles_cdf;
-
+ 
     EmittingTriangleKeyHasher               m_triangle_key_hasher;
     EmittingTriangleHashTable               m_emitting_triangle_hash_table;
-
+ 
     bool                                    m_use_light_tree;
 
+    LightHandlingLambda                     m_cdf_light_handling;
```
dictoon: I'm not sure why we need to store the lambda here?

Edit: Ah, I see, you're trying to factorize the lambda. Not a good idea, just duplicate the lambda code in both samplers.

PR #1576, opened 2017-08-03
Miscellaneous fixes and improvements

dictoon:


src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -323,7 +323,7 @@ QDockWidget* MainWindow::create_dock_widget(const char* dock_name)
 
     const auto& actions = m_ui->menu_view->actions();
     QAction* menu_separator = actions.last();
-    for (size_t i = static_cast<size_t>(actions.size()) - 2; i != 0; --i)
```
glebmish: previously you recommended to use size_t. Why it's not the case here?

src/appleseed.python/CMakeLists.txt
```diff
@@ -44,7 +44,25 @@ endif ()
 # Source files.
 #--------------------------------------------------------------------------------------------------
 
-set (appleseed_module_sources
+set (studio_sources
+    studio/__init__.py
```
glebmish: oh, I forgot to add studio/Qt.py here in my commits. Could you please add it? or I can do it later

src/appleseed.studio/mainwindow/mainwindow.cpp
dictoon: Yes, in general `size_t` is the right type to use when iterating over collections. However, Qt's API uses `int` everywhere so it makes more sense to use `int`, otherwise you'll need casts everywhere.

For instance, here `actions.size()` returns a `int` and `actions[]` requires indexing with an `int` (but right now we're passing a `size_t` so it fails to compile on Windows), so it's just much easier to use an `int` as well in the for loop.

src/appleseed.python/CMakeLists.txt
dictoon: No prob, I'll add it right away.

PR #1577, opened 2017-08-04
Fix most of the warnings in appleseed GCC 7 build 

Biart95:
by omitting gcc warning 'dynamic exception specifications are deprecated' in OSL headers.

### DIFF
--- a/src/appleseed/foundation/platform/_beginoslheaders.h
+++ b/src/appleseed/foundation/platform/_beginoslheaders.h
@@ -34,3 +34,10 @@
     #pragma warning (disable: 4290)     // C++ exception specification ignored except to indicate a function is not __declspec(nothrow)
 
 #endif
+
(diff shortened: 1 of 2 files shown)


src/appleseed/foundation/platform/_beginoslheaders.h
```diff
@@ -34,3 +34,10 @@
     #pragma warning (disable: 4290)     // C++ exception specification ignored except to indicate a function is not __declspec(nothrow)
 
 #endif
+
+#if __GNUC__ >= 7
+
+	#pragma GCC diagnostic push
```
dictoon: Replace tabs by spaces (in both files).

Biart95: Oops, it seems I forgot to change settings in my editor. Thanks.

PR #1578, opened 2017-08-05
Studio python module

glebmish:


cmake/config/linux-gcc.txt
```diff
@@ -309,7 +309,7 @@ macro (add_copy_py_file_to_sandbox_py_module_command py_src module_name)
     get_relative_path_from_module_name (relative_module_path ${module_name})
     set(module_path "${py_path}/${relative_module_path}")
 
-    add_custom_command (TARGET appleseed.python POST_BUILD
+    add_custom_command (TARGET appleseed.python.copy_py_files POST_BUILD
```
est77: you need to make the same changes in cmake/config/windows.txt and the one for OSX too.

src/appleseed.python/studio/ui.py
```diff
@@ -0,0 +1,58 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+
+import Qt
+
+def wrapinstance(addr, type):
```
est77: type is maybe a reserved python keyword. We can rename the argument to cls (short for class) just in case.

src/appleseed.python/studio/ui.py
```diff
@@ -0,0 +1,58 @@
...
+    # Function won't change and will throw Exception for PySide2 and PyQt5.
+    raise Exception("No wrapinstance function defined for " + Qt.__binding__)
```
est77: I think we can do all in wrapinstance. I see that you are trying to import sip or shiboken only once, but importing a module that's already imported is not expensive, just a dictionary lookup, and this function is not going to be used in any kind of performance critical code.

I suggest:

if Qt.__binding__ == 'PyQt4':
     import sip
     return sip.wrapInstance(...)
elif Qt.__binding__ == 'PySide':
    ...
else:
   raise Exception(...)

src/appleseed.python/studio/ui.py
glebmish: I checked docs and made it consistent with what names are used in modules being wrapped
https://shiboken.readthedocs.io/en/latest/shibokenmodule.html
http://pyqt.sourceforge.net/Docs/sip4/python_api.html

src/appleseed.python/studio/ui.py
glebmish: yes, I tried to avoid if-elif chain and repetitive importing. If it's not too expensive I'll make function easier as you suggested

cmake/config/linux-gcc.txt
glebmish: ah yes, forgot about it

glebmish: I don't really like the target name I chose. Maybe you have some ideas?

est77: I don't think it's bad. I don't have anything better now.
It's an internal name that is not visible anywhere.

src/appleseed.python/studio/ui.py
est77: Ok. Lets keep type then.

cmake/config/linux-gcc.txt
glebmish: it is, that's why I wonder
```
$ make 
[  0%] Built target lz4
[ 44%] Built target appleseed
[ 44%] Built target SANDBOX
[ 59%] Built target appleseed.shaders
[ 59%] Built target appleseed.shared
[ 60%] Built target appleseed.cli
[ 63%] Built target appleseed.python
[ 63%] Built target appleseed.python.copy_py_files
[ 98%] Built target appleseed.studio
[ 98%] Built target animatecamera
[ 98%] Built target convertmeshfile
[ 99%] Built target dumpmetadata
[ 99%] Built target makefluffy
[100%] Built target projecttool
```

PR #1582, opened 2017-08-09
Added TXConverter() class and studio-specific function to convert textures

glebmish:


src/appleseed.python/CMakeLists.txt
```diff
@@ -180,8 +180,11 @@ add_custom_target(appleseed.python.copy_py_files ALL)
 add_copy_target_to_sandbox_py_module_command (appleseed.python appleseed)
 add_copy_py_file_to_sandbox_py_module_command (${PROJECT_SOURCE_DIR}/src/appleseed.python/__init__.py appleseed)
 add_copy_py_file_to_sandbox_py_module_command (${PROJECT_SOURCE_DIR}/src/appleseed.python/logtarget.py appleseed)
+add_copy_py_file_to_sandbox_py_module_command (${PROJECT_SOURCE_DIR}/src/appleseed.python/TXConverter.py appleseed)
```
dictoon: We should stick to lower case filenames.

src/appleseed.python/TXConverter.py
```diff
@@ -0,0 +1,38 @@
...
+        return tx_path
+
+    def _change_ext_to_tx(self, path):
+        extension = '.tx'
+        if path.rfind(os.sep) < path.rfind('.'):
+            path_to_tx = path[:path.rfind('.')] + extension
+        else:
+            path_to_tx += extension
+
+        return path_to_tx
```
dictoon: Add newline on the last line.

src/appleseed.python/TXConverter.py
```diff
@@ -0,0 +1,38 @@
...
+            maketx=self.maketx_path, tx_texture=tx_path, texture=path))
+
+        if status != 0:
+            print('maketx failed with error code {}'.format(status))
+            return None
+
+        return tx_path
+
+    def _change_ext_to_tx(self, path):
+        extension = '.tx'
```
dictoon: You're better off using https://docs.python.org/3/library/os.path.html#os.path.splitext instead of doing it manually.

src/appleseed.python/TXConverter.py
```diff
@@ -0,0 +1,38 @@
...
+        tx_path = self._change_ext_to_tx(path)
+        if os.path.exists(tx_path):
+            print('Warning: {tx_texture} exists.'.format(tx_texture=tx_path))
+            return None
+
+        status = os.system('{maketx} -o "{tx_texture}" "{texture}"'.format(
+            maketx=self.maketx_path, tx_texture=tx_path, texture=path))
+
+        if status != 0:
+            print('maketx failed with error code {}'.format(status))
```
dictoon: I didn't know you could use simply `{}` in a string, are you sure this is legal? I couldn't find any docs mentioning this but I may have overlooked it.

src/appleseed.python/TXConverter.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def convert(self, path):
+        if path in self.converted:
+            return self.converted[path]
+        else:
+            path_converted = self._convert(path)
+            self.converted[path] = path_converted
+            return path_converted
+
+    def _convert(self, path):
```
dictoon: I find it slightly confusing to have two methods with the same name but with different visibility. Maybe `_do_convert()`?

src/appleseed.python/studio/helpers.py
```diff
@@ -0,0 +1,35 @@
...
+        if texture_path.endswith('.tx'):
+            continue
+
+        new_texture_path = tx_converter.convert(texture_path)
+        if new_texture_path is None:
+            print('Skipped converting of {texture}'.format(texture=texture_path))
+        else:
+            texture_parameters['filename'] = new_texture_path
+            texture.set_parameters(texture_parameters)
+            print('{} converted to {}'.format(texture_path, new_texture_path))
```
dictoon: Add newline on the file's last line.

src/appleseed.python/studio/helpers.py
```diff
@@ -0,0 +1,35 @@
...
+
+def convert_all_textures_to_tx(maketx_path):
+    scene = studio.current_project().get_scene()
+    textures = get_textures(scene)
+
+    tx_converter = asr.TXConverter(maketx_path)
+
+    for texture in textures:
+        texture_parameters = texture.get_parameters()
+        texture_path = texture_parameters['filename']
```
dictoon: You need to only collect textures of types you know how to deal with, i.e. `texture2d`, because not all texture types with have a `filename` parameter.

src/appleseed.python/TXConverter.py
```diff
@@ -0,0 +1,38 @@
+import os
+
+class TXConverter:
```
est77: Rename to TextureConverter

src/appleseed.python/TXConverter.py
```diff
@@ -0,0 +1,38 @@
...
+            return self.converted[path]
+        else:
+            path_converted = self._convert(path)
+            self.converted[path] = path_converted
+            return path_converted
+
+    def _convert(self, path):
+        tx_path = self._change_ext_to_tx(path)
+        if os.path.exists(tx_path):
+            print('Warning: {tx_texture} exists.'.format(tx_texture=tx_path))
```
est77: lets use the standard logging module instead of prints

src/appleseed.python/__init__.py
```diff
@@ -29,3 +29,4 @@
 
 from _appleseedpython import *
 from logtarget import *
+from TXConverter import *
```
est77: we don't need to import TXConverter here. Most users of appleseed.python will not use it.

src/appleseed.python/studio/__init__.py
```diff
@@ -27,3 +27,4 @@
 #
 
 from _appleseedstudio import *
+from ui import *
```
est77: We don't need to import ui here. Because it imports Qt, it will fail for users without any Qt binding.

src/appleseed.python/TXConverter.py
glebmish: yes, it's a common pattern
there are plenty of examples here: https://docs.python.org/2/library/string.html#format-string-syntax, https://docs.python.org/2/library/string.html#format-examples

src/appleseed.python/studio/helpers.py
glebmish: then how to determine if it's a texture I can deal with?

src/appleseed.python/TXConverter.py
glebmish: change on _convert_with_maketx. ok?

src/appleseed.python/TXConverter.py
glebmish: had to change it because of using logging module

src/appleseed.python/TXConverter.py
glebmish: after changing prints on logging module logs aren't printed to stdout. Do you have any thought why?

src/appleseed.python/textureconverter.py
```diff
@@ -0,0 +1,31 @@
+import os
+import logging
+
+class TextureConverter:
```
est77: Make a new style class:

class TextureConverter(object):

src/appleseed.python/__init__.py
```diff
@@ -29,3 +29,7 @@
 
 from _appleseedpython import *
 from logtarget import *
+
+import logging
```
est77: We should not initialize logging here. Users importing appleseed.python might not need logging and if they need it, they probably want to initialize it themselves with a config that makes sense for them.

For now, lets move this code to each place where we use logging until we find a better way to package the texture converter tool.

src/appleseed.python/TXConverter.py
glebmish: it works now

src/appleseed.python/studio/helpers.py
```diff
@@ -0,0 +1,40 @@
...
+
+        if texture_path.endswith('.tx'):
+            continue
+
+        new_texture_path = tx_converter.convert(texture_path)
+        if new_texture_path is None:
+            logging.info('Skipped converting of {}'.format(texture_path))
```
dictoon: *Skipped conversion of*

src/appleseed.python/textureconverter.py
```diff
@@ -0,0 +1,33 @@
...
+            path_converted = self._convert_with_maketx(path)
+            self.converted[path] = path_converted
+            return path_converted
+
+    def _convert_with_maketx(self, path):
+        base_path, _ = os.path.splitext(path)
+        tx_path = base_path + ".tx"
+
+        if os.path.exists(tx_path):
+            logging.warning('{} exists.'.format(tx_path))
```
dictoon: *already exists*

src/appleseed.python/studio/helpers.py
```diff
@@ -28,7 +28,8 @@ def convert_all_textures_to_tx(maketx_path):
         texture_parameters = texture.get_parameters()
         texture_path = texture_parameters['filename']
 
-        if texture_path.endswith('.tx'):
+        if texture.get_model() != 'disk_texture_2d' or texture_path.endswith('.tx'):
```
est77: we can split the two conditions in two tests. It's natural that we ignore non image textures, we should not log anything if get_model() is not a disk_texture_2d.

src/appleseed.python/bindproject.cpp
```diff
@@ -250,6 +250,11 @@ namespace
         auto_release_ptr<Project> project(reader->load_builtin(project_name));
         return bpy::object(project);
     }
+
+    string qualify_path(const Project* project, const char* filepath) {
```
est77: Move the brace to the next line

src/appleseed.python/studio/helpers.py
```diff
@@ -28,14 +37,23 @@ def convert_all_textures_to_tx(maketx_path):
         texture_parameters = texture.get_parameters()
         texture_path = texture_parameters['filename']
 
-        if texture.get_model() != 'disk_texture_2d' or texture_path.endswith('.tx'):
-            logging.debug('Skipped converting {}'.format(texture_path))
+        logging.debug("Processing {}".format(texture_path))
+        print texture_path
```
est77: remove the print

src/appleseed.studio/python/module.cpp
```diff
@@ -135,4 +136,6 @@ BOOST_PYTHON_MODULE(_appleseedstudio)
 
     bpy::def("main_window", main_window_as_pylong);
     bpy::def("create_dock_widget", create_dock_widget, bpy::args("dock_name"));
+
+    boost::python::def("get_studio_dir", get_executable_directory);
```
est77: I think that the name is of this function is not quite correct. For me the studio directory would be the parent of the bin directory, the root of the appleseed install if you install appleseed using the released versions.

We could rename the function to make it clear that it returns the bin directory or we can make it return the root directory of the appleseed install. I think I prefer the last option myself.

dictoon: What about keeping the same name as the C++ function, `get_executable_directory()`? To me that sounds perfectly clear and unambiguous.

est77: Yes, that's one option I was suggesting, but I don't like that it will point to bin/Ship or bin/Debug when running studio from the sandbox.
If we return the top level studio dir instead, then maketx will be found always in result/bin and we could access other resources inside studio in an independent way.

dictoon: Then I think what you really want is `appleseed::shared::Application::get_root_path()`.

dictoon: `get_executable_directory()` should really return the path to the directory containing the executable. It seems sensible to expose it in the Python API. For everything else (in particular for finding resource files) you should use `appleseed::shared::Application::get_root_path()`.

est77: Ok. Lets expose only get_root_path. We don't need executable_directory as far as I know.

glebmish: so the path to make tx should always be `os.path.join(root_path, 'bin', 'maketx')`?

PR #1584, opened 2017-08-17
Complete color pipeline overhaul

dictoon:
The color pipeline is now configurable to use RGB colors or full spectra. The color pipeline mode ("spectrum mode") is stored in a thread-local static variable. All `DynamicSpectrum<>` objects now always have the same size within a thread. We no longer upgrade RGB colors to spectra or downgrade spectra to RGB colors. `DynamicSpectrum<>::resize()` is gone.

Spectra intent are now only use when converting between RGB colors and spectra, and are no longer stored. Consequently, the size of a 31-channel `DynamicSpectrum<>` is now exactly 128 bytes (31 floats + 1 float of padding).

The result of shading a pixel sample is now always in linear RGB and is represented by a `foundation::Color4f` object. The classes `ShadingFragment`, `ShadingFragmentStack` and `SpectrumStack` are gone.

Cornell Box-based test scenes and scenes from the "sun and sky" test suite have been switched to spectral mode; all other scenes are set to use RGB mode.

All test suite images were thoroughly checked, then updated.


src/appleseed/renderer/kernel/aov/aovaccumulator.h
```diff
@@ -37,14 +37,16 @@
...
 
+// Standard headers.
+#include <cstddef>
+
 // Forward declarations.
-namespace renderer { class Camera; }
-namespace renderer { class ShadingComponents; }
-namespace renderer { class ShadingPoint; }
-namespace renderer { class ShadingResult; }
+namespace renderer  { class Camera; }
```
Narann: Two spaces? I would be interested to know the standard using this. Is it personal taste?

src/appleseed/renderer/kernel/lighting/pathtracer.h
```diff
@@ -768,30 +768,33 @@ bool PathTracer<PathVisitor, VolumeVisitor, Adjoint>::march(
...
         const float s = sampling_context.next2<float>();
-        const size_t channel = foundation::truncate<size_t>(s * channel_count);
-        const bool extinction_is_null = (extinction_coef[channel] < 1.0e-6f);
+        const size_t channel = foundation::truncate<size_t>(s * Spectrum::size());
+        const bool extinction_is_null = extinction_coef[channel] < 1.0e-6f;
 
-        float distance_sample = 0.0f;
-        float distance_pdf = 0.0f;
-        if (!extinction_is_null)
+        // Sample distance.
```
Narann: I suggest to remove the comment or improve it. The original comment position looks clearer IMHO.

src/appleseed/renderer/kernel/lighting/pathtracer.h
```diff
@@ -768,30 +768,33 @@ bool PathTracer<PathVisitor, VolumeVisitor, Adjoint>::march(
...
+        const Spectrum& extinction_coef =
+            volume->extinction_coefficient(vertex.m_volume_data, volume_ray);
         
         sampling_context.split_in_place(1, 2);
 
         // Sample channel uniformly at random.
         const float s = sampling_context.next2<float>();
-        const size_t channel = foundation::truncate<size_t>(s * channel_count);
-        const bool extinction_is_null = (extinction_coef[channel] < 1.0e-6f);
+        const size_t channel = foundation::truncate<size_t>(s * Spectrum::size());
```
Narann: Does it mean in rgb `Spectrum::size()` return `3`?

src/appleseed/renderer/kernel/rendering/shadingresultframebuffer.cpp
```diff
@@ -131,7 +126,7 @@ void ShadingResultFrameBuffer::merge(
     const float* APPLESEED_RESTRICT source_ptr = source.pixel(source_x, source_y);
     float* APPLESEED_RESTRICT dest_ptr = pixel(dest_x, dest_y);
 
-    for (size_t i = 0; i < m_channel_count; ++i)
+    for (size_t i = 0, e = m_channel_count; i < e; ++i)
```
Narann: Why `e`? Why not `for (size_t i = 0; i <  m_channel_count; ++i)`?

src/appleseed/renderer/kernel/rendering/shadingresultframebuffer.cpp
```diff
@@ -152,7 +147,7 @@ void ShadingResultFrameBuffer::develop_to_tile_premult_alpha(
             tile.set_pixel(x, y, color * rcp_weight);
             ptr += 4;
 
-            for (size_t i = 0; i < m_aov_count; ++i)
+            for (size_t i = 0, e = m_aov_count; i < e; ++i)
```
Narann: Same here: Why `e`, Why not `for (size_t i = 0; i < m_aov_count; ++i)`? Access?

src/appleseed/renderer/kernel/rendering/shadingresultframebuffer.cpp
```diff
@@ -168,9 +163,9 @@ void ShadingResultFrameBuffer::develop_to_tile_straight_alpha(
 {
     const float* ptr = pixel(0);
 
-    for (size_t y = 0; y < m_height; ++y)
+    for (size_t y = 0, h = m_height; y < h; ++y)
     {
-        for (size_t x = 0; x < m_width; ++x)
+        for (size_t x = 0, w = m_width; x < w; ++x)
```
Narann: Mmh, you us this trick everywhere. Now I guess there is a reason.

src/appleseed/renderer/modeling/aov/depthaov.cpp
```diff
@@ -80,8 +81,7 @@ namespace
 
         virtual void flush(ShadingResult& result) override
         {
-            result.m_aovs[m_index].m_color.set(m_depth);
-            result.m_aovs[m_index].m_alpha.set(1.0f);
+            result.m_aovs[m_index] = Color4f(m_depth, m_depth, m_depth, 1.0f);
```
Narann: No `set()` method on `m_aovs` instead a copy? I don't know.

src/appleseed/renderer/modeling/aov/diffuseaov.cpp
```diff
@@ -73,18 +75,18 @@ namespace
...
-            m_color = shading_components.m_diffuse;
+            m_color = shading_components.m_diffuse.to_rgb(g_std_lighting_conditions);
             m_color *= multiplier;
         }
 
         virtual void flush(ShadingResult& result) override
         {
-            result.m_aovs[m_index].m_color = m_color;
-            result.m_aovs[m_index].m_alpha = result.m_main.m_alpha;
+            result.m_aovs[m_index].rgb() = m_color;
```
Narann: An assignment on a method `.rgb()`? What is that?

src/appleseed/renderer/utility/dynamicspectrum.h
```diff
@@ -833,32 +631,16 @@ APPLESEED_FORCE_INLINE DynamicSpectrum<float, 31>& operator+=(DynamicSpectrum<fl
...
-    else
-    {
-        DynamicSpectrum<T, N> up_rhs;
-        DynamicSpectrum<T, N>::upgrade(rhs, up_rhs);
-
-        for (size_t i = 0; i < N; ++i)
-            lhs[i] -= up_rhs[i];
-    }
+    for (size_t i = 0; i < DynamicSpectrum<T, N>::size(); ++i)
+        lhs[i] -= rhs[i];
```
Narann: No `_mm_sub_ps` version for 31?

sandbox/tests/test scenes/alpha/01 - opaque red over opaque blue.appleseed
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- File generated by appleseed.dll version 1.7.0-beta (SSE 4.2). -->
+<!-- File generated by appleseed.dll version 1.8.0-beta (SSE 2). -->
```
dictoon: It just so happens that my current build (which I used to automatically update all the test scenes) doesn't have SSE 4.2 enabled. Not a big deal.

src/appleseed/renderer/kernel/aov/aovaccumulator.h
dictoon: There's no standard really, I just press tab to align the opening brace on a tab stop...

src/appleseed/renderer/kernel/lighting/pathtracer.h
dictoon: Yes, correct.

src/appleseed/renderer/kernel/rendering/shadingresultframebuffer.cpp
dictoon: This is to help the reader and especially the compiler: this construction ensures that the loop's upper bound (`m_channel_count`) doesn't change during iteration, so the compiler doesn't have to reload it at each iteration.

src/appleseed/renderer/modeling/aov/depthaov.cpp
dictoon: Yeah, it bothers me a little sometimes too.

`m_avos[m_index]` is a `foundation::Color4f`, which has a `set()` method to set all components to the same value, but if you want to set all components individually, right now there's no other choice than making a copy. I'm pretty sure the compiler will elide this copy, but having a `set(r, g, b, a)` method would be another choice.

src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
```diff
@@ -703,15 +703,15 @@ namespace
         {
             const string widget_base_key = prefix + ".bounces.";
 
-            QSpinBox* max_bounces = create_integer_input(widget_base_key + "max_bounces", 0, 10000, 1);
+            QSpinBox* max_bounces = create_integer_input(widget_base_key + "max_bounces", 0, 100, 1);
```
est77: Would it make sense to make this a constant and reference it everywhere.
Would be useful if we need to adjust the max bounces limit in the future.

src/appleseed/renderer/kernel/aov/aovaccumulator.cpp
```diff
@@ -89,16 +88,12 @@ BeautyAOVAccumulator::BeautyAOVAccumulator()
 
 void BeautyAOVAccumulator::set(const Spectrum& value)
 {
-    m_color_space = value.is_rgb() ? ColorSpaceLinearRGB : ColorSpaceSpectral;
-    m_color = value;
+    m_color = value.to_rgb(g_std_lighting_conditions);
```
est77: Can we make the lighting conditions a thread local variable too?
In our color management PR, lighting conditions are selected as part of the working color space selection and g_std_lighting_conditions does not exist anymore.

src/appleseed/renderer/kernel/lighting/pt/ptlightingengine.cpp
```diff
@@ -1047,14 +1041,10 @@ namespace
...
-                Spectrum precomputed_mis_weights;
-                if (channel_count == scattering_coef.size())
-                    precomputed_mis_weights = scattering_coef;
-                else
-                    Spectrum::upgrade(scattering_coef, precomputed_mis_weights);
+                Spectrum precomputed_mis_weights = scattering_coef;
                 precomputed_mis_weights *= vertex.m_throughput;
                 float max_density = 0.0f;
-                for (size_t i = 0; i < channel_count; ++i)
+                for (size_t i = 0; i < Spectrum::size(); ++i)
```
est77: e = Spectrum::size(); i < e; ++i) ?

src/appleseed/renderer/kernel/lighting/sppm/sppmparameters.cpp
```diff
@@ -96,7 +96,8 @@ namespace
 }
 
 SPPMParameters::SPPMParameters(const ParamArray& params)
-  : m_sampling_mode(get_sampling_context_mode(params))
+  : m_spectrum_mode(get_spectrum_mode(params))
```
est77: Right now, we are collecting all color related parameters in the Frame entity.
Would it make sense to move spectrum_mode there too?

src/appleseed/renderer/modeling/aov/diffuseaov.cpp
dictoon: `rgb()` returns the first three components (R, G, B) as a `foundation::Color3f` reference so assignment works just fine.

src/appleseed/renderer/utility/dynamicspectrum.h
dictoon: Most arithmetic operations are not used (at least not in performance-sensitive contexts) so they haven't been optimized yet.

src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
dictoon: I thought about it but was lazy. But you're right, I'll do it.

src/appleseed/renderer/kernel/lighting/pt/ptlightingengine.cpp
dictoon: I was doing that at first, then reverted to this simpler form... But after more thinking, it appears that maybe you're right and that I should change it everywhere. I'll check if there's any difference in the generated code.

src/appleseed/renderer/modeling/aov/diffuseaov.cpp
Narann: Rhooo that's awesome! :D

src/appleseed/renderer/utility/dynamicspectrum.h
Narann: Ok, thanks!

PR #1591, opened 2017-08-20
Better importance sampling for volumes

Biart95:
Implements technique described in [Importance Sampling Techniques for Path Tracing in Participating Media](https://www.solidangle.com/research/egsr2012_volume).
And fixes a few bugs with volume lighting integration.


src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
```diff
@@ -927,11 +928,17 @@ namespace
             QFormLayout* sublayout = create_form_layout();
             layout->addLayout(sublayout);
 
-            QSpinBox* volume_distance_samples =
-                create_integer_input("advanced.volume_distance_samples", 1, 1000, 1);
-            volume_distance_samples->setToolTip(
-                m_params_metadata.get_path("pt.volume_distance_samples.help"));
-            sublayout->addRow("Distance Samples:", volume_distance_samples);
+            QSpinBox* equiangular_distance_samples =
+                create_integer_input("advanced.equiangular_distance_samples", 0, 1000, 1);
```
dictoon: Is 0 really the smallest valid value? Shouldn't it be 1?

src/appleseed/foundation/math/sampling/mappings.h
```diff
@@ -558,7 +558,7 @@ inline T exponential_distribution_on_segment_pdf(
     const T l,
     const T r)
 {
-    assert (x >= l && x < r);
+    assert (x >= l && x <= r);
```
dictoon: Are you sure the upper bound should be included? If yes, you'll need to update the function's description.

src/appleseed/foundation/math/ray.h
```diff
@@ -83,6 +83,11 @@ class Ray
     template <typename U>
     Ray(const Ray<U, N>& rhs);
 
+    ValueType get_length() const;
```
dictoon: Move after `is_finite()` and add short method description, in particular the fact that the ray must be finite.

src/appleseed/foundation/math/ray.h
```diff
@@ -230,6 +235,19 @@ inline Ray<T, N>::Ray(const Ray<U, N>& rhs)
 }
 
 template <typename T, size_t N>
+inline typename Ray<T, N>::ValueType Ray<T, N>::get_length() const
+{
+    assert(is_finite());
+    return foundation::norm((m_tmax - m_tmin) * m_dir);
```
dictoon: Small optimization: take the norm of `m_dir` *then* multiply by `m_tmax - m_tmin`. This will avoid two multiplications.

src/appleseed/foundation/math/ray.h
```diff
@@ -230,6 +235,19 @@ inline Ray<T, N>::Ray(const Ray<U, N>& rhs)
 }
 
 template <typename T, size_t N>
+inline typename Ray<T, N>::ValueType Ray<T, N>::get_length() const
```
dictoon: `typename Ray<T, N>::ValueType` is just `T`.

src/appleseed/renderer/kernel/shading/shadingpoint.h
```diff
@@ -114,8 +122,15 @@ class ShadingPoint
     // Return the time stored in the ray.
     const ShadingRay::Time& get_time() const;
 
-    // Return true if an intersection was found, false otherwise.
-    bool hit() const;
+    // Return true if the shading point is either
+    // a surface point or a volume point and can be shaded.
+    bool valid() const;
```
dictoon: I would recommend `is_valid()`. The only other place where we have a `valid()` method is on `foundation::CDF` but I'll change it to `is_valid()` as well.

src/appleseed/foundation/math/sampling/mappings.h
Biart95: Mathematically, the segment [l, r) fits better. But programmatically...
Due to imprecise calculations, the actual values are generated in (l - eps, r + eps), and I use `clamp` in `sample` to put values in the required range. Naturally, `clamp` gives me values in [l, r].
Probably, that should be reflected in the description, yes.

src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
Biart95: I made it possible to allow using only exponential/equiangular samples. But as we discussed in Slack, we are not going to use separate controls for different distance samples, so I will change it back to 1.

src/appleseed/foundation/math/sampling/mappings.h
Biart95: @dictoon Any suggestions how can I update the description?

dictoon: Maybe you can just update the description to specify [l, r] instead of [l, r)?

Biart95: Haha indeed

src/appleseed/renderer/kernel/lighting/volumelightingintegrator.cpp
```diff
@@ -0,0 +1,601 @@
...
+                    emission_direction,
+                    light_value,
+                    probability);
+            }
+            else
+                emission_position = light_sample.m_point;
+
+            const Vector3d light_dir = emission_position - volume_ray.m_org;
+
+            m_origin_to_center = dot(light_dir, volume_ray.m_dir);
```
dictoon: `m_origin_to_center` is not a pure angle right, since `light_dir` is not normalized.

src/appleseed/renderer/kernel/lighting/volumelightingintegrator.cpp
```diff
@@ -0,0 +1,601 @@
...
+                    m_sampling_context.next2<Vector2d>(),
+                    emission_position,
+                    emission_direction,
+                    light_value,
+                    probability);
+            }
+            else
+                emission_position = light_sample.m_point;
+
+            const Vector3d light_dir = emission_position - volume_ray.m_org;
```
dictoon: `light_dir` is a bit confusing since it's often *from* the light. What about `to_light`?

src/appleseed/renderer/kernel/lighting/volumelightingintegrator.cpp
dictoon: What about `m_angle_to_light`? `m_origin_to_center` makes it sound like it's a distance.

src/appleseed/renderer/kernel/lighting/volumelightingintegrator.cpp
```diff
@@ -0,0 +1,601 @@
...
+
+            m_origin_to_center = dot(light_dir, volume_ray.m_dir);
+
+            m_center_to_light =
+                std::sqrt(square_norm(light_dir) - square(m_origin_to_center));
+
+            m_near_angle = std::atan2(-m_origin_to_center, m_center_to_light);
+            m_far_angle =
+                (volume_ray.m_tmax == std::numeric_limits<double>::max()) ? HalfPi<double>() :
```
dictoon: You can use `is_finite()` here no?

src/appleseed/renderer/kernel/lighting/volumelightingintegrator.cpp
```diff
@@ -0,0 +1,601 @@
...
+            }
+            else
+                emission_position = light_sample.m_point;
+
+            const Vector3d light_dir = emission_position - volume_ray.m_org;
+
+            m_origin_to_center = dot(light_dir, volume_ray.m_dir);
+
+            m_center_to_light =
+                std::sqrt(square_norm(light_dir) - square(m_origin_to_center));
```
dictoon: Why not add `using namespace std` since you already do it for `foundation`?

src/appleseed/renderer/kernel/lighting/volumelightingintegrator.cpp
```diff
@@ -0,0 +1,601 @@
...
+            SamplingContext child_sampling_context = m_sampling_context.split(1, 1);
+            const double s = child_sampling_context.next2<double>();
+
+            return static_cast<float>(m_origin_to_center +
+                sample_equiangular_distribution(s, m_near_angle, m_far_angle, m_center_to_light));
+        }
+
+        float evaluate(const float distance_sample) const
+        {
+            return static_cast<float>(equiangular_distribution_pdf(
```
dictoon: My advice here: use `float` everywhere. That means changing the type of the four member variables to `float` and adding static casts where you compute them, in the constructor.

src/appleseed/renderer/kernel/lighting/volumelightingintegrator.cpp
```diff
@@ -0,0 +1,601 @@
...
+                m_center_to_light));
+        }
+
+      private:
+        double                  m_origin_to_center;
+        double                  m_center_to_light;
+        double                  m_near_angle;
+        double                  m_far_angle;
+
+        const ShadingContext&   m_shading_context;
```
dictoon: Move these two first. Order of declaration and initialization should match.

src/appleseed/renderer/kernel/lighting/volumelightingintegrator.cpp
```diff
@@ -0,0 +1,601 @@
...
+{
+    // Get full ray transmission.
+    Spectrum ray_transmission;
+    m_volume.evaluate_transmission(
+        m_volume_data, m_volume_ray, ray_transmission);
+
+    const Spectrum& scattering_coef = m_volume.scattering_coefficient(
+        m_volume_data, m_volume_ray);
+
+    Spectrum extinction_coef = m_volume.extinction_coefficient(
```
dictoon: Take const ref instead of making copy.

src/appleseed/renderer/kernel/lighting/volumelightingintegrator.cpp
```diff
@@ -0,0 +1,601 @@
...
+        m_volume_data, m_volume_ray);
+
+    // Precompute MIS weights.
+    // MIS terms are:
+    //  - scattering albedo,
+    //  - one minus total ray transmittance (because we sample distance on segment),
+    //  - throughput of the entire path up to the sampled point.
+    // Reference: "Practical and Controllable Subsurface Scattering
+    // for Production Path Tracing", p. 1 [ACM 2016 Article].
+    for (size_t i = 0; i < Spectrum::size(); ++i)
```
dictoon: We've decided to switch to

    for (size_t i = 0, e = Spectrum::size(); i < e; ++i)

for such loops.

Change in other places.

src/appleseed/renderer/kernel/lighting/volumelightingintegrator.cpp
```diff
@@ -0,0 +1,601 @@
...
+
+            m_near_angle = std::atan2(-m_origin_to_center, m_center_to_light);
+            m_far_angle =
+                (volume_ray.m_tmax == std::numeric_limits<double>::max()) ? HalfPi<double>() :
+                std::atan2(volume_ray.get_length() - m_origin_to_center, m_center_to_light);
+        }
+
+        float sample() const
+        {
+            SamplingContext child_sampling_context = m_sampling_context.split(1, 1);
```
dictoon: I'm a bit concerned by this. Let's pass the sampling context in argument, like we usually do, and see where this leads us.

src/appleseed/renderer/kernel/lighting/volumelightingintegrator.cpp
```diff
@@ -0,0 +1,601 @@
...
+            }
+        }
+    }
+
+    // Add contributions from the light set.
+    if (m_light_sampler.has_lightset())
+    {
+        for (size_t i = 0; i < 2 * m_distance_sample_count; ++i)
+        {
+            for (size_t j = 0; j < m_light_sample_count; ++j)
```
dictoon: Add braces around `for` body.

src/appleseed/renderer/kernel/lighting/volumelightingintegrator.cpp
```diff
@@ -0,0 +1,601 @@
...
+                add_single_distance_sample_contribution_exponential_only(
+                    nullptr,
+                    extinction_coef,
+                    sampling_context,
+                    mis_heuristic,
+                    radiance,
+                    false);
+            if (!ScatteringMode::has_volume(m_scattering_modes))
```
dictoon: Add braces around `if` body.

src/appleseed/renderer/kernel/lighting/volumelightingintegrator.cpp
dictoon: So we're taking `2 * m_light_sample_count * m_distance_sample_count` distance samples when using exponential sampling only...

src/appleseed/renderer/kernel/lighting/volumelightingintegrator.cpp
```diff
@@ -0,0 +1,601 @@
...
+                    sampling_context,
+                    mis_heuristic,
+                    radiance,
+                    true);
+        }
+    }
+}
+
+void VolumeLightingIntegrator::take_single_direction_sample(
+    const bool                  sample_phasefunction,
```
dictoon: `sample_phase_function`

src/appleseed/renderer/kernel/lighting/volumelightingintegrator.cpp
```diff
@@ -0,0 +1,601 @@
...
+            m_time,
+            sampling_context.next2<Vector3f>(),
+            m_shading_point,
+            light_sample);
+
+        integrator.add_emitting_triangle_sample_contribution(
+            sampling_context,
+            light_sample,
+            mis_heuristic,
+            foundation::Dual3d(m_volume_ray.m_dir),
```
dictoon: Remove redundant namespace qualifiers.

src/appleseed/renderer/kernel/lighting/volumelightingintegrator.cpp
```diff
@@ -0,0 +1,601 @@
...
+}
+
+float VolumeLightingIntegrator::draw_exponential_sample(
+    SamplingContext&    sampling_context,
+    const ShadingRay&   volume_ray,
+    const float         extinction) const
+{
+    sampling_context.split_in_place(1, 1);
+
+    if (!volume_ray.is_finite())
```
dictoon: Add braces around `if` body.

src/appleseed/foundation/math/sampling/mappings.h
Biart95: I am clamping at line 551.

src/appleseed/renderer/kernel/lighting/materialsamplers.cpp
```diff
@@ -137,7 +142,7 @@ const ShadingPoint& BSDFSampler::trace(
             m_shading_point,
             ray,
             transmission);
-    if (shading_point.hit()) transmission.set(0.0f);
+
```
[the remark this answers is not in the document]
Biart95: Yes, last time you were right. It was my mistake. You need to check it sometimes manually (environment light), but mostly do not check at all (in-scene lights).

src/appleseed/renderer/kernel/lighting/materialsamplers.cpp
```diff
@@ -67,6 +67,11 @@ const Vector3d& BSDFSampler::get_point() const
     return m_shading_point.get_point();
 }
 
+const ShadingPoint& BSDFSampler::get_shading_point() const
```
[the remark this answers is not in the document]
Biart95: We don't! We are only on halfway to adapt `ShadingPoint` for volume rendering. Now only path joints are made shading points, but points used for direct lighting integration inside `VolumeLightingIntegrator` are still defined by distance along the ray only.
I was thinking to address this in next PRs, but maybe there is no reason to wait.

src/appleseed/renderer/kernel/lighting/materialsamplers.cpp
```diff
@@ -219,6 +231,7 @@ bool VolumeSampler::sample(
 
     m_volume.scattering_coefficient(
         m_volume_data, m_volume_ray, m_distance, value.m_volume);
+    value.m_volume *= pdf;
```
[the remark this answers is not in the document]
Biart95: It is for consistency with `BSDF`. Without it `VolumeSampler::sample` would return just a unit-length direction, multiplied by `scattering_coefficient` which is isotropic.
Why doesn't `Volume::sample` return any `value` like `BSDF::sample`? Because in case of phase functions, value = scattering_coef * pdf always. In most cases it would be a redundancy to return `scattering_coef * pdf` and divide it by `pdf` again for MC sampling.

But I will check this again just in case. Previous time you remark about `if (shading_point.hit()) transmission.set(0.0f);` was very right but I haven't realized it.

src/appleseed/renderer/kernel/lighting/volumelightingintegrator.cpp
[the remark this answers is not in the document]
Biart95: I checked the consistency between equiangular sampling and exponential sampling and even had the test scenes with both (before I merged `m_exponential_samples` and `m_equiangular_samples` together)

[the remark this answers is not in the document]
Biart95: I checked the consistency between equiangular sampling and exponential sampling and even had corresponding test scenes (before I merged `m_exponential_samples` and `m_equiangular_samples` together). So the math is certainly okay.
We still can check consistency between exponential-only sampling and combined, since we have this "Optimize for..." option.
Probably, I just have confusing names and lack of comments...

src/appleseed/renderer/kernel/shading/shadingpoint.h
```diff
@@ -104,6 +105,13 @@ class ShadingPoint
     // Reset the shading point to its initial state (no intersection).
     void clear();
 
+    // Initialize this shading point as a point in participating media,
+    // given a ray through volume and a distance from origin.
+    void create_volume_shading_point(
+        const ShadingPoint&     prev_shading_point,
+        const ShadingRay&       volume_ray,
+        const float             distance);
```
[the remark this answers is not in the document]
Biart95: Good advice!

src/appleseed/renderer/kernel/intersection/intersector.cpp
```diff
@@ -365,7 +365,8 @@ void Intersector::manufacture_hit(
     const TriangleSupportPlaneType&     triangle_support_plane) const
 {
 #ifdef DEBUG
-    // This helps finding bugs if manufacture_hit() is called on a previously used shading point.
+    // This helps finding bugs if make_surface_shading_point()
+    // is called on a previously usedshading point.
```
dictoon: Put back space between *used* and *shading point*.

src/appleseed/renderer/kernel/intersection/intersector.h
```diff
@@ -118,6 +118,14 @@ class Intersector
         const size_t                        primitive_index,
         const TriangleSupportPlaneType&     triangle_support_plane) const;
 
+    // Manufacture a volume shading point "by hand".
+    // There is no restriction placed on the shading point passed to this method.
+    // For instance it may have been previously initialized and used.
+    void make_volume_shading_point(
+        ShadingPoint&               shading_point,
```
dictoon: Vertically align arguments with surrounding methods.

src/appleseed/renderer/kernel/intersection/intersector.cpp
```diff
@@ -391,6 +392,35 @@ void Intersector::manufacture_hit(
     shading_point.m_members = 0;
 }
 
+void Intersector::make_volume_shading_point(
+    ShadingPoint&           shading_point,
```
dictoon: Vertically align arguments with surrounding methods.

src/appleseed/renderer/kernel/lighting/scatteringmode.h
```diff
@@ -70,7 +70,7 @@ class ScatteringMode
     static bool has_volume(const int modes);
     static bool has_diffuse_or_glossy_or_volume(const int modes);
     static bool has_diffuse_and_glossy(const int modes);
-    static bool has_diffuse_or_volumetric(const int modes);
+    static bool has_diffuse_or_Volume(const int modes);
```
dictoon: Should be `has_diffuse_or_volume()` with a lower case *v*.

That smells global search & replace :)

src/appleseed/renderer/kernel/lighting/volumelightingintegrator.cpp
```diff
@@ -88,45 +84,57 @@ namespace
...
+            m_origin_to_projection = dot(origin_to_light, Vector3f(volume_ray.m_dir));
 
-            m_center_to_light =
-                std::sqrt(square_norm(light_dir) - square(m_origin_to_center));
+            // Distance from the projection point to the light center (height).
+            m_projection_to_light =
+                std::sqrt(square_norm(origin_to_light) - square(m_origin_to_projection));
 
-            m_near_angle = std::atan2(-m_origin_to_center, m_center_to_light);
+            m_near_angle = std::atan2(-m_origin_to_projection, m_projection_to_light);
```
dictoon: Maybe you can simplify the math, isn't `m_near_angle` also simply `std::acos(m_origin_to_projection / std::norm(origin_to_light))`? You could compute both and check that they are always equal, in which case you could switch to that simpler formulation.

src/appleseed/renderer/kernel/lighting/volumelightingintegrator.cpp
```diff
@@ -228,10 +228,13 @@ void VolumeLightingIntegrator::add_single_distance_sample_contribution(
         float mis_weights_sum = 0.0f;
         for (size_t i = 0; i < Spectrum::size(); ++i)
         {
-            const float probability =
-                evaluate_exponential_sample(
-                    exponential_sample, m_volume_ray, extinction_coef[i]);
-            mis_weights_sum += foundation::square(probability);
+            if (extinction_coef[i] > 0.0f)
```
dictoon: Are these changes responsible for the very significant brightness difference in the ref render for volume/11, included in this commit? What about the other reference images?

src/appleseed/renderer/kernel/lighting/volumelightingintegrator.cpp
```diff
@@ -0,0 +1,601 @@
...
+            SamplingContext&            sampling_context)
+          : m_shading_context(shading_context)
+          , m_sampling_context(sampling_context)
+        {
+            Vector3d emission_position;
+            if (light_sample.m_light != nullptr)
+            {
+                m_sampling_context.split_in_place(2, 1);
+                Vector3d emission_direction;  // not used
+                Spectrum light_value(Spectrum::Illuminance);
```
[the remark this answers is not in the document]
Biart95: I opened a respective issue: #1650 .

src/appleseed/renderer/kernel/lighting/materialsamplers.cpp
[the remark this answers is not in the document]
Biart95: It is better to keep this in TODO list as an issue: #1651 .

PR #1595, opened 2017-08-21
Color management Part I

est77:
FIrst part of the color management work done with @luisbarrancos 

- Switched to XYZCMFCIE19312Deg color matching functions.
- Removed old Frame color management options.
- Handle chromaticities metadata when saving OpenEXR images.
- Added basic OpenColorIO support to studio.
- Bundled nuke-default OCIO config as studio default config.
- Added multipart OpenEXR support to EXRImageWriter.
- AOV writting improvements.


src/appleseed/foundation/image/colorspace.h
```diff
@@ -236,10 +237,19 @@ Color<T, 3> linear_rgb_to_hsl(const Color<T, 3>& linear_rgb);
 template <typename T>
 Color<T, 3> ciexyz_to_linear_rgb(const Color<T, 3>& xyz);
 
+template <typename T>
+Color<T, 3> ciexyz_to_linear_rgb(
```
dictoon: On the one hand, these two new functions give a semantic name to the operation; on the other hand, they are perfectly identical and all they do is wrap a matrix * color multiplication (+ clamping)... Let's leave them as is, but it's interesting nevertheless.

src/appleseed/foundation/image/exrimagefilewriter.cpp
```diff
@@ -68,26 +68,35 @@ namespace foundation
...
+    const char**            channel_names)
 {
+    assert(channel_names);
+
     initialize_openexr();
 
     try
     {
+        // todo: lift this limitation.
+        assert(channel_count <= 4);
```
dictoon: `props.m_channel_count` is still used in this method (to compute the X stride). I think it's correct, but just checking.

src/appleseed/foundation/image/exrimagefilewriter.h
```diff
@@ -63,10 +63,18 @@ class APPLESEED_DLLSYMBOL EXRImageFileWriter
 {
   public:
     // Write an OpenEXR image file.
-    virtual void write(
+    void write(
```
dictoon: Our convention is to keep the `virtual` keyword.

src/appleseed/foundation/image/exrutils.cpp
```diff
@@ -92,9 +99,37 @@ void add_attributes(
         else if (attr_name == "creation_time")
             addCapDate(header, attr_value);
 
+        else if (attr_name == "chromaticity_wxy")
+        {
+            Vector2f c = from_string<Vector2f>(attr_value);
```
dictoon: Make const (below as well).

src/appleseed/foundation/image/genericimagefilewriter.h
```diff
@@ -52,10 +52,10 @@ class APPLESEED_DLLSYMBOL GenericImageFileWriter
 {
   public:
     // Write an image file.
-    virtual void write(
+    void write(
```
dictoon: Keep `virtual`.

src/appleseed.studio/mainwindow/rendering/renderwidget.cpp
```diff
@@ -299,7 +299,7 @@ void RenderWidget::blit_tile_no_lock(
         m_float_tile_storage->get_storage());
 
     // Transform the tile to the color space of the frame.
-    frame.transform_to_output_color_space(fp_rgb_tile);
+    //frame.transform_to_output_color_space(fp_rgb_tile);
```
dictoon: Remove?

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -297,286 +279,58 @@ size_t Frame::get_pixel_count() const
...
     }
 
     return result;
 }
 
+bool Frame::write_aov_image(const char* file_path, const size_t aov_index) const
+{
+    assert(file_path);
+
+    if (aov_index >= aovs().size())
```
dictoon: Why is this check necessary? Is returning true intended?

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -814,6 +506,82 @@ bool Frame::write_image(
...
+{
+    EXRImageFileWriter writer;
+
+    if (aov)
+    {
+        if (aov->has_color_data())
+        {
+            // If the AOV has color data, assume we can save it as half floats.
+            const CanvasProperties& props = image.properties();
+            Image half_image(image, props.m_tile_width, props.m_tile_height, PixelFormatHalf);
```
dictoon: Make const.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -814,6 +506,82 @@ bool Frame::write_image(
...
+            writer.write(file_path, image, image_attributes, aov->get_channel_count(), aov->get_channel_names());
+    }
+    else
+        writer.write(file_path, image, image_attributes);
+}
+
+namespace
+{
+
+void transform_to_srgb(Tile& tile)
```
dictoon: Indent the content of anonymous namespaces.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -855,22 +623,6 @@ DictionaryArray FrameFactory::get_input_metadata()
 
     metadata.push_back(
         Dictionary()
-            .insert("name", "pixel_format")
```
dictoon: We should probably introduce a migration step to remove the old parameters from frame entities.

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -1432,6 +1436,42 @@ QString MainWindow::get_filter_string(const int filter)
...
+    const string default_ocio_config = (root_path / "ocio" / "config.ocio").string();
+
+    try
+    {
+        m_ocio_config = OCIO::Config::CreateFromFile(default_ocio_config.c_str());
+        RENDERER_LOG_INFO("Using OpenColorIO config: %s", default_ocio_config.c_str());
+        OCIO::SetCurrentConfig(m_ocio_config);
+        return;
+    }
+    catch(...)
```
dictoon: `catch (...)`

Also, this is too wide a net, be more specific in which exceptions you intend to catch.

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -124,6 +125,8 @@ MainWindow::MainWindow(QWidget* parent)
   , m_attribute_editor(0)
   , m_project_file_watcher(0)
 {
+    initialize_ocio();
```
dictoon: Genuine question: is this the right place to initialize OCIO? Shouldn't it be done at startup?

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -1432,6 +1436,42 @@ QString MainWindow::get_filter_string(const int filter)
...
 
+void MainWindow::initialize_ocio()
+{
+    // Try first a user specified OCIO config.
+    if (getenv("OCIO"))
+    {
+        try
+        {
+            m_ocio_config = OCIO::GetCurrentConfig();
+            RENDERER_LOG_INFO("Using OpenColorIO config: %s", getenv("OCIO"));
```
dictoon: Our (weird, admittedly) convention is that log messages are all lower-case. Also we use `oiio` instead of `OpenImageIO` everywhere so let's do the same for now:

    RENDERER_LOG_INFO("using ocio configuration %s", getenv("OCIO"));

Maybe one day I'll fix the case everywhere... Until then, let's try to be consistent.

Same change below.

src/appleseed.studio/mainwindow/mainwindow.h
```diff
@@ -47,6 +47,10 @@
 #include <QMainWindow>
 #include <QObject>
 
+// OpenColorIO headers.
+#include<OpenColorIO/OpenColorIO.h>
```
dictoon: Add space after `#include`. Same change in other places.

src/appleseed.studio/mainwindow/rendering/rendertab.cpp
```diff
@@ -299,6 +306,35 @@ void RenderTab::create_toolbar()
...
+    QLabel* display_label = new QLabel("Display:");
+    display_label->setObjectName("display_label");
+    m_toolbar->addWidget(display_label);
+
+    // Create the display combobox.
+    m_display_transform_combo = new QComboBox();
+    m_display_transform_combo->setObjectName("display_combo");
+    {
+        const char* display_name = m_ocio_config->getDefaultDisplay();
+        std::string default_transform = m_ocio_config->getDefaultView(display_name);
```
dictoon: Make const.

src/appleseed.studio/mainwindow/rendering/rendertab.cpp
```diff
@@ -299,6 +306,35 @@ void RenderTab::create_toolbar()
...
+
+        int default_index = 0;
+        for (int i = 0, e = m_ocio_config->getNumViews(display_name); i < e; ++i)
+        {
+            const char* name = m_ocio_config->getView(display_name, i);
+            m_display_transform_combo->addItem(name);
+
+            if (default_transform == name)
+                default_index = i;
```
dictoon: Shouldn't we `break` here?

src/appleseed.studio/mainwindow/rendering/renderwidget.cpp
```diff
@@ -290,23 +349,38 @@ void RenderWidget::blit_tile_no_lock(
...
+    // Apply OCIO transform.
+    OCIO::PackedImageDesc image_desc(
+        reinterpret_cast<float*>(float_tile.get_storage()),
+        float_tile.get_width(),
+        float_tile.get_height(),
+        float_tile.get_channel_count());
+    m_ocio_processor->apply(image_desc);
 
     // Convert the tile to 8-bit RGB for display.
-    static const size_t ShuffleTable[] = { 0, 1, 2, Pixel::SkipChannel };
```
dictoon: Why renaming this array? It's static const.

src/appleseed.cli/main.cpp
```diff
@@ -617,16 +617,29 @@ namespace
...
+            {
+                const AOV* aov = frame->aovs().get_by_index(i);
+                const string output_filename =
+                    aov->get_parameters().get_optional<string>("output_filename");
 
-                if (frame->get_parameters().get_optional<bool>("output_aovs", false))
-                    frame->write_aov_images(output_filename.c_str());
+                if (!output_filename.empty())
+                {
+                    LOG_INFO(g_logger, "writing %s AOV to disk...", aov->get_model());
```
dictoon: *aov*

src/appleseed/renderer/meta/tests/test_localsampleaccumulationbuffer.cpp
```diff
@@ -50,7 +50,7 @@ using namespace std;
 
 TEST_SUITE(Renderer_Kernel_Rendering_LocalSampleAccumulationBuffer)
 {
-    bool honors_crop_window(const AABB2u crop_window, const bool undo_premultiplied_alpha)
+    bool honors_crop_window(const AABB2u crop_window)
```
dictoon: Pass `crop_window` by const ref (the problem was already there, but let's just fix it now).

src/appleseed/renderer/kernel/rendering/generic/generictilerenderer.cpp
```diff
@@ -199,9 +199,7 @@ namespace
             }
 
             // Develop the framebuffer to the tile.
-            if (frame.is_premultiplied_alpha())
-                framebuffer->develop_to_tile_premult_alpha(tile, aov_tiles);
-            else framebuffer->develop_to_tile_straight_alpha(tile, aov_tiles);
+            framebuffer->develop_to_tile(tile, aov_tiles);
```
dictoon: Not sure why we're removing support for premultiplied alpha?

src/appleseed/foundation/image/exrimagefilewriter.cpp
```diff
@@ -68,6 +71,24 @@ namespace foundation
...
+struct EXRImageFileWriter::Impl
+{
+    vector<const ICanvas*>  m_images;
+    vector<Header>          m_image_headers;
+    vector<size_t>          m_channel_counts;
+    vector<const char**>    m_channel_names;
+};
+
+EXRImageFileWriter::EXRImageFileWriter()
+  : impl(0)
```
dictoon: `nullptr`

src/appleseed/foundation/image/exrimagefilewriter.cpp
```diff
@@ -78,93 +99,128 @@ void EXRImageFileWriter::write(
     write(filename, image, image_attributes, props.m_channel_count, ChannelNames);
 }
 
+namespace
+{
+
+PixelType get_imf_pixel_type(const CanvasProperties& props)
```
dictoon: Indent.

src/appleseed/foundation/image/exrimagefilewriter.h
```diff
@@ -62,6 +62,12 @@ class APPLESEED_DLLSYMBOL EXRImageFileWriter
   : public IImageFileWriter
 {
   public:
+    // Constructor.
+    EXRImageFileWriter();
+
+    // Destructor.
+    ~EXRImageFileWriter() override;
```
dictoon: `virtual`

src/appleseed.studio/mainwindow/rendering/renderwidget.cpp
```diff
@@ -131,6 +131,13 @@ namespace
     }
 }
 
+void RenderWidget::start_render()
+{
+    // Clear the image storage.
+    if (m_image_storage.get())
```
dictoon: Hm, that's normally already taken care of (we're dimming the previous render, on purpose). Why is this necessary?

dictoon:
Awesome!!

Mostly a few small things to fix.

Two questions though:
- Why are we getting rid of premultiplied alpha support?
- Why do we now clear the render widget before rendering starts? This was already taken care of normally.

dictoon:
Nice branch name btw :)

CMakeLists.txt
```diff
@@ -246,6 +246,9 @@ if (WITH_DISNEY_MATERIAL)
     endif ()
 endif ()
 
+if (WITH_STUDIO)
+    find_package (OpenColorIO REQUIRED)
```
dictoon: This should be wrapped in `USE_EXTERNAL_OCIO`.

src/appleseed.studio/CMakeLists.txt
```diff
@@ -546,6 +547,7 @@ link_against_openexr (appleseed.studio)
 target_link_libraries (appleseed.studio
     appleseed
     appleseed.shared
+    ${OPENCOLORIO_LIBRARIES}
```
dictoon: Replace by `link_against_ocio ()`.

CMakeLists.txt
dictoon: I'm taking care of this, as well as the Windows config. I'll make a PR to your repo.

src/appleseed/foundation/image/colorspace.h
luisbarrancos: @dictoon i would just name them CIEXYZ_to_RGB. CIE xyz != CIE XYZ. And the transformation implies linear RGB values anyway. As it is, ciexyz_to_linear_rgb means one needs to read the code to see if we're referring to CIE xyz or XYZ. It's not clear, just my .02c.

dictoon: Yeah I know what CIE xyz and CIE XYZ are two different things. Here is what I propose: when we mean XYZ, then we spell it XYZ. We continue spelling the rest (in particular the rgb part) in lower case to conform to our coding conventions.

The word "linear" is probably redundant but it makes it obvious that we're getting a linear value. It cannot hurt.

What do you think?

luisbarrancos: Ok, so in this case we have to use XYZ. The rest is fine. The "linear" is really redundant since the XYZ<>RGB transformation has to be done with linear values, but i'm totally ok with leaving it.

dictoon: I completely understand that it's redundant, but here like in other places, it's only redundant because of the context or the operation being done. RGB alone doesn't make sense to I'd rather always qualify it with what kind of RGB it is.

luisbarrancos: hmm, i think RGB alone makes sense. In essence we only care about scene-linear values. Ingested material (textures) are to be linearized, but as an operation that wouldn't necessarily be part of the renderer (it can be done with a tool such as maketx for example). And an output render are to be encoded depending on format chosen and standard chosen, also not necessarily part of the renderer (one could save scene-linear files only for example, and leave the conversion to an external tool). I'm not saying that should be done though, just that everywhere RGB is linear, and that linearization and encoding are pre and post render stages not necessarily part of the render, with very well defined contexts. 
For the sake of completion, the convention is usually to use for example, RGB for scene-linear, and R'G'B' for gamma encoded values. (RGB prime).

dictoon: appleseed/foundation/image is meant to have a wider use than just in appleseed the renderer, that's why it contains a bunch of functions that are not (or no longer) used in appleseed, e.g. those dealing with sRGB. Since it has to deal with various RGB flavors, I'd prefer that we are always explicit about which flavor: linear RGB, sRGB, R'G'B'. I'm in favor of introducing `rgb_prime` where necessary.

src/appleseed.studio/CMakeLists.txt
dictoon: Note that you'll need to add `link_against_ocio ()` for Linux and macOS.

est77:
> Why are we getting rid of premultiplied alpha support?

We are removing *unpremultiplied* alpha support.

dictoon:
> We are removing unpremultiplied alpha support.

Ah, ok. Makes more sense.

est77:
> Why do we now clear the render widget before rendering starts? This was already taken care of normally.

We keep a float image where we buffer the rendered image and we convert it to a QImage and apply the OCIO display transform when tiles are rendered or when the display transform changes.

We have to clear the float buffer now when starting a render to avoid having tiles from previous renders resurface when the display transform changes.

src/appleseed/foundation/image/exrimagefilewriter.h
est77: clang-tidy removes virtual when adding overrides, which I think it makes sense.
I wanted to be consistent with that in preparation of start running clang-tidy on appleseed's code base.

src/appleseed.studio/mainwindow/rendering/renderwidget.cpp
est77: If you do a render, stop it, change something and re-render, if you change the display transform while rendering, old tiles from render # 1 appear on the render widget.

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
```diff
@@ -171,6 +171,8 @@ void RenderingManager::start_rendering(
     m_rendering_mode = rendering_mode;
     m_render_tab = render_tab;
 
+    m_render_tab->get_render_widget()->start_render();
```
dictoon: Maybe `on_rendering_start()` to be in line with other similar methods?

dictoon:
> We keep a float image where we buffer the rendered image and we convert it to a QImage and apply the OCIO display transform when tiles are rendered or when the display transform changes.

> We have to clear the float buffer now when starting a render to avoid having tiles from previous renders resurface when the display transform changes.

Ah, I see, thanks.

src/appleseed/foundation/image/exrimagefilewriter.h
dictoon: Ok, then I guess that's our new convention: `override` instead of `virtual`.

src/appleseed.studio/mainwindow/rendering/rendertab.cpp
est77: No, because we are building the menu and getting the default_index at the same time.
If we break we don't add the items after default_index.

dictoon: Ok, I had missed that.

est77:
Fixed linux build and comments addressed.

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -1444,7 +1444,7 @@ void MainWindow::initialize_ocio()
         try
         {
             m_ocio_config = OCIO::GetCurrentConfig();
-            RENDERER_LOG_INFO("Using OpenColorIO config: %s", getenv("OCIO"));
+            RENDERER_LOG_INFO("Using ocio configuration: %s", getenv("OCIO"));
```
dictoon: *using*

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -1459,11 +1459,11 @@ void MainWindow::initialize_ocio()
     try
     {
         m_ocio_config = OCIO::Config::CreateFromFile(default_ocio_config.c_str());
-        RENDERER_LOG_INFO("Using OpenColorIO config: %s", default_ocio_config.c_str());
+        RENDERER_LOG_INFO("Using ocio configuration: %s", default_ocio_config.c_str());
```
dictoon: *using*

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -1459,11 +1459,11 @@ void MainWindow::initialize_ocio()
...
         OCIO::SetCurrentConfig(m_ocio_config);
         return;
     }
-    catch(...)
+    catch(const OCIO::Exception&)
```
dictoon: Add space after `catch`.

dictoon:
Looks good, you can merge after you've fixed the last cosmetic issues.

PR #1599, opened 2017-08-22
Studio python plugins

glebmish:


src/appleseed.python/studio/plugins.py
```diff
@@ -0,0 +1,71 @@
...
+        return
+
+    if not hasattr(plugin_module, 'register'):
+        print "Plugin '{}' has no register function".format(name)
+        return
+
+    try:
+        plugin_module.register()
+    except Exception as e:
+        print "Could not init plugin '{}'".format(plugin_module, e)
```
est77: we need another {} to print the exception

sandbox/studio/plugins/texture_processor/__init__.py
```diff
@@ -1,29 +1,25 @@
...
-    assert isinstance(container, asr.BaseGroup)
+def register():
+    create_convert_textures_action()
 
-    textures = list(container.textures())
 
-    assemblies = container.assemblies()
-    for key in assemblies:
-        textures += get_textures(assemblies[key])
+def create_convert_textures_action():
```
est77: lets move the contents of this function to the register function.

src/appleseed.python/studio/plugins.py
```diff
@@ -0,0 +1,71 @@
...
+import imp
+import traceback
+
+def load_plugins(bundled_plugins_path):
+    load_plugins_from_dir(bundled_plugins_path)
+
+    user_plugins_path = os.environ.get('APPLESEED_STUDIO_PLUGIN_PATH')
+    if user_plugins_path is not None:
+        load_plugins_from_dir(user_plugins_path)
+
```
est77: only one newline. Same between the other functions in this file.

src/appleseed.python/studio/ui.py
```diff
@@ -39,3 +40,21 @@ def wrapinstance(addr, type):
 
     else:
         raise Exception("No wrapinstance function defined for " + Qt.__binding__)
+
```
est77: only one newline

src/appleseed.python/studio/plugins.py
glebmish: why?

src/appleseed.python/studio/plugins.py
glebmish: I added `traceback.print_exc()` instead to get more info out of exception

src/appleseed.studio/python/pythoninterpreter.cpp
```diff
@@ -106,6 +106,26 @@ namespace
 
         return lib_path.string();
     }
+
+    string compute_bundled_plugins_path()
```
dictoon: Can't we use `application::shared::Application::get_root_path()` instead of reimplementing the logic here?

sandbox/studio/plugins/texture_processor/__init__.py
```diff
@@ -1,30 +1,21 @@
...
-    textures = list(container.textures())
-
-    assemblies = container.assemblies()
-    for key in assemblies:
-        textures += get_textures(assemblies[key])
-
-    return textures
-
+def register():
+    menu = ui.find_or_create_menu("Utils")
```
dictoon: What about "Plugins" instead?

dictoon:
Awesome to see this landing!!

PR #1606, opened 2017-08-25
Different fixes

glebmish:


sandbox/studio/plugins/texture_processor/__init__.py
```diff
@@ -28,6 +31,10 @@ def _find_maketx():
             raise Exception('maketx binary is not found')
 
     project = studio.current_project()
+
+    if project is None:
+        raise Exception('No project is opened')
```
est77: I don't think we should throw. If no project is open there are no textures to convert and no work to do.
Maybe simply return?

glebmish: yes, ok

PR #1624, opened 2017-09-12
BDPT work

aytekaman:


src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.cpp
```diff
@@ -78,6 +81,19 @@ namespace
             const ShadingPoint&     shading_point,
             ShadingComponents&      radiance) override      // output radiance, in W.sr^-1.m^-2
         {
+            PathVisitor light_path_visitor;
+            VolumeVisitor volume_visitor;
+
+            PathTracer<PathVisitor, VolumeVisitor, false> light_path_tracer(    // false = not adjoint
```
dictoon: For light tracing, you'll want to set `Adjoint` to `true`. You can find an extended treatment of adjoint tracing in Veach's thesis.

dictoon:
It's a good start.

Regarding how to trace the light path, you can find inspiration in the Light Tracer:
https://github.com/appleseedhq/appleseed/blob/master/src/appleseed/renderer/kernel/lighting/lighttracing/lighttracingsamplegenerator.cpp#L570-L786

dictoon:
I suggest to keep this PR open so that you can progressively add commit until a first working implementation. How does that sound?

aytekaman:
Sounds like a plan:)

src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.h
```diff
@@ -48,7 +49,9 @@ class BDPTLightingEngineFactory
 {
   public:
     // Constructor.
-    explicit BDPTLightingEngineFactory(const ParamArray& params);
+    explicit BDPTLightingEngineFactory(
```
dictoon: You can safely remove `explicit` now that the constructor takes more than one argument. I just wish C++ had explicit constructors by default :)

aytekaman: Ah, you are right. Will do it.

src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.h
```diff
@@ -33,8 +33,13 @@
 #include "renderer/kernel/lighting/ilightingengine.h"
 #include "renderer/utility/paramarray.h"
 
+// appleseed.foundation headers.
+#include "foundation/math/population.h"
```
dictoon: It doesn't look like you need this header here? Maybe only in the .cpp file?

PR #1626, opened 2017-09-13
question to est77

usakhelo:
Hi est77,
here's the question - I've moved constructor call into the body and it doesn't work. While calling init method in your commit works fine.


src/appleseed/renderer/kernel/aov/aovaccumulator.cpp
```diff
@@ -195,8 +195,8 @@ AOVAccumulatorContainer::AOVAccumulatorContainer()
 }
 
 AOVAccumulatorContainer::AOVAccumulatorContainer(const AOVContainer& aovs)
-  : AOVAccumulatorContainer()
 {
+    AOVAccumulatorContainer();
```
est77: This is not calling the default constructor. It's actually creating a new unnamed AOVAccumulatorContainer.

PR #1627, opened 2017-09-13
est77's fix

usakhelo:
Fixes AOVAccumulatorContainer to be compatible with vs2012.


dictoon:
Thanks!

We need to be a bit more descriptive in the commit messages, outside any context it'll be hard to know what "est77's fix" is about :)

Also we have some written rules (for once) about commit messages here, check them out:
https://github.com/appleseedhq/appleseed/wiki/Git-Commits

I'll fix the commit message, no worries.

PR #1629, opened 2017-09-18
Fix bump mapping

dictoon:
- Fix built-in bump mapping algorithm
- Allow using bump mapping and normal mapping with non-texture sources
- Review, check and cleanup code computing screen space derivatives dpdx, dpdy, duvdx and duvdy
- Reimplement appleseed-max's bump mapping OSL shader
- Remove parameter attributes from appleseed-max's normal mapping OSL shader
- Add test scenes


src/appleseed/renderer/modeling/material/material.cpp
```diff
@@ -228,32 +228,25 @@ IBasisModifier* Material::create_basis_modifier(const MessageContext& context) c
...
-    // Retrieve the displacement texture.
-    const TextureSource* displacement_map = static_cast<const TextureSource*>(displacement_source);
-    const Texture& texture = displacement_map->get_texture_instance().get_texture();
+    if (displacement_source == nullptr)
+        return nullptr;
 
-    // Print a warning if the displacement texture is not expressed in the linear RGB color space.
-    if (texture.get_color_space() != ColorSpaceLinearRGB)
+    // Additional checks if a texture instance is used for displacement.
+    const TextureSource* texture_map = dynamic_cast<const TextureSource*>(displacement_source);
```
usakhelo: Franz, I think it is still looking for TextureSource here?

dictoon: Yes that's correct, it's checking if the source is a texture source and if so, does an additional check.

usakhelo: Nevermind, it's something else. I'm still checking.

usakhelo: Something is wrong, still checking..:
image

dictoon: I see. In the appleseed code, try decommenting the line that says

    //#define USE_SCREEN_SPACE_UV_DERIVATIVES

on line 53 of `src\appleseed\renderer\modeling\material\bumpmappingmodifier.cpp`.

dictoon: That's an alternative bump mapping method that may generally work better.

usakhelo:
Franz, I've applied bump material similar to what was already in scene test scene to a sphere and it breaks the smoothing. Attached packed project renamed to .zip.

image


[test_smooth_bump.zip](https://github.com/appleseedhq/appleseed/files/1312695/test_smooth_bump.zip)

dictoon:
Thanks @usakhelo, I'm looking into it.

PR #1637, opened 2017-09-23
Frame image and AOVs saving refactors and improvements

est77:
- Moved cli code to save main and AOV images to Frame so that it can be reused.
- Create directories if needed before saving images and AOVs.
- Renamed Frame's write_image_and_aovs_to_multipart_exr for consistency with similar methods.

### DIFF
--- a/src/appleseed.cli/main.cpp
+++ b/src/appleseed.cli/main.cpp
@@ -608,39 +608,16 @@ namespace
         }
 
         // Write the frame to disk.
+
         if (g_cl.m_output.is_set())
         {
-            LOG_INFO(g_logger, "writing frame to disk...");
             project->get_frame()->write_main_image(g_cl.m_output.value().c_str());
             project->get_frame()->write_aov_images(g_cl.m_output.value().c_str());
         }
         else
         {
             const Frame* frame = project->get_frame();
-
-            // Write the main image.
-            const string output_filename =
-                frame->get_parameters().get_optional<string>("output_filename");
-
-            if (!output_filename.empty())
-            {
-                LOG_INFO(g_logger, "writing frame to disk...");
-                frame->write_main_image(output_filename.c_str());
-            }
-
-            // Write AOVs.
-            for (size_t i = 0, e = frame->aovs().size(); i < e; ++i)
-            {
-                const AOV* aov = frame->aovs().get_by_index(i);
-                const string output_filename =
-                    aov->get_parameters().get_optional<string>("output_filename");
-
-                if (!output_filename.empty())
-                {
-                    LOG_INFO(g_logger, "writing %s aov to disk...", aov->get_model());
-                    frame->write_aov_image(output_filename.c_str(), i);
-                }
-            }
+            frame->write_main_and_aov_images();
         }
 
 #if defined __APPLE__ || defined _WIN32
--- a/src/appleseed/renderer/modeling/frame/frame.cpp
+++ b/src/appleseed/renderer/modeling/frame/frame.cpp
@@ -60,7 +60,7 @@
 #include "foundation/utility/string.h"
 
 // Boost headers.
-#include "boost/filesystem/path.hpp"
+#include "boost/filesystem.hpp"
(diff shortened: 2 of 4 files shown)


src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -274,6 +274,25 @@ size_t Frame::get_pixel_count() const
     return impl->m_crop_window.volume();
 }
 
+namespace
+{
+
+void create_parent_directories(const string& filepath)
```
dictoon: Indent content of anonymous namespaces.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -274,6 +274,25 @@ size_t Frame::get_pixel_count() const
...
+{
+    const bf::path parent_path = bf::path(filepath).parent_path();
+    if (!bf::exists(parent_path))
+    {
+        if (!bf::create_directories(parent_path))
+        {
+            RENDERER_LOG_ERROR(
+                "Could not create directory %s",
```
dictoon: Spell log messages in lower case.

dictoon: No way to print details of the error?

src/appleseed.cli/main.cpp
```diff
@@ -608,39 +608,17 @@ namespace
...
-            // Write AOVs.
-            for (size_t i = 0, e = frame->aovs().size(); i < e; ++i)
-            {
-                const AOV* aov = frame->aovs().get_by_index(i);
-                const string output_filename =
-                    aov->get_parameters().get_optional<string>("output_filename");
-
-                if (!output_filename.empty())
-                {
-                    LOG_INFO(g_logger, "writing %s aov to disk...", aov->get_model());
```
dictoon: AFAICT we've lost this message. Can we put it back (preferably in `write_aov_image()`)?

src/appleseed.cli/main.cpp
```diff
@@ -608,39 +608,17 @@ namespace
         }
 
         // Write the frame to disk.
+        LOG_INFO(g_logger, "writing frame to disk...");
```
dictoon: I would suggest moving this message into `write_main_image()`.

src/appleseed/renderer/modeling/frame/frame.cpp
est77: done

src/appleseed/renderer/modeling/frame/frame.cpp
est77: fixed

src/appleseed.cli/main.cpp
est77: I removed it because I thought it didn't have much value.
We already log all the images we save in Frame::write_image and I added logging to  write_main_and_aov_images_to_multipart_exr in this PR.

src/appleseed.cli/main.cpp
est77: This message was redundant because just after saving the image we log it again, this time with some useful information like filename and timings.

dictoon: Ok!

dictoon: In that case, let's remove it then?

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -363,7 +411,13 @@ void Frame::write_image_and_aovs_to_multipart_exr(const char *file_path) const
             writer.append_part(aov_name.c_str(), image, image_attributes, aov->get_channel_count(), aov->get_channel_names());
     }
 
+    create_parent_directories(file_path);
     writer.write_multipart_exr(file_path);
+
+    RENDERER_LOG_INFO(
+        "wrote multipart EXR image file %s in %s.",
```
dictoon: *exr*

Issue #1421, opened 2017-05-18
Allow setting the distance to sun in Physical Sun model


dictoon:

dictoon:
Done.

PR #1638, opened 2017-09-24
Fix issue #1421

pjessesco:
Add metadata distance
Fix compute_sun_radius()

I'd appreciate any feedback. :)

### DIFF
--- a/src/appleseed/renderer/modeling/light/sunlight.cpp
+++ b/src/appleseed/renderer/modeling/light/sunlight.cpp
@@ -133,8 +133,9 @@ namespace
             else
             {
                 RENDERER_LOG_WARNING(
-                    "distance between sun and scene \"%s\" is not uniform.",
+                    "distance between sun and scene \"%s\" is not uniform, using default value of 149.6 million km.",
                     get_path().c_str());
+                m_values.m_distance = 149.6;
             }
 
             // Compute SunSolidAngle that depends on distance between Sun and scene.
--- a/src/appleseed/renderer/modeling/light/sunlight.cpp
+++ b/src/appleseed/renderer/modeling/light/sunlight.cpp
@@ -85,7 +85,9 @@ namespace
 
     // Solid angle sustained by the Sun, as seen from Earth (in steradians).
     // Reference: http://en.wikipedia.org/wiki/Solid_angle#Sun_and_Moon
-    const float SunSolidAngle = 6.87e-5f;
+    float SunSolidAngle = 6.87e-5f;
+
+    const float SunRadius = 0.6957; // millions of km
 
     class SunLight
       : public Light
@@ -100,7 +102,7 @@ namespace
             m_inputs.declare("turbidity", InputFormatFloat);
             m_inputs.declare("radiance_multiplier", InputFormatFloat, "1.0");
             m_inputs.declare("size_multiplier", InputFormatFloat, "1.0");
-            m_inputs.declare("distance", InputFormatFloat, "149.0");
+            m_inputs.declare("distance", InputFormatFloat, "149.6");
         }
 
         virtual void release() override
@@ -125,6 +127,23 @@ namespace
             // Evaluate uniform inputs.
             m_inputs.evaluate_uniforms(&m_values);
 
+
+            Source* distance_src = get_inputs().source("distance");
+            assert(distance_src != nullptr);
+            if (distance_src->is_uniform())
+                distance_src->evaluate_uniform(m_values.m_distance);
+            else
+            {
+                RENDERER_LOG_WARNING(
+                    "Distance between Sun and scene \"%s\" is not uniform.",
+                    get_path().c_str());
+            }
+
+            // Compute SunSolidAngle that depends on distance between Sun and scene.
+            // angular_diameter = 2 * arctan(sun_radius / (distance))
+            SunSolidAngle = 2 * M_PI * (1 - cos(atan(SunRadius / m_values.m_distance)));
+
+
             // If the Sun light is bound to an environment EDF, let it override the Sun's direction and turbidity.
             const EnvironmentEDF* env_edf = dynamic_cast<EnvironmentEDF*>(m_inputs.get_entity("environment_edf"));
             if (env_edf)
@@ -140,7 +159,7 @@ namespace
             else
             {
                 RENDERER_LOG_WARNING(
-                    "size multiplier of the sun light \"%s\" is not uniform.",
+                    "size multiplier of the Sun light \"%s\" is not uniform.",
                     get_path().c_str());
             }
 
@@ -418,13 +437,11 @@ namespace
         {
             // sun diameter = 1.3914
             // angular_diameter = 2 * arctan(sun_diameter / (2 * distance))
-            // angular_diameter / 2 = arctan(sun_diameter / (2 * distance))
             // tan(angular_diameter / 2) * distance = sun_radius
             // tan(angular_diameter / 2) * scene_diameter = virtual_sun_radius
+            // -> virtual_sun_radius = sun_radius * scene_diameter / distance
 
-            float angular_diameter = 2 * atan(1.3914 / (2 * distance));
-
-            return tan(angular_diameter / 2 ) * scene_diameter;
+            return SunRadius * scene_diameter / distance;
         }
 
         void sample_disk(
--- a/src/appleseed/renderer/modeling/light/sunlight.cpp
+++ b/src/appleseed/renderer/modeling/light/sunlight.cpp
@@ -100,6 +100,7 @@ namespace
             m_inputs.declare("turbidity", InputFormatFloat);
             m_inputs.declare("radiance_multiplier", InputFormatFloat, "1.0");
             m_inputs.declare("size_multiplier", InputFormatFloat, "1.0");
+            m_inputs.declare("distance", InputFormatFloat, "1.0");
         }
 
         virtual void release() override
@@ -251,6 +252,7 @@ namespace
             float           m_turbidity;                // atmosphere turbidity
             float           m_radiance_multiplier;      // emitted radiance multiplier
             float           m_size_multiplier;          // sun size multiplier
+            float           m_distance;                 // distance between sun and earth
         };
 
         Vector3d            m_scene_center;             // world space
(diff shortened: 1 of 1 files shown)


src/appleseed/renderer/modeling/light/sunlight.cpp
```diff
@@ -412,12 +414,13 @@ namespace
             }
         }
 
-        static double compute_sun_radius(const double distance)
+        static double compute_sun_radius(const double distance, const double m_safe_scene_radius)
```
pjessesco: I wonder if there is a better way.

dictoon: The `m_` prefix means *member variable*, it should only be used for member variables, not for arguments.

src/appleseed/renderer/modeling/light/sunlight.cpp
```diff
@@ -100,6 +100,7 @@ namespace
             m_inputs.declare("turbidity", InputFormatFloat);
             m_inputs.declare("radiance_multiplier", InputFormatFloat, "1.0");
             m_inputs.declare("size_multiplier", InputFormatFloat, "1.0");
+            m_inputs.declare("distance", InputFormatFloat, "1.0");
```
dictoon: `1.0` here is the default value of this input. In appleseed, input default values must usually be defined in two places: in input declarations (here) and in input metadata. It's unfortunate, but until we fix this design issue we must specify default values in these two places.

src/appleseed/renderer/modeling/light/sunlight.cpp
```diff
@@ -251,6 +252,7 @@ namespace
             float           m_turbidity;                // atmosphere turbidity
             float           m_radiance_multiplier;      // emitted radiance multiplier
             float           m_size_multiplier;          // sun size multiplier
+            float           m_distance;                 // distance between sun and earth
```
dictoon: In English, Sun is a proper noun so it's capitalized (and so would be Earth). Also, I would suggest using "scene" instead of "Earth" since the entire point of this exercise is to render scenes that take place on other places than the Earth :)

src/appleseed/renderer/modeling/light/sunlight.cpp
```diff
@@ -583,6 +586,23 @@ DictionaryArray SunLightFactory::get_input_metadata() const
...
+                Dictionary()
+                    .insert("value", "0.0")
+                    .insert("type", "hard"))
+            .insert("max",
+                Dictionary()
+                    .insert("value", "500.0")
+                    .insert("type", "soft"))
+            .insert("use", "optional")
+            .insert("default", "149.6")
+            .insert("help", "Distance between sun and earth(millions of km)"));
```
dictoon: *sun* -> *Sun*
*earth* -> *scene*
Insert space before opening parenthesis.

src/appleseed/renderer/modeling/light/sunlight.cpp
```diff
@@ -467,7 +470,7 @@ namespace
 
             const Basis3d basis(outgoing);
             const Vector2d p = sample_disk_uniform(s);
-            const double sun_radius = compute_sun_radius(m_scene_radius) * m_values.m_size_multiplier;
+            const double sun_radius = compute_sun_radius(m_values.m_distance,m_safe_scene_diameter) * m_values.m_size_multiplier;
```
dictoon: Always insert one space after a comma.

src/appleseed/renderer/modeling/light/sunlight.cpp
dictoon: Another very confusing thing: when you're calling `compute_sun_radius()`, you're passing `m_safe_scene_diameter` in 2nd argument, but the argument itself is called `m_safe_scene_radius`.

src/appleseed/renderer/modeling/light/sunlight.cpp
```diff
@@ -412,12 +414,17 @@ namespace
...
-            return 0.0046505 * distance;
+            // sun diameter = 1.3914
+            // angular_diameter = 2 * arctan(sun_diameter / (2 * distance))
+            // angular_diameter / 2 = arctan(sun_diameter / (2 * distance))
+            // tan(angular_diameter / 2) * distance = sun_radius
+            // tan(angular_diameter / 2) * scene_diameter = virtual_sun_radius
+
+            float angular_diameter = 2 * atan(1.3914 / (2 * distance));
+
+            return tan(angular_diameter / 2 ) * scene_diameter;
```
dictoon: And, it turns out that this expression simplifies itself to:

    return 1.3914 / (2 * distance) * scene_diameter

Maybe what you ended up with earlier?

src/appleseed/renderer/modeling/light/sunlight.cpp
```diff
@@ -100,6 +100,7 @@ namespace
             m_inputs.declare("turbidity", InputFormatFloat);
             m_inputs.declare("radiance_multiplier", InputFormatFloat, "1.0");
             m_inputs.declare("size_multiplier", InputFormatFloat, "1.0");
+            m_inputs.declare("distance", InputFormatFloat, "149.0");
```
dictoon: Should be 149.6.

src/appleseed/renderer/modeling/light/sunlight.cpp
dictoon: I'd suggest making 1.3914 a constant, e.g. `SunDiameter`.

pjessesco: then how about make 0.6957 `SunRadius`, rather than `1.3914 / 2`?

dictoon: Yep!

src/appleseed/renderer/modeling/light/sunlight.cpp
```diff
@@ -85,7 +85,9 @@ namespace
 
     // Solid angle sustained by the Sun, as seen from Earth (in steradians).
     // Reference: http://en.wikipedia.org/wiki/Solid_angle#Sun_and_Moon
-    const float SunSolidAngle = 6.87e-5f;
+    float SunSolidAngle = 6.87e-5f;
```
dictoon: `SunSolidAngle` is now a global variable, which will cause all kinds of problems, not being thread-safe being the biggest one. You need to make it a class member (and rename it to `m_sun_solid_angle` as class members should be).

src/appleseed/renderer/modeling/light/sunlight.cpp
```diff
@@ -125,6 +127,23 @@ namespace
...
 
+
+            Source* distance_src = get_inputs().source("distance");
+            assert(distance_src != nullptr);
+            if (distance_src->is_uniform())
+                distance_src->evaluate_uniform(m_values.m_distance);
+            else
+            {
+                RENDERER_LOG_WARNING(
+                    "Distance between Sun and scene \"%s\" is not uniform.",
```
dictoon: For historical reasons all our log messages should be entirely in lower case.

dictoon: Here we should probably use the default Sun-scene distance of 149.6 million km.

src/appleseed/renderer/modeling/light/sunlight.cpp
```diff
@@ -125,6 +127,23 @@ namespace
...
+                    get_path().c_str());
+            }
+
+            // Compute SunSolidAngle that depends on distance between Sun and scene.
+            // angular_diameter = 2 * arctan(sun_radius / (distance))
+            SunSolidAngle = 2 * M_PI * (1 - cos(atan(SunRadius / m_values.m_distance)));
```
dictoon: The problem of `M_PI` is that it's a `double` while we need a `float`. This will fail to compile with Visual Studio (among others). Instead you can use `TwoPi<float>()` in place of `2 * M_PI`.

src/appleseed/renderer/modeling/light/sunlight.cpp
```diff
@@ -140,7 +159,7 @@ namespace
             else
             {
                 RENDERER_LOG_WARNING(
-                    "size multiplier of the sun light \"%s\" is not uniform.",
+                    "size multiplier of the Sun light \"%s\" is not uniform.",
```
dictoon: Make log message all lower case.

src/appleseed/renderer/modeling/light/sunlight.cpp
```diff
@@ -125,6 +127,23 @@ namespace
...
+            {
+                RENDERER_LOG_WARNING(
+                    "Distance between Sun and scene \"%s\" is not uniform.",
+                    get_path().c_str());
+            }
+
+            // Compute SunSolidAngle that depends on distance between Sun and scene.
+            // angular_diameter = 2 * arctan(sun_radius / (distance))
+            SunSolidAngle = 2 * M_PI * (1 - cos(atan(SunRadius / m_values.m_distance)));
+
```
dictoon: Remove one blank line.

src/appleseed/renderer/modeling/light/sunlight.cpp
dictoon: Are you sure about the formula here? The argument of `cos()` should be the radius of the sun, but you passed it the angular radius (which isn't the same thing).

Also, if I plug the default values into your formula (`SunRadius` = 0.6957, `m_values.m_distance` = 149.6) I get `SunSolidAngle` = 6.79e-5 instead of 6.87e-5.

pjessesco: `angular diameter` = 2 * `atan(diameter / (2 * distance))`
`angular diameter` = 2 * atan(radius / distance)` = 2 * `theta` (in wikipedia)
`sun solid angle` = 2 * PI * (1-cos(theta))

this is how i make formula there. i think that difference came because all value(`SunRadius`, `distance`, and so on) are approximate value, not real..

pjessesco: @dictoon 
Value 6.87e-5 assumes that `angular diameter` is 0.00935 (not 0.009301), 
However `angular diameter` in [wikipedia](https://en.wikipedia.org/wiki/Angular_diameter) is 31′31″ – 32′33″ (is similar with 0.00916, 0.00946)

src/appleseed/renderer/modeling/light/sunlight.cpp
```diff
@@ -127,22 +126,20 @@ namespace
...
-
             Source* distance_src = get_inputs().source("distance");
             assert(distance_src != nullptr);
             if (distance_src->is_uniform())
                 distance_src->evaluate_uniform(m_values.m_distance);
             else
             {
                 RENDERER_LOG_WARNING(
-                    "Distance between Sun and scene \"%s\" is not uniform.",
+                    "distance between sun and scene \"%s\" is not uniform.",
```
pjessesco: I can't understand

> Here we should probably use the default Sun-scene distance of 149.6 million km.

franz said.

dictoon: What I meant is this:

Currently, if the distance source is not uniform, you print a warning (which is the right thing to do) but `m_values.m_distance` is left uninitialized. I suggest to do two things:

1. In the case where the distance source is not uniform, use the default distance.
2. Modify the message to indicate that appleseed will use the default distance.

E.g.

    Source* distance_src = get_inputs().source("distance");
    assert(distance_src != nullptr);
    if (distance_src->is_uniform())
        distance_src->evaluate_uniform(m_values.m_distance);
    else
    {
        RENDERER_LOG_WARNING(
            "distance between sun and scene \"%s\" is not uniform, using default value of 149.6 million km.",
            get_path().c_str());
        m_values.m_distance = 149.6;
    }

PR #1648, opened 2017-10-02
Fix vs2012 warning - can't convert from double to float

usakhelo:
Fixes Visual Studio 2012 warning 

> warning C4305: '=' : truncation from 'double' to 'float'


dictoon:
Thanks!

PR #1649, opened 2017-10-03
Cleanups around BRDFs

dictoon:


src/appleseed/renderer/modeling/bsdf/bsdfwrapper.h
```diff
@@ -210,9 +232,30 @@ float BSDFWrapper<BSDFImpl>::evaluate_pdf(
     assert(foundation::is_normalized(outgoing));
     assert(foundation::is_normalized(incoming));
 
+    if (Cull)
```
est77: This code fragment is repeated. Maybe we could refactor it into a function?

Issue #93, opened 2011-10-19
Get rid of std::auto_ptr<> (deprecated in C++11)


dictoon:
The idea is to replace `std::auto_ptr<>` by [`std::unique_ptr<>`](http://en.cppreference.com/w/cpp/memory/unique_ptr) everywhere.

Since `std::unique_ptr<>` supports custom _deleters_, it could also replace `foundation::auto_release_ptr<>` (with an appropriate deleter that would call `release()` to delete the instance).

This task implies that we accept C++11 constructs in the code base. This will require to first check that all the compilers we are using support a reasonable subset of C++11.

Narann:
Following [VFX Reference Platform](http://www.vfxplatform.com/), C++11 products are supposed to be shipped in 2016. Who can't switch to C++11 exactly?

dictoon:
I think Maya is the biggest roadblock right now, as far as I know on Linux it's still compiled with an old version of gcc (4.2? not sure).

Narann:
From [Maya 2016 doc](http://help.autodesk.com/view/MAYAUL/2016/ENU/?guid=__files_Setting_up_your_build_environment_Linux_compiler_requirement_htm):

> Maya is built on RHEL 6.5 which has a system default gcc 4.4.7 compiler.
>
> Use the system default g++ compiler; or, if additional C++11 features are required, use Red Hat’s Developer Toolset 2.1.

est77:
Done in #1655

PR #1655, opened 2017-10-06
clang-tidy: modernize replace auto_ptr

est77:
Automatic changes, using clang-tidy -fix, with only minor manual fixes.

It seems gcc 4.8 has issues with unqualified move calls in some specific cases. Some std:: qualifiers are needed for the code to compile.

Some new headers are also needed to avoid errors when using unique_ptr with incomplete classes.


src/appleseed.python/bindmasterrenderer.cpp
```diff
@@ -50,8 +51,10 @@ namespace
     // A class that wraps MasterRenderer and keeps a Python
     // reference to the project object to prevent it being
     // destroyed by Python before the MasterRenderer is destroyed.
-    struct MasterRendererWrapper
+    class MasterRendererWrapper
+      : NonCopyable
```
dictoon: Inherit publicly?

src/appleseed.python/bindmasterrenderer.cpp
```diff
@@ -85,35 +88,33 @@ namespace
         }
 
         bpy::object                     m_project;
-        std::auto_ptr<MasterRenderer>   m_renderer;
+        std::unique_ptr<MasterRenderer> m_renderer;
     };
 
-    std::auto_ptr<MasterRendererWrapper> create_master_renderer(
+    std::shared_ptr<MasterRendererWrapper> create_master_renderer(
```
dictoon: Why `shared_ptr<>` here?

src/appleseed.studio/mainwindow/project/attributeeditor.cpp
```diff
@@ -36,6 +36,7 @@
 
 // Qt headers.
 #include <QLayout>
+#include <utility>
```
dictoon: Not a Qt header.

src/appleseed.studio/mainwindow/project/disneymateriallayerui.cpp
```diff
@@ -185,19 +186,19 @@ void DisneyMaterialLayerUI::create_input_widgets(const Dictionary& values)
...
                     : create_colormap_input_widgets(im) :
             input_type == "text" ? create_text_input_widgets(im) :
-            auto_ptr<IInputWidgetProxy>(0);
+            unique_ptr<IInputWidgetProxy>(nullptr);
 
         assert(widget_proxy.get());
         connect(widget_proxy.get(), SIGNAL(signal_changed()), SIGNAL(signal_apply()));
 
-        m_widget_proxies.insert(input_name, widget_proxy);
+        m_widget_proxies.insert(input_name, std::move(widget_proxy));
```
dictoon: `std::` qualifier redundant? (In many other places as well.)

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -1736,11 +1736,11 @@ namespace
 
 void MainWindow::slot_clear_render_region()
 {
-    auto_ptr<RenderingManager::IScheduledAction> clear_render_region_action(
+    unique_ptr<RenderingManager::IScheduledAction> clear_render_region_action(
         new ClearRenderRegionAction(m_attribute_editor));
 
     if (m_rendering_manager.is_rendering())
-        m_rendering_manager.schedule(clear_render_region_action);
+        m_rendering_manager.schedule(std::move(clear_render_region_action));
```
dictoon: `std::` qualifier redundant?

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -1762,7 +1762,7 @@ void MainWindow::slot_set_render_region(const QRect& rect)
     }
     else
     {
-        m_rendering_manager.schedule(set_render_region_action);
+        m_rendering_manager.schedule(std::move(set_render_region_action));
```
dictoon: `std::` qualifier redundant?

dictoon:
Builds with VS 2012 so good to merge once review comments have been addressed!

Then we can add a comment to https://github.com/appleseedhq/appleseed/issues/93 that the replacement has finally been done.

src/appleseed.studio/mainwindow/project/disneymateriallayerui.cpp
est77: These qualifiers are needed for gcc 4.8 to compile the code.
Some libraries define move methods and functions, for example Qt and the compiler does not pick std::move.

src/appleseed.python/bindmasterrenderer.cpp
est77: We cannot pass unique_ptr to bpy::make_constructor because unique_ptr is non-copyable.

est77:
Fixed and rebased

PR #1657, opened 2017-10-08
Fix pow_int()

pjessesco:
I run tests and it passed. 

> Could you also add unit tests for the variant of pow_int() taking the exponent as a runtime argument?

How can I make pow_int() take runtime argument?


pjessesco:
I added test

dictoon:
Thank you!

Issue #1645, opened 2017-10-02
Pass renderer::ShadingPoint instead of just UV coordinates to renderer::Source::evaluate()


dictoon:
This will allow appleseed-max to support most 3ds Max procedural maps.

dictoon:
We're now passing a new data structure, `renderer::SourceInputs`.

PR #1660, opened 2017-10-10
Pass shadingpoint to Source::evaluate

KuribohG:
This solves issue #1645.

### DIFF
--- a/src/appleseed/renderer/modeling/input/source.h
+++ b/src/appleseed/renderer/modeling/input/source.h
@@ -42,7 +42,7 @@
 #include "main/dllsymbol.h"
 
 // Forward declarations.
-namespace renderer      { class TextureCache; }
+namespace renderer      { class TextureCache; class ShadingPoint; }
 
 namespace renderer
 {
@@ -66,7 +66,7 @@ class APPLESEED_DLLSYMBOL Source
     // Return true if the source is uniform, false if it is varying.
     bool is_uniform() const;
 
-    // Evaluate the source at a given shading point.
+    // Evaluate the source at a given texture coordinate.
     virtual void evaluate(
         TextureCache&               texture_cache,
         const foundation::Vector2f& uv,
@@ -91,6 +91,34 @@ class APPLESEED_DLLSYMBOL Source
     virtual void evaluate(
         TextureCache&               texture_cache,
         const foundation::Vector2f& uv,
+        Spectrum&                   spectrum,
+        Alpha&                      alpha) const;
+
+    // Evaluate the source at a given shading point.
+    virtual void evaluate(
+        TextureCache&               texture_cache,
+        const ShadingPoint&         shading_point,
+        float&                      scalar) const;
+    virtual void evaluate(
+        TextureCache&               texture_cache,
+        const ShadingPoint&         shading_point,
+        foundation::Color3f&        linear_rgb) const;
+    virtual void evaluate(
+        TextureCache&               texture_cache,
+        const ShadingPoint&         shading_point,
+        Spectrum&                   spectrum) const;
+    virtual void evaluate(
+        TextureCache&               texture_cache,
+        const ShadingPoint&         shading_point,
+        Alpha&                      alpha) const;
+    virtual void evaluate(
+        TextureCache&               texture_cache,
+        const ShadingPoint&         shading_point,
+        foundation::Color3f&        linear_rgb,
+        Alpha&                      alpha) const;
+    virtual void evaluate(
+        TextureCache&               texture_cache,
+        const ShadingPoint&         shading_point,
         Spectrum&                   spectrum,
         Alpha&                      alpha) const;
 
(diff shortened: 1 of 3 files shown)


src/appleseed/renderer/modeling/input/source.h
```diff
@@ -42,7 +42,7 @@
 #include "main/dllsymbol.h"
 
 // Forward declarations.
-namespace renderer      { class TextureCache; }
+namespace renderer      { class TextureCache; class ShadingPoint; }
```
est77: We use one line for each forward declared class, ordered alphabetically.

src/appleseed/renderer/modeling/input/source.h
```diff
@@ -66,7 +66,7 @@ class APPLESEED_DLLSYMBOL Source
     // Return true if the source is uniform, false if it is varying.
     bool is_uniform() const;
 
-    // Evaluate the source at a given shading point.
+    // Evaluate the source at a given texture coordinate.
     virtual void evaluate(
```
est77: I think that you want to pass a ShadingPoint to this method too.

src/appleseed/renderer/modeling/input/source.h
```diff
@@ -66,7 +66,7 @@ class APPLESEED_DLLSYMBOL Source
...
         TextureCache&               texture_cache,
         const foundation::Vector2f& uv,
```
est77: I think that you want to pass ShadingPoint here too

src/appleseed/renderer/modeling/input/source.h
```diff
@@ -183,6 +211,57 @@ inline void Source::evaluate(
     evaluate_uniform(spectrum, alpha);
 }
 
+inline void Source::evaluate(
+	TextureCache&                   texture_cache,
+	const ShadingPoint&             shading_point,
+	float&                          scalar) const
+{
+	evaluate_uniform(scalar);
```
dictoon: Regarding indentation, we use 4 spaces (no tabs).

src/appleseed/renderer/modeling/input/source.h
```diff
@@ -183,6 +211,57 @@ inline void Source::evaluate(
...
+
+inline void Source::evaluate(
+	TextureCache&                   texture_cache,
+	const ShadingPoint&             shading_point,
+	Spectrum&                       spectrum,
+	Alpha&                          alpha) const
+{
+	evaluate_uniform(spectrum, alpha);
+}
+
```
dictoon: Remove one blank line (a single one between methods).

src/appleseed/renderer/modeling/input/source.h
```diff
@@ -94,6 +94,34 @@ class APPLESEED_DLLSYMBOL Source
         Spectrum&                   spectrum,
         Alpha&                      alpha) const;
 
+    // Evaluate the source at a given shading point.
+    virtual void evaluate(
```
est77: You need to modify the existing methods to replace the uv argument by a ShadingPoint, not adding new ones.

KuribohG:
I forgot to change the method for TextureSource, sorry.

What should I do when the uv parameter is not from a ShadingPoint but by explicitly set its coordinate? For example in renderer::ImageSampler::sample. Should I create a ShadingPoint class object (even it has no other values in it)?

dictoon:
@KuribohG That's a good question. In the particular case you're mentioning, we know (I think) that the source is actually a `TextureSource`. We could have an overload taking plain UV coordinates in `TextureSource` only. What do you think?

usakhelo:
Hi,
I've finally looked into this problem and I have few suggestions:
1. I don't see it's possible to replace all UV params in all evaluate methods. Environment and light source produce 'lookup' vectors in several steps and I could not find nice way for them to pass ShadingPoint to `evaluate()`.
2. It's probably possible to make ShadingPoint an additional parameter to `Source::evaluate()` and to `InputArray::evaluate` and then pass dummy value when real ShadingPoint is not accessible (edf-s and lights). In all other places `uv` parameter will be a dummy value. Not a nicest solution either.
3. Another option I think is to add `evaluate` overload taking ShadingPoint to `Source` class and then make `InputArray::evaluate` choose the right one based on `InputFormat` enum. This will probably require adding new formats to `InputFormat` enum and using those formats when declare inputs. I don't know if it's possible and if it's something logical.
4. One other way is to add `evaluate_point` function to `InputArray` and call that function instead of `evaluate` or `evaluate_uniform` in cases where ShadingPoint is available. This method also requires adding new overrides to `Source` method I think. I think this option can be something doable.

These are options that I can think of. If any of these seems logical I can fix it. If not, then not a big deal. Support of 3d procedural textures is probably not the top priority, especially when UV procedurals are supported normally.

Thanks,
Sergo.

dictoon:
Thanks @usakhelo for the detailed analysis of the situation, much appreciated!

Even 2D procedurals may need more inputs than just UVs. For instance they might rely on UV derivatives, etc.

I like solutions 3 and 4. For solution 3, I'm not sure we need to introduce a new `InputFormat`. If we have a `ShadingPoint` when comes the moment of evaluating inputs, then we should pass that. Otherwise we should just pass UVs. This seems reasonable to me. What do you think?

dictoon:
Maybe I'll give this a try today.

usakhelo:
3 may require an overload added to InputArray then?

dictoon:
> 3 may require an overload added to InputArray then?
Yes, absolutely.

usakhelo:
Hi, @dictoon ,
I tried to implement option 3. I didn't do pull request, but here's compare view of the changes: https://github.com/appleseedhq/appleseed/compare/master...usakhelo:test_evaluate_shading_point?expand=1

What I did was to add new methods to source.h (and to texturesource.h for testing) and couple of methods to InputArray. I also changed some of the calls to InputArray::evaluate (edf, bsdf, bssrdf) to use shadingpoint argument. Looks like it's working, in the appleseed.studio. But I couldn't compile max plugin with those changes because of linking error complaining about missing `ShadingPoint::fetch_source_geometry` symbol. I think it's because `get_uv` is the inline function?.. Also it's probably not very good to include files in plugin from kernel folder?

dictoon:
> I couldn't compile max plugin with those changes because of linking error complaining about missing ShadingPoint::fetch_source_geometry symbol

Oh, right... Indeed, I had forgotten about that. We probably don't want to expose `renderer::ShadingPoint` to the outside. The right solution is maybe to introduce a small class with POD ("plain old data") members, and pass that to the `evaluate()` call chain.

usakhelo:
That small class should actually make things much simpler, right? It can be created where shaderpoint is not present and only have uv properties filled in. So all the evaluate function will get that class and there will be no need for overloads.
Where should that class go? Into source.h?

dictoon:
Superseded by #1696.

PR #1661, opened 2017-10-16
Misc work

est77:
- Allow saving custom comments when writting projects. Useful when saving archives for example, 
to save a bounding box for the scene and similar cases.

- Remove old Py3 compat defines.
- Misc fixes.


src/appleseed/renderer/modeling/project/projectfilewriter.cpp
```diff
@@ -943,6 +945,19 @@ bool ProjectFileWriter::write_plain_project_file(
...
+
+        // Write any optional comments.
+        if (extra_comments)
+        {
+            vector<string> lines;
+            split(extra_comments, "\n", lines);
+
+            for (const auto& line : lines)
+            {
+                if (line.empty() == false)
```
dictoon: Or more idiomatically, `!line.empty()`.

PR #1665, opened 2017-10-20
Simpler GGX sampling

est77:
Implemented method described in:
"A Simpler and Exact Sampling Routine for the GGX Distribution of Visible Normals."


est77:
> Nice! Is the sampling better, or just simpler?

I think just simpler. Noise levels seem similar to the previous code.

PR #1668, opened 2017-10-23
Preliminary support for procedural objects & full support for entity plugins

dictoon:


sandbox/samples/cpp/sphereobject/sphereobject.cpp
```diff
@@ -0,0 +1,224 @@
...
+        }
+
+        // Return a string identifying this object model.
+        const char* get_model() const override
+        {
+            return Model;
+        }
+
+        // Compute the local space bounding box of the object over the shutter interval.
+        asr::GAABB3 compute_local_bbox() const override
```
est77: Do we need to pass the shutter interval here?

PR #1675, opened 2017-10-26
Add SDF raymarching object plugin sample

dictoon:


sandbox/samples/cpp/distancefieldobject/distancefieldobject.cpp
```diff
@@ -0,0 +1,315 @@
...
+        asf::Lazy<asr::RegionKit>& get_region_kit() override
+        {
+            return m_lazy_region_kit;
+        }
+
+        // Access materials slots.
+        size_t get_material_slot_count() const override
+        {
+            return 1;
+        }
```
Biart95: Newline?

dictoon: I've left the newline out since these two methods go together under the banner "Access materials slots."

PR #1682, opened 2017-11-03
Add sRGB-to-linear-RGB OSL shader for Max plugin

dictoon:


src/appleseed.shaders/src/max/as_max_srgb_to_linear_rgb.osl
```diff
@@ -0,0 +1,45 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+float srgb_to_linear_rgb(float c)
```
luisbarrancos: @dictoon what about https://github.com/appleseedhq/appleseed/blob/master/src/appleseed.shaders/include/appleseed/color/as_transfer_functions.h#L49
We have the most common transfer functions there besides the sRGB transfer functions (Rec.709, Rec.1886, etc).

luisbarrancos: This is just to do the IDT (input transform), from a sRGB texture to working space (that being scene-linear with sRGB/Rec709 RGB primaries, D65 whitepoint for the time being), correct?

dictoon: Yes, exactly. I looked for existing code but somehow missed it.

@luisbarrancos @est77 Is it OK for Max OSL shaders to include `appleseed/color/as_transfer_functions.h`?

luisbarrancos: @dictoon sure. We also have whitepoint (chromatic adaption transforms) and gamut mapping functions there for different RGB primaries, in case we use other gamuts (i.e, AdobeRGB, etc).

dictoon: We are in 3ds Max. The users plugs a PNG image to the base color input of a Disney material (implemented as a network of OSL nodes). OSL`s `texture()` function doesn't do any color space transformation. What color space transform function should I use between the texture node and the material node?

luisbarrancos: The PNGs usually are sRGB encoded with sRGB/Rec709 RGB primaries, D65 whitepoint, unless you specifically override this, and the PNG format allows other colorimetry. I never encountered anything other than sRGB-sRGB-D65 in the wild though, so assuming a PNG is using this colorimetry is going to cover the vast majority if not all, of the practical cases one would encounter.

As for OSL texture(), that is correct. We have 2 choices here.
 1. do the texture lookup with the PNG file, and transform from input space to scene-linear (sRGB-sRGB-D65 to scene-linear-sRGB-D65)
 2. use a texturing tool, i.e: maketx, to do the conversion to a *.tx file, then pass this to OSL texture().

The best way is actually choice 2, and this option might also read OCIO configs, but we might not be doing texture preprocessing in Max yet, nor Maya for that matter, so choice 1 is the fallback.
For this choice, the transform would be applying the sRGB EOTF to the output color of texture(), before it gets plugged into the material's inputs.
Does Max allow a user to choose the input space for a texture?

dictoon: > Does Max allow a user to choose the input space for a texture?

I'm not sure it does.

In any case, I don't expect most Max users to have any clue what all this means :) All they want is to plug a PNG (or god forbid, JPEG) non-tiled, non-mip-mapped texture into a material's color input and be done with it.

For choice 1, is `sRGB_EOTF()` the right function?

dictoon: Automatic texture preprocessing could be an interesting option (something the user would choose to activate should he care for faster, more accurate results), but we don't have that yet.

luisbarrancos: For choice 1 is sRGB_EOTF() yes.
I think we had test code for using maketx to preprocess textures, but @est77 might need to give his feedback here since i quite honestly forgot what was the state of this.

dictoon: Cool, thanks!

dictoon: PR amended!

PR #1685, opened 2017-11-05
Persist file filters in image/texture file dialogs

### LINKED ISSUE #1634: Persist file filters in image/texture file dialogs

pjessesco:
This fix #1634


dictoon:
Thanks! This looks _a lot_ better :)

One thing I forgot to explain: It's desirable that appleseed compiles cleanly at every commit. This makes finding bugs and regressions much easier since one that checkout any revision and appleseed will compile.

Since it's not the case with your PR, I'll "squash" it, i.e. I'll merge it as a single commit. That means on your end you'll need to fetch `appleseedhq/master`, then reset your local `master` branch to point to `appleseedhq/master`.

Thanks for the good work and for your patience! Let me know if you'd like another issue to tackle.

PR #1688, opened 2017-11-07
Alembic plugin

Narann:
I open this PR to start preliminary discussion about alembic support. As I'm not fluent in cpp, there is certainly many points where the code is badly organized.

This code is WIP and not intends to be merged yet but as we have xform and dform support, I think it can be interesting to start this discussion.


sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,833 @@
+
+//
+// This source file is part of appleseed.
+// Visit http://appleseedhq.net/ for additional information and resources.
+//
+// This software is released under the MIT license.
+//
+// Copyright (c) 2017 Esteban Tovagliari, The appleseedhq Organization
```
dictoon: Put your name in here!

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,833 @@
...
+#include <vector>
+
+namespace asf = foundation;
+namespace asr = renderer;
+
+namespace
+{
+
+// roll into given matrix stack and return the flatten one
+asf::Matrix4d flatten_xform(std::vector<asf::Matrix4d> mtx_stack)
```
dictoon: Pass by const ref to avoid an expensive copy.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,833 @@
...
+{
+    // the matrix we will return
+    auto out_mtx = asf::Matrix4d::identity();
+
+    // mutiply the each matrix of the stack from bottom to top
+    for(auto mtx : mtx_stack)
```
dictoon: Always insert a space after `if`, `for`, `while`, etc.

Coding & formatting guidelines: https://github.com/appleseedhq/appleseed/wiki/Coding-Philosophy-and-Guidelines

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,833 @@
...
+        out_mtx = mtx * out_mtx;
+    }
+
+    return out_mtx;
+}
+
+// roll into given xform sequence stack and return the flatten one
+asr::TransformSequence flatten_xform_seq(const std::vector<asr::TransformSequence>& xform_seq_stack)
+{
+    auto out_xform_seq = asr::TransformSequence();
```
dictoon: I think this creates a copy.

Simply do this:

    asr::TransformSequence out_xform_seq;

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,833 @@
...
+
+    if(xform_seq_stack.size() == 1)
+    {
+        // we don't need to flatten anything
+        // TODO: Maybe we can return here?
```
dictoon: Yes, I think we should.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,833 @@
...
+    out_xform_seq.optimize();
+    out_xform_seq.prepare();
+
+    return out_xform_seq;
+}
+
+// return if given time samples are linearly time spaced
+bool linearly_sampled(const Alembic::Abc::chrono_t start_time,
+                      const Alembic::Abc::chrono_t end_time,
+                      std::vector<Alembic::Abc::chrono_t> samples)
```
dictoon: Pass vector by const ref.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,833 @@
...
+        }
+    }
+    out_xform_seq.optimize();
+    out_xform_seq.prepare();
+
+    return out_xform_seq;
+}
+
+// return if given time samples are linearly time spaced
+bool linearly_sampled(const Alembic::Abc::chrono_t start_time,
```
dictoon: Add a verb to the method's name: `is_linearly_sampled`

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,833 @@
...
+    // start sample
+    auto t_accum = samples[0];
+
+    const Alembic::Abc::chrono_t epsilon = 1.0 / 10000.0;
+
+    // get each sample time and compare with a linearly sampled time
+    for(const auto t : samples)
+    {
+        std::cout << t << " " << t-t_accum << " " << (std::fabs(t-t_accum) > epsilon) << std::endl;
+        if (std::fabs(t-t_accum) > epsilon)
```
dictoon: We have more robust ways to compare floating point values:

To check if `x` and `y` are approximately equal:

    foundation::feq(x, y)

Same thing with a custom epsilon value:

    foundation::feq(x, y, 1.0e-6)

To check if `x` is approximately zero:

    foundation::fz(x)

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
dictoon: Indent contents of anonymous namespaces.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,833 @@
...
+            const auto pos = sample.getPositions()->get();
+            const auto pos_count = sample.getPositions()->size();
+            std::cout << "pos_count " << pos_count << std::endl;
+
+            // reserve space to optimize allocation
+            object->reserve_vertices(pos_count);
+
+            for(auto i = 0; i < pos_count; i++)
+            {
+                const auto v = asr::GVector3(pos[i][0], pos[i][1], pos[i][2]);
```
dictoon: It's really best not to abuse `auto` here, and do this instead:

     const asr::GVector3 v(pos[i][0], pos[i][1], pos[i][2]);

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,833 @@
...
+        // vertex positions
+        ///////////////////////////////////////////////////////////////
+        auto sample_id = 0;
+        for(const auto sample_time : sample_times)
+        {
+            const auto sample_sel = Alembic::AbcGeom::ISampleSelector(sample_time);
+
+            const auto sample = schema.getValue(sample_sel);
+
+            const auto pos = sample.getPositions()->get();
```
dictoon: What does `get()` return? Is this making an expensive copy? It's impossible to tell... I'd suggest not using `auto` when the type is not obvious.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,833 @@
...
+                n_i += face_size;
+                uv_i += face_size;
+            }
+            else if (face_size == 3)
+            {
+                const auto pt0 = face_indices[f_i++];
+                const auto pt1 = face_indices[f_i++];
+                const auto pt2 = face_indices[f_i++];
+
+                auto tri0 = asr::Triangle(pt0, pt2, pt1);
```
dictoon: asr::Triangle tri0(pt0, pt2, pt1);

sandbox/samples/cpp/alembicassembly/alembicassembly.h
```diff
@@ -0,0 +1,148 @@
...
+  : public asr::ProceduralAssembly
+{
+  public:
+    // Constructor.
+    AlembicAssembly(
+        const char*                     name,
+        const asr::ParamArray&          params);
+
+    // Return a string identifying the model of this entity.
+    const char* get_model() const override
```
dictoon: Implement all methods in the .cpp file.

sandbox/samples/cpp/alembicassembly/alembicassembly.h
```diff
@@ -0,0 +1,148 @@
...
+        asf::IAbortSwitch*      abort_switch = 0) override;
+
+  private:
+
+    // retrieve and store assembly parameters
+    void retrieve_params(const asr::Project& project);
+
+    void foo(const asr::Assembly* assembly, Alembic::Abc::IObject o);
+
+    static const char* Model;
```
dictoon: You don't need that. Just declare a _single_ `Model` constant in an anonymous namespace in the cpp file, and use it in all `get_model()` methods.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
est77: I suggest moving the transform stack into its own class.
You can optimize the code if instead of pushing a transform and then looping to flatten it everytime is needed,
you push matrix * stack.top(). Then you don't need to loop to flatten.
You can also simplify it by considering a simple matrix as a transform sequence with only one sample and unify the two cases.

Narann:
Thanks for the review! It actually provide me the obvious things I'm not doing. I will do a global pass on every points.

dictoon:
❤❤❤

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,1525 @@
...
+            {
+                sample_indices.push_back(i);
+            }
+        }
+        return sample_indices;
+    }
+
+    // extract an appleseed xform sequence from given alembic xform schema
+    asr::TransformSequence extract_xform_seq(
+            const AbcGeom::IXformSchema& schema,
```
dictoon: Fix indentation.
Align parameters.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,1525 @@
...
+    }
+
+
+    // return sample indices (with border samples) in given shutter
+    //                           open          close
+    // shutter times       ---------|--------------|------------->
+    //                  (floor)  ,<-'              '-->,  (ceil)
+    // returned samples    ------|-----------|---------|--------->
+    std::vector<AbcIndex_t> schema_to_sample_times(
+        const float shutter_open,
```
dictoon: Align parameters.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,1525 @@
...
+        std::vector<AbcIndex_t> sample_indices;
+
+        if (num_samples < 2)
+        {
+            // no sample, put the main sample (sample 0)
+            sample_indices.push_back(0);
+        }
+        else  // more than 1 sample
+        {
+            // get floor and ceil sample indices to don't miss any
```
dictoon: to _not_ miss any

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,1525 @@
...
+
+// Standard headers.
+#include <cstddef>
+#include <string>
+#include <vector>
+#include <algorithm>
+
+namespace
+{
+    // shortcuts to save space
```
dictoon: Full line comments should start with a capital and end with a period:

    // Shortcuts to save space.

(Fix everywhere.)

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,1525 @@
...
+    namespace AbcUtil    = Alembic::Util;
+
+    typedef Alembic::Abc::index_t  AbcIndex_t;
+    typedef Alembic::Abc::chrono_t AbcChrono_t;
+
+    // roll into given matrix stack and return the flatten one
+    asf::Matrix4d flatten_xform(const std::vector<asf::Matrix4d>& mtx_stack)
+    {
+        // the matrix we will return
+        asf::Matrix4d out_mtx = asf::Matrix4d::identity();
```
dictoon: A small optimization would be to initialize `out_mtx` with the first matrix from the stack (but then you can't use a range-based `for` loop).

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,1525 @@
...
+            out_mtx = mtx * out_mtx;
+        }
+
+        return out_mtx;
+    }
+
+    // roll into given xform sequence stack and return the flatten one
+    asr::TransformSequence flatten_xform_seq(const std::vector<asr::TransformSequence>& xform_seq_stack)
+    {
+        assert(xform_seq_stack.size() > 0);
```
dictoon: Prefer `empty()`:

    assert(!xform_seq_stack.empty());

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,1525 @@
...
+        {
+            // one item, we don't need to flatten anything
+            return xform_seq_stack.back();
+        }
+        else
+        {
+            // xform sequence we will return
+            asr::TransformSequence out_xform_seq;
+
+            for (const asr::TransformSequence& xform_seq : xform_seq_stack)
```
dictoon: That's a place where `auto` would make sense.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,1525 @@
...
+            out_xform_seq.optimize();
+            out_xform_seq.prepare();
+
+            return out_xform_seq;
+        }
+    }
+
+    // return if given time samples are linearly time spaced between start_time
+    // and end_time.
+    const bool are_linearly_sampled(const AbcChrono_t start_time,
```
dictoon: Format like you did with other methods (all parameters go on their own line).

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,1525 @@
...
+        // linearly sampled.
+        const AbcChrono_t increm = (end_time - start_time) / (samples.size()-1);
+
+        // start sample
+        AbcChrono_t t_accum = samples[0];
+
+        // get each sample time and compare with a linearly sampled time
+        for (const AbcChrono_t t : samples)
+        {
+            if (asf::feq(t, t_accum))
```
dictoon: Are you sure about the logic here? Shouldn't it be `!asf::feq()`?

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,1525 @@
...
+            {
+                return false;
+            }
+
+            t_accum += increm;
```
dictoon: In theory errors could accumulate here. It would be more robust (albeit slightly slower) to recompute a linear sampled `t` value at each iteration, for instance using `foundation::fit()`.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,1525 @@
...
+
+        if (num_samples < 2)
+        {
+            // no sample, put the main sample (sample 0)
+            sample_indices.push_back(0);
+        }
+        else  // more than 1 sample
+        {
+            // get floor and ceil sample indices to don't miss any
+            const std::pair<AbcIndex_t, AbcChrono_t>& open_pair =
```
dictoon: Just checking: are you sure that `getFloorIndex()` returns a _reference_ and not a value?

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,1525 @@
...
+                time_sampling->getFloorIndex(shutter_open, num_samples);
+
+            const std::pair<AbcIndex_t, AbcChrono_t>& close_pair =
+                time_sampling->getCeilIndex(shutter_close, num_samples);
+
+            sample_indices.reserve(close_pair.first-open_pair.first+1);
```
dictoon: Add spaces around `+` operator.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,1525 @@
...
+            // time
+            const AbcChrono_t t = time_sampling->getSampleTime(i);
+
+            // alembic xform -> appleseed xform
+            const auto sample = schema.getValue(i);
+            const auto abc_mtx = sample.getMatrix();
+            const auto mtx = asf::Matrix4d(abc_mtx);
+            const auto xform = asf::Transformd::from_local_to_parent(mtx);
+
+            // as all samples are time offsetted, we offset them back
```
dictoon: *offsetted* -> *offset*

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,1525 @@
...
+        }
+
+        return xform_seq;
+    }
+
+    // log information for given alembic archive
+    void log_archive(Abc::IArchive& archive,
+                     AbcCoreFac::IFactory::CoreType core_type)
+    {
+        RENDERER_LOG_INFO("archive name: %s\n"
```
dictoon: Move all arguments to their own line.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,1525 @@
...
+                          "        time sampling count: %d\n"
+                          "        version: %d",
+            archive.getName().c_str(),
+            archive.getNumTimeSamplings(),
+            archive.getArchiveVersion());
+        switch(core_type)
```
dictoon: Add space after `switch`.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,1525 @@
...
+                RENDERER_LOG_INFO("        core type: Ogawa");
+                break;
+            case AbcCoreFac::IFactory::kLayer:
+                RENDERER_LOG_INFO("        core type: Layer");
+                break;
+            case AbcCoreFac::IFactory::kUnknown:
+                RENDERER_LOG_WARNING("        core type: Unknown");
+                break;
+            default:
+                break;  // this should never happen (tm)
```
dictoon: In that case, assert:

    assert(!"This should never happen.");

Or use `assert_otherwise` (defined in `foundation/utility/otherwise.h`).

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,1525 @@
...
+        }
+    }
+
+    // log information for given alembic object
+    void log_obj(const Abc::IObject& obj)
+    {
+        RENDERER_LOG_INFO("%s\n"
+                          "  children count: %zd",
```
dictoon: I'm not sure what's the type returned by `getNumChildren()`, but if it's `std::size_t` then use `FMT_SIZE_T` defined in `foundation/platform/types.h`.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,1525 @@
...
+            obj.getFullName().c_str(),
+            obj.getNumChildren());
+
+        const Abc::MetaData& metadata = obj.getMetaData();
+
+        RENDERER_LOG_INFO("  metadata size: %zd", metadata.size());
+
+        if (metadata.size())
+        {
+            for (const std::pair<std::string, std::string>& it : metadata)
```
dictoon: Ideal case for `const auto&`.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,1525 @@
...
+
+            if (m_verbose)
+                log_obj(root);
+
+            // root children
+            const size_t children_count = root.getNumChildren();
+
+            if (children_count)
+            {
+                for (size_t i = 0; i < root.getNumChildren(); i++)
```
dictoon: The following idiom guarantees that the upper bound is not reevaluated at each iteration:

    for (size_t i = 0, e = root.getNumChildren(); i < e; i++)

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,1525 @@
...
+            else if (AbcGeom::IPolyMesh::matches(h))
+            {
+                const auto polymesh = AbcGeom::IPolyMesh(obj);
+
+                const AbcGeom::IPolyMeshSchema& schema = polymesh.getSchema();
+
+                if (m_verbose)
+                    log_polymesh_schema(schema);
+
+                /*const auto sample_times = schema_to_sample_timesOO(m_shutter_open+m_time_offset,
```
dictoon: Remove dead code.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,1525 @@
...
+                std::vector<size_t> n_idxs;  // normal indices
+
+                const auto n_param = schema.getNormalsParam();
+
+                if (n_param.valid())
+                {
+                    switch(n_param.getScope())
+                    {
+                        case AbcGeom::kConstantScope:
+                            std::cout << "normal kConstantScope" << std::endl;
```
dictoon: Remove debug code.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,1525 @@
...
+        }
+
+        // Return metadata for this camera model.
+        asf::Dictionary get_model_metadata() const override
+        {
+            return
+                asf::Dictionary()
+                    .insert("name", Model)
+                    .insert("label", "Alembic Camera")
+                    .insert("default_model", "true");
```
dictoon: The default camera model is the pinhole camera; remove `default_model` altogether.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,1525 @@
...
+        asf::auto_release_ptr<asr::Assembly> create(
+            const char*             name,
+            const asr::ParamArray&  params) const override
+        {
+            return asf::auto_release_ptr<asr::Assembly>(
+                new AlembicAssembly(name, params));
+        }
+
+      private:
+
```
dictoon: Remove blank line.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,1525 @@
...
+        const char* Model = "alembic_assembly";
+
```
dictoon: Remove blank line.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,1525 @@
...
+        asf::auto_release_ptr<asr::Camera> create(
+            const char*              name,
+            const asr::ParamArray&   params) const override
+        {
+            return asf::auto_release_ptr<asr::Camera>(
+                new AlembicCamera(name, params));
+        }
+
+      private:
+
```
dictoon: Remove blank line.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -0,0 +1,1525 @@
...
+            // generate xform sequence stack from our camera
+            const std::vector<asr::TransformSequence> xform_seq_stack = xform_seq_for_obj(
+                cam_obj,
+                shutter_open,
+                shutter_close,
+                time_offset);
+
+            if (xform_seq_stack.size())
+            {
+                const asr::TransformSequence xform_seq = flatten_xform_seq(xform_seq_stack);
```
dictoon: Get rid of temporary variable?

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
Narann: What about ` assert(xform_seq_stack.size())` so there is no negative operator (less cognitive load).

Narann:
Thanks for the kind words, I will fix those asap.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
dictoon: That has the wrong semantic: `size()` does not return a boolean, it doesn't make sense to check if `size()` is true or false. The fact that 0 evaluates to false and other values to true is inconsequential.

dictoon: Side remark: In C#, `System.Linq` provides `Any()`:

    Debug.Assert(xform_seq_stack.Any());

Narann: Side remark: In Python:

```python
if xform_seq_stack:
```
B-[

dictoon: Even

    assert xform_seq_stack

:) But again not quite happy about the semantics it carries (or the lack thereof)... Maybe it's just me.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
Narann: Done.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
Narann: Done.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
Narann: Done.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
Narann: Done.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
Narann: Done.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
Narann: Done.
Good catch! This code is not used yet as its purpose was to prevent the user of bad sampling. I just don't know yet _when_ I will check and report.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
Narann: Maybe I'm tired but I don't see how a remap can help here.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
Narann: You mean from this:
```c++
    std::vector<AbcIndex_t> schema_to_sample_times(
        const float shutter_open,
        const float shutter_close,
        const size_t num_samples,
        const Abc::TimeSamplingPtr time_sampling)
```
To this:
```c++
    std::vector<AbcIndex_t> schema_to_sample_times(
        const float                shutter_open,
        const float                shutter_close,
        const size_t               num_samples,
        const Abc::TimeSamplingPtr time_sampling)
```
?

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
Narann: Done.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
Narann: No it doesn't. I was doing this to save a copy. Should I remove the ref identifier?

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
Narann: Done.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
Narann: Done.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
Narann: Done (here and everywhere else).

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
Narann: Done.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
Narann: Done.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
Narann: Done.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
Narann: Done.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
Narann: Done.

Nice trick!

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
Narann: Done.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
Narann: Done.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
Narann: Done.

I like to have one line doing one thing and let the compiler optimize. But in this case it's pretty obvious.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
Narann: Done.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
Narann: Done.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
dictoon: Exactly!

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
dictoon: > Should I remove the ref identifier?

Yes, because as the code stands right now, you're keeping a reference to a temporary object that no longer exists outside `getFloorIndex()`, i.e. it's broken (even if it may occasionally work).

Narann:
I think it can be merged now. Rougth instruction:

You need to compile alembic (I did in .so).

Then:
```bash
# to make
export PATH=<alembic_dir>/bin:$PATH
export LD_LIBRARY_PATH=<alembic_dir>/lib:$LD_LIBRARY_PATH
cd <appleseed>/sandbox/samples/cpp/alembicassembly/build
make
```
Then:
```bash
# to render
export PATH=<alembic_dir>/bin:$PATH
export LD_LIBRARY_PATH=<alembic_dir>/lib:$LD_LIBRARY_PATH
cd <appleseed>/sandbox/bin/Release
./appleseed.cli <appleseed>/sandbox/samples/cpp/alembicassembly/cam.appleseed -o <appleseed>/sandbox/samples/cpp/alembicassembly/foo.png
```
The last line can be .studio instead of .cli.

There is a simple scene samples but you would need to provide your own alembic to test. We should do some alembic tests at some points.

This assembly provide both assembly and camera plugin. Documentation is here:

https://github.com/appleseedhq/appleseed/wiki/Alembic-Assembly
https://github.com/appleseedhq/appleseed/wiki/Alembic-Camera

Don't hesitate if you have question but for me, it's ready to go.

dictoon:
Amazing Dorian, thanks for the great work!

I haven't checked the code yet. If I review it and make comments, will you have time to address them? Or should we "adopt" the code and take care of the rest ourselves? Either way is fine, just le me know.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
Narann: I did it, Github diff just didn't catch it.

Narann:
Thanks. You already reviewed the code month ago. I did all modifications and the branch hasn't change that much.

There is one single comment to address (ref usage where I should not). I will do that quickly.

If the whole design is OK, I now have two option:
1) I delete addressed comments to keep this PR clean so you can do more comments and I can address them.
2) Merge then others modify things.

If the design is bad, yes, we should continue to discuss this but as we kinda discussed it at our last meetup, then I inspired myself from other exporters (in the Alembic repo) then you reviewed the code a first type, I think the potential design issues would have raised sooner.

I think this maybe time to let "real" dev do the job now. I suggest to merge, don't worry too much "my" code. Yes, it's "my baby" but I think you guys will take best decisions for it.

If the questions are only related to "you should(n't) use ref here", "align parameters", it will be quicker if you take care of it (I think).

I realize there is a lack not in the code itself but in the test suite. Alembic is too broad to have simple tests, you have to have complex alembics to test.

I think I will focus on writing a script that generate various alembic situation so we can have a "rebuildable" test suite for our alembic support. This script would be runnable in Maya to generate many different alembics.

I think this effort is something not that many appleseed devs can take care easily (while keeping care of my code is something you can do).

With such test suite, we should be able to have regression test for our alembic support.

What do you think ?

dictoon:
That sounds excellent! I will merge then.

Out of curiosity, why is a script generating a test suite better than a bunch of Alembic files?

Narann:
Thanks!

I think having a script generating alembic files is better to explicitly states what's inside alembic for everyone (it's written in the script, no guess). This is something I realized during my tests. Depicts knowing what was supposed to be in the alembic files, I often open them again in Maya to compare/be sure and often wanted "this one with this export option". The idea is to have functions generating scenes, and export presets to export with specific options (different sampling, color set, uv set, etc).

I already have something an I'm working to also generate a subsampled playblast at export time (the maya scene to export) and also subsampled playblast of the alembic file (no interpolation between samples). I also plan to have jpg file of every frame, every sample, etc. And generate a world position value file for every object at every sample, same for vertex 0. So no guess. With that, you can see what the export is supposed to be on the Maya side, on the alembic side, with picture and everything without having Maya. And have values to compare.

dictoon:
Thans for the explanation, makes a lot of sense!

PR #1691, opened 2017-11-11
Add asNoise2D, asNoise3D, lookdev kit sample

luisbarrancos:
And fix small typo for choice of white point, and gamma encoding for DCI.
Test UI changes for Gaffer 0.41 on standard surface shader.


src/appleseed.shaders/include/appleseed/color/as_color_helpers.h
```diff
@@ -188,7 +188,7 @@ float as_luminance(color in_C, string colorspace, string illuminant)
     {
         Y = as_luminance_D60(in_C, colorspace);
     }
-    else if (illuminant = "D65")
```
dictoon: Woops, the OSL compiler doesn't warn in this case?

src/appleseed.shaders/src/appleseed/as_noise2d.osl
```diff
@@ -364,166 +381,196 @@ shader as_noise2d
...
         in_colorOffset,
         in_defaultColor,
         in_alphaGain,
         in_alphaOffset,
-        in_invert,
+        0, // invert
         in_alphaIsLuminance,
         out_outColor,
-        out_outAlpha);
+        out_outAlpha);
```
dictoon: Remove trailing spaces here and above (in whole file probably).

src/appleseed.shaders/src/appleseed/as_noise2d.osl
```diff
@@ -364,166 +381,196 @@ shader as_noise2d
...
 
-        amplitude *= noise_gain;
-        new_st[0] *= lacunarity;
-        new_st[1] *= lacunarity;
+            if (in_ridges)
+            {
+                signal = in_offset - signal;
+                signal *= signal;
+                signal *= weight;
+                weight = clamp(signal * 2.0, 0.0, 1.0);
```
dictoon: Is assigning to `weight` really what you intended to do here?

Also, unless `weight` is negative, `signal` cannot be negative since you square it.

src/appleseed.shaders/src/maya/as_maya_leather.osl
```diff
@@ -1,244 +0,0 @@
-
```
dictoon: What's the rationale behind remove this shader?

luisbarrancos: I don't think *leather* was in already. It's WIP. A Maya node, 3d texture, that uses Worley noise to fake leather, with some extras which i'm having a hard time replicating. Not quite ready yet, but close. It'll be the subject of the next PR together with Gaffer UI updates.

src/appleseed.shaders/src/appleseed/as_noise2d.osl
luisbarrancos: The assignment yes, we start with 1.0 in the first iteration, then store the signal value into weight for the next octave. You're right about the clamp, i wasn't sure which of min() or clamp() would be faster.

src/appleseed.shaders/include/appleseed/color/as_color_helpers.h
luisbarrancos: @dictoon no, at least i didn't noticed it.

PR #1692, opened 2017-11-11
Expose factory registrars to python

est77:
Allows creating plugin entities from python.

### DIFF
--- a/src/appleseed/renderer/modeling/frame/frame.cpp
+++ b/src/appleseed/renderer/modeling/frame/frame.cpp
@@ -386,7 +386,12 @@ namespace
         const string extension = lower_case(filepath.extension().string());
 
         ImageAttributes image_attributes = ImageAttributes::create_default_attributes();
-        // todo: add chromaticity attributes here...
+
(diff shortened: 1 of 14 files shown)


src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -386,7 +386,12 @@ namespace
         const string extension = lower_case(filepath.extension().string());
 
         ImageAttributes image_attributes = ImageAttributes::create_default_attributes();
-        // todo: add chromaticity attributes here...
+
+        // Scene linear sRGB Rec 709 chromaticities.
+        image_attributes.insert("white_xy_chromaticity", Vector2f(0.3127f, 0.3290f));
```
dictoon: Aren't these defined more globally, maybe in `colorspace.h`? Should they be?

est77: For 1.9 we'll make the working color space configurable and these constants are going to be moved to some appropiate place.

dictoon: Ok.

PR #1696, opened 2017-11-13
Pass shading point information to evaluate() functions

usakhelo:
This is another attempt to implement passing `ShadingPoint` to `Source::evaluate() functions`.
POD class is added in order to pass ShadingPoint information.


src/appleseed/renderer/modeling/input/sourceshadingpoint.cpp
```diff
@@ -0,0 +1,14 @@
+// Interface header.
```
dictoon: Add copyright notice.

src/appleseed/renderer/modeling/input/sourceshadingpoint.cpp
```diff
@@ -0,0 +1,14 @@
...
+{
+    SourceShadingPoint::SourceShadingPoint(const foundation::Vector2f uv)
+        : uvx(uv.x)
+        , uvy(uv.y)
+        , pointx(0)
+        , pointy(0)
+        , pointz(0)
+    {
+    }
+}
```
dictoon: Add newline character on the last line of the file.

src/appleseed/renderer/modeling/input/sourceshadingpoint.cpp
```diff
@@ -0,0 +1,14 @@
+// Interface header.
+#include "sourceshadingpoint.h"
+
+namespace renderer
+{
+    SourceShadingPoint::SourceShadingPoint(const foundation::Vector2f uv)
```
dictoon: We don't indent the content of top-level namespaces (since that amounts to indenting the whole file for no gain).

dictoon: Pass `uv` by const ref.

src/appleseed/renderer/modeling/input/sourceshadingpoint.h
```diff
@@ -0,0 +1,23 @@
+#ifndef APPLESEED_RENDERER_MODELING_INPUT_SOURCESHADINGPOINT_H
```
dictoon: Add copyright notice.

src/appleseed/renderer/modeling/input/sourceshadingpoint.h
```diff
@@ -0,0 +1,23 @@
...
+
+// appleseed.foundation headers.
+#include "foundation/math/vector.h"
+
+// appleseed.main headers.
+#include "main/dllsymbol.h"
+
+namespace renderer
+{
+    class APPLESEED_DLLSYMBOL SourceShadingPoint
```
dictoon: I'd suggest `SourceInputs` instead, I think it would be clearer.

Don't forget to update the file names, the header guards...

src/appleseed/renderer/modeling/input/sourceshadingpoint.h
```diff
@@ -0,0 +1,23 @@
...
+    {
+      public:
+        SourceShadingPoint(const foundation::Vector2f uv);
+        float       uvx;           // texture coordinates from UV set #0
```
dictoon: Add `m_` prefix to members.

dictoon: Separate parts with underscores:

    m_uv_x
    m_uv_y

src/appleseed/renderer/modeling/input/sourceshadingpoint.h
```diff
@@ -0,0 +1,23 @@
...
+        float       uvy;
+        double      pointx;        // world space intersection point
```
dictoon: Similarly:

    m_point_x
    m_point_y
    m_point_z

src/appleseed/renderer/kernel/intersection/intersectionfilter.cpp
```diff
@@ -34,6 +34,7 @@
 #include "renderer/global/globaltypes.h"
 #include "renderer/kernel/tessellation/statictessellation.h"
 #include "renderer/modeling/input/source.h"
+#include "renderer/modeling/input/sourceshadingpoint.h"
```
dictoon: Why is this header necessary? I haven't checked precisely, maybe it is.

src/appleseed/renderer/kernel/shading/shadingpoint.cpp
```diff
@@ -33,6 +33,7 @@
 // appleseed.renderer headers.
 #include "renderer/kernel/intersection/intersector.h"
 #include "renderer/modeling/input/source.h"
+#include "renderer/modeling/input/sourceshadingpoint.h"
```
dictoon: Same question in all the .cpp files where this is the only change.

usakhelo: An implicit conversion to shadingpoint pod class happens in all those files, in evaluate() calls. Without that headers conversion and compilation fails. Is implicit conversion good or bad? I'm not sure.

dictoon:
> An implicit conversion to SourceInputs pod class happens in all those files, in evaluate() calls. Without that header conversion and compilation fails. Is implicit conversion good or bad? I'm not sure.

Ah, I see! Well, yeah we usually choose not to enable implicit conversions since it hides computations and make the code harder to follow. I'd prefer an explicit conversion (just add `explicit` in front of your constructor).

The rest looks pretty neat, thanks for the changes!

dictoon:
Looks good, will merge as soon as Travis passes! Thanks man.

usakhelo:
Thank you @dictoon. It does nothing right now since it doesn't carry any additional information but it's easy to add anything.

PR #1703, opened 2017-11-16
Bundle oiiotool

dictoon:


est77:
We need to add oiitool here for travis to pass.
https://github.com/appleseedhq/appleseed/blob/master/travis/build.sh#L45

PR #1704, opened 2017-11-16
Clear AOV images when clearing the main image.

est77:
Fix the issue when doing final renders with AOVs + render regions in studio.
When you re-render with a smaller region, the previous pixels outside the region were not cleared.

### DIFF
--- a/src/appleseed.python/bindframe.cpp
+++ b/src/appleseed.python/bindframe.cpp
@@ -142,7 +142,6 @@ void bind_frame()
         .def("image", &Frame::image, bpy::return_value_policy<bpy::reference_existing_object>())
         .def("aov_images", &Frame::aov_images, bpy::return_value_policy<bpy::reference_existing_object>())
 
-        .def("clear_main_image", &Frame::clear_main_image)
         .def("write_main_image", &Frame::write_main_image)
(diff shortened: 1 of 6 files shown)


src/appleseed/renderer/modeling/aov/aov.h
```diff
@@ -86,6 +86,9 @@ class APPLESEED_DLLSYMBOL AOV
     // Return a reference to the AOV image.
     foundation::Image& get_image() const;
 
+    // Clear the AOV image.
```
dictoon: *to transparent black.*

src/appleseed.python/bindframe.cpp
```diff
@@ -142,7 +142,6 @@ void bind_frame()
         .def("image", &Frame::image, bpy::return_value_policy<bpy::reference_existing_object>())
         .def("aov_images", &Frame::aov_images, bpy::return_value_policy<bpy::reference_existing_object>())
```
dictoon: We're not exposing `Frame::clear_main_and_aov_images()`?

est77: There's no use for it in python I think.
I plan to clean the python bindings for the 1.9 release. There are a number of classes and methods that shouldn't be exposed. This method was one of them.

PR #1709, opened 2017-11-19
Small appleseed.studio fix

dictoon:


src/appleseed.python/studio/plugins.py
```diff
@@ -32,7 +32,8 @@
 
 
 def load_plugins(bundled_plugins_path):
-    load_plugins_from_dir(bundled_plugins_path)
+    if os.path.isdir(bundled_plugins_path):
```
Narann: It's up to the caller to check it's arguments are valid. We should not check caller's arguments, and if we need to, we should raise an exception if they are not valid. We should never ignore things silently.

In this function: If the path doesn't exists, it's not use. So you are in a situation with a function which explicitly ask for a path and do something or not with it. That's not critical but if you let those kind of _approach_ in you're code, you will end with each fonction checking it's arguments (with the core of some function bigger to check arguments than actually do the thing).

_Trust the caller_ is a good code sanity approach. The only reason you check arguments is for security reasons. If the caller is your code, fix it. If the caller is the API user, raise an exception.

My 2 cents.

dictoon: You're preaching to the choir! As in, I've always favored trusting the caller.

It's a bit of an exception here since `load_plugins()` is the entry point called by appleseed.studio on startup. There isn't really any code "above" that.

PR #1716, opened 2017-11-20
3ds Max OSL blend shader

usakhelo:
Hi,

This PR is complement to [3ds max plugin PR](https://github.com/appleseedhq/appleseed-max/pull/82)
Adds OSL blend shader and adapt max disney shader to work it.

Thanks,
Sergo.

### DIFF
--- a/src/appleseed.shaders/CMakeLists.txt
+++ b/src/appleseed.shaders/CMakeLists.txt
@@ -301,12 +301,12 @@ source_group ("src\\gaffer\\vector" FILES
 
 set (src_max_sources
      src/max/as_max_bump_map.osl
+     src/max/as_max_blend_material.osl
      src/max/as_max_color_texture.osl
      src/max/as_max_disney_material.osl
      src/max/as_max_float_texture.osl
      src/max/as_max_normal_map.osl
      src/max/as_max_srgb_to_linear_rgb.osl
-	 src/max/as_max_blend_material.osl
 )
 list (APPEND osl_sources
     ${src_max_sources}
--- a/src/appleseed.shaders/CMakeLists.txt
+++ b/src/appleseed.shaders/CMakeLists.txt
@@ -306,6 +306,7 @@ set (src_max_sources
      src/max/as_max_float_texture.osl
      src/max/as_max_normal_map.osl
      src/max/as_max_srgb_to_linear_rgb.osl
+	 src/max/as_max_blend_material.osl
 )
 list (APPEND osl_sources
     ${src_max_sources}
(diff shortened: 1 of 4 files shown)


src/appleseed.shaders/CMakeLists.txt
```diff
@@ -306,6 +306,7 @@ set (src_max_sources
      src/max/as_max_float_texture.osl
      src/max/as_max_normal_map.osl
      src/max/as_max_srgb_to_linear_rgb.osl
+	 src/max/as_max_blend_material.osl
```
dictoon: Remove tab.

usakhelo:
Hi,
MixColor is never textured, it's just a color. Mask color is a grayscale texture.
I tried to mimic vray blend material, but it's not exactly like this. It's mix color is overridden by texture if texture amount is 100%. In my case both mix amount and texture contribute equally.

est77:
I think that the PR good. We'll merge after the 1.8 release.

Thanks!

src/appleseed.shaders/CMakeLists.txt
```diff
@@ -306,6 +306,7 @@ set (src_max_sources
      src/max/as_max_float_texture.osl
      src/max/as_max_normal_map.osl
      src/max/as_max_srgb_to_linear_rgb.osl
+     src/max/as_max_blend_material.osl
```

src/appleseed.shaders/src/max/as_max_blend_material.osl
```diff
@@ -0,0 +1,90 @@
...
+
+    output	closure color ClosureOut = 0
+)
+{
+    ClosureOut = BaseMtl;
+    
+    if (LayerMtl_1)
+        ClosureOut = mix(ClosureOut, LayerMtl_1, (MixColor_1 * MaskColor_1));
+    if (LayerMtl_2)
+        ClosureOut = mix(ClosureOut, LayerMtl_2, (MixColor_2 * MaskColor_2));
```
dictoon: Is this the right mixing algorithm? Or more precisely, the mixing algorithm we actually want?

Here, the weight of `LayerMtl_1` will be

    (MixColor_1 * MaskColor_1) * (1 - MixColor_2 * MaskColor_2)

It's likely that this is what we want. Just checking.

src/appleseed.shaders/CMakeLists.txt
```diff
@@ -301,6 +301,7 @@ source_group ("src\\gaffer\\vector" FILES
 
 set (src_max_sources
      src/max/as_max_bump_map.osl
+     src/max/as_max_blend_material.osl
```
dictoon: Not quite ordered yet but it's ok :)

dictoon:
Merged, thanks!

src/appleseed.shaders/CMakeLists.txt
usakhelo: That's shame, sorry!

PR #1737, opened 2017-11-27
Add UV transform shader for 3ds max bitmap textures

usakhelo:
Hi,
This PR adds as_max_uv_transform shader to 3ds max shaders. It supports following transformations:
* Offset on U and V axis
* Tiling on U and V axis
* Mirror on U and V axis
* Rotation on W axis
* Crop mode with crop offset and sizes

What's not supported:
* It works slightly different from max bitmap texture when texture is rotated and has some offset
* Place mode
* Real-world uv scale
* Rotation on U or V axis
* VW and WU coords
* Blur
* Any mapping mode other than Explicit Map Channel

Thanks for review!

### DIFF
--- a/src/appleseed.shaders/src/max/as_max_uv_transform.osl
+++ b/src/appleseed.shaders/src/max/as_max_uv_transform.osl
@@ -1,0 +1,131 @@
+
+//
+// This source file is part of appleseed.
+// Visit http://appleseedhq.net/ for additional information and resources.
+//
+// This software is released under the MIT license.
+//
+// Copyright (c) 2017 Sergo Pogosyan, The appleseedhq Organization
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+shader as_max_uv_transform
+(
+    float in_coordU = u,
+    float in_coordV = v,
+
+    float in_offsetU = 0.0,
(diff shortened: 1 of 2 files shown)


src/appleseed.shaders/src/max/as_max_uv_transform.osl
```diff
@@ -0,0 +1,131 @@
...
+    output float out_U = 0.0,
+    output float out_V = 0.0,
+)
+{
+    float out_u = in_coordU;
+    float out_v = in_coordV;
+    float u_center = 0.5;
+    float v_center = 0.5;
+
+    // Apply rotation before tiling/offset which is different from 3ds max native bitmap transforms
```
dictoon: 3ds Max (capital M).

Terminate full line comments with a period (`.`).

src/appleseed.shaders/src/max/as_max_uv_transform.osl
```diff
@@ -0,0 +1,131 @@
...
+    if (in_rotateW != 0.0)
+    {
+        point rot = rotate(
+            point(out_u, out_v, 0.0),
+            in_rotateW,
+            point(u_center, v_center, 0.0),
+            point(u_center, v_center, 1.0));
+
+            out_u = rot[0];
```
dictoon: Fix indentation.

src/appleseed.shaders/src/max/as_max_uv_transform.osl
```diff
@@ -0,0 +1,131 @@
...
+{
+    float out_u = in_coordU;
+    float out_v = in_coordV;
+    float u_center = 0.5;
+    float v_center = 0.5;
+
+    // Apply rotation before tiling/offset which is different from 3ds max native bitmap transforms
+    if (in_rotateW != 0.0)
+    {
+        point rot = rotate(
```
dictoon: It would make sense to use a specialized 2D rotation function, I'm not sure the OSL compiler would be to optimize this.

src/appleseed.shaders/src/max/as_max_uv_transform.osl
```diff
@@ -0,0 +1,131 @@
...
+    float in_coordU = u,
+    float in_coordV = v,
+
+    float in_offsetU = 0.0,
+    float in_offsetV = 0.0,
+
+    float in_tilingU = 1.0,
+    float in_tilingV = 1.0,
+
+    float in_rotateW = 0.0,
```
dictoon: What does the `W` mean here?

src/appleseed.shaders/src/max/as_max_uv_transform.osl
```diff
@@ -0,0 +1,131 @@
...
+    {
+        out_v *= 2;
+        if (mod(out_v, 2) >= 1.0)
+        {
+            out_v = 1 - out_v;
+        }
+    }
+
+    // "Manual" periodic tiling required for correct cropping
+    out_u = fmod (out_u, 1.0);
```
dictoon: Remove space before `(`.

src/appleseed.shaders/src/max/as_max_uv_transform.osl
usakhelo: Rotation around W axis is the only rotation that makes sense, so I only connected that input. It's perpendicular to uv plane. I kept W suffix in case we will need to add U and V rotations.

src/appleseed.shaders/src/max/as_max_uv_transform.osl
usakhelo: Is that the one from `appleseed/gaffer/transform.h` header? Looks like it compiles and renders normally. Had to change plugin a little to pass degrees instead of radians.

src/appleseed.shaders/src/max/as_max_uv_transform.osl
usakhelo: fixed by replacing rotation function

dictoon:
Thanks! Merging.

PR #1752, opened 2017-12-03
Allow more granular installation of appleseed components

danfe:
End-users are often not interested in building sample programs (and thus having header files installed), and/or running unit tests and benchmarks.  These options are left enabled by default though, so those who relied on them won't get their workflow disrupted.

While here, provide a way to disable API Reference generation even when Doxygen is installed: this simplifies things for 3rd-party packagers.


dictoon:
Squashing and merging the PR. Thanks Alexey!

danfe:
Thanks for squashing François.  I was a bit surprised to see fix commits in that pull request while it was only supposed to be one; git had been biting me again. :-(

dictoon:
No worries at all. Thanks for contributing.

PR #1757, opened 2017-12-06
Change MixAmount input type in blend osl shader, code style

usakhelo:
This is complimentary to the 3ds max plugin [PR](https://github.com/appleseedhq/appleseed-max/pull/98).

### DIFF
--- a/src/appleseed.shaders/src/max/as_max_blend_material.osl
+++ b/src/appleseed.shaders/src/max/as_max_blend_material.osl
@@ -42,15 +42,15 @@ surface as_max_blend_material
     closure color LayerMtl_8 = 0,
     closure color LayerMtl_9 = 0,
 
-    color	MixColor_1 = color(0),
-    color	MixColor_2 = color(0),
-    color	MixColor_3 = color(0),
-    color	MixColor_4 = color(0),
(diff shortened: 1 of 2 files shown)


src/appleseed.shaders/src/max/as_max_blend_material.osl
```diff
@@ -68,23 +68,23 @@ surface as_max_blend_material
     ClosureOut = BaseMtl;
     
     if (LayerMtl_1)
-        ClosureOut = mix(ClosureOut, LayerMtl_1, (MixColor_1 * MaskColor_1));
+        ClosureOut = mix(ClosureOut, LayerMtl_1, (MixAmount_1 * MaskColor_1));
```
dictoon: I suggest removing the mask colors if they are never used.

usakhelo: Mask color is the texture, we use them. If no texture then it's 1 so only mix amount affects the mix.

dictoon: Oh ok, sorry.

dictoon:
Thank you!

PR #1760, opened 2017-12-07
Restrict GTR1 roughness domain, fix Student t MDF

luisbarrancos:
We had discussed this for prepare_inputs, but GTR1 might be used by other BxDFs, and is a part of foundation math.

### DIFF
--- a/src/appleseed/foundation/math/microfacet.cpp
+++ b/src/appleseed/foundation/math/microfacet.cpp
@@ -713,9 +713,9 @@ Vector3f GTR1MDF::sample(
     const float         gamma) const
 {
     const float alpha = clamp(alpha_x, 0.001f, 0.999f);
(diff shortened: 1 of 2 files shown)


src/appleseed/foundation/math/microfacet.cpp
```diff
@@ -710,7 +712,8 @@ Vector3f GTR1MDF::sample(
     const float         alpha_y,
     const float         gamma) const
 {
-    const float alpha2 = square(alpha_x);
+    const float alpha = clamp(alpha_x, 0.001f, 0.999f);
+    const float alpha2 = square(alpha);
```
dictoon: `alpha_2` to stay consistent?

luisbarrancos: @dictoon good point.

PR #1762, opened 2017-12-07
Add optional energy compensation for glossy and metal BRDFs.

est77:
Based on [Revisiting Physically Based Shading at Imageworks](http://blog.selfshadow.com/publications/s2017-shadingcourse/imageworks/s2017_pbs_imageworks_slides.pdf)

It's optional and off by default. Also exposed to OSL as an optional param to the as_glossy and as_metal closures. Only works for the GGX and Beckmann distributions. Results are better for GGX because it looses more energy than Beckmann when roughness is high.

Rough copper:
shaderball_rough_copper_ec_comp

Rough glossy:
shaderball_rough_glossy_ec_comp


src/appleseed/renderer/meta/tests/test_energycompensation.cpp
```diff
@@ -0,0 +1,48 @@
...
+//
+
+// appleseed.foundation headers.
+#include "foundation/utility/test.h"
+
+namespace renderer
+{
+
+// From renderer/modeling/bsdf/microfacethelper.cpp.
+void write_microfacet_directional_albedo_tables_to_exr(const char* directory);
```
dictoon: I suggest moving this to `microfacethelper.h`.

src/appleseed/renderer/modeling/bsdf/glossybrdf.cpp
```diff
@@ -307,24 +407,99 @@ namespace
...
+            if (values->m_energy_compensation != 0.0f)
+            {
+                float fms;
+                float eavg;
+                microfacet_energy_compensation_term(
+                    mdf,
+                    values->m_roughness,
+                    cos_in,
+                    cos_on,
+                    fms,
```
dictoon: What do `fms` and `eavg` mean?

src/appleseed/renderer/modeling/bsdf/kelemenbrdf.cpp
```diff
@@ -124,6 +124,7 @@ namespace
     //
     // Reference:
     //
+    //   A Microfacet Based Coupled Specular-Matte BRDF Model with Importance Sampling
```
dictoon: Thanks!

src/appleseed/renderer/modeling/bsdf/metalbrdf.cpp
```diff
@@ -77,13 +77,8 @@ namespace
     //   [1] Microfacet Models for Refraction through Rough Surfaces
     //       http://www.cs.cornell.edu/~srm/publications/EGSR07-btdf.pdf
     //
-    // Todo:
-    //
-    //  This BRDF currently assumes an air | metal interface and will not
```
dictoon: Wasn't this comment helpful?

src/appleseed/renderer/modeling/bsdf/microfacethelper.cpp
```diff
@@ -0,0 +1,370 @@
...
+// Interface header.
+#include "microfacethelper.h"
+
+// appleseed.foundation headers.
+#include "foundation/core/concepts/noncopyable.h"
+#include "foundation/image/image.h"
+#include "foundation/image/tile.h"
+#include "foundation/image/exrimagefilewriter.h"
+#include "foundation/math/sampling/mappings.h"
+#include "foundation/math/microfacet.h"
```
dictoon: Order alphabetically.

src/appleseed/renderer/modeling/bsdf/microfacethelper.cpp
```diff
@@ -0,0 +1,370 @@
...
+
+//
+// References:
+//
+//   [1] A Microfacet Based Coupled Specular-Matte BRDF Model with Importance Sampling
+//       http://sirkan.iit.bme.hu/~szirmay/scook.pdf
+//
+//   [2] Revisiting Physically Based Shading at Imageworks
+//       http://blog.selfshadow.com/publications/s2017-shading-course/imageworks/s2017_pbs_imageworks_slides.pdf
+
```
dictoon: Add end of block `//` line.

src/appleseed/renderer/modeling/bsdf/microfacethelper.cpp
```diff
@@ -0,0 +1,370 @@
...
+{
+  public:
+    template <typename MDF>
+    explicit AlbedoTable(const MDF& mdf)
+    {
+        m_albedo_table = new float[TableSize * TableHeight];
+        m_avg_table = m_albedo_table + TableSize * TableSize;
+
+        float* p = m_albedo_table;
+        for (size_t j = 0; j < TableSize; ++j)
```
dictoon: `i`?

src/appleseed/renderer/modeling/bsdf/microfacethelper.cpp
```diff
@@ -0,0 +1,370 @@
...
+
+    ~AlbedoTable()
+    {
+        delete[] m_albedo_table;
+    }
+
+    void compute_average_albedo_table()
+    {
+        float* p = m_albedo_table;
+        for (size_t j = 0; j < TableSize; ++j)
```
dictoon: `i`?

src/appleseed/renderer/modeling/bsdf/microfacethelper.cpp
```diff
@@ -0,0 +1,370 @@
...
+
+  private:
+    const size_t    TableSize = 32;
+    const size_t    TableHeight = 33;
+    const size_t    SampleCount = 1024;
+
+    float*          m_albedo_table;
+    float*          m_avg_table;
+
+    inline float map_to_index(const float x) const
```
dictoon: `inline` is redundant here.

src/appleseed/renderer/modeling/bsdf/microfacethelper.cpp
```diff
@@ -0,0 +1,370 @@
...
+    {
+        return saturate(x) * static_cast<float>(TableSize - 1);
+    }
+
+    inline float lookup_table(const size_t i , const size_t j) const
```
dictoon: Here too.

src/appleseed/renderer/modeling/bsdf/microfacethelper.cpp
```diff
@@ -0,0 +1,370 @@
...
+    {
+        const size_t ii = min(i, TableSize - 1);
+        const size_t jj = min(j, TableSize - 1);
+        const float *p = m_albedo_table + (jj * TableSize) + ii;
```
dictoon: `const float* p`

src/appleseed/renderer/modeling/bsdf/microfacethelper.cpp
```diff
@@ -0,0 +1,370 @@
...
+struct AlbedoTables
+  : public NonCopyable
+{
+    AlbedoTables()
+      : m_ggx(GGXMDF())
+      , m_beckmann(BeckmannMDF())
+    {
+    }
+
+    AlbedoTable m_ggx;
```
dictoon: Move public data members before methods / constructors.

src/appleseed/renderer/modeling/bsdf/microfacethelper.cpp
```diff
@@ -0,0 +1,370 @@
...
+using namespace std;
+
+namespace bfs = boost::filesystem;
+
+namespace renderer
+{
+namespace
+{
+
+//
```
dictoon: Indent content of anoynmous namespaces.

dictoon:
Looking forward to play with this!

src/appleseed/renderer/modeling/bsdf/metalbrdf.cpp
est77: It is outdated. We removed this limitation some time ago.

est77:
There is still a small amount of energy lost in some cases, it can be seen in the furnace pics.
I tracked this to our BRDFs being double sided. Some details can also be tweaked like the amount of samples used to estimate the integrals.

I think it's still a big improvement over what we had before.

src/appleseed/renderer/modeling/bsdf/glossybrdf.cpp
est77: It's terminology from the paper. fms is the multiscattering part of the BSDF (the energy we want to recover). Eavg is the average amount of energy reflected for a specific roughness value.

src/appleseed/renderer/meta/tests/test_energycompensation.cpp
est77: I had it like that, but because this is only useful in the unit test, I thought it would be cleaner not to add it to the header and use a prototype instead.

dictoon: It's defined in `microfacethelper.cpp` but not exposed in `microfacethelper.h` so it looks odd (and could be removed by accident, only to find out at link time it was actually used). It would also be the first time we do this. Finally, there are many cases already where we expose stuff just for testing purposes, which I think is just fine. In conclusion I'd suggest moving the declaration to `microfacethelper.h` if you agree :)

src/appleseed/renderer/modeling/bsdf/glossybrdf.cpp
dictoon: Cool, thanks.

src/appleseed/renderer/modeling/bsdf/metalbrdf.cpp
dictoon: Cool.

PR #1771, opened 2017-12-13
Add rayswitch, layerShader, vectorField nodes

luisbarrancos:
And icon metadata to the core OSL nodes. It has no effect in Maya, and
is meant only for Gaffer. The main idea here being sharing the core
nodes with all DCC apps.

### DIFF
--- a/src/appleseed.shaders/src/appleseed/as_ray_switch.osl
+++ b/src/appleseed.shaders/src/appleseed/as_ray_switch.osl
@@ -148,7 +148,7 @@ shader as_ray_switch
 #ifdef DEBUG
         string shadername = "";
         getattribute("shader:shadername", shadername);
-        warning("[DEBUG!]: Unknown Ray Type in %s, %s:%d\n",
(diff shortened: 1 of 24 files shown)


luisbarrancos:
Related to https://github.com/appleseedhq/appleseed-maya/pull/107

src/appleseed.shaders/src/appleseed/as_ray_switch.osl
```diff
@@ -0,0 +1,155 @@
...
+    else if (raytype("subsurface"))
+    {
+        out_color = in_color_subsurface;
+    }
+    else
+    {
+#ifdef DEBUG
+        string shadername = "";
+        getattribute("shader:shadername", shadername);
+        warning("[DEBUG!]: Unknown Ray Type in %s, %s:%d\n",
```
dictoon: Is `[DEBUG!]` the "normalized" tag we use everywhere?

luisbarrancos: @dictoon actually no, it should be ```warning("[WARNING]"...)```

PR #1774, opened 2017-12-16
Add metal and plastic osl materials for 3ds max

usakhelo:
Hi,
This PR adds plastic and metal osl shaders for 3ds max. I just copied them from gaffer/surface folder and renamed output closure.


dictoon:
Looks good to me! Are these straight copies of existing shaders?

In any case I'd appreciate a review by @luisbarrancos.

luisbarrancos:
LGTM as well. It's consistent with the shader library and remaining DCC plugins.

dictoon:
Thanks Luis!

usakhelo:
Sorry for late reply - yes it's straight copy. Only output variable name is different.
Thanks for review.

PR #1784, opened 2018-01-10
Add support for shutter opening curves to camera.

pjessesco:
- add Camera::map_sample() method
 - add member variables to renderer::Camera
 - add lerp2() in scalar.h

### DIFF
--- a/src/appleseed/foundation/math/scalar.h
+++ b/src/appleseed/foundation/math/scalar.h
@@ -224,10 +224,10 @@ T mix(const T a, const T b, const U x);
 template <typename T, typename U>
 T lerp(const T a, const T b, const U x);
 
-// lerp2() returns the relative position of c between a and b. 
+// inverse_lerp() returns the relative position of c between a and b. 
 // it is transformation of lerp(). i.e (c - a) / (b - a)
 template <typename T, typename U>
-T lerp2(const T a, const T b, const U c);
+T inverse_lerp(const T a, const T b, const U c);
 
 // fit() remaps a variable x from the range [min_x, max_x] to the
 // range [min_y, max_y]. When x is outside the [min_x, max_x] range,
@@ -825,11 +825,11 @@ inline T lerp(const T a, const T b, const U x)
 }
 
 template <typename T, typename U>
-inline T lerp2(const T a, const T b, const U c)
+inline T inverse_lerp(const T a, const T b, const U x)
 {
     assert(b != a);
     
-    return (c - a) / (b - a);
+    return (x - a) / (b - a);
 }
 
 template <typename T>
--- a/src/appleseed/foundation/math/scalar.h
+++ b/src/appleseed/foundation/math/scalar.h
@@ -224,6 +224,11 @@ T mix(const T a, const T b, const U x);
 template <typename T, typename U>
 T lerp(const T a, const T b, const U x);
 
+// lerp2() returns the relative position of c between a and b. 
+// it is transformation of lerp(). i.e (c - a) / (b - a)
+template <typename T, typename U>
+T lerp2(const T a, const T b, const U c);
+
 // fit() remaps a variable x from the range [min_x, max_x] to the
 // range [min_y, max_y]. When x is outside the [min_x, max_x] range,
 // a linear extrapolation outside the [min_y, max_y] range is used.
@@ -817,6 +822,14 @@ template <typename T, typename U>
 inline T lerp(const T a, const T b, const U x)
 {
     return (U(1.0) - x) * a + x * b;
+}
+
+template <typename T, typename U>
+inline T lerp2(const T a, const T b, const U c)
+{
+    assert(b != a);
+    
+    return (c - a) / (b - a);
 }
 
 template <typename T>
--- a/src/appleseed/foundation/meta/tests/test_scalar.cpp
+++ b/src/appleseed/foundation/meta/tests/test_scalar.cpp
@@ -463,38 +463,38 @@ TEST_SUITE(Foundation_Math_Scalar)
         EXPECT_FEQ(3.0, result);
     }
 
-    TEST_CASE(Lerp2_GivenBlendParameterLessThanFirstValue_ReturnsCorrectlyExtrapoledValue)
+    TEST_CASE(InverseLerp_GivenBlendParameterLessThanFirstValue_ReturnsCorrectlyExtrapoledValue)
     {
-        const double result = lerp2(1.0, 5.0, 0.0);
+        const double result = inverse_lerp(1.0, 5.0, 0.0);
         EXPECT_FEQ(-0.25, result);
     }
 
-    TEST_CASE(Lerp2_GivenBlendParameterEqualToFirstValue_Returns0)
+    TEST_CASE(InverseLerp_GivenBlendParameterEqualToFirstValue_Returns0)
     {
-        const double result = lerp2(1.0, 5.0, 1.0);
+        const double result = inverse_lerp(1.0, 5.0, 1.0);
         EXPECT_FEQ(0.0, result);
     }
 
-    TEST_CASE(Lerp2_GivenBlendParameterEqualToSecondValue_Returns1)
+    TEST_CASE(InverseLerp_GivenBlendParameterEqualToSecondValue_Returns1)
     {
-        const double result = lerp2(1.0, 5.0, 5.0);
+        const double result = inverse_lerp(1.0, 5.0, 5.0);
         EXPECT_FEQ(1.0, result);
     }
 
-    TEST_CASE(Lerp2_GivenBlendParameterGreaterThanSecondValue_ReturnsCorrectlyExtrapoledValue)
+    TEST_CASE(InverseLerp_GivenBlendParameterGreaterThanSecondValue_ReturnsCorrectlyExtrapoledValue)
     {
-        const double result = lerp2(1.0, 5.0, 6.0);
+        const double result = inverse_lerp(1.0, 5.0, 6.0);
         EXPECT_FEQ(1.25, result);
     }
 
-    TEST_CASE(Lerp2_GivenBlendParameterBetweenFirstValueAndSecondValue_ReturnsCorrectlyBlendedValue)
+    TEST_CASE(InverseLerp_GivenBlendParameterBetweenFirstValueAndSecondValue_ReturnsCorrectlyBlendedValue)
     {
-        const double result = lerp2(1.0, 5.0, 3.0);
+        const double result = inverse_lerp(1.0, 5.0, 3.0);
         EXPECT_FEQ(0.5, result);
     }
 
-    TEST_CASE(Lerp2_GivenBlendParameterOfDifferentTypeBetweenFirstValueAndSecondValue_ReturnsCorrectlyBlendedValue){
-        const double result = lerp2(1.0, 5.0, 3.0f);
+    TEST_CASE(InverseLerp_GivenBlendParameterOfDifferentTypeBetweenFirstValueAndSecondValue_ReturnsCorrectlyBlendedValue){
+        const double result = inverse_lerp(1.0, 5.0, 3.0f);
         EXPECT_FEQ(0.5, result);
     }
 
--- a/src/appleseed/foundation/meta/tests/test_scalar.cpp
+++ b/src/appleseed/foundation/meta/tests/test_scalar.cpp
@@ -463,6 +463,41 @@ TEST_SUITE(Foundation_Math_Scalar)
         EXPECT_FEQ(3.0, result);
     }
 
+    TEST_CASE(Lerp2_GivenBlendParameterLessThanFirstValue_ReturnsCorrectlyExtrapoledValue)
+    {
+        const double result = lerp2(1.0, 5.0, 0.0);
+        EXPECT_FEQ(-0.25, result);
+    }
+
+    TEST_CASE(Lerp2_GivenBlendParameterEqualToFirstValue_Returns0)
+    {
+        const double result = lerp2(1.0, 5.0, 1.0);
+        EXPECT_FEQ(0.0, result);
+    }
+
+    TEST_CASE(Lerp2_GivenBlendParameterEqualToSecondValue_Returns1)
+    {
+        const double result = lerp2(1.0, 5.0, 5.0);
+        EXPECT_FEQ(1.0, result);
+    }
+
+    TEST_CASE(Lerp2_GivenBlendParameterGreaterThanSecondValue_ReturnsCorrectlyExtrapoledValue)
+    {
+        const double result = lerp2(1.0, 5.0, 6.0);
+        EXPECT_FEQ(1.25, result);
+    }
+
+    TEST_CASE(Lerp2_GivenBlendParameterBetweenFirstValueAndSecondValue_ReturnsCorrectlyBlendedValue)
+    {
+        const double result = lerp2(1.0, 5.0, 3.0);
+        EXPECT_FEQ(0.5, result);
+    }
+
+    TEST_CASE(Lerp2_GivenBlendParameterOfDifferentTypeBetweenFirstValueAndSecondValue_ReturnsCorrectlyBlendedValue){
+        const double result = lerp2(1.0, 5.0, 3.0f);
+        EXPECT_FEQ(0.5, result);
+    }
+
     TEST_CASE(Fit_GivenIntegerInputRangeAndFloatingPointOutputRange)
     {
         const double result = fit<size_t, double>(4, 0, 19, 0.0, 1.0);
--- a/src/appleseed/renderer/modeling/camera/camera.cpp
+++ b/src/appleseed/renderer/modeling/camera/camera.cpp
@@ -80,8 +80,18 @@ bool Camera::on_render_begin(
     IAbortSwitch*           abort_switch)
 {
     m_shutter_open_time = m_params.get_optional<float>("shutter_open_time", 0.0f);
+    m_shutter_open_end_time = m_params.get_optional<float>("shutter_open_end_time", 0.0f);
+    m_shutter_close_start_time = m_params.get_optional<float>("shutter_close_start_time", 1.0f);
     m_shutter_close_time = m_params.get_optional<float>("shutter_close_time", 1.0f);
     m_shutter_open_time_interval = m_shutter_close_time - m_shutter_open_time;
+    m_normalized_open_end_time = lerp2(m_shutter_open_time, m_shutter_close_time, m_shutter_open_end_time);
+    m_normalized_open_end_time_half = m_normalized_open_end_time / 2;
+    m_normalized_close_start_time = lerp2(m_shutter_open_time, m_shutter_close_time, m_shutter_close_start_time);
+    m_height = 2 / (1 + m_normalized_close_start_time - m_normalized_open_end_time);
+    m_open_linear_curve_slope = m_height / m_normalized_open_end_time;
+    m_close_linear_curve_slope = -m_height / (1 - m_normalized_close_start_time);
+    m_inverse_CDF_open_point = m_height * m_normalized_open_end_time_half;
+    m_inverse_CDF_close_point = m_height * (m_normalized_close_start_time - m_normalized_open_end_time_half);
 
     return true;
 }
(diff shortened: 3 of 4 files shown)


pjessesco:
I assume opening curves are straight line. 
here is my calculation, looks messy. :(

pjessesco:
Here are detailed steps :

pjessesco:
and some previews : 
[0,0,1,1]

[0,1,1,1]

[0,0,0,1]

[0,0.5,0.5,1]

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -80,8 +80,13 @@ bool Camera::on_render_begin(
     IAbortSwitch*           abort_switch)
 {
     m_shutter_open_time = m_params.get_optional<float>("shutter_open_time", 0.0f);
```
pjessesco: should I rename it? (like m_shutter_open_start_time.. maybe too long)

dictoon:
1. The first expression for `h` is incorrect, but the second one is correct:

    h = 2 / (ncs - noe + 1)

2. Expressions of f(x) are all correct.

3. Nice, you didn't forget the integration constants, a mistake I was often doing when I was a student :) However, there's a mistake in your integration: the integral of (x - 1) is not 1/2 (x - 1)^2, but 1/2 x^2 - x + some constant. Consequently, C is not equal to 1, but to 1/2 * h / (ncs - 1). Unless I'm mistaking, that is :)

3. iii. I'm not sure why you need to compute a limit here.

4. i. You didn't forget the +/-, another typical potential mistake, well done!

4. ii. Probably incorrect since the expression of f(x) is incorrect.

Man I feel like a math teacher :)

dictoon:
Another thing:

We will soon want to replace the linear opening/closing by smoother functions (which function to use will be a user setting). So it would make sense to split the problem into two cases:

- Case 1: The time (x in this case) is in the constant section of f(x) -> no remapping is necessary.

- Case 2: The time is in the shutter opening or closing sections of the curve: both cases are instances of the same problem so they can be handled by the same code (first the input is remapped to the shutter opening/closing time interval, then it is remapped using F^-1(x), like you did).

Basically, it makes sense to do things in a more modular manner in prepation to supporting various other opening/closing curves. Do you see what I mean?

pjessesco:
@dictoon  Integral of (x - 1) is 1/2 (x - 1)^2 + C, and it is same with 1/2 x^2 - x + D, and the reason why I compute limit is to use F(x) is continuous...

Anyway, thanks for your detailed review! I agree with you, I'll try to divide case like you said. I'll take a straight line for now, but I think it will be fun to apply another curve.

dictoon:
> Integral of (x - 1) is 1/2 (x - 1)^2 + C, and it is same with 1/2 x^2 - x + D

Yes, you're right.

Note that you don't have to close your PR, you could push more commits to it as you modularize the code and preparing it for merging.

pjessesco:
Hey! I saw your feedback, and I tried to fix my code. 

I thought map_sample() work like a F^-1(x). I've written this code simply with my understand with your feedback, and comparing it with map_sample() in my commit, I could check they work differently, here is result.



I could see dark part between bright part. And here is my new `map_sample()`


    float Camera::new_map_sample(float sample) const
    {
        // shutter is opening
        if(0 <= sample && sample < m_normalized_open_end_time){
            // Opening curve : g(x) = ax + b (a > 0, b < 0)
            // Normalize opening curve : f(x) = 2x
            // Integrate normalized opening curve : F(x) = x^2 (C = 0)
            // Inverse F(x) : F^-1(x) = x^(1/2)
            float new_sample = sqrt(lerp2(0.0f, m_normalized_open_end_time, sample));
            return lerp(0.0f, m_normalized_open_end_time, new_sample);
        }
        // shutter is closing
        else if(m_normalized_close_start_time < sample && sample <= 1){
            // Opening curve : g(x) = ax + b (a < 0, b > 0)
            // Normalize opening curve : f(x) = -2x + 2
            // Integrate normalized opening curve : F(x) = -x^2 + 2x (C = 0)
            // Inverse F(x) : F^-1(x) = 1 - root(1 - x)
            float new_sample = 1 - sqrt(1 - lerp2(m_normalized_close_start_time, 1.0f, sample));
            return lerp(m_normalized_close_start_time, 1.0f, new_sample);
        }
        // shutter is opened
        else{
            return sample;
        }
    }

There is few problem in `new_map_sample()` :
1. `noe` and `ncs` should not be point that seperate cases, because domain and range are changed.

2. Because there is no remapping when shutter is fully opened, F^-1(x) graph is not smooth in x = `ncs` and x = `noe`, and I guess this is related with dark part in picture.

If I fix the problems, it will be same with my previous code. I have no idea to implement better approach...

dictoon:
To be clear, I think your previous logic was fine, I was just suggesting modularizing and factorizing the code to, for instance, share the code for the opening and closing curves (same logic, different parameters).

pjessesco:
Sorry, but I still confused. How can I share code for opening and closing, even they are different equation? 

    if(shutter fully opened)
    {
        // some code
    }
    else
    {
        if (shutter opening){
            // some code
        }
        else {
            // some code
        }
    }

I could think about *modularizing* something like this, but I don't think opening and closing cases sharing the same code in this case. Maybe I have wrong understand about modularization..

dictoon:
Hi, sorry for the delay in answering.

My point is that for the shutter opening and closing cases, although they are different equations, they follow the same structure (a linear curve of the form `a*x+b`) and only the coefficients (`a` and `b`) differ.

So remapping a random value `s` according to a linear curve of the form `a*x+b` could be a function reused twice, once for the shutter opening case and once for the shutter closing case.

Then we can write a function that works with a smoother curve and mix & match linear and smooth curves.

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -380,6 +423,38 @@ DictionaryArray CameraFactory::get_input_metadata()
...
+            .insert("min",
+                Dictionary()
+                    .insert("value", "0.0")
+                    .insert("type", "soft"))
+            .insert("max",
+                Dictionary()
+                    .insert("value", "1.0")
+                    .insert("type", "soft"))
+            .insert("use", "optional")
+            .insert("default", "0.2"));
```
dictoon: The default values must be such that the existing behavior is preserved (infinitely fast opening/closing of the shutter).

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -295,6 +300,44 @@ double Camera::extract_near_z() const
     return near_z;
 }
 
+// Assumed shutter opening/closing curve is straight line.
+float Camera::integrate_inverse_curve(float a, float b, float x) const
```
dictoon: This method isn't really specific to the `Camera` class, it makes sense to transform it into a local free function:

    namespace
    {
        float integrate_inverse_curve(float a, float b, float x) const
        {
            ...
        }
    }

Other remarks:
- We need to find a better name, in particular the name must reflect that it's a linear curve. Maybe `map_sample_to_linear_curve()`?
- make `a`, `b` and `x` const

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -295,6 +300,44 @@ double Camera::extract_near_z() const
...
+{
+    float constant;
+    int root_sign;
+    // opening
+    if(b == 0){
```
dictoon: Opening and closing braces always go on their own lines.

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -295,6 +300,44 @@ double Camera::extract_near_z() const
...
+    // closing
+    else{
+        constant = 1 - (a/2) - b;
+        root_sign = -1;
+    }
+    return root_sign * sqrt(2 * (x + (b * b) / (2 * a) - constant) / a) - b / a;
+}
+
+// Assumed shutter opening/closing curve is straight line.
+float Camera::map_sample(float sample) const
```
dictoon: `map_sample()` is a too vague of a name. Maybe `map_to_shutter_curve()`?

src/appleseed/renderer/modeling/camera/camera.cpp
dictoon: Insert a space between `if` and the opening parenthesis.

You can find a detailed style guide here:
https://github.com/appleseedhq/appleseed/wiki/Coding-Philosophy-and-Guidelines#coding-style

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -295,6 +300,44 @@ double Camera::extract_near_z() const
...
+{
+    // shutter is opening
+    if(0 <= sample && sample < m_height * m_normalized_open_end_time / 2)
```
dictoon: We need to simplify and speed up this code, in particular avoid divisions which are quite expensive. A lot of things (such as `m_height * m_normalized_open_end_time / 2`) can be precomputed.

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -399,6 +384,26 @@ double Camera::get_greater_than_zero(
 }
 
 
+float map_sample_to_linear_curve(const float a, const float b, const float x)
```
dictoon: Place in anonymous namespace.

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -399,6 +384,26 @@ double Camera::get_greater_than_zero(
...
+        constant = 0;
+        root_sign = 1;
+    }
+    // closing
+    else
+    {
+        constant = 1 - (a/2) - b;
+        root_sign = -1;
+    }
+    return root_sign * sqrt(2 * (x + (b * b) / (2 * a) - constant) / a) - b / a;
```
dictoon: Looks like `1 / a` can be computed just once.

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -300,41 +303,22 @@ double Camera::extract_near_z() const
...
     // shutter is opening
-    if(0 <= sample && sample < m_height * m_normalized_open_end_time / 2)
+    if (0 <= sample && sample < m_height * m_normalized_open_end_time_half)
     {
-        return integrate_inverse_curve(m_height / m_normalized_open_end_time, 0, sample);
+        return map_sample_to_linear_curve(m_open_linear_curve_slope, 0, sample);
     }
     // shutter is closing
-    else if(1 >= sample && sample > m_height * (m_normalized_close_start_time - (m_normalized_open_end_time / 2)))
+    else if (1 >= sample && sample > m_height * (m_normalized_close_start_time - m_normalized_open_end_time_half))
```
dictoon: Looks like the entire upper bound here can be precomputed.

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -300,41 +303,22 @@ double Camera::extract_near_z() const
...
-    return root_sign * sqrt(2 * (x + (b * b) / (2 * a) - constant) / a) - b / a;
-}
-
-// Assumed shutter opening/closing curve is straight line.
-float Camera::map_sample(float sample) const
+float Camera::map_to_shutter_curve(float sample) const
 {
     // shutter is opening
-    if(0 <= sample && sample < m_height * m_normalized_open_end_time / 2)
+    if (0 <= sample && sample < m_height * m_normalized_open_end_time_half)
```
dictoon: Looks like the entire upper bound here can be precomputed.

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -300,41 +303,22 @@ double Camera::extract_near_z() const
...
-    else{
-        constant = 1 - (a/2) - b;
-        root_sign = -1;
-    }
-    return root_sign * sqrt(2 * (x + (b * b) / (2 * a) - constant) / a) - b / a;
-}
-
-// Assumed shutter opening/closing curve is straight line.
-float Camera::map_sample(float sample) const
+float Camera::map_to_shutter_curve(float sample) const
```
dictoon: Make `sample` const.

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -80,13 +80,16 @@ bool Camera::on_render_begin(
     IAbortSwitch*           abort_switch)
 {
     m_shutter_open_time = m_params.get_optional<float>("shutter_open_time", 0.0f);
-    m_shutter_open_end_time = m_params.get_optional<float>("shutter_open_end_time", 0.2f);
-    m_shutter_close_start_time = m_params.get_optional<float>("shutter_close_start_time", 0.8f);
+    m_shutter_open_end_time = m_params.get_optional<float>("shutter_open_end_time", 0.0f);
```
dictoon: Do you confirm that instantaneous shutter opening and closing are properly handled by `map_to_shutter_curve()` (ideally without much overhead)?

pjessesco:
Oh I mistake commit, and I'm trying to fix my commit. Sorry.

dictoon:
No worries at all :)

pjessesco:
Should I add test for `lerp2()`?

dictoon:
> Should I add test for lerp2()?

Yes, that's a great idea.

In fact, we should find a better name than `lerp2()`, one that better reflects what the function does. Any idea?

src/appleseed/renderer/modeling/camera/camera.cpp
pjessesco: Yes. When shutter open instantly, sample doesn't map to 1st case of `map_to_shutter_curve()`, and it is similar with 2nd case and closing instantly case. I could check it works well.

src/appleseed/renderer/modeling/camera/camera.cpp
pjessesco: Oh, I worried about too many member variables(although I already made quite many member variables)..

pjessesco:
Naming is the hardest one :( 
I'll keep its name for now, and try to add some test.

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -295,6 +305,44 @@ double Camera::extract_near_z() const
...
+        if (a > 0)
+        {
+            constant = 0;
+        }
+        // closing
+        else
+        {
+            constant = 1 - (a / 2) - b;
+        }
+        return (sqrt(2 * a * (x - constant) + b * b) - b) / a;
```
pjessesco: I couldn't optimize better then this. :(

src/appleseed/foundation/math/scalar.h
```diff
@@ -224,6 +224,11 @@ T mix(const T a, const T b, const U x);
 template <typename T, typename U>
 T lerp(const T a, const T b, const U x);
 
+// lerp2() returns the relative position of c between a and b. 
+// it is transformation of lerp(). i.e (c - a) / (b - a)
+template <typename T, typename U>
+T lerp2(const T a, const T b, const U c);
```
dictoon: Rename `c` to `x` for symmetry with `lerp()` since it looks like they're both interpolation parameters?

I wonder if `inverse_lerp()` would make more sense than `lerp2()`?

src/appleseed/foundation/meta/tests/test_scalar.cpp
```diff
@@ -463,6 +463,41 @@ TEST_SUITE(Foundation_Math_Scalar)
         EXPECT_FEQ(3.0, result);
     }
 
+    TEST_CASE(Lerp2_GivenBlendParameterLessThanFirstValue_ReturnsCorrectlyExtrapoledValue)
```
dictoon: Very nice tests!

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -80,8 +80,18 @@ bool Camera::on_render_begin(
...
 {
     m_shutter_open_time = m_params.get_optional<float>("shutter_open_time", 0.0f);
+    m_shutter_open_end_time = m_params.get_optional<float>("shutter_open_end_time", 0.0f);
+    m_shutter_close_start_time = m_params.get_optional<float>("shutter_close_start_time", 1.0f);
     m_shutter_close_time = m_params.get_optional<float>("shutter_close_time", 1.0f);
     m_shutter_open_time_interval = m_shutter_close_time - m_shutter_open_time;
+    m_normalized_open_end_time = lerp2(m_shutter_open_time, m_shutter_close_time, m_shutter_open_end_time);
+    m_normalized_open_end_time_half = m_normalized_open_end_time / 2;
+    m_normalized_close_start_time = lerp2(m_shutter_open_time, m_shutter_close_time, m_shutter_close_start_time);
+    m_height = 2 / (1 + m_normalized_close_start_time - m_normalized_open_end_time);
```
dictoon: We need to find a better name than `m_height`...

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -80,8 +80,18 @@ bool Camera::on_render_begin(
...
+    m_open_linear_curve_slope = m_height / m_normalized_open_end_time;
+    m_close_linear_curve_slope = -m_height / (1 - m_normalized_close_start_time);
+    m_inverse_CDF_open_point = m_height * m_normalized_open_end_time_half;
```
dictoon: Use all lower case for member variables: `m_inverse_cdf_open_point`.

src/appleseed/renderer/modeling/camera/camera.cpp
dictoon: Well this looks pretty good already!

src/appleseed/renderer/modeling/camera/camera.h
```diff
@@ -158,8 +158,18 @@ class APPLESEED_DLLSYMBOL Camera
   protected:
     TransformSequence   m_transform_sequence;
     float               m_shutter_open_time;
+    float               m_shutter_open_end_time;
+    float               m_shutter_close_start_time;
     float               m_shutter_close_time;
     float               m_shutter_open_time_interval;
+    float               m_normalized_open_end_time;
```
dictoon: Good names!

src/appleseed/renderer/modeling/camera/camera.h
```diff
@@ -184,6 +194,9 @@ class APPLESEED_DLLSYMBOL Camera
         SamplingContext&                sampling_context,
         ShadingRay&                     ray) const;
 
+    // Map a sample using inverse of CDF calculated from camera shutter graph. Works in initialize_ray().
```
dictoon: *Works in* -> *Used in*

src/appleseed/renderer/modeling/camera/camera.h
```diff
@@ -194,6 +207,15 @@ class APPLESEED_DLLSYMBOL Camera
 };
 
 
+//
+// Shutter curves. Works in Camera::map_to_shutter_curve().
+//
+
+// Integrate and inverse opening/closing curve y = a * x + b to transform samples.
+namespace
```
dictoon: It doesn't make much sense to forward-declare a free function from an anonymous namespace since it can't be used anyway (being in an anonymous namespace).

You can simply remove the whole declaration here.

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -295,6 +305,44 @@ double Camera::extract_near_z() const
...
+        {
+            constant = 1 - (a / 2) - b;
+        }
+        return (sqrt(2 * a * (x - constant) + b * b) - b) / a;
+    }
+}
+
+float Camera::map_to_shutter_curve(const float sample) const
+{
+    // shutter is opening
```
dictoon: Full line comments should start with a capital and end with a period.

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -295,6 +305,44 @@ double Camera::extract_near_z() const
     return near_z;
 }
 
+namespace
+{   
+    float map_sample_to_linear_curve(const float a, const float b, const float x)
+    {
+        float constant;
+        // opening
+        if (a > 0)
```
dictoon: Type all floating point literals appropriately, i.e. 0.0f for float and 0.0 for double.

src/appleseed/foundation/math/scalar.h
pjessesco: I think `inverse_lerp()` is good too!

src/appleseed/renderer/modeling/camera/camera.cpp
pjessesco: How about... `shutter_pdf_max_height`?

dictoon: Sounds good!

pjessesco:
Thanks!
And I'd like to try add other curves, maybe bezier curve that mentioned in issue page? I guess it has similar tasks with linear curve..

dictoon:
Yes, definitely! Not sure which curve(s) exactly to support, maybe talk to people on Slack and check what other renderers offer...

fedormatantsev:
@pjessesco have linear shutter interpolation, I'm going to implement curve shutter. Please, assign this one to me.

PR #1785, opened 2018-01-10
BCD improvements

est77:
- Removed OpenMP dependency.
- Removed unused BCD code.
- BCD VS2012 compatibility.
- Added progress reporting and aborting.
- Correct pixels with invalid values after denoising.
- Small code style changes and formatting.
- Fix moc parse errors with boost predef.


src/appleseed/renderer/kernel/denoising/denoiser.h
```diff
@@ -34,6 +34,7 @@
 
 // Forward declarations.
 namespace foundation    { class Image; }
+namespace foundation    { class IAbortSwitch; }
```
dictoon: Order alphabetically.

src/bcd/bcd/IDenoiser.h
```diff
@@ -13,15 +13,29 @@
...
+    IProgressReporter();
+    virtual ~IProgressReporter();
+
+    virtual void progress(const float i_progress) const = 0;
+
+    virtual bool isAborted() const = 0;
+
+  private:
+    // Non-copyable.
+    IProgressReporter(const IProgressReporter&);
```
dictoon: Make public and mark as `= delete` now that we're using C++11?

src/bcd/bcd/IDenoiser.h
```diff
@@ -13,15 +13,29 @@
...
-#include <functional>
-#include <memory>
 
 namespace bcd
 {
 
 template<class T>
 class DeepImage;
 
+class IProgressReporter
```
dictoon: Should probably be moved to appleseed.foundation no?

In which case `isAborted()` should be renamed to `is_aborted()`.

src/bcd/bcd/IDenoiser.h
```diff
@@ -13,15 +13,29 @@
...
+{
+  public:
+    IProgressReporter();
+    virtual ~IProgressReporter();
+
+    virtual void progress(const float i_progress) const = 0;
```
dictoon: What does the `i_` prefix mean? Input? Is that a convention of the BCD code?

src/bcd/bcd/MultiscaleDenoiser.cpp
```diff
@@ -29,6 +29,62 @@ using namespace std;
...
+                m_progress_reporter->progress(
+                    i_progress / float(((1 << (2 * m_nbOfScales)) - 1) / 3));
+            }
+            else
+            {
+                int s = m_nbOfScales - 1 - m_scale;
+                // next higher definition scale is 4 times slower
+                // 1 + 4 + 4^2 + ... 4^s = (4^(s+1) - 1) / (4 - 1) = (2^(2*(s+1)) - 1) / 3
+                // = ((1 << (2*(s+1))) - 1) / 3
+                float factor = 1.f / float(((1 << (2 * m_nbOfScales)) - 1) / 3);
```
dictoon: May I suggest to move `float(((1 << (2 * m_nbOfScales)) - 1) / 3)` to a free function and call it four times here?

src/bcd/bcd/MultiscaleDenoiser.cpp
```diff
@@ -95,12 +151,16 @@ bool MultiscaleDenoiser::denoise()
...
+        MultiScaleProgressReporter progressReporter(
+            m_nbOfScales - 1,
+            m_nbOfScales,
+            m_progressReporter);
+        denoiser.setProgressReporter(&progressReporter);
 
-        denoiser.denoise();
+        const bool success = denoiser.denoise();
+
+        if (success == false)
```
dictoon: `!success`

src/bcd/bcd/MultiscaleDenoiser.cpp
```diff
@@ -109,19 +169,20 @@ bool MultiscaleDenoiser::denoise()
...
+
+        MultiScaleProgressReporter progressReporter(
+            scale,
+            m_nbOfScales,
+            m_progressReporter);
+        denoiser.setProgressReporter(&progressReporter);
+
+        const bool success = denoiser.denoise();
+
+        if (success == false)
```
dictoon: `!success`

src/bcd/bcd/MultiscaleDenoiser.cpp
```diff
@@ -29,6 +29,62 @@ using namespace std;
 
 namespace bcd
 {
+namespace
+{
+
+class MultiScaleProgressReporter
```
dictoon: Indent content of anonymous namespace.

Our rule is pretty simple:

We only make an exception for top-level namespaces that span entire files. Indenting those would amount to indenting a whole file for no clear benefit. We always indent everything else.

src/bcd/bcd/MultiscaleDenoiser.cpp
```diff
@@ -29,6 +29,62 @@ using namespace std;
...
+  : public IProgressReporter
+{
+  public:
+    MultiScaleProgressReporter(
+        int                 i_scale,
```
dictoon: Make const.

src/bcd/bcd/MultiscaleDenoiser.cpp
dictoon: Make const.

src/bcd/bcd/IDenoiser.h
est77: I didn't want to introduce a dependency between BCD and foundation.
This is BCD's progress reporter and there is an adaptor to use our abort switch and possibly report progress.

src/bcd/bcd/IDenoiser.h
est77: input param. I kept the original style of the code.

src/bcd/bcd/IDenoiser.h
est77: I thought = delete was not supported by the VC version needed for some Max plugins.
Can I use it?

dictoon: Sorry, you're right, it's not supported by VS 2012.

src/bcd/bcd/IDenoiser.h
dictoon: Didn't we already introduce dependencies to appleseed in BCD?

PR #1794, opened 2018-01-17
Add new nodes, refactor CMS code

luisbarrancos:
The nodes being subsurface, composite color, blend color, texture lookup, texture info, switch texture, blackbody.
The luminance node is now aware of user-set RGB primaries and whitepoint with respective code in headers being refactored for an upcoming colorspace node able to build RGB<>XYZ matrices for any RGBW chromaticity coordinates. Note that the blend node uses the Photoshop *lightness*, *hue*, *saturation*, *color* blend modes as well, and these are quantities from LCh_ab(CIELAB) color space, which requires a transformation from RGB to CIEXYZ. So the blend node has to be aware of the working/rendering space primaries and whitepoint.


src/appleseed.shaders/include/appleseed/color/as_color_blend_modes.h
```diff
@@ -0,0 +1,545 @@
...
+    {
+        rgb = mix(A, B, A_alpha);
+        alpha = mix(A_alpha, B_alpha, A_alpha);
+    }
+    else
+    {
+#ifdef DEBUG
+        string shadername = "";
+        getattribute("shader:shadername", shadername);
+        warning("[DEBUG]: Unknown blend mode %s in %s, %s:%d\n",
```
dictoon: `Unknown compositing mode` maybe?

src/appleseed.shaders/include/appleseed/math/as_math_helpers.h
```diff
@@ -4,7 +4,7 @@
 //
 // This software is released under the MIT license.
 //
-// Copyright (c) 2016-2017 Luis Barrancos, The appleseedhq Organization
+// Copyright (c) 2016-2018 Luis Barrancos, The appleseedhq Organization
```
dictoon: I'll soon do a pass over all files to update the copyright notices.

src/appleseed.shaders/include/appleseed/maya/as_maya_cms_syncolor_idt.h
```diff
@@ -60,18 +51,21 @@ color transform_color_space_to_Rec709(
     {
         transformed_color = sRGB_EOTF(input_color);
     }
-    else if (color_space == "scene-linear Rec 709/sRGB")
+    else if (color_space == "scene-linear Rec 709/sRGB" ||
```
dictoon: Where are these strings coming from, exactly? They look human-readable, i.e. they don't look like identifiers; how come we can expect exact matches?

src/appleseed.shaders/src/appleseed/as_switch_texture.osl
```diff
@@ -0,0 +1,201 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+#include "appleseed/transform/as_transform_helpers.h"
+
+shader as_SwitchTexture
```
dictoon: Why is this shader starting with an upper case letter (after the `as_` prefix)?

src/appleseed.shaders/src/appleseed/as_subsurface.osl
```diff
@@ -0,0 +1,370 @@
...
+                    point(Nn));
+            }
+        }
+
+        out_outColor += in_specular_weight * as_glossy(
+            "ggx",
+            Nn,
+            vector(0), // tangent
+            in_specular_roughness,
+            0.0, // specular spread for Student-T MDF
```
dictoon: Align comments vertically? Maybe more readable. Also, `Student's t-MDF` is probably more correct :)

src/appleseed.shaders/include/appleseed/color/as_color_blend_modes.h
luisbarrancos: @dictoon exactly, fixing this.

src/appleseed.shaders/include/appleseed/maya/as_maya_cms_syncolor_idt.h
luisbarrancos: We have the SynColor system, that provides the working/rendering space strings, they are passed via https://github.com/appleseedhq/appleseed/blob/master/src/appleseed.shaders/include/appleseed/maya/as_maya_helpers.h#L194 , the attribute for working space.
So when using SynColor we just check the options provided in maya 2017.
These are:
 * "ACES2065-1"
 * "ACEScg"
 * "scene-linear CIE XYZ"
 * "scene-linear Rec 2020"
 * "scene-linear Rec 709/sRGB"
 * "Sharp RGB"

I just noticed a typo in a couple of identifiers in a related function, so i'll be updating that.

src/appleseed.shaders/src/appleseed/as_switch_texture.osl
luisbarrancos: typo, it should be ordinary camel case to match Maya.

PR #1798, opened 2018-01-20
Explicit library linking

philliptee:
Add two cmake macros that explicitly link against openexr and pthread libraries.
This change enables compiling and linking on the ubuntu linux distributions


dictoon:
Thanks! Glad that you found the problem!

src/bcd/CMakeLists.txt
```diff
@@ -61,6 +61,15 @@ add_library (bcd STATIC ${sources})
 apply_static_lib_linker_flags (bcd)
 
 
+#--------------------------------------------------------------------------------------------------
+# Static libraries.
+#--------------------------------------------------------------------------------------------------
+
+link_against_platform (bcd)
+
```
dictoon: You can remove the blank line here.

philliptee: Are you sure about this? The spacing I used in this commit matches the style of this file and other files where I see these macros used.

dictoon: Your code is perfectly formatted, very much appreciated. I was just suggesting grouping the two linking commands but it's such a minute detail, forget about it :)

philliptee:
Commit formatted and rebased properly! :)

dictoon:
Perfect, thank you!

PR #1799, opened 2018-01-20
Denoiser improvements

est77:
- Ignore invalid samples in histograms and covariances.
- Added Spike prefiltering.
- Fixed crashes with BCD and crop windows.
- Set the number of threads used by BCD.
- Exposed more BCD options and tweaked parameter ranges and defaults.


src/bcd/bcd/Denoiser.cpp
```diff
@@ -271,15 +271,12 @@ void Denoiser::reorderPixelSetJumpNextChunk(
...
-void reorderPixelSetShuffleCPP11(vector<PixelPosition>& io_rPixelSet)
-{
-    unsigned seed = static_cast<unsigned>(std::chrono::system_clock::now().time_since_epoch().count());
-    shuffle(io_rPixelSet.begin(), io_rPixelSet.end(), std::default_random_engine(seed));
-}
-
 void Denoiser::reorderPixelSetShuffle(vector<PixelPosition>& io_rPixelSet)
 {
-    reorderPixelSetShuffleCPP11(io_rPixelSet);
+    unsigned seed =
```
dictoon: I would prefer if we used a fixed seed so that everything remains deterministic. Would that be possible?

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -743,28 +743,35 @@ void Frame::extract_parameters()
...
-    }
-
     // Fill denoiser options.
     DenoiserOptions options;
+
+    string prefilter_spikes = m_params.get_optional<string>("prefilter_spikes", "off");
+    options.m_prefilter_spikes = prefilter_spikes == "on";
+
+    options.m_prefilter_threshold_stdev_factor =
+        get_parameters().get_optional<float>(
```
dictoon: You can continue using `m_params` here and below.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -899,15 +909,49 @@ DictionaryArray FrameFactory::get_input_metadata()
             .insert("default", "off")
             .insert("on_change", "rebuild_form"));
 
+    metadata.push_back(
+        Dictionary()
+            .insert("name", "prefilter_spikes")
+            .insert("label", "Prefilter Spikes")
+            .insert("type", "enumeration")
```
dictoon: Wouldn't `"boolean"` make more sense?

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -899,15 +909,49 @@ DictionaryArray FrameFactory::get_input_metadata()
...
+            .insert("type", "text")
+            .insert("use", "optional")
+            .insert("min",
+                Dictionary()
+                    .insert("value", "0.1")
+                    .insert("type", "hard"))
+            .insert("max",
+                Dictionary()
+                    .insert("value", "4.0")
+                    .insert("type", "hard"))
```
dictoon: `"hard"` means that there's absolutely no way to go above that value. Is that what we want?

src/appleseed/renderer/kernel/rendering/generic/genericframerenderer.cpp
```diff
@@ -350,6 +350,7 @@ namespace
                 if (!m_abort_switch.is_aborted())
                 {
                     m_frame.denoise(
+                        m_tile_renderers.size(),
```
dictoon: Although that's technically correct, wouldn't `m_params.m_thread_count` make more sense?

src/appleseed/renderer/kernel/denoising/denoiser.h
```diff
@@ -43,15 +43,15 @@ class DenoiserOptions
 {
   public:
     float   m_histogram_patch_distance_threshold; //  Histogram patch distance threshold
-    int     m_patch_radius;                       //  Patch has (1 + 2 x m_patchRadius)^2 pixels
-    int     m_search_window_radius;               //  Search windows (for neighbors) spreads across (1 + 2 x m_patchRadius)^2 pixels
+    size_t  m_patch_radius;                       //  Patch has (1 + 2 x m_patchRadius)^2 pixels
```
dictoon: Our convention for end-of-line comments is to start them with a lower case character and no period at the end.

src/appleseed/renderer/modeling/aov/denoiseraov.cpp
```diff
@@ -341,11 +351,40 @@ auto_release_ptr<AOVAccumulator> DenoiserAOV::create_accumulator() const
...
+{
+    const int w = impl->m_histograms.getWidth();
+    const int h = impl->m_histograms.getHeight();
+    const int samples_channel_index = static_cast<int>(impl->m_num_bins * 3);
+
+    for (int j = 0; j < h; ++j)
+    {
+        for (int i = 0; i < w; ++i)
+        {
+            int num_samples = impl->m_histograms.get(j, i, samples_channel_index);
```
dictoon: Make const.

src/bcd/bcd/Denoiser.cpp
est77: I'm not sure about it, but we never enable the random pixel order in appleseed.

est77:
Fixed and rebased

src/bcd/bcd/Denoiser.cpp
dictoon: Ah you mean that this code is not used? Ok then.

PR #1802, opened 2018-01-20
Type conversion errors fixed.

Blackhart:
Incorrect type conversion produces an error message in visual studio with the "Treat Warning As Errors" flag.


dictoon:
Thanks for the fix! They're fixed in #1801 which I just merged.

PR #1805, opened 2018-01-22
CIE L*u*v and L*a*b* range fixes

luisbarrancos:
When remapping to [0,1], plus some minor doc fixes.

### DIFF
--- a/src/appleseed.shaders/include/appleseed/color/as_color_transforms.h
+++ b/src/appleseed.shaders/include/appleseed/color/as_color_transforms.h
@@ -660,7 +660,7 @@ color transform_CIEXYZ_to_CIELAB(
     vector reference_white_CIExyY)
 {
     color white_XYZ = transform_CIExyY_to_CIEXYZ(reference_white_CIExyY);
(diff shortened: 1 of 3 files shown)


src/appleseed.shaders/include/appleseed/color/as_color_transforms.h
```diff
@@ -1045,36 +1045,44 @@ color transform_CIELCh_uv_to_linear_RGB(
 }
 
 //
-// CIELAB, CIELUV, have values in range L [0,100], a,b [-100,100]
-// and u,v [-100,100].
+// CIELAB, CIELUV, have values in range L [0,100], a,b [-128,128]
+// and u,v ([-134,220],[-140,122])
```
dictoon: Kind of weird no? What's the rationale?

luisbarrancos: I wholeheartedly agree, but that seems to be the case. It is based on u, v coordinates, http://scc.ustc.edu.cn/zlsc/sugon/intel/ipp/ipp_manual/IPPI/ippi_ch6/ch6_color_models.htm
The entire rationale for this space eludes me so far.

PR #1808, opened 2018-01-22
Refactorings in preparation to render stamp implementation

dictoon:


src/appleseed/renderer/kernel/rendering/generic/genericframerenderer.cpp
```diff
@@ -349,13 +353,27 @@ namespace
...
-                        m_tile_callbacks.empty() ? nullptr : m_tile_callbacks[0],
-                        &m_abort_switch);
+                    m_is_rendering = false;
+                    return;
+                }
+
+                // Denoising pass.
+                if (m_frame.get_denoising_mode() == Frame::DenoisingMode::Denoise)
+                {
+                    if (m_pass_count > 1)
```
est77: Why do we need this test?

dictoon: It's just cosmetic: we don't want to print these messages if there is only one pass.

PR #1813, opened 2018-01-25
New appleseed documentation

luisbarrancos:
Base configuration, skeleton, for unified documentation with plugins, intersphinx


luisbarrancos:
@dictoon regarding code documentation, do we follow conventions for python code (docstrings), and do we use doxygen for C++? I was wondering if when adding the Python, C++ API sections there was a way to automate more or less these sections.

dictoon:
> Do we follow conventions for python code (docstrings)

Yes we do. Also, PEP8.

> do we use doxygen for C++

We do have doxygen docs enabled but we haven't started using doxygen's markup. I'm still slightly ambivalent about it, but it's probably a good idea.

dictoon:
> I was wondering if when adding the Python, C++ API sections there was a way to automate more or less these sections.

It probably makes sense to automate that with doxygen for the C++ API, and I-don't-know-what-tool for Python docstrings.

docs/source/features/features.rst
```diff
@@ -0,0 +1,233 @@
...
+ * Sheen BRDF
+ * Arbitrary mixtures of BRDFs
+ * Energy compensation in microfacet BxDFs.
+
+.. _label_features_subsurface_scattering:
+
+Subsurface Scattering
+---------------------
+
+Appleseed provides state of the art raytraced, animation-friendly subsurface scattering, fully exposed to OSL.
```
dictoon: *appleseed* (lower case)

*state-of-the-art*

docs/source/features/features.rst
```diff
@@ -0,0 +1,233 @@
...
+ * Unidirectional Path Tracing [#]_
+ * Stochastic Progressive Photon Mapping [#]_
+ * Light Tracing
+
+.. _label_features_reflection_models:
+
+Reflection Models
+-----------------
+
+ * Lambertian BRDF (purely diffuse)
```
dictoon: I've slightly edited this list on the web site since the time you copied it. Might be worth synchronizing the two.

dictoon:
Ah it's because that PR hasn't been merged yet probably :)

docs/source/features/features.rst
luisbarrancos: Even in the beginning of a sentence *appleseed* remains without capitalization of the 1st letter?

luisbarrancos:
@dictoon that is correct, RTD is monitoring only the *master* and what they deem the *latest* branch (whatever that might be) to do the documentation builds.

docs/source/features/features.rst
luisbarrancos: in terms of features, i think the website generalizes a bit more without going into details, and the docs here should be a bit more specific. For instance, we have no mention about other MDFs, we only refer to the GGX MDF, and yet we support GTR and Student's t-MDF, albeit energy conservation is active only for GGX, Beckmann, just to name an example.
What do you think? Have in mind the crux of this PR is providing the base initial configuration :)

dictoon: > Even in the beginning of a sentence appleseed remains without capitalization of the 1st letter?

So far that's what we've done indeed. The only exception is https://www.blender.org/download/external-renderers/ because Ton kindly insisted :)

luisbarrancos: Ok, done. I'll review the appleseed-maya docs as well to make sure we're consistent in this regard.

dictoon: > we have no mention about other MDFs, we only refer to the GGX MDF, and yet we support GTR and Student's t-MDF

That's an omission, it should be fixed on the website's feature list.

> energy conservation is active only for GGX, Beckmann, just to name an example.

We probably should avoid having nearly identical feature lists in two distinct places as this will inevitably lead to them becoming partially out-of-date and/or out-of-sync.

However specific aspects such as compatibility between features (the example you gave) totally have their place in the docs.

PR #1825, opened 2018-02-01
Fixes, refactorings and minor improvements

dictoon:


src/appleseed/renderer/modeling/scene/scene.h
```diff
@@ -147,7 +147,7 @@ class APPLESEED_DLLSYMBOL Scene
     };
 
     // Return render-time data of this entity.
-    // Render-time data are available between on_frame_begin() and on_frame_end() calls.
+    // Render-time data are available between on_render_begin() and on_renderend() calls.
```
est77: Missing _

dictoon: Good catch, fixed & rebased.

src/appleseed.studio/mainwindow/project/entityeditor.cpp
```diff
@@ -729,6 +729,7 @@ void EntityEditor::slot_open_file_picker(const QString& widget_name)
         const QString file_picker_type = metadata.get<QString>("file_picker_type");
         const QString filter =
             file_picker_type == "image" ? get_oiio_image_files_filter() :
```
est77: Maybe we could convert this code into a classic if else?
I don't have a very strong opinion about it but I find chained ternary operators 
harder to read and it's quite likely that we have to extend this list at some point.

dictoon: I find this pattern quite readable but I don't have any strong opinion about it either. You can find another case of multiple chained ternary operators above in the file, on lines 225-235. Let me know what you think.

PR #1826, opened 2018-02-02
Add initial Fresnel node

luisbarrancos:
I'm still a bit ambivalent about this. On one side it gives creative freedom and it is a tool that is useful for an artist. On the other side it introduces viewer dependency into the shader network, with all that it entails in terms of light transport (BDPT, etc).
The documentation adds a warning about this node, but the main question is to what extent should we allow (or restrict) creative freedom. Feedback welcome. Neither this nor https://github.com/appleseedhq/appleseed-maya/pull/141 are merged yet. If we agree it's in the best interest of everyone to disregard this node, we can just discard these PRs.

### DIFF
--- a/src/appleseed.shaders/CMakeLists.txt
+++ b/src/appleseed.shaders/CMakeLists.txt
@@ -431,6 +431,7 @@ set (src_appleseed_sources
      src/appleseed/as_disney_material.osl
      src/appleseed/as_double_shade.osl
      src/appleseed/as_falloff_angle.osl
+     src/appleseed/as_fresnel.osl
      src/appleseed/as_glass.osl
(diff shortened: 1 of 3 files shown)


luisbarrancos:
screenshot_2018-02-02_21-09-35

There is this warning in the documentation, but it ends up not resolving the question definitively on way or the other.

dictoon:
Creative freedom is more important than most other things, in my opinion. That's why I'm often against hard limits (e.g. for numeric sliders).

We just need to make it clear that this node will only work in special cases, and I feel that your orange warning box does this adequately.

luisbarrancos:
@dictoon thanks for the feedback. I'll be merging this PR then.

PR #1829, opened 2018-02-04
Misc work

est77:
- Bind artist friendly fresnel functions to python for convenience.
- Refactored Glass BSDF to allow code reuse.

### DIFF
--- a/src/appleseed/renderer/modeling/bsdf/glassbsdf.cpp
+++ b/src/appleseed/renderer/modeling/bsdf/glassbsdf.cpp
@@ -88,6 +88,8 @@ namespace
     class GlassBSDFImpl
       : public BSDF
     {
+        typedef GlassBSDFInputValues InputValues;
+
       public:
         GlassBSDFImpl(
             const char*                 name,
@@ -143,11 +145,11 @@ namespace
                     context);
 
             if (mdf == "ggx")
-                m_mdf.reset(new GGXMDF());
+                m_mdf_type = GGX;
             else if (mdf == "beckmann")
-                m_mdf.reset(new BeckmannMDF());
+                m_mdf_type = Beckmann;
             else if (mdf == "std")
-                m_mdf.reset(new StdMDF());
+                m_mdf_type = Std;
             else return false;
 
             const string volume_parameterization =
@@ -209,10 +211,10 @@ namespace
             const int                   modes,
             BSDFSample&                 sample) const override
         {
+            const InputValues* values = static_cast<const InputValues*>(data);
+
             if (!ScatteringMode::has_glossy(modes))
                 return;
-
-            const InputValues* values = static_cast<const InputValues*>(data);
 
             const Basis3f basis(
                 values->m_precomputed.m_backfacing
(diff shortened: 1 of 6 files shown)


src/appleseed/renderer/modeling/bsdf/glassbsdf.cpp
```diff
@@ -88,6 +88,8 @@ namespace
     class GlassBSDFImpl
       : public BSDF
     {
+        typedef GlassBSDFInputValues InputValues;
```
dictoon: Move to `private` section at the end of the class.

src/appleseed/renderer/modeling/bsdf/glassbsdf.cpp
```diff
@@ -228,70 +230,183 @@ namespace
...
+                    const GGXMDF mdf;
+                    is_refraction = do_sample(
+                        mdf,
+                        s,
+                        adjoint,
+                        basis,
+                        alpha_x,
+                        alpha_y,
+                        0.0f,
+                        values->m_precomputed.m_reflection_color,
```
dictoon: Would it make sense to simply pass `values->m_precomputed` and `sample` instead of the individual fields? Right now the method takes 16 samples, which seems a bit excessive (and potentially inefficient since it almost certainly won't all fit in registers).

src/appleseed/renderer/modeling/bsdf/glassbsdf.cpp
etovagliari: I tried, but gcc 4.8 does not compile if I move the line down.

src/appleseed/renderer/modeling/bsdf/glassbsdf.cpp
etovagliari: I want to use do_sample(), which is a template now, outside of the normal rendering loop, 
without InputValues or BSDF samples and with floats, not Spectrums.
I plan to use it to compute EC tables. It could also be useful for random walk SSS.

src/appleseed/renderer/modeling/bsdf/glassbsdf.cpp
est77: I tried, but gcc 4.8 does not compile if I move the line down.

src/appleseed/renderer/modeling/bsdf/glassbsdf.cpp
est77: I want to use do_sample(), which is a template now, outside of the normal rendering loop,
without InputValues or BSDF samples and with floats, not Spectrums.
I plan to use it to compute EC tables. It could also be useful for random walk SSS.

src/appleseed/renderer/modeling/bsdf/glassbsdf.cpp
dictoon: Interesting.

src/appleseed/renderer/modeling/bsdf/glassbsdf.cpp
dictoon: Alright.

Issue #1830, opened 2018-02-04
Assertion assert(m_has_render_data) fails when rendering Orhotgraphic camera


usakhelo:
Assertion `assert(m_has_render_data)`  fails during material preview in 3ds max.

`OrthographicCamera::on_render_begin()` calls `Scene::get_render_data()` and render_data is initialized later in `Scene::on_render_begin()`.

PR #1831, opened 2018-02-05
Fix unavailable render_data issue #1830

usakhelo:


usakhelo:
Hi,
I'm wondering if this fixes issue #1830. Seems too easy, but I'm not sure, I've just moved `on_render_begin` calls on camera after render_data initialized.

PR #1834, opened 2018-02-07
Add f3d/tex3d node, fix specular in SSS

luisbarrancos:
Since the optional token was incorrectly named in asSubsurface.
We also need a set of UDIM textures made on Mudbox, Zbrush, to test the atlas type in the texture() call.

### DIFF
--- a/src/appleseed.shaders/include/appleseed/texture/as_texture_helpers.h
+++ b/src/appleseed.shaders/include/appleseed/texture/as_texture_helpers.h
@@ -30,7 +30,7 @@
 
 #include "appleseed/math/as_math_helpers.h"
 
-string set_wrap_mode(int wrap_mode)
+string get_wrap_mode(int wrap_mode)
 {   
     if (wrap_mode == 0)
     {
@@ -58,8 +58,8 @@ void set_wrap_mode(
     int wrap_mode[2],
     output string mode[2])
 {
-    mode[0] = set_wrap_mode(wrap_mode[0]);
-    mode[1] = set_wrap_mode(wrap_mode[1]);
+    mode[0] = get_wrap_mode(wrap_mode[0]);
+    mode[1] = get_wrap_mode(wrap_mode[1]);
 }
 
 void set_wrap_mode(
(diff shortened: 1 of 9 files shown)


src/appleseed.shaders/include/appleseed/texture/as_texture_helpers.h
```diff
@@ -71,6 +71,26 @@ void set_wrap_mode(
     mode[1] = set_wrap_mode(wrap_tmode);
 }
 
+void set_wrap_mode(
+    int wrap_mode[3],
+    output string mode[3])
+{
+    mode[0] = set_wrap_mode(wrap_mode[0]);
```
dictoon: Something is wrong here. `set_wrap_mode()` doesn't return anything.

src/appleseed.shaders/src/appleseed/as_texture3d.osl
```diff
@@ -0,0 +1,327 @@
...
+        string as_maya_attribute_name = "filename",
+        string as_maya_attribute_short_name = "fil",
+        string label = "Filename",
+        string page = "3D Texture",
+        string widget = "filename",
+        int as_maya_attribute_connectable = 0,
+        int as_maya_attribute_keyable = 0,
+        int as_maya_attribute_hidden = 1,
+        int gafferNoduleLayoutVisible = 0,
+        string help = "The 3d texture file (field3d, f3d, etc)",
```
dictoon: *3D*

src/appleseed.shaders/src/appleseed/as_texture3d.osl
```diff
@@ -0,0 +1,327 @@
...
+        int gafferNoduleLayoutVisible = 0,
+        int divider = 1
+    ]],
+    color in_color = color(0)
+    [[
+        string as_maya_attribute_name = "color",
+        string as_maya_attribute_short_name = "c",
+        string label = "Default Color",
+        string page = "3D Texture",
+        string help = "Default Fill color to use if the texture lookup fails."
```
dictoon: *fill*

src/appleseed.shaders/src/appleseed/as_texture3d.osl
```diff
@@ -0,0 +1,327 @@
...
+    ]],
+    float in_channel_fill = 0.0
+    [[
+        string as_maya_attribute_name = "channelFill",
+        string as_maya_attribute_short_name = "chf",
+        float softmin = 0.0,
+        float softmax = 1.0,
+        string label = "Default Channel Fill",
+        string page = "3D Texture",
+        string help = "Default Fill value for any channels requested but not present",
```
dictoon: *fill*

src/appleseed.shaders/include/appleseed/texture/as_texture_helpers.h
luisbarrancos: There is a string return type set above. We could bypass this completely if we had the ability to associate a string value with another string (or any other value really) into the OSL token for the *mapper* options, but although there is apparently nothing forbidding it in the spec, it's not working in appleseed-maya so far.

dictoon: I see. I would suggest changing the name of the first `set_wrap_mode()` function (the one that returns a string): it doesn't set anything, instead it retrieves the string version of a wrap mode. Maybe `get_wrap_mode_name()`?

luisbarrancos: Good point, done.

PR #1835, opened 2018-02-07
initialized color variable in blit_bitmap

psy2d:
This will prevent the 'maybe-uninitialized' warnings while building appleseed.


dictoon:
Thank you!

dictoon:
I've slightly changed the commit title. You can find guidelines for commit titles here: https://github.com/appleseedhq/appleseed/wiki/Git-Commits

Narann:
No performance cost for this?

dictoon:
Nope, the code is very slow anyway (and is documented as such: simple but slow). It's only used to generate render stamps.

Issue #1666, opened 2017-10-22
Add support for nullptr to foundation::to_string()


dictoon:
The following code currently fails to compile in VS 2015:

    void* p = nullptr;
    ASSERT_NEQ(nullptr, p);

dictoon:
Done.

PR #1836, opened 2018-02-07
to_string method now supports nullptr parameter.

Blackhart:
From issue #1666 

to_string now returns the word `"<null>"` when `nullptr` or null pointer are passed as argument.


src/appleseed/foundation/utility/string.h
```diff
@@ -379,12 +382,21 @@ namespace impl
 }
 
 // General entry-point.
-template <typename T>
+template <typename T, typename std::enable_if_t<!IsPointer<T>::R>* = 0>
```
dictoon: What's the purpose of the default template param value here?

Also, if it turns out to be necessary/useful, could `nullptr` be used instead of `0` (here and below)?

src/appleseed/foundation/utility/string.h
```diff
@@ -404,6 +416,13 @@ inline std::string to_string(const uint8& value)
     return to_string(static_cast<unsigned int>(value));
 }
 
+// Handle nullptr object
+template <>
+inline std::string to_string(const std::nullptr_t& value)
```
dictoon: Oh, I had forgotten about `std::nullptr_t`. Do we actually need the `std::enable_if_t<>` logic then, can't we just provide an overload of `to_string()` for `std::nullptr_t`?

Blackhart: This specialization is only available for `to_string(nullptr)`. It doesn't works if I make:
```cpp
void* p = nullptr; 
to_string(p);
```

src/appleseed/foundation/utility/string.h
Blackhart: This specification is a mistake ...
Can't use that to distinguish pointer and non-pointer.

From the reference: http://en.cppreference.com/w/cpp/types/enable_if

> ### Notes
>
> A common mistake is to declare two function templates that differ only in their default template arguments. This is illegal because default template arguments are not part of function template's signature, and declaring two different function templates with the same signature is illegal.
>
```cpp
struct T {
    enum { int_t,float_t } m_type;
    template <typename Integer,
              typename = std::enable_if_t<std::is_integral<Integer>::value>
    >
    T(Integer) : m_type(int_t) {}
 
    template <typename Floating,
              typename = std::enable_if_t<std::is_floating_point<Floating>::value>
    >
    T(Floating) : m_type(float_t) {} // error: cannot overload
};
```

Blackhart: I deleted this specialization because you did it already in your implementation structure, which, I modified in order to return a `"<null>"` string when the `value == nullptr`.

Blackhart:
Fixed

src/appleseed/foundation/utility/string.h
```diff
@@ -366,14 +366,20 @@ namespace impl
...
-            std::stringstream sstr;
-            sstr << "0x"
-                 << std::hex
-                 << std::uppercase
-                 << std::setw(2 * sizeof(void*))
-                 << std::setfill('0')
-                 << reinterpret_cast<uintptr_t>(value);
-            return sstr.str();
+            const uintptr_t ptr = reinterpret_cast<uintptr_t>(value);
+            if (((void*)ptr) == nullptr)
```
dictoon: Or simply `if (ptr == 0)`.

src/appleseed/foundation/utility/string.h
```diff
@@ -366,14 +366,20 @@ namespace impl
...
+                return std::string("<null>");
```
dictoon: `return "<null>"` should be enough.

Blackhart:
Fixed

dictoon:
Squashed, rebased and merged.

Here are our guidelines for commit messages: 
https://github.com/appleseedhq/appleseed/wiki/Git-Commits

Thanks!

PR #1842, opened 2018-02-09
3ds Max metadata in appleseed shaders, fix subsurface, fix uv_transform, add closure2surface

usakhelo:
Hi,
This PR:
1. adds 3ds max metadata to all the shaders in order to build osl plugins by parsing shader files.
2. fixes some bugs in uv_transform shader and fixes a bug in as_subsurface closure.
3. updates existing 3ds max osl shaders and this requires modifications of the plugin. I'll post PR to support this change later.

Thanks,
Sergo.

### DIFF
--- a/src/appleseed.shaders/CMakeLists.txt
+++ b/src/appleseed.shaders/CMakeLists.txt
@@ -314,6 +314,7 @@ source_group ("src\\gaffer\\vector" FILES
 set (src_max_sources
      src/max/as_max_blend_material.osl
      src/max/as_max_bump_map.osl
+     src/max/as_max_closure2surface.osl
      src/max/as_max_color_texture.osl
      src/max/as_max_disney_material.osl
      src/max/as_max_float_texture.osl
@@ -324,7 +325,6 @@ set (src_max_sources
      src/max/as_max_plastic_material.osl
      src/max/as_max_srgb_to_linear_rgb.osl
      src/max/as_max_sss_material.osl
-     src/max/as_max_standard_material.osl
      src/max/as_max_uv_transform.osl
 )
 list (APPEND osl_sources
--- a/src/appleseed.shaders/src/max/as_max_uv_transform.osl
+++ b/src/appleseed.shaders/src/max/as_max_uv_transform.osl
@@ -56,6 +56,9 @@ shader as_max_uv_transform
 
     output float out_U = 0.0,
     output float out_V = 0.0,
+
+    output float out_outUV[2] = {0.0, 0.0},
+    output float out_outUvFilterSize[2] = {-1, -1},
 )
 {
     float out_u = in_coordU;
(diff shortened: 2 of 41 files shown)


luisbarrancos:
LGTM

usakhelo:
Max PR is ready for review. I'll merge this once plugin is approved.

src/appleseed.shaders/CMakeLists.txt
```diff
@@ -324,8 +324,8 @@ set (src_max_sources
      src/max/as_max_plastic_material.osl
      src/max/as_max_srgb_to_linear_rgb.osl
      src/max/as_max_sss_material.osl
-     src/max/as_max_standard_material.osl
      src/max/as_max_uv_transform.osl
+     src/max/as_max_closure2surface.osl
```

src/appleseed.shaders/src/max/as_max_uv_transform.osl
```diff
@@ -66,7 +69,14 @@ shader as_max_uv_transform
     // Apply rotation before tiling/offset which is different from 3ds Max native bitmap transforms.
     if (in_rotateW != 0.0)
     {
-        rotate2d(out_u, out_v, in_rotateW, out_u, out_v);
+        point rot = rotate(
+                point(out_u, out_v, 0.0),
```
dictoon: Fix indentation.

src/appleseed/renderer/kernel/shading/closures.cpp
```diff
@@ -1241,6 +1241,7 @@ namespace
             OSL::Color3     mean_free_path;
             float           ior;
             float           fresnel_weight;
+            float           zero_scattering_weight;
```
est77: There are a few missing bits:

You need to initialize zero_scattering_weight with a default value here:
https://github.com/appleseedhq/appleseed/pull/1842/commits/08bb52777280d2dda2608b068704f99b1af62435#diff-84fe0468350bc438ced5b170a9197da1L1263

In case the user did not specify it as an argument.

And you need to copy the value to the BSSDF inputs here:

https://github.com/appleseedhq/appleseed/pull/1842/commits/08bb52777280d2dda2608b068704f99b1af62435#diff-84fe0468350bc438ced5b170a9197da1L1384

est77: But before that, we should first check with Artem if this parameter will still exist in his improved RandomWalkBSSRDF.

est77: Artem told me that he is going to remove the zero scattering weight param from the bssrdf.
We could just remove the changes in this file.

luisbarrancos: @est77 i'll remove this from as_subsurface as well.

usakhelo:
Removed zero_scattering_weight changes and rebased.

usakhelo:
Sorry, fixed code style and pushed again.

luisbarrancos:
If this was being widely used, shader groups using those two shaders would now be invalid and would need their shader graphs updated, but this is not the case since the two shaders were added recently to git and weren't included in any release yet.

usakhelo:
travis failed the commit with message `The job exceeded the maximum time limit for jobs, and has been terminated.`

dictoon:
@usakhelo Once in a while a Travis job takes too long to finish and fails. It's not clear why. Merging this anyway. Thanks!

Issue #1722, opened 2017-11-22
Move appleseed.python tests to subdirectory


dictoon:

RajaVamsi11:
@dictoon  I have a question?
This issue is just about  moving appleseed.python  test to sub directory.
(simply can i cut tests from appleseed.python and paste into subdirectory ?)

dictoon:
It's about moving the tests, updating the CMake files, making sure the tests continue running, etc.

RajaVamsi11:
Ok @dictoon  I will try solving this.

codesavory:
I see that there is a **test** folder inside appleseed.python, is there something more specific expected for this issue?

PR #1849, opened 2018-02-18
Move appleseed.python tests to subdirectory

RISHAV98:
closing #1722 

I have moved the appleseed.python tests to subdirectory .
please have a look


dictoon:
You're commiting binary and object files, that makes no sense.

I highly recommend you start by getting familiar with the concept of source control and reading about best practices regarding Git and GitHub.

PR #1851, opened 2018-02-19
Add a project migration step 22.

pjessesco:
- It generate `shutter_open_end_time` and `shutter_close_start_time` as `shutter_open_time` and `shutter_close_time`, if they are not exist.

### DIFF
--- a/src/appleseed/renderer/modeling/project/projectfileupdater.cpp
+++ b/src/appleseed/renderer/modeling/project/projectfileupdater.cpp
@@ -1585,6 +1585,47 @@ namespace
             }
         }
     };
+
+    //
+    // Update from revision 21 to revision 22.
+    //
+    
+    class UpdateFromRevision_21
+      : public Updater
+    {
+      public:
+        explicit UpdateFromRevision_21(Project& project)
+          : Updater(project, 21)
+        {
+        }
+
+        void update() override
+        {
+            if (Scene* scene = m_project.get_scene())
+            {
+                for (each<CameraContainer> i = scene->cameras(); i; ++i)
+                {
+                    Dictionary& camera_params = i->get_parameters();
+
+                    copy_in_same_path(
+                        camera_params,
+                        "shutter_open_end_time",
+                        "shutter_open_time");
+
+                    copy_in_same_path(
+                        camera_params,
+                        "shutter_close_start_time",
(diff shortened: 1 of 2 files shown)


src/appleseed/renderer/modeling/project/projectfileupdater.cpp
```diff
@@ -1585,6 +1585,47 @@ namespace
...
+                    copy_in_same_path(
+                        camera_params,
+                        "shutter_close_start_time",
+                        "shutter_close_time");
+                }
+            }
+        }
+
+      private:        
+        void copy_in_same_path(Dictionary& dict, const char* dest_key, const char* src_key){
```
pjessesco: I think it can be defined in `Updater` class..

src/appleseed/renderer/modeling/project/projectfileupdater.cpp
```diff
@@ -1585,6 +1585,47 @@ namespace
...
+        explicit UpdateFromRevision_21(Project& project)
+          : Updater(project, 21)
+        {
+        }
+
+        void update() override
+        {
+            if (Scene* scene = m_project.get_scene())
+            {
+                for (each<CameraContainer> i = cameras; i; ++i)
```
dictoon: `scene.cameras()`, as is the code doesn't compile.

By the way, we now switched to C++11 so feel free to use range-based for loops:

    for (Camera& camera : scene->cameras())

src/appleseed/renderer/modeling/project/projectfileupdater.cpp
dictoon: Yes it makes sense to move it next to the other similar methods.

The name is not very clear, maybe `copy_if_exist_no_overwrite()`?

Also make static, and move the opening brace to its own line.

src/appleseed/renderer/modeling/project/projectfileupdater.cpp
pjessesco: Sorry for PR before check :(
Then shall I change other loops to range-based?

dictoon:
Thanks, merged!

> Then shall I change other loops to range-based?

We try to use the new range-based notation for new code but both notations are fine so I merged.

PR #1855, opened 2018-02-21
GCC +7.0:  fix warnings

oktomus:
- Fix throw specifications errors by ignoring them before including the headers using them
- Ignore the deprecated errors from OpenImageIO
- Fix bool enum context in Eigen


src/appleseed/foundation/image/color.h
```diff
@@ -40,7 +40,14 @@
 // Imath headers.
 #ifdef APPLESEED_ENABLE_IMATH_INTEROP
 #include "foundation/platform/_beginexrheaders.h"
+#if __GNUC__ >= 7
```
biart: I am terribly sorry, but I forgot to mention that we have special headers specifically for this use case!

Please, have a look at the `#include "foundation/platform/_beginexrheaders.h"` at line 42 and `#include "foundation/platform/_endexrheaders.h"` at line 51. These headers work as "brackets" in which we place all OpenEXR includes.

If you put lines 43-46 in the above header and lines 48-50 in the header below, we can avoid further duplication these pragmas and the code will become cleaner.

src/appleseed/foundation/image/exrimagefilewriter.cpp
```diff
@@ -41,6 +41,10 @@
 
 // OpenEXR headers.
 #include "foundation/platform/_beginexrheaders.h"
+#if __GNUC__ >= 7
```
biart: You can put all pragmas to `"foundation/platform/_beginexrheaders.h"` once, as I mentioned above.
If you do this, we don't need to repeat these pragmas every time.

src/appleseed/foundation/image/exrimagefilewriter.cpp
```diff
@@ -53,6 +57,9 @@
 #include "OpenEXR/ImfTileDescription.h"
 #include "OpenEXR/ImfTiledOutputPart.h"
 #include "OpenEXR/ImfTiledOutputFile.h"
+#if __GNUC__ >= 7
```
biart: Same here, move to `"foundation/platform/_endexrheaders.h"`

src/appleseed/foundation/image/exrutils.h
```diff
@@ -32,7 +32,14 @@
 
 // OpenEXR headers.
 #include "foundation/platform/_beginexrheaders.h"
+#if __GNUC__ >= 7
```
biart: Same here

src/thirdparty/bcd/bcd/ImageIO.cpp
```diff
@@ -21,11 +21,20 @@
     #pragma warning (disable: 4800)     // 'int': forcing value to bool 'true' or 'false' (performance warning)
 #endif
 
+#if __GNUC__ >= 7
```
biart: Since this is not a part of src/appleseed, it is okay to leave it like this without breaking the code consistency.

src/appleseed/foundation/image/genericprogressiveimagefilereader.cpp
```diff
@@ -39,8 +39,15 @@
 
 // OpenImageIO headers.
 #include "foundation/platform/_beginoiioheaders.h"
+#if __GNUC__ >= 7
```
biart: This is OpenImageIO headers. It seems that OIIO depends on OpenEXR and includes its headers, therefore we can place `#include "foundation/platform/_beginexrheaders.h` at the beginning of `_beginoiioheaders.h` and `#include "foundation/platform/_endexrheaders.h` at the end of `_endoiioheaders.h`.
So it will be beginEXRheaders beginOIIOheaders ... endOIIOheaders endEXRheaders

src/appleseed/foundation/math/matrix.h
```diff
@@ -45,7 +45,18 @@
 // Imath headers.
 #ifdef APPLESEED_ENABLE_IMATH_INTEROP
 #include "foundation/platform/_beginexrheaders.h"
+
```
biart: Same here. Please, remove newly added pragmas for all `src/appleseed` headers below.

src/thirdparty/bcd/bcd/ImageIO.cpp
biart: So no need to put it to separate headers =)

src/appleseed/foundation/image/genericprogressiveimagefilereader.cpp
```diff
@@ -38,17 +38,12 @@
 #include "foundation/image/tile.h"
 
 // OpenImageIO headers.
+#include "foundation/platform/_beginexrheaders.h"
 #include "foundation/platform/_beginoiioheaders.h"
```
dictoon: If OpenImageIO includes OpenEXR header files, should `_beginoiioheaders.h` include `_beginexrheaders.h`?

oktomus: Yes, that make sense. I will try to add `#include "foundation/platform/_beginexrheaders.h"` inside `_beginoiioheaders .h`.

oktomus:
Great ! I am glad I could have helped :)

Issue #1484, opened 2017-06-24
Proper handling of media priorities in Tracer


biart:
We need to ignore the intersections that don't pass medium priority check in `Tracer::do_trace` and `Tracer::do_trace_between` just like we already do that in [pathtracer.h](https://github.com/appleseedhq/appleseed/blob/0b372f6842e3c3c073da4602815550514746f14e/src/appleseed/renderer/kernel/lighting/pathtracer.h#L279).

dictoon:
Good point!

psy2d:
I'd like to work on this issue.

psy2d:
I wanted to confirm my understanding:  for the intersections that don't pass medium priority check, we simply
1. Continue the ray in the same direction.
2. Update the medium list.
3. Trace the ray further.
 Is that right?

biart:
Yes, it is absolutely correct.

psy2d:
Ok thanx!

biart:
See https://github.com/appleseedhq/appleseed/pull/1857 for more details.

PR #1857, opened 2018-02-25
Handle false intersections in Tracer::do_trace() and Tracer::do_trace_between()

psy2d:
A fix for #1484

### DIFF
--- a/src/appleseed/renderer/kernel/lighting/tracer.cpp
+++ b/src/appleseed/renderer/kernel/lighting/tracer.cpp
@@ -129,12 +129,45 @@ const ShadingPoint& Tracer::do_trace(
             }
             break;
         }
-
+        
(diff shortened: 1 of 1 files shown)


biart:
Also, I quickly crafted a scene that might be helpful to test whether shading rays work properly with priorities:

[psy.zip](https://github.com/appleseedhq/appleseed/files/1764878/psy.zip)

psy2d:
Thanks! I'll make the required changes and test it.

biart:
I investigated this and it seems that the result is already correct in every possible case. Shading rays don't interact with the surface in any way (except for alpha), so ignoring some of them won't change anything. Sorry for the non-existent issue.

But I think it is worth to keep in mind that our current media handling is "simplified" in case we want to add some approximations for our shading rays in the future (like thin glass).

Issue #1860, opened 2018-02-25
Provide an EditorConfig file


dictoon:
See http://editorconfig.org/

dictoon:
Done.

PR #1861, opened 2018-02-25
Add an editorconfig file, feature #1860

oktomus:


dictoon:
Awesome!

We could also add the following properties:
- `charset` (utf-8)
- `trim_trailing_whitespace` (true)
- `insert_final_newline` (true)

oktomus:
Sure :)

dictoon:
Perfect, thank you!

Issue #1858, opened 2018-02-25
Fix saving of adaptive sampler's diagnostic AOVs


dictoon:
They exist as "extra AOV images" created by `Frame::create_extra_aov_image()` but they aren't saved to disk by `Frame::write_aov_images()` or `Frame::write_main_and_aov_images()` since they aren't backed by AOV entities.

oktomus:
You can assign me if that's ok :)

oktomus:
This can be closed :)

dictoon:
Yes, thanks!

PR #1862, opened 2018-02-25
Don't reset trackers when sampling is finished

oktomus:
This is not a fix for #1858 but a fix for a mistake in the adaptive sampler. The trackers were resat before writing them in an image when diagnostic is enabled.

### DIFF
--- a/src/appleseed/renderer/kernel/rendering/final/adaptivepixelrenderer.cpp
+++ b/src/appleseed/renderer/kernel/rendering/final/adaptivepixelrenderer.cpp
@@ -241,7 +241,7 @@ namespace
                     trackers[1].insert(shading_result.m_main[1]);
                     trackers[2].insert(shading_result.m_main[2]);
                 }
-                
+
                 // Stop if the variation criterion is met.
                 if (trackers[0].get_variation() <= m_params.m_max_variation &&
                     trackers[1].get_variation() <= m_params.m_max_variation &&
--- a/src/appleseed/renderer/kernel/rendering/final/adaptivepixelrenderer.cpp
+++ b/src/appleseed/renderer/kernel/rendering/final/adaptivepixelrenderer.cpp
@@ -193,13 +193,6 @@ namespace
                 if (remaining_samples == 0)
                     break;
 
-                // Stop if the variation criterion is met.
-                if (trackers[0].get_size() > 0 &&
-                    trackers[0].get_variation() <= m_params.m_max_variation &&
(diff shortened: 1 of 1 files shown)


src/appleseed/renderer/kernel/rendering/final/adaptivepixelrenderer.cpp
```diff
@@ -187,16 +187,16 @@ namespace
...
-                trackers[1].reset_variation();
-                trackers[2].reset_variation();
-
                 // Don't exceed 'max' samples in total.
                 assert(trackers[0].get_size() <= m_params.m_max_samples);
                 const size_t remaining_samples = m_params.m_max_samples - trackers[0].get_size();
                 if (remaining_samples == 0)
                     break;
 
+                trackers[0].reset_variation();
```
dictoon: Yes, that looks correct actually.

I would go even further: I would move the other termination checks just after the check on the remaining number of samples:

    // Stop if the variation criterion is met.
    if (trackers[0].get_variation() <= m_params.m_max_variation &&
        trackers[1].get_variation() <= m_params.m_max_variation &&
        trackers[2].get_variation() <= m_params.m_max_variation)
        break;

oktomus: If you do so, sampling will stop on the first step. Just have to check the size at the same time.

dictoon: Oh you're right, sorry. Maybe it's better to leave the test where it was then. Sorry for the troubles!

oktomus: Ok no problem :)

src/appleseed/renderer/kernel/rendering/final/adaptivepixelrenderer.cpp
```diff
@@ -241,7 +241,7 @@ namespace
                     trackers[1].insert(shading_result.m_main[1]);
                     trackers[2].insert(shading_result.m_main[2]);
                 }
-
+
```
dictoon: Remove trailing spaces.

oktomus: That's what happen when you edit the code directly on github :(

dictoon: Yeah, not recommended especially since it's easy to introduce mistakes that will break compilation :)

PR #1865, opened 2018-02-27
Save extra AOVs

oktomus:
We are now able to:
- Successfully save extra AVOS as exr
- Save extra AOVs as png if possible

Fix for #1858 

I am not sure if you are ok regarding changes on ` Frame::create_extra_aov_image`.
I haven't been able to test futher with png because I have a linking issue (possible version mismatch) which cause that `pngptr` is null in `PNGImageFileWriter::write`.

I was also thinking of a different approach:
Create new AVO classes for data images and color images. Is this possible ?

### DIFF
--- a/src/appleseed/renderer/kernel/rendering/final/adaptivepixelrenderer.cpp
+++ b/src/appleseed/renderer/kernel/rendering/final/adaptivepixelrenderer.cpp
@@ -91,8 +91,8 @@ namespace
         {
             if (m_params.m_diagnostics)
             {
-                m_variation_aov_index = frame.create_extra_aov_image("variation");
-                m_samples_aov_index = frame.create_extra_aov_image("samples");
+                m_variation_aov_index = frame.create_extra_aov_image("variation", PixelFormatHalf);
+                m_samples_aov_index = frame.create_extra_aov_image("samples", PixelFormatHalf);
 
                 if ((thread_index == 0) && (m_variation_aov_index == size_t(~0) || m_samples_aov_index == size_t(~0)))
                 {
--- a/src/appleseed/renderer/kernel/rendering/final/adaptivepixelrenderer.cpp
+++ b/src/appleseed/renderer/kernel/rendering/final/adaptivepixelrenderer.cpp
@@ -91,8 +91,8 @@ namespace
         {
             if (m_params.m_diagnostics)
             {
-                m_variation_aov_index = frame.create_extra_aov_image("variation", PixelFormatHalf);
-                m_samples_aov_index = frame.create_extra_aov_image("samples", PixelFormatHalf);
+                m_variation_aov_index = frame.create_extra_aov_image("variation");
+                m_samples_aov_index = frame.create_extra_aov_image("samples");
 
                 if ((thread_index == 0) && (m_variation_aov_index == size_t(~0) || m_samples_aov_index == size_t(~0)))
                 {
--- a/src/appleseed/renderer/modeling/frame/frame.cpp
+++ b/src/appleseed/renderer/modeling/frame/frame.cpp
@@ -122,6 +122,7 @@ struct Frame::Impl
     AOVContainer            m_aovs;
     AOVContainer            m_internal_aovs;
     DenoiserAOV*            m_denoiser_aov;
+	vector<size_t>          m_extra_aovs;
 };
 
 Frame::Frame(
@@ -283,9 +284,12 @@ const AOVContainer& Frame::internal_aovs() const
 
 size_t Frame::create_extra_aov_image(const char* name) const
 {
-    const size_t index = aov_images().get_index(name);
+    size_t index = aov_images().get_index(name);
     if (index == size_t(~0) && aov_images().size() < MaxAOVCount)
-        return aov_images().append(name, 4, PixelFormatFloat);
+    {
+        index = aov_images().append(name, 4, PixelFormatFloat);
+        impl->m_extra_aovs.push_back(index);
+    }
 
     return index;
 }
@@ -693,6 +697,7 @@ namespace
 
     bool write_extra_aovs(
         const ImageStack&       images,
+            const vector<size_t>&   aov_indices,
         const bf::path&         directory,
         const string&           base_file_name,
         const string&           extension)
@@ -705,12 +710,15 @@ namespace
             return false;
         }
 
-        for (size_t i = 0, e = images.size(); i < e; ++i)
+        for (size_t i = 0, e = aov_indices.size(); i < e; ++i)
         {
-            const Image & image = images.get_image(i);
+            size_t image_index = aov_indices[i];
+            assert(image_index < images.size());
 
-            // Compute AOV image file path.
-            const string aov_name = images.get_name(i);
+            const Image & image = images.get_image(image_index);
+
+            // Compute image file path.
+            const string aov_name = images.get_name(image_index);
             const string safe_aov_name = make_safe_filename(aov_name);
             const string aov_file_name = base_file_name + "." + safe_aov_name + extension;
             const string aov_file_path = (directory / aov_file_name).string();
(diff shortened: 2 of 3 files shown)


est77:
Thanks for the PR!
I'll check it today after work.

dictoon:
Thanks! Small remark: it's spelled AOV for Arbitrary Output Variables, not AVO.

oktomus:
Wooups, I know and still I made the mistake a lot

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -776,6 +804,30 @@ bool Frame::write_main_and_aov_images() const
...
 
+    if (!aov_images().empty())
+    {
+        const bf::path boost_file_path(filepath);
+        const bf::path directory = boost_file_path.parent_path();
+        const string base_file_name = boost_file_path.stem().string();
+        const string extension = boost_file_path.extension().string();
+
+        // Extra AOVs
+        for (size_t i = 0, e = aov_images().size(); i < e; ++i)
```
est77: This fragment of code is repeated. Refactor into a function in an unnamed namespace if possible, 
if not, refactor to a private method of the frame.

src/appleseed/renderer/kernel/rendering/final/adaptivepixelrenderer.cpp
```diff
@@ -91,8 +91,8 @@ namespace
         {
             if (m_params.m_diagnostics)
             {
-                m_variation_aov_index = frame.create_extra_aov_image("variation");
-                m_samples_aov_index = frame.create_extra_aov_image("samples");
+                m_variation_aov_index = frame.create_extra_aov_image("variation", PixelFormatHalf);
```
est77: With these changes, the range and precision of the images used by the adaptive pixel renderer is much lower than before. This is going to affect the results of the adaptive sampling. I think we want to keep them as floats.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -278,11 +278,13 @@ const AOVContainer& Frame::internal_aovs() const
     return impl->m_internal_aovs;
 }
 
-size_t Frame::create_extra_aov_image(const char* name) const
+size_t Frame::create_extra_aov_image(
+    const char* name,
+    const PixelFormat pixel_format) const
```
est77: Personally, I prefer to undo this change and remove the option of saving extra aov images to pngs.
The images in the extra aov images are always non-color images and in my opinion, it doesn't make much sense to save them as pngs.

src/appleseed/renderer/kernel/rendering/final/adaptivepixelrenderer.cpp
oktomus: I chanded the format because the tile_end maps values to colors.

oktomus:
Thanks for the review @est77, I will commit asap :)

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -720,7 +765,7 @@ bool Frame::write_aov_images(const char* file_path) const
 
     bool success = true;
 
-    if (!aovs().empty())
+    if (!aovs().empty() || !aov_images().empty())
```
est77: Remove the !aov_images().empty() check

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -687,6 +700,38 @@ namespace
 
         return true;
     }
+
+    bool write_extra_aovs(
+        const ImageStack & images,
```
est77: Remove space between the types and the '&' and align parameter names vertically.

est77:
Thanks for the changes.

There is a problem that I missed yesterday.
In write_extra_aovs(), we save all the images in the image stack:

for (size_t i = 0, e = images.size(); i < e; ++i)

but many AOVs also create images on the image stack.
As a result, we will save them twice, once when the AOVs are written and a second time in write_extra_aovs.

Maybe the easiest way to avoid saving some images twice is to keep a vector of the image indices created by Frame::create_extra_aov_image and in write_extra_aovs, write only those images.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -646,6 +649,16 @@ namespace
             }
             else if (extension == ".png")
             {
+                // Test if the image contains appropriate data
```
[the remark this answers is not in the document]
oktomus: Yes, I forgot that

src/appleseed/renderer/modeling/frame/frame.cpp
oktomus: Can you explain me why ?

oktomus:
Thanks for the review, I will do that :)

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -720,27 +763,34 @@ bool Frame::write_aov_images(const char* file_path) const
 
     bool success = true;
 
-    if (!aovs().empty())
```
oktomus: I do not check for the size anymore because if I do, I would have to check `aovs().empty()` and `aov_images().empty()`. Otherwise extra AOVs wouldn't be saved if there isn't any main AOV.

oktomus: Maybe I'm wrong, there is always an AOV, beauty, right ? @est77

est77: beauty is not an AOV in appleseed. There's an image in the frame to store it.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -114,13 +114,15 @@ struct Frame::Impl
...
     DenoisingMode           m_denoising_mode;
+    bool                    m_save_extra_aovs;
 
     // Images.
     unique_ptr<Image>       m_image;
     unique_ptr<ImageStack>  m_aov_images;
     AOVContainer            m_aovs;
     AOVContainer            m_internal_aovs;
     DenoiserAOV*            m_denoiser_aov;
+	vector<size_t>          m_extra_aovs;
```
est77: align the type and the var member name with the previous lines.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -280,9 +284,12 @@ const AOVContainer& Frame::internal_aovs() const
 
 size_t Frame::create_extra_aov_image(const char* name) const
 {
-    const size_t index = aov_images().get_index(name);
+    size_t index = aov_images().get_index(name);
```
est77: index can be const. We don't modify it in the method.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -687,6 +694,42 @@ namespace
 
         return true;
     }
+
+    bool write_extra_aovs(
+        const ImageStack&       images,
+            const vector<size_t>&   aov_indices,
```
est77: align with the rest of the arguments.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -687,6 +694,42 @@ namespace
...
+
+        if (extension != ".exr")
+        {
+            RENDERER_LOG_ERROR("extra AOVs can only be saved as exr.");
+            return false;
+        }
+
+        for (size_t i = 0, e = aov_indices.size(); i < e; ++i)
+        {
+            size_t image_index = aov_indices[i];
```
est77: make const

est77:
Thanks for the changes!
Only a few very small details and we can merge.

oktomus:
Thanks for the review, I'm really sorry for all these align issues.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -720,27 +764,34 @@ bool Frame::write_aov_images(const char* file_path) const
...
+        // Write AOV image.
+        if (!write_image(aov_file_path.c_str(), aov->get_image(), aov))
+            success = false;
+    }
 
-            // Write AOV image.
-            if (!write_image(aov_file_path.c_str(), aov->get_image(), aov))
-                success = false;
-        }
+    if (impl->m_save_extra_aovs && !aov_images().empty())
```
dictoon: Why checking `!aov_images().empty()`? Isn't this test redundant?

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -687,6 +695,42 @@ namespace
...
+    bool write_extra_aovs(
+        const ImageStack&       images,
+        const vector<size_t>&   aov_indices,
+        const bf::path&         directory,
+        const string&           base_file_name,
+        const string&           extension)
+    {
+        bool success = true;
+
+        if (extension != ".exr")
```
dictoon: Is `extension` guaranteed to always be a lower case?

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -821,6 +888,21 @@ void Frame::write_main_and_aov_images_to_multipart_exr(const char* file_path) co
...
     }
 
+    if (impl->m_save_extra_aovs)
+    {
+        for (size_t i = 0, e = impl->m_extra_aovs.size(); i < e; ++i)
+        {
+            const size_t image_index = impl->m_extra_aovs[i];
+            assert(image_index < aov_images().size());
+
+            const Image & image = aov_images().get_image(image_index);
```
dictoon: Remove space before `&`.

dictoon:
Very nice work @oktomus! Appreciated.

dictoon:
@est77 When merging this, make sure to squash the PR into a single commit, and to remove the individual commit titles from the commit body (which GitHub fills automatically).

est77:
Done!

PR #1866, opened 2018-02-28
Add triplanar normal mode, update references

luisbarrancos:
Add [Colour](https://github.com/colour-science/colour) reference since it was used to precompute the RGB<>XYZ and RGB<>RGB matrices.

### DIFF
--- a/src/appleseed.shaders/include/appleseed/color/as_color_transforms.h
+++ b/src/appleseed.shaders/include/appleseed/color/as_color_transforms.h
@@ -135,6 +135,7 @@ vector get_illuminant_CIExyz(string illuminant, float white_CIExy[2])
         : transform_CIExy_to_CIExyz(white_CIExy);
 }
 
+
 //
 // Reference:
 //
@@ -169,6 +170,7 @@ vector transform_CIEXYZ_to_CIExyY(color XYZ, string illuminant)
     return transform_CIEXYZ_to_CIExyY(XYZ, white_CIExy);
 }
 
+
 //
 // Reference:
 //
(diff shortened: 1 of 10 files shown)


src/appleseed.shaders/include/appleseed/color/as_color_transforms.h
```diff
@@ -749,7 +749,6 @@ color transform_CIELAB_to_CIEXYZ(
         get_illuminant_CIExyY(illuminant));
 }
 
-
```
dictoon: In source code, comment blocks of the form

    //
    // ...
    // ...
    //

should always be preceded by two blank lines.

It's the only case where we use two consecutive blank lines.

src/appleseed.shaders/src/appleseed/as_triplanar.osl
```diff
@@ -37,9 +37,24 @@ shader as_triplanar
...
+        string as_maya_attribute_short_name = "blm",
+        string widget = "popup",
+        string options = "Color|Tangent Normal",
+        string label = "Blend Mode",
+        string page = "Projection",
+        int as_maya_attribute_connectable = 0,
+        int as_maya_attribute_keyable = 0,
+        int as_maya_attribute_hidden = 1,
+        int gafferNoduleLayoutVisible = 0,
+        string help = "Blend colours, or tangent space normal maps using reoriented normal mapping.",
```
dictoon: Use American spelling, *color*.

src/appleseed.shaders/include/appleseed/color/as_color_transforms.h
luisbarrancos: ok, i wasn't sure about that, will fix now.

PR #1867, opened 2018-03-01
Add camera shutter times check

fedormatantsev:
Please, review;
Also in the scope of this issue I suggest to change m_shutter_open_end_time default value to m_shutter_open_time and m_shutter_close_start default value to m_shutter_close_time; In this case there won't be any warnings if start/close values are set and open end/ close start values are missing;
screenshot from 2018-03-01 14-02-38


dictoon:
Nice work, thank you!

PR #1870, opened 2018-03-01
Preserve environment_edf transform after edit.

pjessesco:
I added edit_entity() when Entity = EnvironmentEDF and ParentEntity = Scene.

### DIFF
--- a/src/appleseed.studio/mainwindow/project/projectbuilder.h
+++ b/src/appleseed.studio/mainwindow/project/projectbuilder.h
@@ -165,15 +165,11 @@ Entity* ProjectBuilder::edit_entity(
     return new_entity_ptr;
 }
 
-template <>
-inline renderer::EnvironmentEDF* ProjectBuilder::edit_entity(
-    renderer::EnvironmentEDF*           old_entity,
-    renderer::Scene&                    parent,
-    const foundation::Dictionary&       values) const
+template <typename OldEntity, typename NewEntity>
+inline void copy_transform_sequence(
+    const OldEntity&                   old_entity,
+    const NewEntity&                   new_entity)
 {
-    foundation::auto_release_ptr<renderer::EnvironmentEDF> new_entity(create_entity<renderer::EnvironmentEDF>(values));
-    renderer::EnvironmentEDF* new_entity_ptr = new_entity.get(); 
-
     new_entity->transform_sequence().clear();
 
     for (size_t i = 0, e = old_entity->transform_sequence().size(); i < e; ++i)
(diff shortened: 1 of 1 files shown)


src/appleseed.studio/mainwindow/project/projectbuilder.h
```diff
@@ -165,15 +165,11 @@ Entity* ProjectBuilder::edit_entity(
     return new_entity_ptr;
 }
 
-template <>
-inline renderer::EnvironmentEDF* ProjectBuilder::edit_entity(
-    renderer::EnvironmentEDF*           old_entity,
-    renderer::Scene&                    parent,
-    const foundation::Dictionary&       values) const
+template <typename OldEntity, typename NewEntity>
+inline void copy_transform_sequence(
```
pjessesco: Is it okay to define `copy_transform_sequence()` between `edit_entity()` definitions?

pjessesco:
Similarly with this PR, same happens in `test scene/cameras/01` : blurry effect doesn't preserved with edit. I'd like to take it too, what value is related with blur effect?

src/appleseed.studio/mainwindow/project/projectbuilder.h
```diff
@@ -164,15 +165,11 @@ Entity* ProjectBuilder::edit_entity(
     return new_entity_ptr;
 }
 
-template <>
-inline renderer::Camera* ProjectBuilder::edit_entity(
-    renderer::Camera*                   old_entity,
-    renderer::Scene&                    parent,
-    const foundation::Dictionary&       values) const
+template <typename OldEntity, typename NewEntity>
```
dictoon: I would expect `OldEntity` and `NewEntity` to always be the same type; let's unify them in a single `Entity` param?

src/appleseed.studio/mainwindow/project/projectbuilder.h
```diff
@@ -164,15 +165,11 @@ Entity* ProjectBuilder::edit_entity(
...
+inline void copy_transform_sequence(
+    const OldEntity&                   old_entity,
```
dictoon: In the strict context of this function, I think `source_entity` and `destination_entity` would make more sense.

dictoon:
> Similarly with this PR, same happens in test scene/cameras/01 : blurry effect doesn't preserved with edit. I'd like to take it too, what value is related with blur effect?

Good catch! That's a little surprising though. The blurring effect is due to depth of field which is entirely controlled by camera parameters (which are properly preserved across edits, normally):

    <parameter name="diaphragm_blades" value="5" />
    <parameter name="diaphragm_tilt_angle" value="0.0" />
    <parameter name="f_stop" value="1" />
    <parameter name="focal_distance" value="0.5" />

Investigation appreciated!!

PR #1872, opened 2018-03-03
Build and install blenderseed shaders

est77:


src/appleseed.shaders/CMakeLists.txt
```diff
@@ -154,6 +154,18 @@ source_group ("include\\appleseed\\transform" FILES
 # Source files.
 #--------------------------------------------------------------------------------------------------
 
+
```
dictoon: Remove one blank line.

PR #1874, opened 2018-03-03
Fix compile error

pjessesco:
So sorry about this :(


dictoon:
No problem, thanks for the fix.

Issue #1789, opened 2018-01-14
Typing values in color entity's Color field doesn't update color preview


dictoon:

jaykay12:
![screenshot from 2018-02-14 22-59-15](https://user-images.githubusercontent.com/13948542/36218569-d88bd50e-11da-11e8-91a8-404ab72f1ab8.png)


If I am not wrong, the color preview containing bug is this one, its not updating when I am changing the RGB value in the input field.

dictoon:
@jaykay12 That's correct.

jaykay12:
Okayy @dictoon thanks for confirming.
I will be trying to resolve this one once my sessionals are over in just couple of days.

dictoon:
Done.

PR #1875, opened 2018-03-03
Fix color field update issue

rgirish28:
Fixes issue #1789 and changes the picker button color preview when the text changes. Only updates the value once return is pressed as with the other fields.

### DIFF
--- a/src/appleseed.studio/utility/inputwidgetproxies.cpp
+++ b/src/appleseed.studio/utility/inputwidgetproxies.cpp
@@ -239,7 +239,7 @@ ColorPickerProxy::ColorPickerProxy(QLineEdit* line_edit, QToolButton* picker_but
   , m_picker_button(picker_button)
 {
     connect(m_line_edit, SIGNAL(returnPressed()), SIGNAL(signal_changed()));
-    connect(m_line_edit, SIGNAL(textChanged(const QString &)), SLOT(slot_set(const QString &)));
+    connect(m_line_edit, SIGNAL(textChanged(const QString&)), SLOT(slot_set(const QString&)));
 }
 
 namespace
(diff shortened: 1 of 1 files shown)


src/appleseed.studio/utility/inputwidgetproxies.cpp
```diff
@@ -239,6 +239,7 @@ ColorPickerProxy::ColorPickerProxy(QLineEdit* line_edit, QToolButton* picker_but
   , m_picker_button(picker_button)
 {
     connect(m_line_edit, SIGNAL(returnPressed()), SIGNAL(signal_changed()));
+    connect(m_line_edit, SIGNAL(textChanged(const QString &)), SLOT(slot_set(const QString &)));
```
dictoon: We always put the `&` on the type's side: `const QString&`.

dictoon:
Nice, thanks! I'm (positively) surprised that this is a single line fix.

dictoon:
One last detail: Could you change the commit title to use present tense, i.e. "Fix" instead of "Fixed"? Our commit guidelines are detailed here: https://github.com/appleseedhq/appleseed/wiki/Git-Commits

rgirish28:
Thanks. I've made the required changes. Yeh, I was surprised it was a one line fix as well. I like how the IInputWidgetProxy class is setup and it definitely helped with keeping the code clean 👍

oktomus:
@rgirish28 You should rename also your commit message :)

rgirish28:
Done 👍

dictoon:
Thanks @rgirish28. I will actually squash the PR into a single commit as it doesn't make sense to have two in this case (we try to have one commit per feature, fix, refactoring or change; we merge commits that address PR review comments into the relevant commits).

Thanks again!

PR #1880, opened 2018-03-04
Minor fix on save extra AOVs (originally #1865)

oktomus:
Original PR: #1865

### DIFF
--- a/src/appleseed/renderer/modeling/frame/frame.cpp
+++ b/src/appleseed/renderer/modeling/frame/frame.cpp
@@ -706,7 +706,7 @@ namespace
     {
         bool success = true;
 
-        if (extension != ".exr")
+        if (extension != ".exr" && extension != ".EXR")
         {
             RENDERER_LOG_ERROR("extra AOVs can only be saved as exr.");
             return false;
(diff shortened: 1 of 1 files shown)


src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -706,7 +706,7 @@ namespace
     {
         bool success = true;
 
-        if (extension != ".exr")
+        if (extension != ".exr" && extension != ".EXR")
```
dictoon: That's not robust enough (wouldn't work with ".Exr" for instance). I suggest converting the extension to lower case before doing the check.

oktomus: Ok, then we should also do that in write_image

dictoon: It's already properly done in `write_image()`:

    string extension = lower_case(bf_file_path.extension().string());

dictoon: That said, the problem appears to also exist in `write_main_image()`, maybe that's what you meant.

oktomus: No it wasn't, I haven't seen the lower_case code. But thanks for pointing out this, I will fix that

dictoon:
Thanks!

Issue #1617, opened 2017-09-08
Wire up OSL's printf() to the renderer's logger


dictoon:

oktomus:
You mean mapping `osl_printf ` to the logger ? Or something else ?

dictoon:
Yes exactly, such as `print()` calls in OSL shaders appear in the console or in appleseed.studio's Log panel.

oktomus:
I haven't found much in the OSL source code. It seems that shader print calls are mapped to osl_printf which itself use the default ErrorHandler. I tried to override the default ErrorHandler provider but it hasn't worked. I posted on the OSL mailing list to see if a developer can tell me how they deal with shader functions.

dictoon:
Done.

PR #1881, opened 2018-03-04
Wire up correctly OSL's printf() to the renderer's logger (#1617)

oktomus:
Issue #1617 
It now works correctly.
Tested on the scene `osl/01 - lambertian ....` with the shader `matte.osl` edited like this:
```

surface matte(float Kd = 1, color Cs = 0)
{
	printf("This is a printf that should appear in the log\n");
	error("This error should appear in the log");
	warning("This warning should appear");
    Ci = Kd * Cs * diffuse(N);
}
```

And here is the output on a 1 by 1 image:
```
2018-03-04T11:25:07.167967Z <004>   147 MB info    | osl: This is a printf that should appear in the log
2018-03-04T11:25:07.168097Z <004>   147 MB error   | osl: Shader error [matte]: This error should appear in the log
2018-03-04T11:25:07.168142Z <004>   147 MB warning | osl: Shader warning [matte]: This warning should appear
2018-03-04T11:25:07.169069Z <004>   147 MB info    | osl: This is a printf that should appear in the log
2018-03-04T11:25:07.169466Z <004>   147 MB info    | osl: This is a printf that should appear in the log
```

I am wondering if we should do something to avoid concurrence. As you can see, the printf appears more than once.

### DIFF
--- a/src/appleseed/renderer/kernel/rendering/oiioerrorhandler.cpp
+++ b/src/appleseed/renderer/kernel/rendering/oiioerrorhandler.cpp
@@ -45,11 +45,16 @@ namespace renderer
 // OIIOErrorHandler class implementation.
 //
 
+OIIOErrorHandler::OIIOErrorHandler()
+{
+	verbosity(VERBOSE);
+}
+
 void OIIOErrorHandler::operator()(int errcode, const string& msg)
 {
     const string modified_msg = prefix_all_lines(trim_both(msg), "osl: ");
 
-    switch (errcode)
+    switch (errcode & ErrorCodeMask)
     {
       case EH_WARNING:
         RENDERER_LOG_WARNING("%s", modified_msg.c_str());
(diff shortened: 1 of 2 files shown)


src/appleseed/renderer/kernel/rendering/oiioerrorhandler.cpp
```diff
@@ -49,7 +49,7 @@ void OIIOErrorHandler::operator()(int errcode, const string& msg)
 {
     const string modified_msg = prefix_all_lines(trim_both(msg), "osl: ");
 
-    switch (errcode)
+    switch (errcode & 0xffff0000)
```
est77: Make 0xffff0000 a constant with a name, like const int ErrorCodeMask = 0xffff0000; or similar.

oktomus: No problem.

dictoon:
> I am wondering if we should do something to avoid concurrence. As you can see, the printf appears more than once.

We can live with that for now. Avoiding repeatitions is going to require a bit more engineering (probably keeping a table of message hashes in memory).

src/appleseed/renderer/kernel/rendering/oiioerrorhandler.h
```diff
@@ -51,6 +51,11 @@ class OIIOErrorHandler
   : public OIIO::ErrorHandler
 {
   public:
+    OIIOErrorHandler() : OIIO::ErrorHandler()
```
dictoon: No need to call the base class' constructor, it's done automatically.

Also, can you move the constructor definition to the .cpp file?

oktomus: Ok :)

dictoon:
Perfect, thanks!

PR #1883, opened 2018-03-04
Delete default autofocus_target value in thin lens camera

pjessesco:
Delete default value because studio make default `autofocus_target` after edit even if there is no `autofocus_target ` in .studio file.

This makes studio preserve blurry effect in `test scene/cameras/01`.

### DIFF
--- a/src/appleseed/renderer/modeling/camera/camera.cpp
+++ b/src/appleseed/renderer/modeling/camera/camera.cpp
@@ -240,7 +240,7 @@ double Camera::extract_f_stop() const
 }
 
 void Camera::extract_focal_distance(
-    const bool&             autofocus_enabled,
+    const bool              autofocus_enabled,
     Vector2d&               autofocus_target,
     double&                 focal_distance) const
 {
@@ -249,35 +249,37 @@ void Camera::extract_focal_distance(
 
     if (autofocus_enabled)
     {
-        if (!has_param("autofocus_target"))
+        if (has_param("autofocus_target"))
+            autofocus_target = m_params.get_required<Vector2d>("autofocus_target", DefaultAFTarget);
+        
+        else
         {
             RENDERER_LOG_ERROR(
                 "while defining camera \"%s\": no \"autofocus_target\" parameter found; "
-                "using default autofocus_target value \"%f, %f\".",
+                "using default value \"%f, %f\".",
                 get_path().c_str(),
                 DefaultAFTarget[0],
                 DefaultAFTarget[1]);
             autofocus_target = DefaultAFTarget;
         }
-        else
-            autofocus_target = m_params.get_required<Vector2d>("autofocus_target", DefaultAFTarget);
 
-        focal_distance = 0.0;
+        focal_distance = DefaultFocalDistance;
     }
     else
     {
-        if (!has_param("focal_distance"))
+        if (has_param("focal_distance"))
+            focal_distance = m_params.get_required<double>("focal_distance", DefaultFocalDistance);
+
+        else
         {
             RENDERER_LOG_ERROR(
                 "while defining camera \"%s\": no \"focal_distance\" parameter found; "
-                "using default focal distance value \"%f\".",
+                "using default value \"%f\".",
                 get_path().c_str(),
                 DefaultFocalDistance);
             focal_distance = DefaultFocalDistance;
         }
-        else
-            focal_distance = m_params.get_required<double>("focal_distance", DefaultFocalDistance);
-            
+
         autofocus_target = DefaultAFTarget;
     }
 }
--- a/src/appleseed/renderer/modeling/camera/camera.cpp
+++ b/src/appleseed/renderer/modeling/camera/camera.cpp
@@ -251,7 +251,6 @@ void Camera::extract_focal_distance(
     {
         if (has_param("autofocus_target"))
             autofocus_target = m_params.get_required<Vector2d>("autofocus_target", DefaultAFTarget);
-        
         else
         {
             RENDERER_LOG_ERROR(
@@ -269,7 +268,6 @@ void Camera::extract_focal_distance(
     {
         if (has_param("focal_distance"))
             focal_distance = m_params.get_required<double>("focal_distance", DefaultFocalDistance);
-
         else
         {
             RENDERER_LOG_ERROR(
--- a/src/appleseed/renderer/modeling/camera/camera.cpp
+++ b/src/appleseed/renderer/modeling/camera/camera.cpp
@@ -240,50 +240,45 @@ double Camera::extract_f_stop() const
 }
 
 void Camera::extract_focal_distance(
-    bool&                   autofocus_enabled,
+    const bool&             autofocus_enabled,
     Vector2d&               autofocus_target,
     double&                 focal_distance) const
 {
     const Vector2d DefaultAFTarget(0.5);        // in NDC
     const double DefaultFocalDistance = 1.0;    // in meters
 
-    if (has_param("focal_distance"))
+    if (autofocus_enabled)
     {
-        if (has_param("autofocus_target"))
+        if (!has_param("autofocus_target"))
         {
-            RENDERER_LOG_WARNING(
-                "while defining camera \"%s\": autofocus is enabled; \"focal_distance\" parameter "
-                "will be ignored.",
-                get_path().c_str());
-
-            autofocus_enabled = true;
-            autofocus_target = m_params.get_required<Vector2d>("autofocus_target", DefaultAFTarget);
-            focal_distance = 0.0;
+            RENDERER_LOG_ERROR(
+                "while defining camera \"%s\": no \"autofocus_target\" parameter found; "
+                "using default autofocus_target value \"%f, %f\".",
+                get_path().c_str(),
+                DefaultAFTarget[0],
+                DefaultAFTarget[1]);
+            autofocus_target = DefaultAFTarget;
         }
         else
-        {
-            autofocus_enabled = false;
-            autofocus_target = DefaultAFTarget;
-            focal_distance = m_params.get_required<double>("focal_distance", DefaultFocalDistance);
-        }
-    }
-    else if (has_param("autofocus_target"))
-    {
-        autofocus_enabled = true;
-        autofocus_target = m_params.get_required<Vector2d>("autofocus_target", DefaultAFTarget);
+            autofocus_target = m_params.get_required<Vector2d>("autofocus_target", DefaultAFTarget);
+
         focal_distance = 0.0;
     }
     else
     {
-        RENDERER_LOG_ERROR(
-            "while defining camera \"%s\": no \"focal_distance\" or \"autofocus_target\" parameter found; "
-            "using default focal distance value \"%f\".",
-            get_path().c_str(),
-            DefaultFocalDistance);
-
-        autofocus_enabled = false;
+        if (!has_param("focal_distance"))
+        {
+            RENDERER_LOG_ERROR(
+                "while defining camera \"%s\": no \"focal_distance\" parameter found; "
+                "using default focal distance value \"%f\".",
+                get_path().c_str(),
+                DefaultFocalDistance);
+            focal_distance = DefaultFocalDistance;
+        }
+        else
+            focal_distance = m_params.get_required<double>("focal_distance", DefaultFocalDistance);
+            
         autofocus_target = DefaultAFTarget;
-        focal_distance = DefaultFocalDistance;
     }
 }
 
--- a/src/appleseed/renderer/modeling/camera/camera.h
+++ b/src/appleseed/renderer/modeling/camera/camera.h
@@ -182,7 +182,7 @@ class APPLESEED_DLLSYMBOL Camera
 
     // Utility function to retrieve the focal distance (in meters) from the camera parameters.
     void extract_focal_distance(
-        const bool&                     autofocus_enabled,
+        const bool                      autofocus_enabled,
         foundation::Vector2d&           autofocus_target,
         double&                         focal_distance) const;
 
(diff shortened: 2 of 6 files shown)


pjessesco:
I inserted this print statement in `Camera::extract_focal_distance()`.

    std::cout<<has_param("focal_distance")<<"\n";
    std::cout<<has_param("autofocus_target")<<"\n";

And here is the result.



First 2 lines (1 and 0) printed when I render before any edit, and others printed when I edit camera while interactive rendering. So I could check `has_param(autofocus_target)` becomes `true` after edit camera. 

Then, should I fix `edit_entity()` in `projectbuilder.h`?

dictoon:
Thanks for the explanation. I think I start to understand.

I wonder if the problem isn't simply that the entity editor uses the input's default value (as defined in the input metadata) when the entity being edited lacks any value for that input.

pjessesco:
Yes, so I thought when there is no `autofocus_target` in project xml file,  `autofocus_target` should not be default value after edit(unless we edit `autofocus_target` in GUI), it should be `None`(or something), so project renders blurry...

My understand is if user set `autofocus_target` in project file, it means project will use autofocus, and if not, project won't use autofocus default.

dictoon:
> My understand is if user set autofocus_target in project file, it means project will use autofocus, and if not, project won't use autofocus default.

That's correct.

I think the fix needs to be done in the entity editor rather than changing the default autofocus target value.

pjessesco:
Thanks! I'll try it

src/appleseed/renderer/modeling/camera/thinlenscamera.cpp
```diff
@@ -747,8 +747,7 @@ DictionaryArray ThinLensCameraFactory::get_input_metadata() const
             .insert("name", "autofocus_target")
             .insert("label", "Autofocus Target")
             .insert("type", "text")
-            .insert("use", "optional")
-            .insert("default", "0.5 0.5"));
+            .insert("use", "optional"));
```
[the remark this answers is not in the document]
pjessesco: For understand clearly, is it intended for situation when designer(modeler) forget to set `autofocus_target` in project file?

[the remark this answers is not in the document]
dictoon: Not really. It's the default value to use when _creating_ a thin lens camera from appleseed.studio, i.e. by default the camera will use autofocus with the focus set on the center of the image. Autofocus is not mandatory and can be disabled, in which case the focus distance is adjusted manually.

pjessesco:
How about control `autofocus_enabled` in edit window? Maybe I can set it true or false depends on whether project file contains `autofocus_target` in `projectfileupdater.cpp`

dictoon:
While that could be a solution for this particular scenario, I believe the bug is more general: the entity editor will set some inputs to default values if they weren't set in the project, which is a bug.

pjessesco:
Hi. 
I found another case similarly with this. `shutter_open_end_time` is not defined in the project file, but defaults to 0 in the edit window, like `autofocus_target`. I guess it's because its default value is defined in `camera.cpp`. However, I think it is hard to say that `shutter_open_end_time` case is also a bug. (The default value of `shutter_open_end_time` in camera.cpp is not displayed in the edit window after delete it in code) 

> the entity editor will set some inputs to default values if they weren't set in the project

So, my opinion is it's not a general bug, and we need some process for `autofocus_target` only. Any suggestions?

dictoon:
I think I understand the issue a little bit better. The root of the problem is differing defaults in the entity editor (default focal distance = empty, default autofocus target = 0.5, 0.5) and in the thin lens camera (one -- and only one -- of focal distance or autofocus target must be defined).

It's not clear how this can be fixed easily, apart from changing the entity editor to no longer fill in inputs with default values when those inputs have no value.

What do you think?

(On a side note, I think we should check if `Camera::extract_f_stop()` and `Camera::extract_focal_distance()` could be moved to the `ThinLensCamera` class.)

pjessesco:
Separating `autofocus_target` and `autofocus_enabled` comes to me as a quite good way (like checkbox I mentioned, or something)..

It doesn't seems good for me that `autofocus_enabled` is controlled by`autofocus_target`, also user doesn't know that autofocus has higher priority than custom focal distance before rendering.

For clear, does entity editor has to be fixed?

dictoon:
Let's try introducing `autofocus_enabled` then. A project migration step will need to be introduced (`projectfileupdater.cpp`).

If you search for `"visible_if"` in the source code, you'll find that it's possible to show some inputs conditionally: you could show `autofocus_target` only if `autofocus_enabled` is true.

pjessesco:
I'll try to move functions after these commits are reviewed.

src/appleseed/renderer/modeling/camera/thinlenscamera.cpp
```diff
@@ -735,20 +737,34 @@ DictionaryArray ThinLensCameraFactory::get_input_metadata() const
...
     metadata.push_back(
         Dictionary()
             .insert("name", "focal_distance")
             .insert("label", "Focal Distance")
             .insert("type", "text")
-            .insert("use", "optional"));
+            .insert("use", "optional")
+            .insert("visible_if",
+                Dictionary()
+                    .insert("autofocus_enabled", "false")));
```
pjessesco: I think input box appears/disappears interactively, but it doesn't seems good : for example, `focal_distance` checkbox appears after I uncheck `autofocus_enabled` -> press 'OK' -> and reopen edit window. It was quite inconvenient for me.

dictoon: It's definitely supposed to show/hide widgets immediately, without having to close and reopen the window. You can see this in action by editing the frame entity (Output section in the Project Explorer) and changing the value in the Denoiser dropdown.

src/appleseed/renderer/modeling/project/projectfileupdater.cpp
```diff
@@ -1630,6 +1630,38 @@ namespace
             }
         }  
     };
+
+    //
+    // Update from revision 22 to revision 23.
```
dictoon: Perfect!

src/appleseed/renderer/modeling/camera/thinlenscamera.cpp
```diff
@@ -735,20 +737,34 @@ DictionaryArray ThinLensCameraFactory::get_input_metadata() const
...
 
     metadata.push_back(
         Dictionary()
-            .insert("name", "autofocus_target")
+            .insert("name", "extract_focal_distance")
```
dictoon: Why not keep `autofocus_target` here?

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -240,50 +240,45 @@ double Camera::extract_f_stop() const
...
-            autofocus_enabled = true;
-            autofocus_target = m_params.get_required<Vector2d>("autofocus_target", DefaultAFTarget);
-            focal_distance = 0.0;
-        }
-        else
-        {
-            autofocus_enabled = false;
+            RENDERER_LOG_ERROR(
+                "while defining camera \"%s\": no \"autofocus_target\" parameter found; "
+                "using default autofocus_target value \"%f, %f\".",
```
dictoon: `using default autofocus_target value` -> `using default \"autofocus_target\" value`

or simply

`using default value`

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -240,50 +240,45 @@ double Camera::extract_f_stop() const
...
     double&                 focal_distance) const
 {
     const Vector2d DefaultAFTarget(0.5);        // in NDC
     const double DefaultFocalDistance = 1.0;    // in meters
 
-    if (has_param("focal_distance"))
+    if (autofocus_enabled)
     {
-        if (has_param("autofocus_target"))
+        if (!has_param("autofocus_target"))
```
dictoon: While this is completely correct, I'd find it even easier to read if the `if` clause was reversed.

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -240,50 +240,45 @@ double Camera::extract_f_stop() const
...
     else
     {
-        RENDERER_LOG_ERROR(
-            "while defining camera \"%s\": no \"focal_distance\" or \"autofocus_target\" parameter found; "
-            "using default focal distance value \"%f\".",
-            get_path().c_str(),
-            DefaultFocalDistance);
-
-        autofocus_enabled = false;
+        if (!has_param("focal_distance"))
```
dictoon: Same thing here, reverse clause.

src/appleseed/renderer/modeling/camera/camera.h
```diff
@@ -182,7 +182,7 @@ class APPLESEED_DLLSYMBOL Camera
 
     // Utility function to retrieve the focal distance (in meters) from the camera parameters.
     void extract_focal_distance(
-        bool&                           autofocus_enabled,
+        const bool&                     autofocus_enabled,
```
dictoon: You no longer need to pass a reference here, simply pass a value.

src/appleseed/renderer/modeling/camera/thinlenscamera.cpp
dictoon: Would appreciate further investigating what's going on :)

src/appleseed/renderer/modeling/camera/thinlenscamera.cpp
pjessesco: oops, my mistake. :(

src/appleseed/renderer/modeling/camera/thinlenscamera.cpp
pjessesco: Then how about delete contition for now? I can try to fix it later.

dictoon: I agree, let's do this in a separate PR.

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -240,50 +240,47 @@ double Camera::extract_f_stop() const
...
-            RENDERER_LOG_WARNING(
-                "while defining camera \"%s\": autofocus is enabled; \"focal_distance\" parameter "
-                "will be ignored.",
-                get_path().c_str());
-
-            autofocus_enabled = true;
             autofocus_target = m_params.get_required<Vector2d>("autofocus_target", DefaultAFTarget);
-            focal_distance = 0.0;
-        }
+
```
dictoon: Remove this blank line, the `else` block goes with the `if` block.

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -240,50 +240,47 @@ double Camera::extract_f_stop() const
...
     else
     {
-        RENDERER_LOG_ERROR(
-            "while defining camera \"%s\": no \"focal_distance\" or \"autofocus_target\" parameter found; "
-            "using default focal distance value \"%f\".",
-            get_path().c_str(),
-            DefaultFocalDistance);
+        if (has_param("focal_distance"))
+            focal_distance = m_params.get_required<double>("focal_distance", DefaultFocalDistance);
+
```
dictoon: Same here.

src/appleseed/renderer/modeling/camera/thinlenscamera.cpp
```diff
@@ -750,21 +750,15 @@ DictionaryArray ThinLensCameraFactory::get_input_metadata() const
             .insert("name", "focal_distance")
             .insert("label", "Focal Distance")
             .insert("type", "text")
-            .insert("use", "optional")
-            .insert("visible_if",
-                Dictionary()
-                    .insert("autofocus_enabled", "false")));
+            .insert("use", "optional"));
```
pjessesco: Maybe focal_distance can have default value, compared with `autofocus_target`?

dictoon: I was thinking about that. The default value would be empty, correct? I think empty is the default value for default values :)

pjessesco: Sorry, I can't understand what you saying. Poor English :(

dictoon: No problem, your English is fine.

You said:

> Maybe focal_distance can have default value, compared with autofocus_target?

Maybe I misunderstood you. What did you have in mind?

pjessesco: Oh I just wondering why `focal_distance` has no default value, not like `autofocus_target`.

dictoon: Ok, that's what I understood.

The default value of `focal_distance` should be `""` (empty string), do you agree? In that case, the empty string is already the default value (I think) so there's no need to specify it. But you can add it if you want, maybe it would make things clearer.

pjessesco: In my understand, both `m_focal_distance` and `m_autofocus_target` is set to some value, whether they are defined or not in edit window (in `extract_focal_distance()`), and I guess `.insert("default", "0.0")` make text that shows in edit window, when they are not defined in project file.

..So what I intend is, as you said, I'd like insert default value same with `DefaultFocalDistance`. Then I'll commit it soon.

dictoon: Oh, I think I understand what you mean. Ok then, go ahead with adding a default value for `focal_distance`. 1.0 is a good default value.

dictoon:
Looks good! I'll merge this, and you can move the functions and investigate/fix the `visible_if` thing in a separate PR. Is that ok?

pjessesco:
Ok, first I'll try to move functions.

Issue #1815, opened 2018-01-27
Add diagnostic AOV for invalid samples


luisbarrancos:
Sometimes, specially with bigger shading networks and bigger scenes, it can be quite tricky to find out which shader(s) is generating NaN/Inf/invalid samples.
An diagnostic pass for this explicit purpose would help tremendously in pinpointing the misbehaving culprit.

oktomus:
In the final exported AOVs, how do you know which shader is failling ? Do we export an AOv per shader ?

dictoon:
@oktomus We don't export an AOV per shader. There is no easy way to know which shader caused invalid samples, just having a map (an AOV) that shows in which pixels did NaN/inf/negative samples occur would be enough I think. @luisbarrancos can you confirm this?

luisbarrancos:
That is correct, just having fast visual feedback for inf/NaNs would
suffice.

luisbarrancos:
Currently we have a object ID shader. What do you think about using this as a base or something similar, but rescaled so that it's within [0, 0.25] or [0,0.33], in order to provide some measure of background on which to overlap the invalid samples? This would give immediate feedback on which invalid samples were occurring, and from here we can go to any DCC app and inspect the shader networks (and resp. shaders) associated with it. Does something like this makes sense?

dictoon:
It's an interesting idea yes, so the original render in the background (darkened) and invalid samples marked in e.g. bright pink on top?

oktomus:
Why adding the original render in the background ? We can save both the rendered image and the diagnostic AOV.

dictoon:
I agree and I'm slightly split about this, but I guess it's out of convenience: with a single image you can see where there are invalid samples in the render, you don't have to overlay two images (original render + diagnostic AOV). I'm fine with either solution: one is more convenient, the other is more orthogonal.

luisbarrancos:
It was for convenience really, but then again, we can just toggle the aov
back to object id, or beauty pass, back to invalid samples. Is there
anything else we could include in such a samples AOV that would warrant
further attention? I.e., normalized samples budget used on G channel,
invalid samples on R channel? Or any other diagnostic quantity of use?

oktomus:
I have another question @luisbarrancos, how do you get the shader network from only the object id? Sorry if it sounds stupid, but I'm wondering..

luisbarrancos:
[...]

PR #1888, opened 2018-03-06
Create a diagnostic AOV for invalid samples

oktomus:
This implements #1815.

### DIFF
--- a/src/appleseed/renderer/kernel/rendering/final/adaptivepixelrenderer.cpp
+++ b/src/appleseed/renderer/kernel/rendering/final/adaptivepixelrenderer.cpp
@@ -86,10 +86,11 @@ namespace
             ISampleRendererFactory*     factory,
             const ParamArray&           params,
             const size_t                thread_index)
-          : m_params(params)
+          : PixelRendererBase(frame, thread_index, params)
+          , m_params(params)
           , m_sample_renderer(factory->create(thread_index))
         {
-            if (m_params.m_diagnostics)
+            if (are_diagnostics_enabled())
             {
                 m_variation_aov_index = frame.create_extra_aov_image("variation");
                 m_samples_aov_index = frame.create_extra_aov_image("samples");
@@ -116,7 +117,7 @@ namespace
                 m_params.m_min_samples,
                 m_params.m_max_samples,
                 m_params.m_max_variation,
-                m_params.m_diagnostics ? "on" : "off");
+                are_diagnostics_enabled() ? "on" : "off");
         }
 
         void release() override
@@ -125,10 +126,12 @@ namespace
         }
 
         void on_tile_begin(
-            const Frame&                frame,
-            Tile&                       tile,
-            TileStack&                  aov_tiles) override
+            const Frame&            frame,
+            Tile&                   tile,
+            TileStack&              aov_tiles) override
         {
+            PixelRendererBase::on_tile_begin(frame, tile, aov_tiles);
+
             m_scratch_fb_half_width = truncate<int>(ceil(frame.get_filter().get_xradius()));
             m_scratch_fb_half_height = truncate<int>(ceil(frame.get_filter().get_yradius()));
 
@@ -139,16 +142,21 @@ namespace
                     frame.aov_images().size(),
                     frame.get_filter()));
 
-            if (m_params.m_diagnostics)
-                m_diagnostics.reset(new Tile(tile.get_width(), tile.get_height(), 2, PixelFormatFloat));
+            if (are_diagnostics_enabled())
+            {
+                m_diagnostics.reset(new Tile(
+                    tile.get_width(), tile.get_height(), 2, PixelFormatFloat));
+            }
         }
 
         void on_tile_end(
-            const Frame&                frame,
-            Tile&                       tile,
-            TileStack&                  aov_tiles) override
+            const Frame&            frame,
+            Tile&                   tile,
+            TileStack&              aov_tiles) override
         {
-            if (m_params.m_diagnostics)
+            PixelRendererBase::on_tile_end(frame, tile, aov_tiles);
+
+            if (are_diagnostics_enabled())
             {
                 const size_t width = tile.get_width();
                 const size_t height = tile.get_height();
@@ -183,7 +191,7 @@ namespace
         {
             const size_t aov_count = frame.aov_images().size();
 
-            on_pixel_begin(pi, aov_accumulators);
+            on_pixel_begin(pi, pt, tile_bbox, aov_accumulators);
 
             m_scratch_fb->clear();
 
@@ -285,7 +293,7 @@ namespace
             }
 
             // Store diagnostics values in the diagnostics tile.
-            if (m_params.m_diagnostics && tile_bbox.contains(pt))
+            if (are_diagnostics_enabled() && tile_bbox.contains(pt))
             {
                 Color<float, 2> values;
 
@@ -309,7 +317,7 @@ namespace
                 m_diagnostics->set_pixel(pt.x, pt.y, values);
             }
 
-            on_pixel_end(pi, aov_accumulators);
+            on_pixel_end(pi, pt, tile_bbox, aov_accumulators);
         }
 
         StatisticsVector get_statistics() const override
@@ -329,14 +337,12 @@ namespace
             const size_t                    m_min_samples;
             const size_t                    m_max_samples;
             const float                     m_max_variation;
-            const bool                      m_diagnostics;
 
             explicit Parameters(const ParamArray& params)
               : m_sampling_mode(get_sampling_context_mode(params))
               , m_min_samples(params.get_required<size_t>("min_samples", 16))
               , m_max_samples(params.get_required<size_t>("max_samples", 256))
               , m_max_variation(pow(10.0f, -params.get_optional<float>("quality", 2.0f)))
-              , m_diagnostics(params.get_optional<bool>("enable_diagnostics", false))
             {
             }
         };
@@ -391,7 +397,7 @@ IPixelRenderer* AdaptivePixelRendererFactory::create(
 
 Dictionary AdaptivePixelRendererFactory::get_params_metadata()
 {
-    Dictionary metadata;
+    Dictionary metadata = PixelRendererBaseFactory::get_params_metadata();
 
     metadata.dictionaries().insert(
         "min_samples",
@@ -416,16 +422,6 @@ Dictionary AdaptivePixelRendererFactory::get_params_metadata()
             .insert("default", "2.0")
             .insert("label", "Quality")
             .insert("help", "Quality factor"));
-
-    metadata.dictionaries().insert(
-        "enable_diagnostics",
-        Dictionary()
-            .insert("type", "bool")
-            .insert("default", "false")
-            .insert("label", "Enable Diagnostics")
-            .insert(
-                "help",
-                "Enable adaptive sampling diagnostics"));
 
     return metadata;
 }
--- a/src/appleseed/renderer/kernel/rendering/final/adaptivepixelrenderer.h
+++ b/src/appleseed/renderer/kernel/rendering/final/adaptivepixelrenderer.h
@@ -31,7 +31,7 @@
 #define APPLESEED_RENDERER_KERNEL_RENDERING_FINAL_ADAPTIVEPIXELRENDERER_H
 
 // appleseed.renderer headers.
-#include "renderer/kernel/rendering/ipixelrenderer.h"
+#include "renderer/kernel/rendering/pixelrendererbase.h"
 #include "renderer/utility/paramarray.h"
 
 // appleseed.foundation headers.
@@ -53,7 +53,7 @@ namespace renderer
 //
 
 class AdaptivePixelRendererFactory
-  : public IPixelRendererFactory
+  : public PixelRendererBaseFactory
 {
   public:
     // Constructor.
--- a/src/appleseed/renderer/kernel/rendering/final/uniformpixelrenderer.cpp
+++ b/src/appleseed/renderer/kernel/rendering/final/uniformpixelrenderer.cpp
@@ -59,12 +59,7 @@
 // Standard headers.
 #include <cmath>
 
-// Forward declarations.
-namespace foundation    { class Tile; }
-namespace renderer      { class TileStack; }
-
 using namespace foundation;
-using namespace std;
 
 namespace renderer
 {
@@ -80,10 +75,12 @@ namespace
     {
       public:
         UniformPixelRenderer(
+            const Frame&                frame,
             ISampleRendererFactory*     factory,
             const ParamArray&           params,
             const size_t                thread_index)
-          : m_params(params)
+          : PixelRendererBase(frame, thread_index, params)
+          , m_params(params)
           , m_sample_renderer(factory->create(thread_index))
           , m_sample_count(m_params.m_samples)
           , m_sqrt_sample_count(round<int>(sqrt(static_cast<double>(m_params.m_samples))))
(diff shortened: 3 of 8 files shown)


src/appleseed/renderer/kernel/rendering/final/adaptivepixelrenderer.cpp
```diff
@@ -73,6 +73,10 @@ namespace renderer
 
 namespace
 {
+
```
oktomus: I will remove this line

src/appleseed/renderer/kernel/rendering/pixelrendererbase.h
```diff
@@ -32,20 +32,26 @@
...
 #include <cstddef>
+#include <memory>
 
 // Forward declarations.
 namespace foundation    { class Tile; }
 namespace renderer      { class AOVAccumulatorContainer; }
 namespace renderer      { class Frame; }
 namespace renderer      { class TileStack; }
 
+using namespace foundation;
```
dictoon: We completely forbid `using` declarations in header files since this defeats the point of namespaces entirely. Please use fully qualified names.

src/appleseed/renderer/kernel/rendering/pixelrendererbase.h
```diff
@@ -32,20 +32,26 @@
 
 // appleseed.renderer headers.
 #include "renderer/kernel/rendering/ipixelrenderer.h"
+#include "renderer/modeling/frame/frame.h"
 
 // appleseed.foundation headers.
+#include "foundation/image/tile.h"
```
dictoon: I presume you need this for `std::unique_ptr<>`? If you do, then please remove the `Tile` forward declaration below.

src/appleseed/renderer/kernel/rendering/pixelrendererbase.h
```diff
@@ -32,20 +32,26 @@
 
 // appleseed.renderer headers.
 #include "renderer/kernel/rendering/ipixelrenderer.h"
+#include "renderer/modeling/frame/frame.h"
```
dictoon: There's already a `Frame` forward declaration so this seems redundant.

src/appleseed/renderer/kernel/rendering/pixelrendererbase.h
```diff
@@ -60,6 +66,8 @@ class PixelRendererBase
     // Constructor.
```
dictoon: Change to plural now that there are two constructors.

src/appleseed/renderer/kernel/rendering/pixelrendererbase.h
```diff
@@ -60,6 +66,8 @@ class PixelRendererBase
...
     PixelRendererBase();
```
dictoon: Remove this blank line so that the comment applies to both constructors.

src/appleseed/renderer/kernel/rendering/pixelrendererbase.cpp
```diff
@@ -51,22 +54,66 @@ PixelRendererBase::PixelRendererBase()
 {
 }
 
+PixelRendererBase::PixelRendererBase(
+    const Frame&        frame,
+    const size_t        thread_index)
+  : m_invalid_pixel_count(0)
+{
+    m_invalid_sample_aov_index = frame.create_extra_aov_image("invalid_samples");
```
dictoon: The invalid samples AOV should be optional (since most of the time we don't need it, but it uses significant amounts of memory).

src/appleseed/renderer/kernel/rendering/pixelrendererbase.cpp
```diff
@@ -51,22 +54,66 @@ PixelRendererBase::PixelRendererBase()
...
+
+    if ((thread_index == 0) && (m_invalid_sample_aov_index == size_t(~0)))
+    {
+        RENDERER_LOG_WARNING(
```
dictoon: *invalid samples diagnostic* -> *invalid samples aov*

src/appleseed/renderer/kernel/rendering/pixelrendererbase.cpp
```diff
@@ -51,22 +54,66 @@ PixelRendererBase::PixelRendererBase()
...
+                Color<uint8, 1> sample_state;
+                m_invalid_sample_diagnostic->get_pixel(x, y, sample_state);
+
+                Color4f color(1.0f, 0.0f, 1.0f, 1.0f);
+
+                if (sample_state[0] == CorrectSampleHint)
+                {
+                    tile.get_pixel(x, y, color);
+                    // 20% of luminance
+                    color.rgb().set(((color.r + color.g + color.b) / 3.0f) * 0.2f);
```
dictoon: You can use `luminance(color)` which takes into account human perception of colors.

src/appleseed/renderer/kernel/rendering/pixelrendererbase.cpp
```diff
@@ -41,6 +42,8 @@ using namespace foundation;
 
 namespace renderer
 {
+const uint8 InvalidSampleHint = 1;
```
dictoon: Move after the "PixelRendererBase class implementation" block comment since this is technically part of the implementation of `PixelRendererBase`.

src/appleseed/renderer/kernel/rendering/final/uniformpixelrenderer.h
```diff
@@ -32,6 +32,7 @@
 
 // appleseed.renderer headers.
 #include "renderer/kernel/rendering/ipixelrenderer.h"
+#include "renderer/modeling/frame/frame.h"
```
dictoon: Prefer forward decls whenever possible (to reduce coupling and compile times).

src/appleseed/renderer/kernel/rendering/final/uniformpixelrenderer.cpp
```diff
@@ -181,7 +183,7 @@ namespace
                     m_total_sampling_dim.insert(child_sampling_context.get_total_dimension());
 
                     // Merge the sample into the framebuffer.
-                    if (shading_result.is_valid())
+                    if (shading_result.is_valid() && shading_result.m_main.r < 0.8f)
```
dictoon: Shouldn't be part of the PR :)

src/appleseed/renderer/kernel/rendering/final/adaptivepixelrenderer.cpp
```diff
@@ -148,6 +150,7 @@ namespace
             Tile&                       tile,
             TileStack&                  aov_tiles) override
         {
+            PixelRendererBase::on_tile_end(frame, tile, aov_tiles);
```
dictoon: I'd suggest adding a blank line after this to make it stand out.

src/appleseed/renderer/kernel/rendering/final/adaptivepixelrenderer.cpp
```diff
@@ -129,6 +130,7 @@ namespace
             Tile&                       tile,
             TileStack&                  aov_tiles) override
         {
+            PixelRendererBase::on_tile_begin(frame, tile, aov_tiles);
```
dictoon: I'd suggest adding a blank line after this to make it stand out.

src/appleseed/renderer/kernel/rendering/pixelrendererbase.h
oktomus: That makes sense, we should change that also on the adaptive pixel renderer

src/appleseed/renderer/kernel/rendering/pixelrendererbase.h
oktomus: Yes, my bad

src/appleseed/renderer/kernel/rendering/pixelrendererbase.cpp
oktomus: @dictoon Is this possible to share some Parameters along pixel renderer base and sub classes ?

src/appleseed/renderer/kernel/rendering/final/uniformpixelrenderer.cpp
oktomus: Woops, that's weird, I used `--patch` `to add and was careful.

oktomus:
Thanks for the review @dictoon. Regarding different colors for different kind of errors, I was thinking of:
- Create another ShadingResult::is_valid so that we can get a error code (an enum telling what is the error)
- Don't extend the original one to avoid useless tests when we don't want the AOV

src/appleseed/renderer/kernel/rendering/pixelrendererbase.cpp
oktomus: Or, I can use Frame::save_extra_aovs ? And we could do the same in the adaptive renderer ? 
Because `use diagnostics` and `save extra aovs` work only if they are both enabled.

dictoon: I think this coupling would be difficult to understand. In fact, do we agree that we have the same issue with the adaptive pixel sampler's diagnostic AOVs? Even if they are enabled, they aren't save to disk unless Save Extra AOVs is checked in the frame? If so then let's do the same for the invalid samples AOV, for now. (By the way, a checkbox must be added to the render settings to let users enable the invalid sample AOV.)

dictoon:
> Create another ShadingResult::is_valid so that we can get a error code (an enum telling what is the error)

That sounds good, my only concerns are about complexity and especially performance. Let's try it (in a separate PR) and see how it goes!

src/appleseed/renderer/kernel/rendering/pixelrendererbase.cpp
oktomus: Ok I'll do that !

src/appleseed/renderer/kernel/rendering/pixelrendererbase.h
```diff
@@ -58,7 +62,7 @@ class PixelRendererBase
 {
   public:
     // Constructor.
-    PixelRendererBase();
+    PixelRendererBase(const Frame& frame, const size_t thread_index, const ParamArray& params);
```
dictoon: You can make this more readable by having one parameter per line.

src/appleseed/renderer/kernel/rendering/pixelrendererbase.h
```diff
@@ -74,18 +78,48 @@ class PixelRendererBase
...
+
+    size_t                                  m_invalid_sample_count;
+    size_t                                  m_invalid_pixel_count;
+    size_t                                  m_invalid_sample_aov_index;
+    std::unique_ptr<foundation::Tile>       m_invalid_sample_diagnostic;
+
+  protected:
+    const Parameters                m_base_params;
+};
+
```
dictoon: Add a second blank line before block comments. That's the only case where we use two blank lines.

src/appleseed/renderer/kernel/rendering/pixelrendererbase.h
```diff
@@ -74,18 +78,48 @@ class PixelRendererBase
...
+        AOVAccumulatorContainer&            aov_accumulators);
 
     void signal_invalid_sample();
 
   private:
-    size_t                          m_invalid_sample_count;
-    size_t                          m_invalid_pixel_count;
+    struct Parameters
+    {
+        const bool                  m_diagnostics;
```
dictoon: It would look _nicer_ to forget about trying to align this to anything. There's no hard rule here, we're clearly in the domain of aesthetics :)

src/appleseed/renderer/kernel/rendering/final/adaptivepixelrenderer.cpp
```diff
@@ -86,10 +83,11 @@ namespace
             ISampleRendererFactory*     factory,
             const ParamArray&           params,
             const size_t                thread_index)
-          : m_params(params)
+          : PixelRendererBase(frame, thread_index, params)
+          , m_params(params)
           , m_sample_renderer(factory->create(thread_index))
         {
-            if (m_params.m_diagnostics)
+            if (m_base_params.m_diagnostics)
```
dictoon: It's quite problematic to have parameters in two places, `m_base_params` and `m_params`. I would suggest simply keeping `m_base_params` in the base class, and renaming it to `m_params`.

src/appleseed/renderer/kernel/rendering/final/adaptivepixelrenderer.cpp
```diff
@@ -126,11 +124,13 @@ namespace
 
         void on_tile_begin(
             const Frame&                frame,
-            Tile&                       tile,
+            foundation::Tile&           tile,
```
dictoon: You're in the .cpp file and there's a `using namespace foundation` statement so you can remove the qualification here (and everywhere in the file).

src/appleseed/renderer/kernel/rendering/final/adaptivepixelrenderer.cpp
```diff
@@ -139,16 +139,19 @@ namespace
                     frame.aov_images().size(),
                     frame.get_filter()));
 
-            if (m_params.m_diagnostics)
-                m_diagnostics.reset(new Tile(tile.get_width(), tile.get_height(), 2, PixelFormatFloat));
+            if (m_base_params.m_diagnostics)
```
dictoon: Wrap the body of multiline `if` statements with braces:

    if (m_base_params.m_diagnostics)
    {
        ...
    }

src/appleseed/renderer/kernel/rendering/pixelrendererbase.cpp
```diff
@@ -33,48 +33,103 @@
...
+                m_invalid_sample_diagnostic->get_pixel(x, y, sample_state);
+
+                foundation::Color4f color(1.0f, 0.0f, 1.0f, 1.0f);
+
+                if (sample_state[0] == CorrectSampleHint)
+                {
+                    tile.get_pixel(x, y, color);
+                    // 20% of luminance
+                    color.rgb().set(foundation::luminance(color.rgb()));
+                    color.rgb().set(color.r * 0.2f);
```
dictoon: That looks a bit weird. Why not simply

    color.rgb().set(0.2f * luminance(color.rgb()));

src/appleseed/renderer/kernel/rendering/pixelrendererbase.cpp
```diff
@@ -107,4 +176,25 @@ void PixelRendererBase::signal_invalid_sample()
     ++m_invalid_sample_count;
 }
 
+//
```
dictoon: Add another blank line here too.

src/appleseed/renderer/kernel/rendering/pixelrendererbase.cpp
```diff
@@ -99,6 +157,20 @@ void PixelRendererBase::on_pixel_end(
         {
             RENDERER_LOG_WARNING("more invalid samples found, omitting warning messages for brevity.");
         }
+
+    }
+
+    // Invalid samples diagnostic
+    if (tile_bbox.contains(pt) && m_params.m_diagnostics)
```
dictoon: It would make sense to swap the conditions to avoid calling `AABB::contains()` if `m_params.m_diagnostics` is false.

src/appleseed/renderer/kernel/rendering/final/uniformpixelrenderer.h
```diff
@@ -52,11 +52,12 @@ namespace renderer
 //
 
 class UniformPixelRendererFactory
-  : public IPixelRendererFactory
+  : public PixelRendererBaseFactory
 {
   public:
     // Constructor.
     UniformPixelRendererFactory(
+        const Frame&                frame,
```
dictoon: Need to forward declare `Frame`.

src/appleseed/renderer/kernel/rendering/final/adaptivepixelrenderer.h
```diff
@@ -41,8 +41,7 @@
 #include <cstddef>
 
 // Forward declarations.
-namespace foundation    { class Dictionary; }
-namespace renderer      { class Frame; }
```
dictoon: You need to restore the forward decl of `Frame`.

src/appleseed/renderer/kernel/rendering/final/adaptivepixelrenderer.cpp
```diff
@@ -139,16 +142,19 @@ namespace
                     frame.aov_images().size(),
                     frame.get_filter()));
 
-            if (m_params.m_diagnostics)
-                m_diagnostics.reset(new Tile(tile.get_width(), tile.get_height(), 2, PixelFormatFloat));
+            if (PixelRendererBase::m_params.m_diagnostics)
```
dictoon: Add braces around multiline body.

oktomus:
Thanks, I'm happy to know that my code is of quality :smile:

Issue #1848, opened 2018-02-17
Change default number of diffuse bounces to 3 in appleseed.studio


dictoon:

abhayjindal51:
Hello @dictoon , 
I would like to work on this issue

rgirish28:
So do you want the maximum diffuse bounces to be set to 3 as default in render settings? Just making sure. Thanks.

dictoon:
That's correct. There are unfortunately a couple places where the default value appears: in the render settings, but also when retrieving max bounce parameters from a render configuration.

rgirish28:
This issue can be closed now :)

dictoon:
Let's keep it open until the default is _indeed_ changed to 3; right now the default is still unlimited, it only becomes 3 (instead of 8) when unlimited is unchecked.

rgirish28:
Default changed for real! #1891

juancarlosgzrz:
Done!

dictoon:
It doesn't appear to be working as expected, default is still unlimited:

![image](https://user-images.githubusercontent.com/321290/37553114-a16b4524-29c1-11e8-9a5f-c7a09e91a220.png)

dictoon:
It's not done:

![image](https://user-images.githubusercontent.com/321290/37553126-d573d534-29c1-11e8-8510-6c1b27fc5111.png)

It will be done when https://github.com/appleseedhq/appleseed/pull/1891 is merged.

PR #1889, opened 2018-03-06
Change default max_diffuse_bounces to 3

rgirish28:
Changed default max_diffuse_bounces to 3 as required in #1848 
I'm not sure if I have found every instance of where the default value is set. 
Would you want the default setting to be unlimited bounces or just 3 bounces? I can make the change to the commit if that's the case.

### DIFF
--- a/src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
+++ b/src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
@@ -790,13 +790,19 @@ namespace
             const bool              allow_unlimited = true)
         {
             const int DefaultMaxBounces = 8;
+            const int DefaultMaxDiffuseBounces = 3;
 
             const int max_bounces =
                 get_config<int>(config, construct_bounce_param_path(bounce_type), -1);
 
+            const string widget_max_bounce_key = widget_key_prefix + ".bounces.max_" + bounce_type + "_bounces";
+
             if (allow_unlimited)
                 set_widget(widget_key_prefix + ".bounces.unlimited_" + bounce_type + "_bounces", max_bounces == -1);
-            set_widget(widget_key_prefix + ".bounces.max_" + bounce_type + "_bounces", max_bounces == -1 ? DefaultMaxBounces : max_bounces);
+            if (bounce_type == "diffuse") 
+                set_widget(widget_max_bounce_key, max_bounces == -1 ? DefaultMaxDiffuseBounces : max_bounces);
+            else
+                set_widget(widget_max_bounce_key, max_bounces == -1 ? DefaultMaxBounces : max_bounces);
         }
 
         void save_separate_bounce_settings(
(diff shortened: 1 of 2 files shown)


src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
```diff
@@ -790,14 +790,18 @@ namespace
...
 
             const int max_bounces =
                 get_config<int>(config, construct_bounce_param_path(bounce_type), -1);
 
             if (allow_unlimited)
                 set_widget(widget_key_prefix + ".bounces.unlimited_" + bounce_type + "_bounces", max_bounces == -1);
-            set_widget(widget_key_prefix + ".bounces.max_" + bounce_type + "_bounces", max_bounces == -1 ? DefaultMaxBounces : max_bounces);
-        }
+            if (bounce_type == "diffuse")
+                set_widget(widget_key_prefix + ".bounces.max_" + bounce_type + "_bounces", max_bounces == -1 ? DefaultMaxDiffuseBounces : max_bounces);
```
dictoon: I would suggest factorizing `widget_key_prefix + ".bounces.max_" + bounce_type + "_bounces"`.

rgirish28:
Made the changes and will update the rest in another PR. Thanks!

src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
```diff
@@ -790,13 +790,27 @@ namespace
...
-
+	    
+            const string widget_max_bounce_key = widget_key_prefix + ".bounces.max_" + bounce_type + "_bounces";
+            const string widget_unlimited_max_bounce_key = widget_key_prefix + ".bounces.unlimited_" + bounce_type + "_bounces";
+            
             if (allow_unlimited)
-                set_widget(widget_key_prefix + ".bounces.unlimited_" + bounce_type + "_bounces", max_bounces == -1);
-            set_widget(widget_key_prefix + ".bounces.max_" + bounce_type + "_bounces", max_bounces == -1 ? DefaultMaxBounces : max_bounces);
+            {   
+                if (bounce_type == "diffuse")
```
dictoon: I think we misunderstood each others. The logic should not change, it was right before (before your PR, and in your initial commit).

src/appleseed/renderer/kernel/lighting/pt/ptlightingengine.cpp
```diff
@@ -1043,8 +1043,8 @@ Dictionary PTLightingEngineFactory::get_params_metadata()
         "max_diffuse_bounces",
         Dictionary()
             .insert("type", "int")
-            .insert("default", "8")
-            .insert("unlimited", "true")
+            .insert("default", "3")
+            .insert("unlimited", "false")
```
dictoon: That should not change as well.

That said, I've just realized that the `unlimited` property is completely unused to I'll remove it altogether.

dictoon: To clarify: what it's supposed to mean that the property's value _can_ be unlimited. The name is less than ideal, I agree.

rgirish28:
I have made the required changes and will address the next phase in another PR. Thanks.

dictoon:
Thank you!

PR #1891, opened 2018-03-08
Set all max bounce types to limited values and increase project revision

### LINKED ISSUE #1848: Change default number of diffuse bounces to 3 in appleseed.studio

rgirish28:
Fixes #1848 by setting all max_bounces to limited default values (8 in every case except max_diffuse_bounces, which is 3) and increases the project file revision to 23. I also changed the file revision to 23 in the projectfilereader unit test input as it was failing the identical assertion.


sandbox/tests/unit tests/inputs/test_projectfilereader_configurationblocks.appleseed
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project format_revision="22">
+<project format_revision="23">
```
dictoon: You are two GSoC candidates adding a new project file migration step, they will conflict. Let's see who gets his PR merged first :)

src/appleseed/renderer/modeling/project/projectfileupdater.cpp
```diff
@@ -1630,6 +1630,45 @@ namespace
             }
         }  
     };
+
+    //
+    // Update from revision 22 to revision 23.
```
dictoon: Excellent!

src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
```diff
@@ -798,7 +798,7 @@ namespace
             const int DefaultMaxDiffuseBounces = 3;
 
             const int max_bounces =
-                get_config<int>(config, construct_bounce_param_path(bounce_type), -1);
+                get_config<int>(config, construct_bounce_param_path(bounce_type), bounce_type == "diffuse" ? DefaultMaxDiffuseBounces : DefaultMaxBounces);
```
dictoon: While I _think_ this is correct, this function is getting really complicated and hard to follow.

What about adding a default value parameter to `load_separate_bounce_settings()`? That way checking the bounce type and deciding what default to use based on it would no longer be necessary.

src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
```diff
@@ -769,7 +769,7 @@ namespace
         {
             const int DefaultMaxBounces = 8;
 
-            const int max_bounces = get_config<int>(config, param_path, -1);
+            const int max_bounces = get_config<int>(config, param_path, DefaultMaxBounces);
```
dictoon: To clarify things further, I would suggest to:

- Rename `load_bounce_settings()` to `load_global_max_bounce_settings()`.
- Pass the default value in argument to the function.

What do you think?

rgirish28: Yes I agree. I think that way it is also more clear what default value is being set in load_separate_bounce_settings as well. I'll make the necessary changes and commit again.

rgirish28: The only issue is that load_bounce_settings is used  for photon mapping as well. So, I am going to change it to load_max_bounce_settings as global won't be applicable for photon mapping.

rgirish28:
> The only issue is that load_bounce_settings is used for photon mapping as well. So, I am going to change it to load_max_bounce_settings as global won't be applicable for photon mapping. I hope that is alright.

So as explained above, I've not changed it to global_max_bounce, but I've added a default argument and the flow of code is much better now. Thank you for such positive feedback!

dictoon:
> The only issue is that load_bounce_settings is used for photon mapping as well. So, I am going to change it to load_max_bounce_settings as global won't be applicable for photon mapping.

Actually I think "global" is also applicable to the photon mapping case (since at the moment we only expose a global max bounces setting in the photon mapper).

rgirish28:
Ah, actually you're right. I just thought because the parameter is named max_bounces, the function should still be called max_bounces, but logically it refers to global bounces in both cases as you said. I've made the corrections. I hope we can merge before the other PR leads to a conflict :)

rgirish28:
So, the issue with the code and the reason it was written like that before was because there are two possible scenarios: 

- Default value is unlimited: Then we need to check the box as well as set the default value for the text box. This is the reason why the photon mapping default value has changed, because now we are using just one default value to set two values. So, I suggest keeping the current function definition and going back to the previous logic. That way it is just as complex as it was before, but with an extra parameter. 
- Default value is limited: Then logic in current version of PR works. 

I hope I'm clear. I'll make the changes in another commit so that you can understand what I'm talking about and we can squash them together later. Thanks.

rgirish28:
I've made the changes in a new commit. It's back to the old logic and the merge conflicts have been resolved. Next time, I'll skip lunch and socialising to get ahead in the PR conflict game ;) :)

src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
```diff
@@ -762,14 +762,15 @@ namespace
...
             const Configuration&    config,
             const string&           widget_key_prefix,
-            const string&           param_path)
+            const string&           param_path,
+            const int               default_max_bounces)
         {
             const int DefaultMaxBounces = 8;
-
-            const int max_bounces = get_config<int>(config, param_path, -1);
+
```
dictoon: Remove trailing spaces.

dictoon:
> Next time, I'll skip lunch and socialising to get ahead in the PR conflict game ;)

Ahah, tough choice! Lunch and socializing still seems a better option though :)

rgirish28:
I think that should fix every issue. Sorry about the delay but I had to spend the whole of last week writing a paper. Back to PR duty now :)

dictoon:
The changes look good, thanks! All you need to do now is rebase your PR because as is, GitHub says it's unmergable.

> Sorry about the delay but I had to spend the whole of last week writing a paper. Back to PR duty now

No worries at all!

rgirish28:
Rebased and should be alright for merging.

dictoon:
Looks good indeed, will merge as soon as Travis passes.

rgirish28:
Sweet! Thanks!

PR #1893, opened 2018-03-10
Fix motion blur related crash in debug builds. Small tweaks

est77:
- Fix crash in debug builds when shutter time interval is zero.
- Enabled coating over glass in std surface.

### DIFF
--- a/src/appleseed/renderer/modeling/camera/camera.cpp
+++ b/src/appleseed/renderer/modeling/camera/camera.cpp
@@ -87,14 +87,20 @@ bool Camera::on_render_begin(
     check_shutter_times_for_consistency();
 
     m_shutter_open_time_interval = m_shutter_close_time - m_shutter_open_time;
-    m_normalized_open_end_time = inverse_lerp(m_shutter_open_time, m_shutter_close_time, m_shutter_open_end_time);
-    m_normalized_open_end_time_half = m_normalized_open_end_time / 2;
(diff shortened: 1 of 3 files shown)


src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -355,23 +361,19 @@ void Camera::initialize_ray(
...
-    if (m_shutter_open_time == m_shutter_close_time)
-    {
-        ray.m_time =
-            ShadingRay::Time::create_with_normalized_time(
-                0.0f,
-                m_shutter_open_time,
-                m_shutter_close_time);
-    }
-    else
+    float sample_time = 0.0f;
```
pjessesco: Small question : is it better make `sample_time` and pass it to `create_with_normalized_time()`, than pass `0.0` directly to function?

est77: Code-wise it is cleaner to have only one call instead of two which shared almost all the parameters.
I don't think there will be any significant difference in terms of speed.

pjessesco: Um.. my question was compare about `create_with_norm_time(0.0, open_time, close_time)` and `create_with_norm_time(sample_time, open_time, close_time)`, very minor.

And I think this variable can be `const`.

PR #1894, opened 2018-03-10
Show focal distance or autofocus target input form interactively.

pjessesco:
`autofocus_target` shows when `autofocus_enabled` is checked, and `focal_distance` shows when unchecked.


src/appleseed/renderer/modeling/camera/thinlenscamera.cpp
```diff
@@ -743,23 +743,30 @@ DictionaryArray ThinLensCameraFactory::get_input_metadata() const
             .insert("label", "Enable autofocus")
             .insert("type", "boolean")
             .insert("use", "optional")
-            .insert("default", "true"));
+            .insert("default", "true")
+            .insert("on_change", "rebuild_form"));
```
dictoon: Good catch! I had forgotten about this.

dictoon:
Thanks man!

PR #1902, opened 2018-03-11
Add max_color_balance shader, remove color and multiplier from color and float shaders

usakhelo:
I've added max_color_balance shader to support Output rollout of 3ds max bitmap textures.
Same shader also mixes textures and color inputs based on in_multiplier parameter.


usakhelo:
>Out of curiosity, where does the peculiar color mixing/balancing logic come from?
Color balance part comes from maya shader, and mixing part is an attempt to replicate how 3ds max's blending works.

usakhelo:
Comments fixed, I'll merge it when checks are complete.

Issue #1892, opened 2018-03-10
Fix cameras' print_settings()


pjessesco:
It should print `shutter_open_end` and `shutter_close_start`.

pjessesco:
@dictoon Isn't it related with other cameras?

dictoon:
You're completely right.

slongle:
Does it is just the change from "shutter open" to "shutter_open_end" and "shutter close" to "shutter close start"? 
If so, it need to change the code in src/appleseed/renderer/modeling/camera/

- "thinlencamera.cpp" at 558, 559 line
- "sphericalcamera.cpp" at 213, 214 line
- "orthographiccamera.cpp" at 253, 254 line
- "pinholecamera.cpp" at 287, 288 line

slongle:
You mean `shutter_open_end` is different from `shutter open`?

pjessesco:
Yes it is!

PR #1905, opened 2018-03-12
Fix cameras' print_settings()

slongle:
I add 2 extra lines to print `shutter open end` and `shutter close start` which are in src/appleseed/renderer/modeling/camera/

- "thinlencamera.cpp" at 560, 561 line
- "sphericalcamera.cpp" at 215, 216 line
- "orthographiccamera.cpp" at 255, 256 line
- "pinholecamera.cpp" at 289, 290 line
#1892

### DIFF
--- a/src/appleseed/renderer/modeling/camera/orthographiccamera.cpp
+++ b/src/appleseed/renderer/modeling/camera/orthographiccamera.cpp
@@ -251,14 +251,18 @@ namespace
                 "  film height                   %f\n"
                 "  near z                        %f\n"
                 "  shutter open                  %f\n"
-                "  shutter close                 %f",
+                "  shutter close                 %f\n"
+                "  shutter open end              %f\n"
+                "  shutter close start           %f",
                 get_path().c_str(),
                 Model,
                 m_film_dimensions[0],
                 m_film_dimensions[1],
                 m_near_z,
                 m_shutter_open_time,
-                m_shutter_close_time);
+                m_shutter_close_time,
+                m_shutter_open_end_time,
+                m_shutter_close_start_time);
         }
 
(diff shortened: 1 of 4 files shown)


src/appleseed/renderer/modeling/camera/orthographiccamera.cpp
```diff
@@ -251,14 +251,18 @@ namespace
                 "  film height                   %f\n"
                 "  near z                        %f\n"
                 "  shutter open                  %f\n"
-                "  shutter close                 %f",
+                "  shutter close                 %f\n"
```
pjessesco: Nice! But I think it will be better if `close` and `open_end` order changed, same with input form.

slongle: Really thanks for your patience!
I see it and should I also change `shutter open`  and `shutter close` to  `shutter open start` and `shutter close start` respectively?

pjessesco: It doesn't seem necessary for me, but I'm not sure.

src/appleseed/renderer/modeling/camera/thinlenscamera.cpp
```diff
@@ -570,7 +572,9 @@ namespace
                 m_diaphragm_tilt_angle,
                 m_near_z,
                 m_shutter_open_time,
-                m_shutter_close_time);
+                m_shutter_close_time,
```
pjessesco: I think you should change order here too.

pjessesco:
It seems good to me.

est77:
Thanks for the PR!

Issue #1882, opened 2018-03-04
Allow sending AOV tiles to stdout


dictoon:

oktomus:
Being able to print tiles boundaries while rendering or printing each tile rendered output ?

dictoon:
The goal here is to send the actual AOV tiles _pixels_ to stdout (configured to use binary format). Check the file appleseed.cli/stdouttilecallback.cpp.

oktomus:
Thanks, I will have a look!

dictoon:
Fantastic! The goal is to send AOV tiles to blenderseed, our Blender plugin. Feel free to discuss the Blender side of things with @jdent02 on #blenderseed.

PR #1906, opened 2018-03-13
Allow sending AOV tiles to stdout

oktomus:
Issue #1882 

Since there is a need to refactor a lot of things, I am doing this step by step.

I will need to update the protocol so that it supports AOVs. For that, I was thinking of doing as in `houdinitilecallbacks`.

Needs to be merged along with [ Implement protocol v2 decoding and reading #208 ](https://github.com/appleseedhq/blenderseed/pull/208)

### DIFF
--- a/src/appleseed.cli/stdouttilecallback.cpp
+++ b/src/appleseed.cli/stdouttilecallback.cpp
@@ -30,8 +30,8 @@
 #include "stdouttilecallback.h"
 
 // appleseed.renderer headers.
-#include "renderer/api/frame.h"
 #include "renderer/api/aov.h"
+#include "renderer/api/frame.h"
 
 // appleseed.foundation headers.
 #include "foundation/image/canvasproperties.h"
@@ -68,9 +68,9 @@ namespace
       : public TileCallbackBase
     {
       public:
-        explicit StdOutTileCallback(TileOutputOptions export_options)
+        explicit StdOutTileCallback(StdOutTileCallbackFactory::TileOutputOptions export_options)
           : m_header_sent(false)
-          , m_single_plane(export_options == BeautyOnly)
+          , m_export_options(export_options)
         {
         }
 
@@ -129,7 +129,7 @@ namespace
         boost::mutex m_mutex;
 
         bool m_header_sent;
-        const bool m_single_plane;
+        const StdOutTileCallbackFactory::TileOutputOptions m_export_options;
 
         void send_header(const Frame& frame)
         {
@@ -137,8 +137,9 @@ namespace
 
             // Build and write tiles header.
             // This header is sent only once and can contains AOVs and frame informations.
+            const bool beauty_only = (m_export_options == StdOutTileCallbackFactory::TileOutputOptions::BeautyOnly);
             const size_t chunk_size = 1 * sizeof(uint32);
-            const size_t plane_count = m_single_plane ? 1 : 1 + frame.aov_images().size();
+            const size_t plane_count = beauty_only ? 1 : 1 + frame.aov_images().size();
             const uint32 header[] =
             {
                 static_cast<uint32>(ChunkTypeTilesHeader),
@@ -149,7 +150,7 @@ namespace
 
             send_plane_definition(frame.image(), "beauty", 0);
 
-            if (!m_single_plane)
+            if (!beauty_only)
             {
                 for (size_t i = 0, e = frame.aov_images().size(); i < e; ++i)
                 {
@@ -199,7 +200,7 @@ namespace
             const size_t h = tile.get_height();
 
             // Build and write highlight tile header.
-            // This header is sent to allow highlighting tiles being renderer.
+            // This header is sent to allow highlighting tiles being rendered.
             const size_t chunk_size = 4 * sizeof(uint32);
             const uint32 header[] =
             {
@@ -229,7 +230,7 @@ namespace
                 tile_y,
                 0);
 
-            if (!m_single_plane)
+            if (m_export_options == StdOutTileCallbackFactory::TileOutputOptions::AllAOVs)
             {
                 // Send AOV tiles.
                 for (size_t i = 0, e = frame.aov_images().size(); i < e; ++i)
--- a/src/appleseed.cli/stdouttilecallback.cpp
+++ b/src/appleseed.cli/stdouttilecallback.cpp
@@ -67,6 +67,11 @@ namespace
       : public TileCallbackBase
     {
       public:
+        StdOutTileCallback(const bool single_plane = false)
+          : m_single_plane(single_plane)
+        {
+        }
+
         void release() override
         {
             // The factory always return the same tile callback instance.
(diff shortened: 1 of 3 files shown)


src/appleseed.cli/stdouttilecallback.cpp
```diff
@@ -124,14 +129,37 @@ namespace
...
+#ifdef _WIN32
+            _setmode(_fileno(stdout), old_stdout_mode);
+#endif
+        }
 
+      private:
+        // Do not change the values of the enumerators as this WILL break client compabitility.
+        enum ChunkType
+        {
+            ChunkTypeTileData       = 1,
```
fedormatantsev: Why don't you start with 0?

src/appleseed.cli/stdouttilecallback.cpp
```diff
@@ -162,21 +199,7 @@ namespace
             }
 
             fflush(stdout);
```
dictoon: That should be done outside of `send_tile()`.

src/appleseed.cli/stdouttilecallback.cpp
```diff
@@ -124,14 +129,37 @@ namespace
...
+        // Do not change the values of the enumerators as this WILL break client compabitility.
+        enum ChunkType
+        {
+            ChunkTypeTileData       = 1,
+            ChunkTypeTileHighlight  = 2
+        };
+
+        boost::mutex m_mutex;
+
+        const bool m_single_plane;
```
dictoon: It's not clear to me what this means and what's its purpose, but I suppose it'll become clear later?

oktomus: Yes, it's like in the houdini callback. When `true`, aovs would be sent in stdout

oktomus:
I implemented aov output from the cli. I used other version so that blenderseed keeps working with older versions of appleseed. 
Should I make aov output optional ? Or maybe we can do this later with another ticket.
Next step is to implement the blenderseed side.

src/appleseed.cli/stdouttilecallback.cpp
```diff
@@ -31,6 +31,7 @@
 
 // appleseed.renderer headers.
 #include "renderer/api/frame.h"
+#include "renderer/kernel/aov/imagestack.h"
```
dictoon: "External" applications such as appleseed.cli, appleseed.studio, etc. must only include API headers, in this case you want to include `renderer/api/aov.h`.

src/appleseed.cli/stdouttilecallback.cpp
```diff
@@ -67,6 +68,12 @@ namespace
       : public TileCallbackBase
     {
       public:
+        StdOutTileCallback(const bool single_plane = false)
```
dictoon: Mark the constructor as `explicit` to avoid implicit conversions from `bool`.

The code would be even clearer with two changes:
- Use an enum rather than a boolean parameter
- Remove the default value

src/appleseed.cli/stdouttilecallback.cpp
```diff
@@ -83,18 +90,116 @@ namespace
...
+            fwrite(header, sizeof(header), 1, stdout);
+
+            send_plane_definition(frame.image(), "beauty", 0);
+
+            if (!m_single_plane)
+            {
+                for (size_t i = 0, e = frame.aov_images().size(); i < e; ++i)
+                {
+                    send_plane_definition(
+                                frame.aov_images().get_image(i),
```
dictoon: Fix indentation.

src/appleseed.cli/stdouttilecallback.cpp
```diff
@@ -83,18 +90,116 @@ namespace
...
+                                frame.aov_images().get_name(i),
+                                i + 1);
+                }
+            }
+
+            m_header_sent = true;
+        }
 
+        void send_plane_definition(
+            const Image&            img,
```
dictoon: Align parameter names vertically with other methods.

src/appleseed.cli/stdouttilecallback.cpp
```diff
@@ -83,18 +90,116 @@ namespace
...
+            const char*             name,
+            const size_t            index) const
+        {
+            // Build and write aov header.
```
dictoon: *AOV*

src/appleseed.cli/stdouttilecallback.cpp
```diff
@@ -83,18 +90,116 @@ namespace
...
 
             // Retrieve the source tile and its dimensions.
-            const Tile& tile = frame->image().tile(tile_x, tile_y);
+            const Tile& tile = frame.image().tile(tile_x, tile_y);
             const size_t w = tile.get_width();
             const size_t h = tile.get_height();
 
-            // Build and write tile header.
+            // Build and write highlight tile header.
+            // This header is sent for allowing to displaying marks arround the tile being renderer.
```
dictoon: *This header is sent to allow highlighting tiles being rendered.*

src/appleseed.cli/stdouttilecallback.cpp
```diff
@@ -106,42 +211,62 @@ namespace
...
+            const size_t            tile_x,
+            const size_t            tile_y,
+            const size_t            plane_index) const
+        {
+            const size_t x = tile_x * properties.m_tile_width;
+            const size_t y = tile_y * properties.m_tile_height;
 
-            // Retrieve the source tile and its dimensions.
-            const Tile& tile = frame->image().tile(tile_x, tile_y);
+            // Retrieve the tile dimensions
```
dictoon: End sentence with a point.

src/appleseed.cli/stdouttilecallback.cpp
```diff
@@ -106,42 +211,62 @@ namespace
...
             const size_t w = tile.get_width();
             const size_t h = tile.get_height();
             const size_t c = tile.get_channel_count();
 
             // Build and write tile header.
-            const size_t chunk_size = 5 * sizeof(uint32) + w * h * c * sizeof(float);
+            // This header contains information about the tile aov that will be written.
```
dictoon: *AOV*

src/appleseed.cli/stdouttilecallback.cpp
dictoon: I'm confused about another thing: are there cases where `single_plane` is set to false? I can only see a single use of this constructor and `single_plane` is set to true.

oktomus: No there isn't any such case for the moment. I have to modify the factory. And also, find a way to make this optional from the cli.

oktomus:
Hey @dictoon, thanks for your review, and sorry for all my english mistakes !

dictoon:
Thanks for the PR, and no worries at all about the English mistakes, that's also what reviews are for :)

oktomus:
We need to create a ticked to expose the output options in the cli.

src/appleseed.cli/stdouttilecallback.cpp
```diff
@@ -199,7 +199,7 @@ namespace
             const size_t h = tile.get_height();
 
             // Build and write highlight tile header.
-            // This header is sent for allowing to displaying marks arround the tile being renderer.
+            // This header is sent to allow highlighting tiles being renderer.
```
dictoon: being *rendered* :)

src/appleseed.cli/stdouttilecallback.cpp
```diff
@@ -31,7 +31,7 @@
 
 // appleseed.renderer headers.
 #include "renderer/api/frame.h"
-#include "renderer/kernel/aov/imagestack.h"
+#include "renderer/api/aov.h"
```
dictoon: Order alphabetically.

src/appleseed.cli/stdouttilecallback.cpp
```diff
@@ -68,9 +68,9 @@ namespace
       : public TileCallbackBase
     {
       public:
-        StdOutTileCallback(const bool single_plane = false)
-          : m_single_plane(single_plane)
-          , m_header_sent(false)
+        explicit StdOutTileCallback(TileOutputOptions export_options)
+          : m_header_sent(false)
+          , m_single_plane(export_options == BeautyOnly)
```
dictoon: I'd go as far as storing export options directly instead of having `m_single_plane` that doesn't much to me.

src/appleseed.cli/stdouttilecallback.h
```diff
@@ -38,11 +38,17 @@
 namespace appleseed {
 namespace cli {
 
+enum TileOutputOptions
```
dictoon: Exactly. Can you just move it inside `StdOutTileCallbackFactory` and use C++11's scoped enum?

oktomus: Yes!

dictoon:
Thanks!

Issue #1733, opened 2017-11-26
Optimize plugins discovery


dictoon:
We're currently scanning all files in all search paths for every entity type. This is quite slow.

firedranzer:
@dictoon Is this issue still open? I want to take this issue.

dictoon:
It is!

firedranzer:
How can I proceed to solve this issue? I am relatively new to this organization.

dictoon:
The code to optimize is in `EntityFactoryRegistrar::register_factories_from_plugins()`, in file `appleseed/renderer/modeling/entity/registerentityfactories.h`.

PR #1907, opened 2018-03-13
new technique to search for plugins 

MohamedAli1995:
issue: Optimize plugins discovery #1733
brief about the proposed technique : 
1- store all plugins information ( entry_point, name,register_function)
2- search once in all paths and in each path search if  one of the plugins exists  , if so , then load it to the cache and call  the plugin's register_factory_function.


dictoon:
Something is off with your PR, it includes 4078 changed files! You'll need to clean things up and open a new one.

PR #1912, opened 2018-03-16
Optimization and refactoring

### LINKED ISSUE #1910: Switch to `= default` syntax
With VS 2015 (I haven't tried with VS 2017), the following code is problematic:

    class float4
    {
        float4() {}
        float4(float a) { ... }
        ~float4() {}
    };

    float4 texels[2][2];

The problem is that VS 2015 (and earlier versions, probably) doesn't realize, when exception handling is enabled, that the default constructor and the destructor do nothing and in particular don't throw. Because of that, it actually generates code to initialize all 2x2=4 `float4` objects in the line

    float4 texels[2][2];

Specifically, it generates a call to `__vec_ctor` and `__vec_dtor` which are compiler-generated functions, as described here:

https://www.geoffchappell.com/studies/msvc/language/compgen/index.htm?tx=12

Marking them `__forceinline` doesn't solve the problem.

What _does_ solve the problem though is using C++11's way of telling to the compiler that we want the default implementation:

    class float4
    {
        float4() = default;
        float4(float a) { ... }
        ~float4() = default;
    };

More details:

https://stackoverflow.com/questions/1025313/will-an-empty-constructor-or-destructor-do-the-same-thing-as-the-generated-one/2322314#2322314

---

We need to switch to this syntax for all performance-sensitive classes (`foundation::Vector`, `foundation::Color`, etc.).

Note that `= default` is not supported by VS 2012, which we still need to use to build some versions of the 3ds Max and Maya plugins.

dictoon:
This fixes #1910.

### DIFF
--- a/src/appleseed/foundation/image/color.h
+++ b/src/appleseed/foundation/image/color.h
@@ -66,7 +66,11 @@ class Color
     static const size_t Components = N;
 
     // Constructors.
-    Color();                                    // leave all components uninitialized
+#if !defined(_MSC_VER) || _MSC_VER >= 1800
+    Color() = default;                          // leave all components uninitialized
+#else
+    Color() {}                                  // leave all components uninitialized
+#endif
     explicit Color(const ValueType val);        // set all components to `val`
 
     // Construct a color from another color of a different type.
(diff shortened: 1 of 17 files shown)


src/appleseed/foundation/image/color.h
```diff
@@ -66,7 +66,11 @@ class Color
     static const size_t Components = N;
 
     // Constructors.
-    Color();                                    // leave all components uninitialized
+#if _MSC_VER >= 1800
```
fedormatantsev: What if _MSC_VER is undefined? E.g. linux?

fedormatantsev: Also what abour dtors?

dictoon: > What if _MSC_VER is undefined? E.g. linux?

Wow, good point! Will fix.

> Also what abour dtors?

None of these classes have destructors.

dictoon:
@fedormatantsev Thanks! I'll merge once Travis passes.

PR #1914, opened 2018-03-17
[WIP] Optimize algorithm to load plugins

MohamedAli1995:
issue: Optimize plugins discovery #1733
brief about the proposed technique :
1- store all plugins information ( entry_point, name,register_function)
2- search once in all paths for every plugin ( plugins are loaded once per path instead of once per file)


dictoon:
Superseded by #1920.

PR #1916, opened 2018-03-17
OSL tweaks

luisbarrancos:


src/appleseed.shaders/src/appleseed/as_standard_surface.osl
```diff
@@ -35,10 +35,7 @@ shader as_standard_surface
     string as_maya_classification = "drawdb/shader/surface:rendernode/appleseed/surface:shader/surface:swatch/AppleseedRenderSwatch",
     string help = "Standard Surface Shader",
     string icon = "asStandardSurface",
-    int as_maya_type_id = 0x001279d4,
-    string as_max_class_id = "651059681 943276521",
```
est77: Do we want to remove these lines?

luisbarrancos: No we don't, i'm not sure how that ended up there.

src/appleseed.shaders/src/appleseed/as_standard_surface.osl
```diff
@@ -672,7 +686,7 @@ shader as_standard_surface
 
         if (in_fresnel_type == 0)
         {
-            if (compute_transmission)
+            if (!compute_coating && compute_transmission)
```
dictoon: Isn't that a regression? @est77 removed this check that was causing things to become black when coating was applied, IIRC.

src/appleseed.shaders/src/appleseed/as_standard_surface.osl
```diff
@@ -477,7 +491,7 @@ shader as_standard_surface
         string as_maya_attribute_name = "maximumRayDepth",
         string as_maya_attribute_short_name = "mrd",
         int min = 0,
-        int max = 100,
+        int max = 32,
```
dictoon: Why this change? appleseed allows up to 100 bounces.

src/appleseed.shaders/src/appleseed/as_standard_surface.osl
```diff
@@ -97,6 +97,20 @@ shader as_standard_surface
         int as_maya_attribute_hidden = 1,
         int gafferNoduleLayoutVisible = 0
     ]],
+    int in_sss_profile = 0
+    [[
+        string as_maya_attribute_name = "sssProfile",
+        string as_maya_attribute_short_name = "ssp",
+        string widget = "mapper",
+        string options = "Better Dipole:0|Directional Dipole:1|Gaussian:2|Normalized Diffusion:3|Standard Dipole:4|Random Walk:5",
```
dictoon: In terms of profiles, in my opinion all plugins should only expose (in this order):

- "Gaussian"
- Better Dipole with the name "Dipole"
- "Normalized Diffusion"
- "Random Walk"

src/appleseed.shaders/src/appleseed/as_standard_surface.osl
luisbarrancos: When this shader was done, coating on glass would generate a huge amount of noise hence it was disabled since it was unusable. This was kind of suspended pending work on a mechanism to layer BxDFs, we then had as_layer() closure. I haven't revisited this particular case yet, but if @est77 removed this condition then i assume the coating on BTDF case is already correct.

est77: We did some testing after removing the condition and results looked correct.
Do you have any particular scene where the result was incorrect?

src/appleseed.shaders/src/appleseed/as_blackbody.osl
```diff
@@ -271,14 +271,14 @@ shader as_blackbody
...
-        out_color += in_specular_amount *
+        out_color += in_specular_weight *
             as_glossy(
                 "ggx",
                 Nn,
                 vector(0), // tangent vector
                 in_specular_roughness,
                 0.0, // specular spread
-                0.0, // anisotropy amount
+                0.0, // anisotropy weight
```
dictoon: That's an example of _amount_ being a more proper term than _weight_, in my opinion.

src/appleseed.shaders/src/appleseed/as_standard_surface.osl
```diff
@@ -97,30 +97,44 @@ shader as_standard_surface
...
         int gafferNoduleLayoutVisible = 0
     ]],
-    int in_sss_maximum_ray_depth = 2
+    int in_subsurface_profile = 0
     [[
-        string as_maya_attribute_name = "sssMaximumRayDepth",
+        string as_maya_attribute_name = "subsurfaceProfile",
+        string as_maya_attribute_short_name = "ssp",
+        string widget = "mapper",
+        string options = "Gaussian:0|Dipole:1|Normalized Diffusion:2|:Random Walk:3",
```
dictoon: There's an extra `:` before "Random Walk", is that OK?

src/appleseed.shaders/src/appleseed/as_standard_surface.osl
```diff
@@ -788,15 +803,34 @@ shader as_standard_surface
...
+            }
+            else
+            {
+                subsurface_profile = "randomwalk";
+            }
+
             // Disable Fresnel weight since transmittance is computed earlier,
             // but ideally, the BSSRDF should be used.
-
+
```
dictoon: Remove trailing spaces.

src/appleseed.shaders/src/appleseed/as_texture.osl
```diff
@@ -424,69 +426,79 @@ shader as_texture
...
+                
 #ifdef DEBUG
-            string shadername = "";
-            getattribute("shader:shadername", shadername);
-            warning("[WARNING!]: Unknown OETF mode %d, in %s, %s:%d\n",
-                    in_eotf, shadername, __FILE__, __LINE__);
+                string shadername = "";
+                
+                getattribute("shader:shadername", shadername);
+                warning("[WARNING!]: Unknown OETF mode %d, in %s, %s:%d\n",
```
dictoon: Remove exclamation mark.

PR #1922, opened 2018-03-20
Refactor renderer::Camera and fix warnings on Windows

dictoon:


src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -64,6 +64,40 @@ namespace
...
+
+    // Normalization factor to make the area under the curve equal to 1.
+    float                       m_shutter_curve_bezier_normalization_factor;
+
+    // Polynomials of open/close CDFs.
+    Vector<float, 7>            m_shutter_curve_bezier_open_cdf;
+    Vector<float, 7>            m_shutter_curve_bezier_close_cdf;
+
+    Impl()
+      : m_use_bezier_shutter_curve(false)
```
fedormatantsev: This probably should go to on_render_begin()

src/appleseed/renderer/modeling/camera/camera.h
```diff
@@ -171,26 +177,6 @@ class APPLESEED_DLLSYMBOL Camera
     float               m_inverse_cdf_close_point;
```
fedormatantsev: I suggest to move inverse_cdf_close_point/inverse_cdf_open_point to impl as well, those two are internal vars used only for inverse sampling. Normalized times seem to be internal vars as well.

src/appleseed/renderer/modeling/camera/camera.cpp
dictoon: Good point!

src/appleseed/renderer/modeling/camera/camera.h
dictoon: Yes, makes sense.

dictoon:
Thanks for the review, pushing a new commit in a minute (I'll squash at the end).

src/appleseed/renderer/modeling/camera/camera.h
```diff
@@ -173,8 +173,6 @@ class APPLESEED_DLLSYMBOL Camera
     float               m_shutter_open_time_interval;
     float               m_normalized_open_end_time;
     float               m_normalized_close_start_time;
```
fedormatantsev: I also mentioned normalized times, they are implementation specific too. Sorry if that wasn't clear.

dictoon: Hadn't read your message until the end :) Pushing a new commit with a few renamings as well in an attempt to reduce verbosity somewhat.

dictoon:
Cool, thanks!

There are two final renamings I want to do:
- `m_shutter_open_time` -> `m_shutter_open_start_time`
- `m_shutter_close_time` -> `m_shutter_close_end_time`

I'll also replace "start" by "begin" in order to be consistent with "end".

fedormatantsev:
Sounds good. This will make naming a lot more consistent.

PR #1923, opened 2018-03-20
Microfacet BRDF refactorings

est77:


src/appleseed/foundation/math/microfacet.cpp
```diff
@@ -154,25 +154,25 @@ MDF::~MDF()
 //
 
 float BlinnMDF::D(
-    const Vector3f&     h,
+    const Vector3f&     m,
```
oktomus: Maybe I'm wrong but parameters aren't correctly aligned

src/appleseed/foundation/math/microfacet.cpp
```diff
@@ -224,12 +224,12 @@ float BlinnMDF::pdf(
 //
 
 float BeckmannMDF::D(
-    const Vector3f&     h,
+    const Vector3f&     m,
```
oktomus: Here too

dictoon:
> I'm not sure if I'm well suited to review your code ^^

We welcome everybody, including and especially GSoC candidates and students, to have a look and provide feedback on PRs!

src/appleseed/foundation/math/microfacet.cpp
est77: The first part of this file is code that's in an unnamed namespace.
It doesn't need to have the same parameters alignment as the rest of the code in the file.

oktomus: Ok :)

src/appleseed/foundation/math/dual.h
```diff
@@ -50,7 +50,12 @@ class Dual
     typedef T ValueType;
 
     // Constructors.
-    Dual();
+    #if !defined(_MSC_VER) || _MSC_VER >= 1800
```
dictoon: That's debatable, but in the rest of the code we don't indent preprocessor statements and for consistency I would recommend doing the same here.

dictoon: A more important point: I actually hadn't touched this code on purpose. This changes the behavior as `m_has_derivates` is no longer set to `false` by the default constructor. Are you sure this is safe in all cases?

src/appleseed/renderer/modeling/bsdf/microfacethelper.h
```diff
@@ -275,6 +295,14 @@ class MicrofacetBRDFHelper
 
         return false;
     }
+
+    static void flip_directions(
```
dictoon: Looking at what this method does, I feel the name is not entirely adequate.

src/appleseed/foundation/math/dual.h
est77: Dual is inmutable. When you construct a Dual with a vector or with a vector with derivatives, m_has_derivatives is correctly set.

The change in the PR changes only the behaviour of default constructed Duals.
If we happen to use uninitialized Duals somewhere in the code that would be bug.

src/appleseed/renderer/modeling/bsdf/microfacethelper.h
est77: I removed it an inlined the code manually.

src/appleseed/foundation/math/dual.h
```diff
@@ -50,7 +50,12 @@ class Dual
     typedef T ValueType;
 
     // Constructors.
-    Dual();
+#if !defined(_MSC_VER) || _MSC_VER >= 1800
+        Dual() = default;                         // leave all components uninitialized
```
dictoon: Fix indentation.

src/appleseed/foundation/math/dual.h
```diff
@@ -66,6 +71,8 @@ class Dual
     const T& get_dx() const;
     const T& get_dy() const;
 
+    void set_derivatives(const T& dx, const T& dy);
```
dictoon: Don't we need a `set_value()` has well then?

src/appleseed/renderer/modeling/bsdf/microfacethelper.h
```diff
@@ -166,26 +174,36 @@ class MicrofacetBRDFHelper
...
-        const float                     cos_on,
         Spectrum&                       value)
     {
-        const foundation::Vector3f h = foundation::normalize(incoming + outgoing);
-        const foundation::Vector3f m = shading_basis.transform_to_local(h);
-        const foundation::Vector3f wo = shading_basis.transform_to_local(outgoing);
+        foundation::Vector3f wo = shading_basis.transform_to_local(outgoing);
+        foundation::Vector3f wi = shading_basis.transform_to_local(incoming);
+
+        if (Flip)
```
dictoon: I'm still slightly unsure about what `Flip` means. Why use `std::abs()` instead of negating Y components?

src/appleseed/foundation/math/dual.h
est77: I added this method to avoid copying m_value when we compute the ray differentials.
I don't think we need set_value(), at least for now.

src/appleseed/renderer/modeling/bsdf/microfacethelper.h
est77: I moved the flipping code from the glossy brdf here. The names come from there.
What this does is place wi and wo in the same hemisphere as the shading normal.

src/appleseed/foundation/math/dual.h
dictoon: Ok

src/appleseed/renderer/modeling/bsdf/microfacethelper.h
dictoon: Ok. Maybe you can add this in a comment?

src/appleseed/renderer/modeling/bsdf/glassbsdf.cpp
```diff
@@ -1255,8 +1255,8 @@ namespace
             for (size_t i = 0; i < SampleCount; ++i)
             {
                 // Generate a uniform sample in [0,1)^4.
```
dictoon: Fix comment.

PR #1924, opened 2018-03-21
Miscellaneous fixes and improvements

dictoon:


src/appleseed/renderer/kernel/rendering/oiioerrorhandler.cpp
```diff
@@ -56,21 +56,18 @@ void OIIOErrorHandler::operator()(int errcode, const string& msg)
...
       case EH_WARNING:
         RENDERER_LOG_WARNING("%s", modified_msg.c_str());
         break;
 
       case EH_ERROR:
-        RENDERER_LOG_ERROR("%s", modified_msg.c_str());
-        break;
-
       case EH_SEVERE:
-        RENDERER_LOG_FATAL("%s", modified_msg.c_str());
```
oktomus: Why not using log_fatal ?

src/appleseed/foundation/utility/bufferedfile.h
```diff
@@ -203,6 +205,29 @@ class APPLESEED_DLLSYMBOL BufferedFile
 };
 
 
+//
+// Utility free functions.
+//
+
+struct ExceptionEOF : public Exception {};
```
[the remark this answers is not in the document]
dictoon: It would have made sense to. So far the approach is to promote exception types to `foundation/core/exceptions/` only when they're commonly used, in multiple places.

src/appleseed/renderer/kernel/rendering/oiioerrorhandler.cpp
dictoon: `LOG_FATAL()` is pretty drastic. In appleseed.cli (the command line interface to appleseed) it will actually _terminate_ execution. Unless proven otherwise, I think it's safe to assume that no OSL message deserves such radical treatment.

dictoon:
Thanks for the review!

src/appleseed/renderer/kernel/rendering/oiioerrorhandler.cpp
oktomus: Ok sure :)

PR #1929, opened 2018-03-24
Fix #1885: missing extension in Open Recent menu

### LINKED ISSUE #1885: Missing extension in recent projects
Try this:
- Save the project to a location without specifying manually the `.appleseed` extension
- Your project is correctly saved as a `.appleseed` file
- Checkout `File > Open recent`
- You will see your project but without the `.appleseed` extension, and that can't be opened.

If you manually specify the extension, there isn't any problem.

Idloj:


Idloj:
Is it OK now? (I want to keep "Fix" and the issue number adjacent to leverage Github issue-auto-closing feature). I have tested it and it works fine.

oktomus:
It is for me, but I think it's better to have clear messages rather than automatic closing issues.

dictoon:
Looks great, thank you!

Issue #1908, opened 2018-03-14
Experiment with pow()-less sRGB conversions


est77:
`std::pow()` can be incredibly horribly slow on linux, to the point of appearing in profiles high on top, 
with ray tracing and texture lookups.

Reference:
   http://chilliant.blogspot.com.es/2012/08/srgb-approximations-for-hlsl.html

oktomus:
Also needs to find out if

PR #1931, opened 2018-03-25
Use fast colorspace conversion

oktomus:
#1908 
I just use `fast_...` where it wasn't used before. I'm sure there is some places where it's not correct, but I would like to know what is your opinion.
Btw, `convert_tile_srgb_to_linear_rgb` in `texturestore.cpp` seems to never be used.

### DIFF
--- a/src/appleseed/foundation/image/text/textrenderer.cpp
+++ b/src/appleseed/foundation/image/text/textrenderer.cpp
@@ -255,13 +255,13 @@ void TextRenderer::draw_string(
                 if (image_color_space == ColorSpaceSRGB)
                 {
                     // Convert background color from sRGB to linear RGB.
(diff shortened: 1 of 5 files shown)


dictoon:
> Btw, convert_tile_srgb_to_linear_rgb in texturestore.cpp seems to never be used.

Yes it's used on line 209 of that file:

    case ColorSpaceSRGB:
      convert_tile_srgb_to_linear_rgb(*record.m_tile);
      break;

oktomus:
@dictoon But this function is never called (the function load). I am not sure but my IDE tells me that Load is never used when I'm looking for usages of this function.

dictoon:
Yes it's called (by `foundation::LRUCache`) otherwise we would never have any textures :)

oktomus:
Ahah, works for me ! :)

PR #1932, opened 2018-03-27
Add test for curve self-intersection

rgirish28:
This adds a test case to check for self intersection for curve ribbons and fixes the earlier issue.

### DIFF
--- a/src/appleseed/foundation/math/beziercurve.h
+++ b/src/appleseed/foundation/math/beziercurve.h
@@ -887,12 +887,12 @@ bool BezierCurveIntersector<BezierCurveType>::converge(
         // Compute point on curve.
         const VectorType p = curve.evaluate_point(w);
 
-        // Compare Z distances.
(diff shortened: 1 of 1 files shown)


src/appleseed/foundation/math/beziercurve.h
```diff
@@ -887,13 +887,13 @@ bool BezierCurveIntersector<BezierCurveType>::converge(
...
 
-        // Compare Z distances.
-        if (p.z < ValueType(0.0) || p.z > t)
-            return false;
-
         // Compute curve width.
         const ValueType width = curve.evaluate_width(w);
 
+        // Compare Z distances.
+        if (p.z < ValueType(0.0) || p.z > t || p.z <= width)
```
dictoon: Can be simplified to

    if (p.z <= width || p.z > t)

dictoon:
Thank you!

PR #1938, opened 2018-03-29
Always save AOVs to EXR files.

est77:
Fix AOV related bugs.

### DIFF
--- a/src/appleseed/renderer/modeling/aov/aov.cpp
+++ b/src/appleseed/renderer/modeling/aov/aov.cpp
@@ -151,7 +151,7 @@ void UnfilteredAOV::create_image(
     const size_t tile_height,
     ImageStack& aov_images)
 {
-    // One extra channel to keep track of the distance
+    // Add one extra channel to keep track of the distance
     // to the nearest sample for each pixel.
     const size_t num_channels = get_channel_count() + 1;
 
@@ -163,6 +163,10 @@ void UnfilteredAOV::create_image(
             tile_height,
             num_channels,
             PixelFormatFloat);
+
+    // We need to clear the image because the default channel value
+    // might not be zero and also to initialize the pixel distance channel.
+    clear_image();
 }
 
 }   // namespace renderer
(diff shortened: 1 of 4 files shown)


src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -751,10 +745,10 @@ bool Frame::write_main_image(const char* file_path) const
...
     if (impl->m_denoising_mode == DenoisingMode::WriteOutputs)
     {
-        if (ends_with(lower_case(file_path), ".exr"))
-            impl->m_denoiser_aov->write_images(file_path);
-        else
-            RENDERER_LOG_ERROR("denoiser outputs can only be saved to exr images.");
+        bf::path boost_file_path(file_path);
+        boost_file_path.replace_extension(".exr");
+
+        impl->m_denoiser_aov->write_images(file_path);
```
dictoon: You're not using `boost_file_path` here. Intentional?

src/appleseed/renderer/modeling/aov/aov.cpp
```diff
@@ -163,6 +163,8 @@ void UnfilteredAOV::create_image(
             tile_height,
             num_channels,
             PixelFormatFloat);
+
+    clear_image();
```
dictoon: Whyu does `UnfilteredAOV::create_image()` clears the image but `AOV::create_image()` doesn't? Can we make the behavior more consistent?

est77: There are 2 reasons. One is that for some AOVs, 0 is not the default channel value (normals for example).
The other is that there is an extra hidden channel in in the images of unfiltered AOVs.
The accumulators use it to keep track of the nearest sample to the pixel center seen so far.

We can clear all the images, but it is not needed for the others AOVs and by clearing them we will allocate all the tiles in the image.

dictoon: Thanks for the explanation. Can you add it as a comment in the code?

est77: Fixed and rebased

est77:
Fixed and rebased

Issue #1719, opened 2017-11-20
Minor UI artifacts when squeezing Tests dialog


dictoon:
The button at the bottom left of the dialog are cropped:

![image](https://user-images.githubusercontent.com/321290/33046088-c49b86ee-ce4f-11e7-9e14-4f09ff12884a.png)

qq456cvb:
This can be resolved by setting minimum size of the left panel or automatic hide button texts and replace them with "..." like "Check onl...". What do you think?

dictoon:
The solution of setting a minimum size makes much more sense, and I see that it's the one you implemented. Thanks!

dictoon:
Fixed.

qq456cvb:
Although the minimum widths for left(350) and right(300) panels specified in the patch is safe for MacOS, it's not safe for windows.

![tim 20180329121736](https://user-images.githubusercontent.com/7703989/38069847-31340974-334b-11e8-909a-75c293802fd6.png)

I think 400 and 320 are safe widths for both Windows and MacOS. I do not have a build for Linux, if it is also safe, I will issue a PR.

dictoon:
Can you open a PR and set widths to the values you suggested?

qq456cvb:
Sure.

PR #1939, opened 2018-03-30
fix bug #1719 with a safer minimum width

qq456cvb:


dictoon:
Thanks!

@est77 can you test this on Linux?

PR #1942, opened 2018-03-31
Light paths recording, querying and visualization

dictoon:


src/appleseed/renderer/modeling/camera/sphericalcamera.cpp
```diff
@@ -215,6 +216,14 @@ namespace
             return true;
         }
 
+        RasterizationCamera get_rasterization_camera() const override
+        {
+            RasterizationCamera rc;
+            rc.m_aspect_ratio = 1024.0 / 576.0;
```
fedormatantsev: Why?

src/appleseed/renderer/modeling/object/meshobject.cpp
```diff
@@ -144,6 +145,53 @@ Lazy<RegionKit>& MeshObject::get_region_kit()
...
+        const auto& v2 = impl->m_tess.m_vertices[prim.m_v2];
+
+        // todo: check that vertex normals are available.
+        const auto& n0 = impl->m_tess.m_vertex_normals[prim.m_n0];
+        const auto& n1 = impl->m_tess.m_vertex_normals[prim.m_n1];
+        const auto& n2 = impl->m_tess.m_vertex_normals[prim.m_n2];
+
+        ObjectRasterizer::Triangle triangle;
+
+        triangle.m_v0[0] = v0[0];
```
fedormatantsev: Please, move this into triangle constructor (e.g. triangle(v0, v1, v2, n0, n1, n2))
Or to another "factory" function (e.g. init_triangle(triangle, vo, vq, v2, n0, n1, n2) or even triangle = construct_triangle(v0, v1, v2, n0, n1, n2))

src/appleseed/renderer/modeling/object/object.cpp
```diff
@@ -107,4 +107,8 @@ void Object::on_frame_end(
     ConnectableEntity::on_frame_end(project, parent);
 }
 
+void Object::rasterize(ObjectRasterizer& rasterizer) const
+{
```
fedormatantsev: assert(!"Not implemented")? Or even throw

src/appleseed/renderer/modeling/object/object.h
```diff
@@ -110,6 +111,9 @@ class APPLESEED_DLLSYMBOL Object
     // Return the alpha map cached by on_frame_begin().
     const Source* get_alpha_map() const;
 
+    // Send this object to an object rasterizer.
+    virtual void rasterize(ObjectRasterizer& rasterizer) const;
```
fedormatantsev: Maybe It would be clearer to introduce IRasterizable interface?
e.g.
class IRasterizable
{
  public:
    virtual void rasterize(ObjectRasterizer& rasterizer) const = 0
}

src/appleseed.studio/mainwindow/rendering/lightpathswidget.cpp
```diff
@@ -0,0 +1,382 @@
...
+
+            glNormal3d(triangle.m_n1[0], triangle.m_n1[1], triangle.m_n1[2]);
+            glVertex3d(triangle.m_v1[0], triangle.m_v1[1], triangle.m_v1[2]);
+            
+            glNormal3d(triangle.m_n2[0], triangle.m_n2[1], triangle.m_n2[2]);
+            glVertex3d(triangle.m_v2[0], triangle.m_v2[1], triangle.m_v2[2]);
+        }
+    };
+
+    void draw(const ObjectInstance& object_instance)
```
fedormatantsev: Here we could ask for IRasterizable instead

src/appleseed.studio/mainwindow/rendering/lightpathswidget.cpp
```diff
@@ -0,0 +1,382 @@
...
+
+    const auto& rc = m_camera.get_rasterization_camera();
+
+    glMatrixMode(GL_PROJECTION);
+    glLoadIdentity();
+    gluPerspective(
+        rad_to_deg(rc.m_hfov) / rc.m_aspect_ratio,
+        rc.m_aspect_ratio,
+        0.01,
+        1000.0);
```
fedormatantsev: Far and near z should be configurable

src/appleseed/renderer/modeling/object/object.h
fedormatantsev: Argh. If it was intended to skip object instances with no rasterize() implemented, please leave a comment here.

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -295,16 +295,6 @@ Vector2d Camera::extract_film_dimensions() const
     return film_dimensions;
 }
 
-namespace
-{
-    // Compute the focal length (in meters), given the film width (in meters)
-    // and the horizontal field of view (in degrees).
```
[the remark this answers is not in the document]
dictoon: The comment has already been moved to the header file. For comments like for code, we try to avoid duplication so method descriptions should only appear in the public ("interface") part of the code, usually in the header files.

src/appleseed/renderer/modeling/object/object.cpp
dictoon: By default the method does nothing on purpose: most objects don't know how to rasterize themselves yet but we don't want this to be blocking.

src/appleseed.studio/mainwindow/rendering/lightpathswidget.cpp
dictoon: I completely agree but I needed to get this feature out of the door early. Over time we will add ways to configure OpenGL rendering including near and far Z, shading mode, etc.

src/appleseed/renderer/modeling/object/object.h
dictoon: I'm not sure to understand. As soon as `renderer::Object` has a `rasterize()` method, all objects (and thus all object instances) become rasterizable, although they might currently not yet override the default implementation which does nothing. We cannot use an `IRasterizable` interface to detect which objects implement it and which don't.

fedormatantsev: Yes, I understood your intention later.

src/appleseed/renderer/modeling/camera/orthographiccamera.cpp
```diff
@@ -251,6 +252,14 @@ namespace
             return true;
         }
 
+        RasterizationCamera get_rasterization_camera() const override
+        {
+            RasterizationCamera rc;
+            rc.m_aspect_ratio = 1024.0 / 576.0;
```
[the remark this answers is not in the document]
dictoon: You're right, for the orthographic camera we can compute the film's aspect ratio (will push a fix) but for the spherical camera we cannot, that's why we return a reasonable camera. All of this will definitely be refined over time.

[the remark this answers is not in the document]
fedormatantsev: Ah, those are going to be configurable as well. Nice then.

PR #1944, opened 2018-04-01
Implement simple BDPT lighting engine

zentojamorn:
a bdpt implementation
still doesn't support
- shading normal
- volumetric
- specular surface


src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
```diff
@@ -659,6 +659,7 @@ namespace
             QComboBox* combobox = create_combobox("engine");
             combobox->setToolTip(m_params_metadata.get_path("lighting_engine.help"));
             combobox->addItem("Unidirectional Path Tracer", "pt");
+			combobox->addItem("Bidirectional Path Tracing", "bdpt");
```
dictoon: Replace tabs by 4 spaces here and everywhere else.

src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.cpp
```diff
@@ -36,6 +36,9 @@
 #include "renderer/modeling/edf/edf.h"
 #include "renderer/modeling/material/material.h"
 #include "renderer/modeling/project/project.h"
+#include "renderer/kernel/shading/shadingcomponents.h"
```
dictoon: Order alphabetically.

src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.cpp
```diff
@@ -51,10 +54,45 @@ namespace
...
     //
 
+    enum class BDPTVertexType
+    {
+        Camera, Light, Surface, Medium
+    };
+
+    struct BDPTVertex
+    {
+        foundation::Vector3d    m_position;
```
dictoon: You can remove all `foundation::` prefixes since there's already a using clause.

I'd also suggest adding a `using namespace std` clause to make the code terser.

src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.cpp
```diff
@@ -51,10 +54,45 @@ namespace
...
+            pdf *= std::max(foundation::dot(vertex.m_geometric_normal, w * std::sqrt(invDist2)), 0.0);
+            return pdf * invDist2;
+        }
+    };
+
     class BDPTLightingEngine
       : public ILightingEngine
     {
       public:
+        const unsigned int num_max_vertices = 9;
```
dictoon: I would recommend using `size_t` instead of `unsigned int`.

src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.cpp
```diff
@@ -83,37 +123,317 @@ namespace
         {
         }
 
+        Spectrum ComputeGeometryTerm(
```
dictoon: `compute_geometry_term()` according to our coding conventions:

    https://github.com/appleseedhq/appleseed/wiki/Coding-Philosophy-and-Guidelines#naming-conventions

src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.cpp
```diff
@@ -83,37 +123,317 @@ namespace
...
+                    shading_point.get_ray().m_depth,
+                    result);
+            }
+
+            return result * (float)g;
+        }
+
+        /// TODO:: precompute path density using fwd_pdf and rev_pdf.
+        float ComputePathDensity(
+            std::vector<BDPTVertex>             &light_vertices,
```
dictoon: Move `&` to the type's side:

    std::vector<BDPTVertex>&             light_vertices,

src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.cpp
```diff
@@ -83,37 +123,317 @@ namespace
...
         void compute_lighting(
             SamplingContext&        sampling_context,
             const PixelContext&     pixel_context,
             const ShadingContext&   shading_context,
             const ShadingPoint&     shading_point,
             ShadingComponents&      radiance) override      // output radiance, in W.sr^-1.m^-2
         {
-            PathVisitor light_path_visitor;
-            VolumeVisitor volume_visitor;
+            std::vector<BDPTVertex> camera_vertices;
```
dictoon: We can't use `std::vector<>` during rendering as that causes memory allocations. You could use dedicated arenas for instance.

src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.cpp
```diff
@@ -83,37 +123,317 @@ namespace
...
+            // camera_vertices[0] is on shading_point
+            camera_vertices.reserve(num_max_vertices - 1); 
+
+            // light_vertices[0] is on light source
+            light_vertices.reserve(num_max_vertices);
+
+            trace_light(sampling_context, shading_context, &light_vertices);
+            trace_camera(sampling_context, shading_context, shading_point, &camera_vertices);
+
+            for (int s = 0;s < light_vertices.size() + 1;s++)
```
dictoon: Insert space after `;`.

dictoon: Type mismatch, `s` must be of type `size_t`.

Basically `size_t` is the default integer type to use everywhere, unless you need signed integers in which case the default is `int`.

src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.cpp
```diff
@@ -83,37 +123,317 @@ namespace
...
+            {
+                for (unsigned int t = 2;t < camera_vertices.size() + 2;t++)
```
dictoon: Use `size_t`.

src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.h
```diff
@@ -40,6 +40,7 @@
 namespace foundation    { class Dictionary; }
 namespace renderer      { class ForwardLightSampler; }
 namespace renderer      { class Project; }
+namespace renderer      { class BackwardLightSampler; }
```
dictoon: Order alphabetically.

src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
```diff
@@ -659,6 +659,7 @@ namespace
             QComboBox* combobox = create_combobox("engine");
             combobox->setToolTip(m_params_metadata.get_path("lighting_engine.help"));
             combobox->addItem("Unidirectional Path Tracer", "pt");
+            combobox->addItem("Bidirectional Path Tracing", "bdpt");
```
aytekaman: I think this should be named `"Bidirectional Path Tracer"` for consistency.

dictoon: Good point.

src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.cpp
```diff
@@ -90,52 +134,324 @@ namespace
...
+                        static_cast<Vector3f>(normalize(prev2_vertex.m_position - prev_vertex.m_position)),
+                        ScatteringMode::All);
+                    result *= static_cast<float>(prev_vertex.convertDensity(pdf, vertex));
+                }
+            }
+
+            // start from camera
+            for (size_t i = 2; i <= q; i++)
+            {
+                if (i == 2) // on shading point (first hit from camera ray)
```
dictoon: Why not start iterating from 3 then?

src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.cpp
```diff
@@ -51,10 +56,47 @@ namespace
...
+        Spectrum                m_Le = Spectrum(0.0f);
+        ShadingPoint            m_shading_point;
+        bool                    m_is_light_vertex = false;
+
+        float                   m_fwd_pdf = 0.0f;
+        float                   m_rev_pdf = 0.0f;
+
+        ///TODO:: create a proper constructor
+
+        double convertDensity(double pdf, const BDPTVertex& vertex) const
```
dictoon: Should be called `convert_density()`.

src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.cpp
```diff
@@ -90,52 +134,324 @@ namespace
...
+                    static_cast<Vector3f>(light_vertex.m_dir_to_prev_vertex),
+                    ScatteringMode::All,
+                    light_eval_bsdf);
+
+                Spectrum geometry = compute_geometry_term(shading_context, shading_point, camera_vertex, light_vertex);
+                result = geometry * camera_eval_bsdf.m_beauty * light_eval_bsdf.m_beauty * camera_vertex.m_beta * light_vertex.m_beta;
+            }
+
+            // check if connection is near black or not
+            bool nearBlack = true;
```
dictoon: You can use `fz(result, 1.0e-9)`.

src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.cpp
```diff
@@ -90,52 +134,324 @@ namespace
...
+            float denominator = 0.0;
+            /// TODO:: unhandled case where denominator = 0 (specular surface).
+
+            // [p = 0, q = s + t], [p = 1, q = s + t - 1] ... [p = s + t - 2, q = 2]
+            for (size_t i = 0; i <= s + t - 2; i++)
+            {
+                denominator += compute_path_density(light_vertices, camera_vertices, s, t, i, s + t - i);
+            }
+
+            float miWeight = numerator / denominator;
```
dictoon: Variables should be spelled in snake case: `mi_weight`. Shouldn't it be `mis_weight` actually? What does "mi" stands for?

src/appleseed/renderer/kernel/lighting/forwardlightsampler.cpp
```diff
@@ -155,6 +156,24 @@ void ForwardLightSampler::sample(
     else sample_emitting_triangles(time, s, light_sample);
 }
 
+float ForwardLightSampler::evaluate_pdf(const ShadingPoint & light_shading_point) const
```
dictoon: Remove space before `&` (fix everywhere).

src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.cpp
```diff
@@ -90,52 +134,324 @@ namespace
...
-                shading_context.get_max_iterations());
+            size_t num_light_vertices = trace_light(sampling_context, shading_context, light_vertices);
+            size_t num_camera_vertices = trace_camera(sampling_context, shading_context, shading_point, camera_vertices);
+
+            for (size_t s = 0; s < num_light_vertices + 1; s++)
+                for (size_t t = 2; t < num_camera_vertices + 2; t++)
+                    if (s + t <= num_max_vertices)
+                        connect(shading_context, shading_point, light_vertices, camera_vertices, s, t, radiance);
+
+            shading_context.get_arena().clear();
```
dictoon: `shading_context.get_arena()` returns the per-pixel-sample arena. I don't think it should be cleared anywhere but in `GenericSampleRenderer::render_sample()`.

In fact, even the clearing of that arena in `pathtracer.h` looks suspicious to me (it was introduced by myself but I think we need to revisit this decision. `PathTracer` might want to provide a separate per-path-vertex arena instead).

zentojamorn: If I remember correctly, the code was there since the beginning of my development but I'll remove according to your request.

src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.cpp
```diff
@@ -388,34 +386,23 @@ namespace
...
             {
                 denominator += compute_path_density(light_vertices, camera_vertices, s, t, i, s + t - i);
             }
 
-            float miWeight = numerator / denominator;
+            float mis_weight = numerator / denominator;
 
-            assert(miWeight <= 1.0f);
-            radiance.m_beauty += miWeight * result;
+            assert(mi_weight <= 1.0f);
```
dictoon: Watch out, it should be `mis_weight` here.

src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.cpp
```diff
@@ -90,52 +136,309 @@ namespace
...
+            g *= max(dot(v, a.m_geometric_normal), 0.0);
+
+            Spectrum result(0.0);
+
+            // do the visibility test
+            if (g > 0.0)
+            {
+                // compute inverse dist2 and normalize 2 dot products
+                const double dist2 = square_norm(v);
+                g /= (dist2 * dist2);
```
dictoon: You're effectively dividing `g` by `dist^4` here, is that intended?

dictoon: Oh, ok. I think a comment explaining this will go a long way! And/or renaming `g` to `g2`?

src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.cpp
```diff
@@ -90,52 +136,309 @@ namespace
             const ShadingPoint&     shading_point,
             ShadingComponents&      radiance) override      // output radiance, in W.sr^-1.m^-2
         {
-            PathVisitor light_path_visitor;
-            VolumeVisitor volume_visitor;
+            BDPTVertex camera_vertices[8];
```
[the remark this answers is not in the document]
zentojamorn: 8 = num_max_vertices - 1
9 = num_max_vertices
I'll change this to new and delete in the next commit. (I couldn't get allocator to work properly.)

src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.cpp
zentojamorn: g (geometry term) was correct. I think the source code might be confusing. It was the trade-off between readability and performance. I'll rewrite the code so that it's more understandable in the next commit.

src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.cpp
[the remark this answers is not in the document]
dictoon: Yes, but why is num_max_vertices 9 in the first place? Note that appleseed supports up to 100 bounces.

PR #1948, opened 2018-04-03
Adaptively split tiles

aytekaman:


aytekaman:
Here is the plan:

1. I think it is always the last tile that needs to be subdivided. This tile can be divided into four quadrants. We can repeat this step recursively till some predefined threshold is reached.

2. This means that we can generate `TileJob`s according to this at the beginning.

3. We need to find a way to index adaptively subdivided tiles. I plan to use `tile_level` for this purpose. `tile_level = 0` means we don't want to address subdivided tiles. When `tile_level = 1`, this is the first subdivision: tile dimensions are halved and tile indices are doubled.

src/appleseed/renderer/kernel/rendering/generic/generictilerenderer.cpp
```diff
@@ -131,15 +132,15 @@ namespace
             // Retrieve tile properties.
             Tile& tile = frame.image().tile(tile_x, tile_y);
             TileStack aov_tiles = frame.aov_images().tiles(tile_x, tile_y);
-            const int tile_origin_x = static_cast<int>(frame_properties.m_tile_width * tile_x);
-            const int tile_origin_y = static_cast<int>(frame_properties.m_tile_height * tile_y);
+            const int tile_origin_x = static_cast<int>(frame_properties.m_tile_width * tile_x / ((tile_level + 1) * (tile_level + 1)));
```
dictoon: You can use `square(tile_level + 1)` for brevity (`foundation::square()` is defined in `foundation/math/scalar.h`).

dictoon:
I'm not sure that only the last tile needs to be subdivided. What we need is to keep all rendering threads busy until the frame is completed. It could even be that the initial number of tiles is too low compared to the number of rendering threads, in which case tiles need to be split right away into smaller subtiles.

aytekaman:
Some good news: a simple adaptive split implementation reduced the render time of the default Cornell Box scene by ~5% (8 cores). There are lots of issues to be addressed though. I will do a code cleanup and update this PR soon.

aytekaman:
I have pushed the latest changes on my branch. Here is what we have now:

1. We have one level of subdivision in an octree-like manner. Each tile have pointers to four sub-tiles. We can subdivide/merge tiles when needed. `Tile` and `ShadingResultFrameBuffer` are subdivided/merged together. When a `Tile` is subdivided/merged, contents are not copied to parent/children because `Tile` contents don't have to be preserved between passes. When `ShadingResultFrameBuffer` is subdivided/merged, contents are copied to parent/children. Right now, it is bit inefficient because contents are copied one pixel at a time. I will improve this by copying whole stride of the image at once.

2. Split decision doesn't work properly right now. Some times, it splits tiles right away at the beginning of the pass because sometimes, there are very few scheduled jobs right at the start.

3. Now, tiles are need to be adressed using three parameters (`tile_x`, `tile_y`, `tile_level`). This calls for a lots of changes at different places (tilecallbacks etc.) and I am not entirely happy with that. I am open to suggestions.

4. Since the `TileRenderer` compute the pixel locations before the rendering starts using the default tile dimensions, majority of the pixels can be wasted if the tile assigned to a `TileRenderer` is split. Impact on the performance can be ignored though.

dictoon:
I believe split decisions should take into account the number of tiles that remain to be rendered, not just the number of jobs that are currently in the queue.

dictoon:
> Since the TileRenderer compute the pixel locations before the rendering starts using the default tile dimensions, majority of the pixels can be wasted if the tile assigned to a TileRenderer is split. Impact on the performance can be ignored though.

I think so too.

dictoon:
> Now, tiles are need to be adressed using three parameters (tile_x, tile_y, tile_level). This calls for a lots of changes at different places (tilecallbacks etc.) and I am not entirely happy with that. I am open to suggestions.

I think these changes are unavoidable.

aytekaman:
Hi @FrancoisBeaune, yes you are totally right. I think it is time I should close this PR. Sorry that I couldn't finish this on time. I am having trouble finding sufficient time to work on appleseed these days.

PR #1949, opened 2018-04-03
Small fixes and cleanups

dictoon:


src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -775,71 +746,49 @@ bool Frame::write_aov_images(const char* file_path) const
...
+            return false;
 
-        success = success && write_extra_aovs(
-            aov_images(),
-            impl->m_extra_aovs,
-            directory,
-            base_file_name,
-            ".exr");
+        // Write AOV images.
+        if (!write_aov_images(filepath.c_str()))
```
est77: Instead of using the beauty "output_filename" parameter, you need to get the filename for each AOV:
aov->get_parameters().get_optional<string>("output_filename") and save each AOV there.
We use this method in Gaffer, where AOVs can be saved anywhere with any filename.

PR #1951, opened 2018-04-04
Fixed incorrect option formatting for OSL node.

jdent02:
in_y_tex_eotf is a string yet it used the (name):(integer) formatting in the options list.  This is unnecessary (the integers are not used) and breaks the blenderseed parser.


src/appleseed.shaders/src/appleseed/as_triplanar.osl
```diff
@@ -456,7 +456,7 @@ shader as_triplanar
         string label = "Input Transfer Function",
         string page = "Projection.Y Axis.Color Management",
         string widget = "mapper",
```
est77: widget has to be "popup" for strings. mapper is only for integers.

jdent02:
Updated

dictoon:
@est77 If it looks good please squash the PR into a single commit and fix the commit title (use present rather than past tense).

Thanks guys!

PR #1952, opened 2018-04-06
Update opengl.h to build in macOS

pjessesco:
Fix `gl.h` and `glu.h` path in macOS. Maybe this patch should be tested in other OS.


dictoon:
Thank you!

PR #1953, opened 2018-04-06
appleseed.python C++ consistency changes.

est77:


src/appleseed.python/bindmeshobject.cpp
```diff
@@ -140,19 +140,18 @@ void bind_mesh_object()
...
-        .def_readwrite("n1", &Triangle::m_n1)
-        .def_readwrite("n2", &Triangle::m_n2)
-        .def_readwrite("a0", &Triangle::m_a0)
-        .def_readwrite("a1", &Triangle::m_a1)
-        .def_readwrite("a2", &Triangle::m_a2)
-        .def_readwrite("pa", &Triangle::m_pa)
-
-        .def("has_vertex_attributes", &Triangle::has_vertex_attributes)
-        ;
+        .def_readwrite("m_v0", &Triangle::m_v0)
```
dictoon: While I'm totally OK with this change, I have the feeling that the `m_` prefix is not very Pythonic. Are we sure we want to improve consistency at the cost of making the Python API less natural to Python coders?

est77: I prefer being consistent with the C++ API in this case.
Our Python API is not very Pythonic for a start, but I find it very useful to prototype code that can be ported to C++ later if needed.

dictoon: Ok.

PR #1955, opened 2018-04-07
Add missing denoiser options

est77:


src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -1117,6 +1122,28 @@ DictionaryArray FrameFactory::get_input_metadata()
...
+                Dictionary()
+                    .insert("denoiser", "on")));
+
+    metadata.push_back(
+        Dictionary()
+            .insert("name", "random_pixel_order")
+            .insert("label", "Random Pixel Order")
+            .insert("type", "boolean")
+            .insert("use", "optional")
+            .insert("default", "false")
```
dictoon: The default value here is not consistent with the one used in `get_optional()`.

est77:
Fixed

PR #1956, opened 2018-04-07
MasterRenderer refactor

est77:
- Remove unused BaseRenderer base class.
- Move implementation to private implementation.


src/appleseed/renderer/kernel/rendering/masterrenderer.cpp
```diff
@@ -75,416 +80,563 @@ namespace renderer
...
+
+    // Initialize OSL's shading system.
+    bool initialize_osl_shading_system(
+        TextureStore&               texture_store,
+        foundation::IAbortSwitch&   abort_switch)
     {
-        Stopwatch<DefaultWallclockTimer> stopwatch;
-        stopwatch.start();
+        // Initialize OIIO.
+        const ParamArray& params = m_params.child("texture_store");
```
dictoon: This variable is confusing. AFAICT, it's used only once so I would suggest inlining it.

src/appleseed/renderer/kernel/rendering/masterrenderer.cpp
```diff
@@ -75,416 +80,563 @@ namespace renderer
...
+            {
+                for (size_t tx = 0; tx < frame_props.m_tile_count_x; ++tx)
+                    tile_callback->on_tile_end(frame, tx, ty);
+            }
+        }
     }
 
-    return true;
-}
+    Project&                    m_project;
```
dictoon: Move data members to the top of the struct.

est77:
Fixed and rebased

dictoon:
Thanks!

PR #1969, opened 2018-04-12
Small camera refactoring and bug fixes.

fedormatantsev:
- Refactor camera initialization & fix bad naming
- Initialize motion blur related variables only if they are required
- Linear shutter curve now uses normalized time for time mapping

### DIFF
--- a/src/appleseed/renderer/modeling/camera/camera.cpp
+++ b/src/appleseed/renderer/modeling/camera/camera.cpp
@@ -63,12 +63,15 @@ namespace renderer
 namespace
 {
     const UniqueID g_class_uid = new_guid();
+
+    const float ShutterOpenBeginTimeDefaultValue = 0.0f;
+    const float ShutterCloseEndTimeDefaultValue = 1.0f;
 }
 
 struct Camera::Impl
 {
     float               m_normalized_open_end_time;
-    float               m_normalized_close_start_time;
+    float               m_normalized_close_begin_time;
 
     float               m_inverse_cdf_open_point;
     float               m_inverse_cdf_close_point;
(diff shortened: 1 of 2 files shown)


src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -123,94 +126,162 @@ bool Camera::on_render_begin(
...
+    impl->m_inverse_cdf_open_point = evaluate_polynomial(impl->m_shutter_curve_bezier_open_cdf, 1.0f);
+    impl->m_inverse_cdf_close_point = evaluate_polynomial(impl->m_shutter_curve_bezier_close_cdf, 0.0f);
+}
+
+void Camera::initialize_shutter_curve_linear()
+{
+    float cbt_minus_oet_plus_one =
+        impl->m_normalized_close_begin_time - impl->m_normalized_open_end_time + 1.0f;
+
+    if (cbt_minus_oet_plus_one == 0.0f)
```
pjessesco: How about `ncbt`, or `n_cbt` to contain `normalized` meaning?

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -123,94 +126,162 @@ bool Camera::on_render_begin(
...
+            ShutterCloseEndTimeDefaultValue,
+            ShutterCloseEndTimeDefaultValue);
+
+        m_shutter_open_begin_time = ShutterOpenBeginTimeDefaultValue;
+        m_shutter_open_end_time = ShutterOpenBeginTimeDefaultValue;
+        m_shutter_close_begin_time = ShutterCloseEndTimeDefaultValue;
+        m_shutter_close_end_time = ShutterCloseEndTimeDefaultValue;
+
+        impl->m_normalized_open_end_time =
+            inverse_lerp(
```
pjessesco: I think `inverse_lerp()` is not needed since it is always 0.

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -123,94 +126,162 @@ bool Camera::on_render_begin(
...
+                m_shutter_open_begin_time,
+                m_shutter_close_end_time,
+                m_shutter_open_end_time);
+
+        impl->m_normalized_close_begin_time =
+            inverse_lerp(
```
pjessesco: Same here.

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -123,94 +126,162 @@ bool Camera::on_render_begin(
...
+    if (cbt_minus_oet_plus_one == 0.0f)
+    {
+        RENDERER_LOG_ERROR(
+            "while defining camera \"%s\": invalid values \"%f\" \"%f\" for parameters \"%s\", \"%s\" "
+            "using default values for parameters \"%s\", \"%s\", \"%s\", \"%s\" "
+            "\"%f\", \"%f\", \"%f\", \"%f\".",
+            get_path().c_str(),
+            m_shutter_open_end_time,
+            m_shutter_close_begin_time,
+            "shutter_open_end_time",
```
pjessesco: In curiosity, is it better that pass strings (e.g `"shutter_open_end_time"`) as parameter than put it in statement? I mean,

`invalid values "%f" "%f" for parameters "shutter_close_begin_time", "shutter_open_begin_time"`

src/appleseed/renderer/modeling/camera/camera.cpp
fedormatantsev: Yep

fedormatantsev: Yep

src/appleseed/renderer/modeling/camera/camera.cpp
fedormatantsev: I mimic how error reporting is done in other places. This approach decouples error message and parameters. E.g. it would be easier to rename this parameters

Issue #1937, opened 2018-03-29
Implement Correlated Multi-Jittered Sampling for pixel rendering


fedormatantsev:
Here is a good [paper](https://graphics.pixar.com/library/MultiJitteredSampling/paper.pdf)

Now there is Halton sequence sampling, which isn't the best option.

PR #1975, opened 2018-04-15
Implement Correlated Multi-Jittered Sampling

oktomus:
Sorry for the new PR, it was getting really dirty
#1937 

cmj_comp

[Paper](https://graphics.pixar.com/library/MultiJitteredSampling/paper.pdf)

## Testing it

1. In the appleseed fork directory, get the last state of master.
```
git checkout master # If not already
git pull appleseed_remote master
```
2. Run cmake and compile, this will be the build using Halton sampling.
3. `cp -r` `the whole fork directory somewhere
4. Get the CMJ code
```
git remote add oktomus git@github.com:oktomus/appleseed.git
git pull oktomus cmj
git checkout cmj
```
`5. Run cmake and compile, this will be the build using CMJ sampling.

# What still needs to be done

1. Compare convergence rates of QMC and CMJ

# Next ?

Implement a progressive mode using CMJ https://graphics.pixar.com/library/ProgressiveMultiJitteredSampling/

### DIFF
--- a/src/appleseed/foundation/math/correlatedmultijitter.h
+++ b/src/appleseed/foundation/math/correlatedmultijitter.h
@@ -1,0 +1,197 @@
+
+//
+// This source file is part of appleseed.
+// Visit http://appleseedhq.net/ for additional information and resources.
+//
+// This software is released under the MIT license.
+//
+// Copyright (c) 2018 Kevin Masson, The appleseedhq Organization
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+#ifndef APPLESEED_FOUNDATION_MATH_CORRELATEDMULTIJITTER_H
+#define APPLESEED_FOUNDATION_MATH_CORRELATEDMULTIJITTER_H
+
+// appleseed.foundation headers.
+#include "foundation/math/vector.h"
+
+// Standard headers.
+#include <cassert>
+#include <cmath>
+
+namespace foundation
+{
+
+//
+// Correlated Multi-Jittered Sampling.
+//
+// Reference:
+//
+//   Andrew Kensler, Correlated Multi-Jittered Sampling
+//   https://graphics.pixar.com/library/MultiJitteredSampling/paper.pdf
+//
+
+template <typename T>
+inline void cmj_compute_mn(
+    T&                  m,
+    T&                  n,
+    const unsigned int  sample_count);
+
+template <typename T>
+inline Vector<T, 2> cmj_generate_sample(
+    unsigned int        instance, // number of the sample to generate [0, sample_count)
+    const unsigned int  m,
+    const unsigned int  n,
+    const unsigned int  sample_count, // number of sample in the sequence
+    const unsigned int  pattern); // pattern to use for generating the full sequence
+
+template <typename T>
+inline Vector<T, 2> cmj_generate_sample(
+    unsigned int        instance,
+    const unsigned int  sample_count,
+    const unsigned int  pattern);
+
+inline unsigned int cmj_permute(
+    unsigned int        i,
+    const unsigned int  l,
+    const unsigned int  p);
+
+inline float cmj_randfloat(
+    unsigned int        i,
+    const unsigned int  p);
+
+
+//
+// Correlated Multi-Jittered Sampling implementation.
+//
+
+template <typename T>
+void cmj_compute_mn(
+    T&                  m,
+    T&                  n,
+    const unsigned int  sample_count)
+{
+    assert(sample_count > 1);
+    m = static_cast<int>(std::sqrt(static_cast<float>(sample_count)));
+    n = (sample_count + m - 1) / m;
+}
+
+template <typename T>
+Vector<T, 2> cmj_generate_sample(
+    unsigned int        instance,
+    const unsigned int  m,
+    const unsigned int  n,
+    const unsigned int  sample_count,
+    const unsigned int  pattern)
+{
+    assert(m >= 1 && n >= 1 && sample_count > 1);
+    assert(instance >= 0 && instance < sample_count);
+    instance = cmj_permute(instance, sample_count, pattern * 0x51633e2d);
+    const int sx = cmj_permute(instance % m, m, pattern * 0x68bc21eb);
+    const int sy = cmj_permute(instance / m, n, pattern * 0x02e5be93);
+    const T jx = cmj_randfloat(instance, pattern * 0x967a889b);
+    const T jy = cmj_randfloat(instance, pattern * 0x368cc8b7);
+    Vector<T, 2> r;
+    r[0] = (sx + (sy + jx) / n) / m;
+    r[1] = (instance + jy) / sample_count;
+    return r;
+}
+
+template <typename T>
+Vector<T, 2> cmj_generate_sample(
+    unsigned int        instance,
+    const unsigned int  sample_count,
+    const unsigned int  pattern)
+{
+    unsigned int m, n;
+    cmj_compute_mn(m, n, sample_count);
+    return cmj_generate_sample<T>(instance, m, n, sample_count, pattern);
+}
+
+unsigned int cmj_permute(
+    unsigned int        i,
+    const unsigned int  l,
+    const unsigned int  p)
+{
+    unsigned int w = l - 1;
+    if ((l & w) == 0)
+    {
+        // l is a power of two (fast).
+        i ^= p;
+        i *= 0xe170893d;
+        i ^= p >> 16;
+        i ^= (i & w) >> 4;
+        i ^= p >> 8;
+        i *= 0x0929eb3f;
+        i ^= p >> 23;
+        i ^= (i & w) >> 1;
+        i *= 1 | p >> 27;
+        i *= 0x6935fa69;
+        i ^= (i & w) >> 11;
+        i *= 0x74dcb303;
+        i ^= (i & w) >> 2;
+        i *= 0x9e501cc3;
+        i ^= (i & w) >> 2;
+        i *= 0xc860a3df;
+        i &= w;
+        i ^= i >> 5;
+        return (i + p) & w;
+    }
+    else
+    {
+        w |= w >> 1;
+        w |= w >> 2;
+        w |= w >> 4;
+        w |= w >> 8;
+        w |= w >> 16;
+        do
+        {
+            i ^= p; i *= 0xe170893d;
+            i ^= p >> 16;
+            i ^= (i & w) >>  4;
+            i ^= p >>  8;  i *= 0x0929eb3f;
+            i ^= p >> 23;
+            i ^= (i & w) >> 1; i *= 1 | p >> 27;
+            i *= 0x6935fa69;
+            i ^= (i & w) >> 11;  i *= 0x74dcb303;
+            i ^= (i & w) >>  2;  i *= 0x9e501cc3;
+            i ^= (i & w) >>  2;  i *= 0xc860a3df;
+            i &= w;
+            i ^= i >> 5;
+        } while (i >= l);
+        return (i + p) % l;
+    }
+}
+
+float cmj_randfloat(unsigned int i, const unsigned int p)
+{
+    i ^= p;
+    i ^= i >> 17;
+    i ^= i >> 10; i *= 0xb36534e5;
+    i ^= i >> 12;
+    i ^= i >> 21; i *= 0x93fc4795;
+    i ^= 0xdf6e307f;
+    i ^= i >> 17; i *= 1 | p >> 18;
+    return i * (1.0f / 4294967808.0f);
+}
+
+}       // namespace foundation
+
+#endif  // !APPLESEED_FOUNDATION_MATH_CORRELATEDMULTIJITTER_H
--- a/src/appleseed/foundation/math/sampling/qmcsamplingcontext.h
+++ b/src/appleseed/foundation/math/sampling/qmcsamplingcontext.h
@@ -31,6 +31,7 @@
 #define APPLESEED_FOUNDATION_MATH_SAMPLING_QMCSAMPLINGCONTEXT_H
 
 // appleseed.foundation headers.
+#include "foundation/math/correlatedmultijitter.h"
 #include "foundation/math/permutation.h"
 #include "foundation/math/primes.h"
 #include "foundation/math/qmc.h"
@@ -40,6 +41,7 @@
 
 // Standard headers.
 #include <cassert>
+#include <cmath>
 #include <cstddef>
 
 // Unit test case declarations.
@@ -55,6 +57,7 @@ namespace foundation
 // A sampling context featuring:
 //
 //   - deterministic sampling based on Halton sequences
+//   - deterministic sampling based on modified Chiu et al's multi-jittered sampling method
 //   - Faure digit scrambling
 //   - Cranley-Patterson rotation
 //   - Monte Carlo padding
@@ -63,6 +66,9 @@ namespace foundation
 //
 //   Kollig and Keller, Efficient Multidimensional Sampling
 //   www.uni-kl.de/AG-Heinrich/EMS.pdf
+//
+//   Andrew Kensler, Correlated Multi-Jittered Sampling
+//   graphics.pixar.com/library/MultiJitteredSampling/paper.pdf
 //
 
 template <typename RNG>
@@ -128,17 +134,47 @@ class QMCSamplingContext
 
     typedef Vector<double, 4> VectorType;
 
-    RNG&        m_rng;
-    Mode        m_mode;
+    struct CMJParameters
+    {
+        bool            m_enabled; // tells CMJ should be used or not
+        size_t          m_m; // deduced from the sample count
+        size_t          m_n; // also deduced from the sample count, n * m = sample count
+        size_t          m_sequence; // index of the next sample to generate [-1, sample_count)
+        int             m_pattern; // pattern to use for generating samples of the full sequence
 
-    size_t      m_base_dimension;
-    size_t      m_base_instance;
+        CMJParameters(
+            const size_t    sample_count,
+            const size_t    dimension,
+            const int       pattern,
+            const Mode      mode)
 
-    size_t      m_dimension;
-    size_t      m_sample_count;
+          : m_enabled(false)
+          , m_m(0)
+          , m_n(0)
+          , m_sequence(0)
+          , m_pattern(pattern)
+        {
+            if (sample_count > 1 && dimension == 2 && mode == QMCMode)
+            {
+                m_enabled = true;
+                cmj_compute_mn(m_m, m_n, sample_count);
+            }
+        }
+    };
 
-    size_t      m_instance;
-    VectorType  m_offset;
+    RNG&            m_rng;
+    Mode            m_mode;
+
+    size_t          m_base_dimension;
+    size_t          m_base_instance;
+
+    size_t          m_dimension;
+    size_t          m_sample_count;
+
+    size_t          m_instance;
+    VectorType      m_offset;
+
+    CMJParameters   m_cmj_params;
 
     // Cranley-Patterson rotation.
     template <typename T>
@@ -154,10 +190,23 @@ class QMCSamplingContext
 
     void compute_offset();
 
-    template <typename T> struct Tag {};
+    enum MethodSpecifier
+    {
+        M_RNG,
+        M_Halton,
+        M_CMJ
+    };
 
-    template <typename T> T next2(Tag<T>);
-    template <typename T, size_t N> Vector<T, N> next2(Tag<Vector<T, N>>);
+    template <typename T, int M> struct Tag {};
+
+    // Dispatchers.
+    template <typename T, int M> T next2(Tag<T, M>);
+    template <typename T, size_t N, int M> Vector<T, N> next2(Tag<Vector<T, N>, M>);
+
+    // Different methods implementations.
+    template <typename T, size_t N> Vector<T, N> next2(Tag<Vector<T, N>, M_RNG>);
+    template <typename T, size_t N> Vector<T, N> next2(Tag<Vector<T, N>, M_Halton>);
+    template <typename T> Vector<T, 2> next2(Tag<Vector<T, 2>, M_CMJ>);
 };
 
 
@@ -177,6 +226,7 @@ inline QMCSamplingContext<RNG>::QMCSamplingContext(
   , m_sample_count(0)
   , m_instance(0)
   , m_offset(0.0)
+  , m_cmj_params(0, 0, 0, mode)
 {
 }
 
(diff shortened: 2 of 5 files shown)


src/appleseed/renderer/kernel/rendering/final/uniformpixelrenderer.cpp
```diff
@@ -58,6 +58,7 @@
 
 // Standard headers.
 #include <cmath>
+#include <vector>
```
dictoon: Remove, unless it's necessary and it was missing.

src/appleseed/foundation/math/sampling/qmcsamplingcontext.h
```diff
@@ -55,6 +57,7 @@ namespace foundation
 // A sampling context featuring:
 //
 //   - deterministic sampling based on Halton sequences
+//   - deterministic sampling based on modified Chiu et al’s multi-jittered sampling method
```
dictoon: Unless I'm mistaking, `’` is not a standard quote, and the file is no longer encoded in ANSI. Use a standard quote. (Are you using Sublime Text by any chance?)

src/appleseed/foundation/math/sampling/qmcsamplingcontext.h
```diff
@@ -63,6 +66,7 @@ namespace foundation
 //
 //   Kollig and Keller, Efficient Multidimensional Sampling
 //   www.uni-kl.de/AG-Heinrich/EMS.pdf
+//   graphics.pixar.com/library/MultiJitteredSampling/paper.pdf
```
dictoon: Insert a blank line to separate papers, and add the title of the paper.

src/appleseed/foundation/math/correlatedmultijitter.h
```diff
@@ -0,0 +1,186 @@
...
+            i ^= (i & w) >>  2;  i *= 0x9e501cc3;
+            i ^= (i & w) >>  2;  i *= 0xc860a3df;
+            i &= w;
+            i ^= i >> 5;
+        } while (i >= l);
+        return (i + p) % l;
+    }
+}
+
+inline float cmj_randfloat (unsigned int i, const unsigned int p)
```
dictoon: Remove space before `(`.

src/appleseed/foundation/math/correlatedmultijitter.h
```diff
@@ -0,0 +1,186 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+#ifndef APPLESEED_FOUNDATION_MATH_CMJ_H
```
dictoon: Fix header guard, it should match the file name.

src/appleseed/foundation/math/correlatedmultijitter.h
```diff
@@ -0,0 +1,186 @@
...
+//
+// Correlated Multi-Jittered Sampling.
+//
+// References:
+//
+//   https://graphics.pixar.com/library/MultiJitteredSampling/paper.pdf
+//
+
+template <typename T>
+inline void cmj_compute_mn(T& m, T& n, const unsigned int N);
```
dictoon: Remove `inline` qualifiers from declarations.

src/appleseed/foundation/math/correlatedmultijitter.h
```diff
@@ -0,0 +1,186 @@
...
+#include <cassert>
+#include <cmath>
+
+namespace foundation
+{
+
+//
+// Correlated Multi-Jittered Sampling.
+//
+// References:
```
dictoon: Singular.

src/appleseed/foundation/math/correlatedmultijitter.h
```diff
@@ -0,0 +1,186 @@
...
+
+template <typename T>
+inline void cmj_compute_mn(T& m, T& n, const unsigned int N);
+
+template <typename T>
+inline Vector<T, 2> cmj_generate_sample(
+    unsigned int        s,
+    const unsigned int  m,
+    const unsigned int  n,
+    const unsigned int  N,
```
dictoon: I'm not too fond of having two parameters differing only by their case (`n` and `N`). Also it would be nice to add comments stating what each parameter is/does.

src/appleseed/foundation/math/correlatedmultijitter.h
```diff
@@ -0,0 +1,186 @@
...
+    const unsigned int  p);
+
+template <typename T>
+inline Vector<T, 2> cmj_generate_sample(
+    unsigned int        s,
+    const unsigned int  N,
+    const unsigned int  p);
+
+inline unsigned int cmj_permute(
+    unsigned int i,
```
dictoon: Align.

src/appleseed/foundation/math/correlatedmultijitter.h
```diff
@@ -0,0 +1,186 @@
...
+    const unsigned int l,
+    const unsigned int p);
+
+inline float cmj_randfloat (unsigned int i, const unsigned int p);
```
dictoon: Remove space before `(`.

src/appleseed/foundation/math/correlatedmultijitter.h
```diff
@@ -0,0 +1,186 @@
...
+
+
+//
+// Correlated Multi-Jittered Sampling implementation.
+//
+
+template <typename T>
+inline void cmj_compute_mn(T& m, T& n, const unsigned int N)
+{
+    m = static_cast<int>(std::sqrt((float) N));
```
dictoon: Use `static_cast<>()` instead of C-style cast.

src/appleseed/foundation/math/correlatedmultijitter.h
```diff
@@ -0,0 +1,186 @@
...
+inline Vector<T, 2> cmj_generate_sample(
+    unsigned int        s,
+    const unsigned int  m,
+    const unsigned int  n,
+    const unsigned int  N,
+    const unsigned int  p)
+{
+    assert(m > 0 && n > 0 && N > 0);
+    s = cmj_permute(s, N, p * 0x51633e2d);
+    int sx = cmj_permute(s % m, m, p * 0x68bc21eb);
```
dictoon: Make `const` everything that can be.

src/appleseed/foundation/math/correlatedmultijitter.h
```diff
@@ -0,0 +1,186 @@
...
+    const unsigned int  N,
+    const unsigned int  p)
+{
+    unsigned int m, n;
+    cmj_compute_mn(m, n, N);
+    return cmj_generate_sample<T>(s, m, n, N, p);
+}
+
+inline unsigned int cmj_permute(
+    unsigned int i,
```
dictoon: Align.

src/appleseed/foundation/math/sampling/qmcsamplingcontext.h
```diff
@@ -128,17 +132,45 @@ class QMCSamplingContext
...
+    struct CMJParameters
+    {
+        bool            m_enabled;
+        size_t          m_m;
+        size_t          m_n;
+        size_t          m_sequence;
+        int             m_pattern;
+
+        explicit CMJParameters(
+            const size_t    samples_count,
```
dictoon: `sample_count`

src/appleseed/foundation/math/correlatedmultijitter.h
```diff
@@ -0,0 +1,186 @@
...
+
+inline float cmj_randfloat (unsigned int i, const unsigned int p);
+
+
+//
+// Correlated Multi-Jittered Sampling implementation.
+//
+
+template <typename T>
+inline void cmj_compute_mn(T& m, T& n, const unsigned int N)
```
dictoon: Why is the type of `m` and `n` not fixed?

src/appleseed/foundation/math/sampling/qmcsamplingcontext.h
```diff
@@ -128,17 +132,45 @@ class QMCSamplingContext
...
-    Mode        m_mode;
+    struct CMJParameters
+    {
+        bool            m_enabled;
+        size_t          m_m;
+        size_t          m_n;
+        size_t          m_sequence;
+        int             m_pattern;
+
+        explicit CMJParameters(
```
dictoon: Remove `explicit` as it's redundant here.

src/appleseed/foundation/math/sampling/qmcsamplingcontext.h
```diff
@@ -128,17 +132,45 @@ class QMCSamplingContext
...
+        }
+    };
 
-    size_t      m_dimension;
-    size_t      m_sample_count;
+    CMJParameters   m_cmj_params;
 
-    size_t      m_instance;
-    VectorType  m_offset;
+    RNG&            m_rng;
```
dictoon: Keep these two data members first as they are the construction parameters.

src/appleseed/foundation/math/sampling/qmcsamplingcontext.h
```diff
@@ -256,7 +293,8 @@ inline void QMCSamplingContext<RNG>::split_in_place(
     const size_t        dimension,
     const size_t        sample_count)
 {
-    assert(m_sample_count == 0 || m_instance == m_sample_count);    // can't split in the middle of a sequence
+    // can't split in the middle of a sequence
```
dictoon: Since this is now a full line comment, it must start with a capital and end with a dot.

src/appleseed/foundation/math/sampling/qmcsamplingcontext.h
```diff
@@ -256,7 +293,8 @@ inline void QMCSamplingContext<RNG>::split_in_place(
...
+    assert(m_sample_count == 0 || m_instance == m_sample_count || m_cmj_params.m_enabled);
```
dictoon: That assert originally checked that:
- Either we didn't know how many samples would be required (`m_sample_count == 0`)
- Or that we know how many samples would be required, and we would have computed all of them (`m_instance == m_sample_count`).

Now, it seems that the check is entirely bypassed if `m_cmj_params.m_enabled` is enabled. That doesn't seem right: even when CMJ is enabled, it doesn't change the fact that splitting in place can only occur once all samples have been drawn.

src/appleseed/foundation/math/sampling/qmcsamplingcontext.h
```diff
@@ -344,7 +384,7 @@ inline Vector<T, N> QMCSamplingContext<RNG>::next2(Tag<Vector<T, N>>)
 {
     Vector<T, N> v;
 
-    assert(m_sample_count == 0 || m_instance < m_sample_count);
+    assert(m_sample_count == 0 || m_instance < m_sample_count || m_cmj_params.m_enabled);
```
dictoon: Same remark here.

src/appleseed/foundation/math/sampling/qmcsamplingcontext.h
```diff
@@ -381,6 +421,61 @@ inline Vector<T, N> QMCSamplingContext<RNG>::next2(Tag<Vector<T, N>>)
     return v;
 }
 
+template <typename RNG>
+template <typename T>
+inline Vector<T, 2> QMCSamplingContext<RNG>::next2(Tag<Vector<T, 2>>)
+{
+    Vector<T, 2> v;
+
+    assert(m_sample_count == 0 || m_instance < m_sample_count || m_cmj_params.m_enabled);
```
dictoon: Same remark here.

src/appleseed/foundation/math/sampling/qmcsamplingcontext.h
```diff
@@ -381,6 +421,61 @@ inline Vector<T, N> QMCSamplingContext<RNG>::next2(Tag<Vector<T, N>>)
...
+                    v[i] = static_cast<T>(PrecomputedHaltonSequence[m_instance * 4 + i]);
+                    v[i] = rotate(v[i], static_cast<T>(m_offset[i]));
+                }
+            }
+            else
+            {
+                v[0] = radical_inverse_base2<T>(m_instance);
+                v[0] = rotate(v[0], static_cast<T>(m_offset[0]));
+
+                for (size_t i = 1; i < 2; ++i)
```
dictoon: That `for` loop is no longer necessary since the only possible value is `i` == 1.

src/appleseed/foundation/meta/tests/test_correlatedmultijitter.cpp
```diff
@@ -0,0 +1,117 @@
...
+        sample = cmj_generate_sample<float>(1, 64, 3);
+        EXPECT_FEQ(0.335739,        sample.x);
+        EXPECT_FEQ(0.430547,        sample.y);
+        sample = cmj_generate_sample<float>(16, 17, 5);
+        EXPECT_FEQ(0.972767,        sample.x);
+        EXPECT_FEQ(0.430547,        sample.y);
+    }
+
+    void generate_cmj_image(
+        const size_t    samples_count,
```
dictoon: `sample_count`

src/appleseed/foundation/math/sampling/qmcsamplingcontext.h
oktomus: No, vim ^^

src/appleseed/foundation/math/correlatedmultijitter.h
oktomus: Since it's a reference, you have to exactly match the type when using it. It can be tedious when you are using size_t or int instead of unsigned int.

src/appleseed/foundation/math/sampling/qmcsamplingcontext.h
oktomus: I agree with you. But in `uniformpixelrenderer`, a large number is given to the context as long as a sample count. Since their are not compatible together, should I set the instance to 0 ?

src/appleseed/foundation/math/sampling/qmcsamplingcontext.h
oktomus: It was just to void copied lines.

src/appleseed/foundation/math/sampling/qmcsamplingcontext.h
```diff
@@ -381,6 +421,61 @@ inline Vector<T, N> QMCSamplingContext<RNG>::next2(Tag<Vector<T, N>>)
...
+            v = cmj_generate_sample<T>(
+                m_cmj_params.m_sequence++,
+                m_cmj_params.m_m,
+                m_cmj_params.m_n,
+                m_sample_count,
+                m_cmj_params.m_pattern);
+        }
+        else
+        {
+            if (m_instance < PrecomputedHaltonSequenceSize)
```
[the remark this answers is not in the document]
oktomus: Wow, very clever, thanks !

oktomus:
Thanks for the review @fedormatantsev and @dictoon :) I commented some of your remarks.

src/appleseed/foundation/math/correlatedmultijitter.h
```diff
@@ -0,0 +1,186 @@
...
+    T jx = cmj_randfloat(s, p * 0x967a889b);
+    T jy = cmj_randfloat(s, p * 0x368cc8b7);
+    Vector<T, 2> r;
+    r[0] = (sx + (sy + jx) / n) / m;
+    r[1] = (s + jy) / N;
+    return r;
+}
+
+template <typename T>
+inline Vector<T, 2> cmj_generate_sample(
```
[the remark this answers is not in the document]
oktomus: This code is used in the tests

src/appleseed/foundation/math/sampling/qmcsamplingcontext.h
oktomus: Then what should I do ? Can I create 2 context ?

oktomus: Or why not using base_instance when cmj ?

src/appleseed/foundation/math/sampling/qmcsamplingcontext.h
```diff
@@ -279,7 +333,7 @@ template <typename RNG>
 template <typename T>
 inline T QMCSamplingContext<RNG>::next2()
 {
-    return next2(Tag<T>());
+    return next2(TypeTag<T>());
```
fedormatantsev: My initial idea was to have one "generic" next2 that is used most of the time and some "specific" next2 such as RNG or CMJ. This can be achieved by having one "generic" ```T next2(tag<T, Method>)``` that will consequently call Halton and three specific:
```Vector<T, N> next2(tag<Vector<T, N>, M_Halton>)```
```Vector<T, 2> next2(tag<Vector<T, 2>, M_CMJ>)``` and 
```T next2(tag<T, M_RNG>)```; 
When you pass CMJ with improper signature it fallback to generic next2 automagically since there is no specialization for it; 
So you basically could check if CMJ is enabled and call next2 with M_CMJ right here no matter if the signature is ok;

```inline T QMCSamplingContext<RNG>::next2(Tag<T, Method>)``` should only (dumb) dispatch execution flow to (as it does now)
```inline Vector<T, N> QMCSamplingContext<RNG>::next2(Tag<Vector<T, N>, Method>)``` which dispatches any calls to 
```inline Vector<T, N> QMCSamplingContext<RNG>::next2(Tag<Vector<T, N>, M_Halton>)```

So, there is how it is supposed to work:
1) Call ```next2(<double>)``` -> [if CMJ is enabled] ```next2(<tag<double, M_CMJ>>)``` -> ```(next2<tag<Vector<double, N>, CMJ>>)``` -> [there is no such specialization, fallback to generic]   ```next2(<tag<Vector<double, N>, Method>>)``` -> [Generic will dispatch to Halton]```next2(<tag<Vector<double, N>, M_Halton>>)```
2)  Call ```next2(<double>)``` -> [if CMJ is disabled] ```next2(<tag<double, M_Halton>>)``` -> [there is specialization for Halton, use it] ```next2(<tag<Vector<double, N>, M_Halton>>)```
3)  Call ```next2(<double>)``` -> [mode is RNG] ```next2(<tag<double, M_RNG>>)``` -> [here we go, specialization for RNG] ```next2(<tag<Vector<double, N>, M_RNG>>)```
4) Call ```next2(<Vector2d>) ``` -> [if CMJ is enabled, there is specialization] ```next2(<tag<Vector<double, 2>, M_CMJ>>)```
5) Call ```next2(<Vector2d>) ``` -> [if CMJ is disabled, there is specialization] ```next2(<tag<Vector<double, 2>, M_Halton>>)```
e.t.c

You can notice there is only 1 tag instead of 2; The whole number of functions is 6:
```cpp 
T next2(T) // we set method here
T next2(tag<T, Method>) // here we ONLY dispatch tag<T, already_defined_method> to tag<Vector<T, 1>, already_defined_method>```
Vector<T, N> next2(tag<Vector<T, N>, Method>) // generic, this simply dispatch to Halton
Vector<T, N> next2(tag<Vector<T, N>, M_Halton>)
Vector<T, N> next2(tag<Vector<T, N>, M_RNG>)
Vector<T, 2> next2(tag<Vector<T, 2>, M_CMJ>)
```

fedormatantsev: Another good point of my approach is that you can simply add new methods. Just implement ```cppVector<T, 1> next2(<Vector<T, 1>, M_CMJ>)``` and you get 1d cmj with no hustle:)

src/appleseed/foundation/math/correlatedmultijitter.h
```diff
@@ -0,0 +1,197 @@
...
+Vector<T, 2> cmj_generate_sample(
+    unsigned int        instance,
+    const unsigned int  m,
+    const unsigned int  n,
+    const unsigned int  sample_count,
+    const unsigned int  pattern)
+{
+    assert(m > 1 && n > 1 && sample_count > 1);
+    assert(instance >= 0 && instance < sample_count);
+    instance = cmj_permute(instance, sample_count, pattern * 0x51633e2d);
```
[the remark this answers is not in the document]
oktomus: s isn't really explicit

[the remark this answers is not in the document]
fedormatantsev: Agree

src/appleseed/foundation/math/sampling/qmcsamplingcontext.h
```diff
@@ -279,7 +335,18 @@ template <typename RNG>
 template <typename T>
 inline T QMCSamplingContext<RNG>::next2()
 {
-    return next2(Tag<T>());
+    if (m_cmj_params.m_enabled)
```
fedormatantsev: Can we somehow get m_cmj_params.m_enabled to be true while in RNG sampling mode?

src/appleseed/foundation/math/sampling/qmcsamplingcontext.h
```diff
@@ -332,55 +399,90 @@ inline void QMCSamplingContext<RNG>::compute_offset()
...
+        v[i] = rand2<T>(m_rng);
+
+    ++m_instance;
+
+    return v;
+}
+
+template <typename RNG>
+template <typename T, size_t N>
+inline Vector<T, N> QMCSamplingContext<RNG>::next2(Tag<Vector<T, N>, M_CMJ>)
```
fedormatantsev: No method should be specified here

src/appleseed/foundation/math/sampling/qmcsamplingcontext.h
oktomus: No, you can see that in the constructor of CMJParameters

src/appleseed/foundation/math/sampling/qmcsamplingcontext.h
oktomus: I think it make more sense to fallback here instead of next2(Vec<T, N>, M).
Because if we fallback to Halton in Vec<T, N>, M, then we assume this will never be called if we set the mode to RNG?

fedormatantsev: If RNG is enabled, we explicitly call RNG specialization of next2

oktomus: Correct, but it's just in the code. I guess it's a matter of taste. Unless it's actually better to fallback in Vec<T, N>, M ?

fedormatantsev: Okay; let's think we have another sampling strategy "awesome" that works only with "unicorn" type:) In this scenario one will have to define another "fallback" for the "awesome" sampling strategy;

You are right, it works as it should now, and my suggestion is not about your code - it's good, but for future programmers that probably would implement new sampling strategies as it was done before, so they would define a new excessive function; It's true since the code itself is quite complicated;

I suggest saying "If we don't have a good specialization for next2 call, we use Halton", you saying "If we don't have a good specialization for CMJ next2 call, we use Halton". That's the difference.

But yes, it's a matter of taste anyway.

oktomus: I agree with you that your solution is simpler when it comes to implement a new method :)

src/appleseed/foundation/math/sampling/qmcsamplingcontext.h
```diff
@@ -256,7 +309,8 @@ inline void QMCSamplingContext<RNG>::split_in_place(
     const size_t        dimension,
     const size_t        sample_count)
 {
-    assert(m_sample_count == 0 || m_instance == m_sample_count);    // can't split in the middle of a sequence
+    // Can't split in the middle of a sequence.
```
dictoon: Revert.

src/appleseed/foundation/math/sampling/qmcsamplingcontext.h
```diff
@@ -154,10 +190,23 @@ class QMCSamplingContext
 
     void compute_offset();
 
-    template <typename T> struct Tag {};
+    enum MethodSpecifier
```
dictoon: Use `enum class` and remove `M_` prefixes.

src/appleseed/foundation/math/sampling/qmcsamplingcontext.h
```diff
@@ -265,6 +319,8 @@ inline void QMCSamplingContext<RNG>::split_in_place(
     m_sample_count = sample_count;
     m_instance = 0;
 
+    m_cmj_params = CMJParameters(sample_count, dimension, m_base_instance, m_mode);
```
dictoon: I'm not quite sure to understand how this can work given that in the vast majority of the cases, sample_count is just 1 (grep for `split_in_place` in the code). In that cases CMJ degenerates to pure random no?

Overall I doubt that CMJ is a good fit for our style of general purpose sampler. Maybe it would make sense to use it only for image plane sampling.

oktomus: Yes, I intended to use it only on image plane sampling but at the time i wasn't sure on how to achieve this. I think starting over isn't a bad idea. 

What do you think of the idea of having 2 different samplers ?

oktomus: up @dictoon

PR #1979, opened 2018-04-16
OSL tweaks, texture, luminance, labels

luisbarrancos:
This should fix the parameter name changes. Only the labels were changed, but it'll be a good idea to be thorough to make sure we're on the same wavelenght here (pun intended).
The luminance function was updated, the texture fixes incorporated again.

### DIFF
--- a/src/appleseed.shaders/src/appleseed/as_standard_surface.osl
+++ b/src/appleseed.shaders/src/appleseed/as_standard_surface.osl
@@ -97,6 +97,20 @@ shader as_standard_surface
         int as_maya_attribute_hidden = 1,
         int gafferNoduleLayoutVisible = 0
     ]],
+    int in_subsurface_profile = 0
+    [[
+        string as_maya_attribute_name = "subsurfaceProfile",
+        string as_maya_attribute_short_name = "ssp",
+        string widget = "mapper",
+        string options = "Gaussian:0|Dipole:1|Normalized Diffusion:2|Random Walk:3",
+        string label = "Subsurface Profile",
+        string page = "Subsurface.Advanced",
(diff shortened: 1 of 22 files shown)


src/appleseed.shaders/src/appleseed/as_standard_surface.osl
```diff
@@ -97,14 +97,28 @@ shader as_standard_surface
         int as_maya_attribute_hidden = 1,
         int gafferNoduleLayoutVisible = 0
     ]],
+    int in_subsurface_profile = 0
```
dictoon: Does that mean the default is now Gaussian, whereas it was Normalized Diffusion before? If so, we need to change it back to Normalized Diffusion (2).

src/appleseed.shaders/src/appleseed/as_standard_surface.osl
```diff
@@ -298,7 +312,7 @@ shader as_standard_surface
         float max = 1.0,
         string label = "Coating Reflectivity",
         string page = "Coating",
-        string help = "Coating specular amount."
+        string help = "Coating specular reflectivity."
```
dictoon: A very minor thing, but some help strings end with a dot, some others not.

luisbarrancos: You're right, now it's as good a time to fix this as any other. I'll update the remaining nodes affected by this as well.

src/appleseed.shaders/src/appleseed/as_standard_surface.osl
luisbarrancos: Ok.

PR #1984, opened 2018-04-17
Bundle python on linux.

est77:


scripts/appleseed.package.py
```diff
@@ -724,6 +768,13 @@ def is_system_lib(self, lib):
                 return True
         return False
 
+    def add_python_to_stage(self):
+        progress("Linux-specific: Adding Python 2.7 to staging directory")
+
+        merge_tree(os.path.join(self.settings.python_path, "bin"), "appleseed/bin", symlinks=True)
```
dictoon: Is that really all there is to it? What about Python's standard library, etc.?

est77: The standard library is inside settings.python_path / lib, we copy it in the next line

dictoon: Oh ok, nice.

PR #1994, opened 2018-04-21
Fix shader metadata for Gaffer

luisbarrancos:
It was displaying incorrectly and had missing icons, help messages. It
still has issues dealing with special characters "&" and "/" in the
metadata string values.


src/appleseed.shaders/src/appleseed/as_standard_surface.osl
```diff
@@ -33,8 +33,8 @@ shader as_standard_surface
 [[
     string as_maya_node_name = "asStandardSurface",
     string as_maya_classification = "drawdb/shader/surface:rendernode/appleseed/surface:shader/surface:swatch/AppleseedRenderSwatch",
-    string help = "Standard Surface Shader",
-    string icon = "asStandardSurface",
+    string help = "Appleseed's standard material shader.",
```
dictoon: Lowercase: *appleseed*

PR #2004, opened 2018-04-23
Small Microfacet BRDF fixes and optimizations

est77:
- Move special cases tests as early as possible.
- Prevent possible negative pdfs in evaluate and pdf.


dictoon:
Please fix typo in commit and remove period at the end.

est77:
Fixed

Issue #1971, opened 2018-04-14
Rename Mean Free Path to Depth in SSS shaders


dictoon:
- Rename only the user-visible labels, not the input names.
- If tooltips are available, add "Mean Free Path" in the tooltip.

dictoon:
Done.

PR #2013, opened 2018-04-30
Add as_blender_input_socket metadata

luisbarrancos:
In order to hide the affected node's input, preventing connections.


luisbarrancos:
It also addresses https://github.com/appleseedhq/appleseed/issues/1971

PR #2019, opened 2018-05-03
EC parameter for 3ds max Metal OSL shader

usakhelo:
Hi,
I added energy compensation parameter to max metal OSL shader.
I'll make max plugin PR to support this shortly.

### DIFF
--- a/src/appleseed.shaders/src/max/as_max_metal_material.osl
+++ b/src/appleseed.shaders/src/max/as_max_metal_material.osl
@@ -66,12 +66,6 @@ shader as_max_metal_material
     [[
         float min = 1.0
     ]],
-    float                EnergyCompensation = 0.0
-    [[
-        string help = "Energy Compensation",
-        float  min = 0.0,
-        float  max = 1.0
-    ]],
     float                Anisotropic = 0.0
     [[
         string help = "Anisotropy",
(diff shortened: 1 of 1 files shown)


src/appleseed.shaders/src/max/as_max_metal_material.osl
```diff
@@ -66,6 +66,12 @@ shader as_max_metal_material
     [[
         float min = 1.0
     ]],
+    float                EnergyCompensation = 0.0
```
dictoon: Default value should be 1, I believe. Can you check what's the default value in other shaders supporting EC?

usakhelo: I checked the studio and if you add MetalBSDF EC parameter is 0

usakhelo: in as_metal.osl it's 1.0 by default. I can make max_metal 1 by default too. But what should I do with max materials? Should they be 1 or 0?

luisbarrancos: i.e
```
    out_outColor += as_metal(
        distribution,
        Nn,
        tangent,
        in_face_reflectance,
        in_edge_reflectance,
        in_roughness,
        0.0, // specular spread, used in Student's t-MDF only (unused)
        in_anisotropy_amount,
        "energy_compensation", 1.0);   
```
Also, why not use https://github.com/appleseedhq/appleseed/blob/master/src/appleseed.shaders/src/appleseed/as_metal.osl instead?

luisbarrancos: @usakhelo ok. About the EC, i really regret not having removed that once i did the PR. If you think for instance what happens with Glass without the EC tables, the amount of energy loss is really dramatic.

PR #2021, opened 2018-05-05
appleseed.python Python 3 support

### LINKED ISSUE #2016: Add support for building appleseed.python with Python 3

est77:
By default appleseed.python3 is not built.

closes #2016


CMakeLists.txt
```diff
@@ -115,6 +115,8 @@ option (WITH_HEADERS                        "Install header files (to build samp
 option (WITH_SAMPLES                        "Install sample files (requires headers)"               ON)
 option (WITH_TESTS                          "Install unit tests and benchmarks"                     ON)
 option (WITH_PYTHON                         "Build Python bindings"                                 ON)
```
dictoon: I suggest updating the help message to clarify that this will build Python 2 bindings only.

src/appleseed.python/CMakeLists.txt
```diff
@@ -125,17 +129,37 @@ if (CMAKE_SYSTEM_NAME STREQUAL "Windows")
...
+    endif ()
+endif()
 
 #--------------------------------------------------------------------------------------------------
 # Include paths.
 #--------------------------------------------------------------------------------------------------
 
 include_directories (
-    .
+    .
```
dictoon: Revert?

src/appleseed.python/CMakeLists.txt
```diff
@@ -173,6 +212,9 @@ add_copy_py_file_to_sandbox_py_module_command (${PROJECT_SOURCE_DIR}/src/applese
 add_copy_py_file_to_sandbox_py_module_command (${PROJECT_SOURCE_DIR}/src/appleseed.python/textureconverter.py appleseed)
 add_copy_dir_to_sandbox_py_module_command (${PROJECT_SOURCE_DIR}/src/appleseed.python/test appleseed)
 
+if (WITH_PYTHON3_BINDINGS)
+    add_copy_target_to_sandbox_py_module_command (appleseed.python3 appleseed)
+endif()
```
dictoon: Add space after `endif`.

src/appleseed.python/CMakeLists.txt
```diff
@@ -183,3 +225,7 @@ install (FILES __init__.py DESTINATION ${py_module_dst})
 install (FILES logtarget.py DESTINATION ${py_module_dst})
 install (DIRECTORY test DESTINATION ${py_module_dst})
 install (TARGETS appleseed.python DESTINATION ${py_module_dst})
+
+if (WITH_PYTHON3_BINDINGS)
+    install (TARGETS appleseed.python3 DESTINATION ${py_module_dst})
+endif()
```
dictoon: Add newline character on last line.

est77:
> What's the impact of changing the declaration order of OCIO profiles?

There should be no change, but after updating to OCIO 1.1.0, my new default is always the first colorspace, instead of sRGB as with previous OCIO versions. Changing the order fixes the issue.

CMakeLists.txt
```diff
@@ -114,7 +114,9 @@ option (WITH_DOXYGEN                        "Generate API Reference with Doxygen
 option (WITH_HEADERS                        "Install header files (to build samples)"               ON)
 option (WITH_SAMPLES                        "Install sample files (requires headers)"               ON)
 option (WITH_TESTS                          "Install unit tests and benchmarks"                     ON)
-option (WITH_PYTHON                         "Build Python bindings"                                 ON)
+option (WITH_PYTHON2_BINDINGS               "Build Python bindings for Python 2"                    ON)
```
dictoon: I was thinking of just changing the help text, not the option. This is cleaner but we must update build instructions now.

est77: I checked. We don't use WITH_PYTHON in the instruction. We use the default ON value.

dictoon: Ok cool, thanks.

PR #2024, opened 2018-05-06
Include scenepicker.h only where needed.

est77:
The header indirectly includes OSL headers and it is only used inside studio.
After removing the include from api/rendering.h, the Maya plugin can be built without OSL dependencies.

### DIFF
--- a/src/appleseed.studio/mainwindow/rendering/cameracontroller.h
+++ b/src/appleseed.studio/mainwindow/rendering/cameracontroller.h
@@ -32,6 +32,7 @@
 
 // appleseed.renderer headers.
 #include "renderer/api/rendering.h"
+#include "renderer/api/scenepicker.h"
 
 // appleseed.foundation headers.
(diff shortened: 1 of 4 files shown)


src/appleseed.studio/mainwindow/rendering/cameracontroller.h
```diff
@@ -32,6 +32,7 @@
 
 // appleseed.renderer headers.
 #include "renderer/api/rendering.h"
+#include "renderer/kernel/rendering/scenepicker.h"
```
dictoon: In that case, I would recommend introducing `renderer/api/scenepicker.h` instead. Otherwise the rule that only `renderer/api/` headers can be included in client applications is no longer always true and things get confusing.

An alternative would be to keep `scenepicker.h` in `renderer/api/rendering.h` and try to move as much things as possible to the cpp file, but this may not be possible.

PR #2026, opened 2018-05-07
Add Clamp Roughness parameter to Rendering Settings dialog

usakhelo:
Hi,
I added Clamp Roughness parameter to Rendering Settings/Unidirectional Path Tracer/Advanced/Next Event Estimation.


src/appleseed/renderer/kernel/lighting/pt/ptlightingengine.cpp
```diff
@@ -269,6 +272,8 @@ namespace
             const bool      m_has_max_ray_intensity;
             const float     m_max_ray_intensity;
 
+            const bool      m_clamp_roughness;
```
dictoon: Move after volume parameters to keep the order consistent.

src/appleseed/renderer/kernel/lighting/pt/ptlightingengine.cpp
```diff
@@ -293,6 +298,7 @@ namespace
               , m_distance_sample_count(params.get_optional<size_t>("volume_distance_samples", 2))
               , m_enable_equiangular_sampling(!params.get_optional<bool>("optimize_for_lights_outside_volumes", false))
               , m_record_light_paths(params.get_optional<bool>("record_light_paths", false))
+              , m_clamp_roughness(params.get_optional<bool>("clamp_roughness", false))
```
dictoon: The order of initialization _must_ match the order of declaration.

src/appleseed/renderer/kernel/lighting/pt/ptlightingengine.cpp
```diff
@@ -1161,6 +1167,14 @@ Dictionary PTLightingEngineFactory::get_params_metadata()
             .insert("label", "Next Event Estimation")
             .insert("help", "Explicitly connect path vertices to light sources to improve efficiency"));
 
+    metadata.dictionaries().insert(
+        "clamp_roughness",
+        Dictionary()
+        .insert("type", "bool")
```
dictoon: Indent once more.

src/appleseed/renderer/kernel/lighting/pt/ptlightingengine.cpp
```diff
@@ -1161,6 +1167,14 @@ Dictionary PTLightingEngineFactory::get_params_metadata()
...
+        .insert("default", "false")
+        .insert("label", "Clamp BSDF roughness")
+        .insert("help", "Clamp BSDF roughness paramater to maximum level to avoid fireflies in glossy reflections"));
```
dictoon: *paramater* -> *parameter*

*to **a** maximum level*

*to avoid* -> *to reduce*

src/appleseed/renderer/kernel/lighting/sppm/sppmlightingengine.cpp
```diff
@@ -183,6 +183,7 @@ namespace
                 ~0, // max glossy bounces
                 ~0, // max specular bounces
                 ~0, // max volume bounces
+                false,
```
dictoon: Add a comment explaining what this `false` is for (here and in the other places).

src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
```diff
@@ -956,6 +957,7 @@ namespace
             create_pt_advanced_nee_dl_settings(layout);
             create_pt_advanced_nee_ibl_settings(layout);
             create_pt_advanced_nee_max_ray_intensity_settings(layout);
+            create_pt_advanced_nee_clamp_roughness_settings(layout);
```
dictoon: I'm not sure Clamp Roughness should go inside NEE. I think that even without NEE, Clamp Roughness is effective. Can you confirm?

src/appleseed/renderer/kernel/lighting/sppm/sppmlightingengine.cpp
usakhelo: `// don't clamp roughness` will work? Or `// don't use clamp roughness`?

src/appleseed/renderer/kernel/lighting/lighttracing/lighttracingsamplegenerator.cpp
```diff
@@ -862,6 +865,8 @@ namespace
                 ~0, // max diffuse bounces
                 ~0, // max glossy bounces
                 ~0, // max specular bounces
+                ~0, // max volume bounces
+                false,
```
usakhelo: @dictoon I wanted to point out this change - I had to add `max volume bounces` which means it didn't have enough arguments and `false` was converted to `size_t max_volume_bounces`. Is it possible it was a bug?

src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
usakhelo: Sorry, I'm not sure about that. May be @est77 can answer?

dictoon: What I meant is this: Is roughness clamping working when you disable NEE?

dictoon: I'm going to merge this as it looks good, and we can move Roughness Clamping outside of NEE params afterward.

usakhelo:
@dictoon, there's a bug in the code. Checkbox works the other way around.

PR #2027, opened 2018-05-07
Random-walk SSS enhancements

### LINKED ISSUE #1779: SSS taken into account after diffuse bounce even when diffuse bounce set to 0

biart:
This PR adds random-walk BSSRDF option that uses glass BSDF for surface scattering. (High-order scattering still uses diffuse approximation though)

Another changes:
1) Add volume anisotropy parameter for random-walk BSSRDF (and corresponding reparametrization)
2) Move random-walk SSS to the separate OSL closures: `randomwalk_diffuse` and `randomwalk_glass`
3) Unfortunately, there was a bug in Dwivedi sampling and it is disabled for now
4) Fix #1779 

About Dwivedi sampling:
When Dwivedi sampling is used instead of classical sampling, it gives severe fireflies on low-probability paths. When it is accurately mixed with classical sampling using MIS, some fireflies still remain.

In the previous build, the pdf of Dwivedi sampling was not taken into account properly, which removed fireflies, but made renders incorrect (slightly brighter).

Now, if we want to fix Dwivedi, we need to compare the renders carefully against the references from this PR.

TODO: Fix shaders that depend on random-walk SSS and provide docs.


src/appleseed/renderer/modeling/bssrdf/randomwalkbssrdf.cpp
```diff
@@ -118,18 +125,55 @@ namespace
...
+            return true;
+        }
+
+        void on_frame_end(
+            const Project&              project,
+            const BaseGroup*            parent) override
+        {
+            BSSRDF::on_frame_end(project, parent);
+
+            if (m_use_glass_bsdf) m_glass_bsdf->on_frame_end(project, parent);
```
dictoon: Move `if` body to new line.

src/appleseed/renderer/modeling/bssrdf/randomwalkbssrdf.cpp
```diff
@@ -213,80 +296,102 @@ namespace
...
+                {
+                    return false;
+                }
 
-            // Choose color channel used for distance sampling.
-            sampling_context.split_in_place(1, 2);
-            const float s = sampling_context.next2<float>();
-            const size_t channel = truncate<size_t>(s * Spectrum::size());
-            Spectrum channel_pdf(1.0f / Spectrum::size());
+                if (!volume_scattering_occured) return true;
```
dictoon: Move `if` body to new line (everywhere).

sandbox/tests/test scenes/bssrdf/shaders/subsurface_randomwalk.osl
```diff
@@ -4,7 +4,7 @@ surface subsurface_randomwalk(vector Tn = 0 [[int lockgeom = 0]])
     closure color diff = 0;
 
     diff =
-        0.3 * as_randomwalk_diffuse(N, color(0.5, 0.0, 0.0),  0.10, 1.3) +
+        0.3 * as_subsurface("randomwalk", N, color(0.5, 0.0, 0.0), 0.10, 1.3) +
         0.2 * as_randomwalk_glass(N, color(0.5, 0.3, 0.05), 0.05, 1.3, 1.0, 0.4) +
```
dictoon: Need to fix here too?

dictoon: Oh nevermind, not the same closure!

PR #2037, opened 2018-05-10
Fix warnings with VS on warning level 4

dictoon:


sandbox/samples/cpp/basic/basic.cpp
```diff
@@ -50,6 +50,7 @@
 #include "foundation/math/matrix.h"
 #include "foundation/math/scalar.h"
 #include "foundation/math/transform.h"
+#include "foundation/math/vector.h"
```
oktomus: Did you add this on purpose ?

sandbox/samples/cpp/basic/basic.cpp
```diff
@@ -269,8 +270,7 @@ asf::auto_release_ptr<asr::Project> build_project()
             "beauty",
             asr::ParamArray()
                 .insert("camera", "camera")
-                .insert("resolution", "640 480")
-                .insert("color_space", "srgb")));
+                .insert("resolution", "640 480")));
```
oktomus: I think this is not related to the PR

PR #2040, opened 2018-05-14
Add Albedo AOV

usakhelo:
Hi, this is an attempt to add Albedo AOV.
Albedo component added to the ShadingComponents. Component and a flag added to PathVertex class. Path radiance member is passed to the vertex through new PathTracer::trace() function. AOV component is saved in sample() function of certain shaders, and only copied into PathVertex::m_aov when flag is not set, e.g. copied only once.


src/appleseed/renderer/kernel/lighting/pathtracer.h
```diff
@@ -523,6 +550,9 @@ size_t PathTracer<PathVisitor, VolumeVisitor, Adjoint>::trace(
             const bool continue_path =
                 process_bounce(sampling_context, vertex, bsdf_sample, next_ray);
 
+            if (vertex.m_albedo_aov_saved)
+                radiance.m_albedo = vertex.m_albedo;
```
est77: We could save the albedo inside the PathVertex, as m_albedo.

src/appleseed/renderer/kernel/lighting/pathtracer.h
```diff
@@ -684,6 +714,11 @@ bool PathTracer<PathVisitor, VolumeVisitor, Adjoint>::process_bounce(
             sample);
 
         next_ray.m_max_roughness = m_clamp_roughness ? sample.m_max_roughness : 0.0f;
+        if (sample.m_mode == ScatteringMode::Diffuse && !vertex.m_albedo_aov_saved)
+        {
+            vertex.m_albedo = sample.m_value.m_albedo;
```
est77: What do you think of introducing PathVisitor::on_first_diffuse_bounce(const PathVertex& vertex) and the visitor can then copy the albedo from the vertex to ShadingComponents?
We could undo the changes in PathTracer::trace

src/appleseed/renderer/kernel/lighting/pt/ptlightingengine.cpp
```diff
@@ -835,6 +837,7 @@ namespace
                 clamp_contribution(radiance.m_glossy);
                 clamp_contribution(radiance.m_volume);
                 clamp_contribution(radiance.m_emission);
+                clamp_contribution(radiance.m_albedo);
```
est77: We don't need to clamp the albedo. It should be between 0 and 1 from a start.

src/appleseed/renderer/kernel/lighting/pathtracer.h
usakhelo: I tried to add a method to PathVisitor but then all the lighting engines required to add it to their implementation of PathVisitor. So I had to add dummy methods there. Is that better than modifying PathTracer?
I'll try to add it and add a commit.

src/appleseed/renderer/kernel/lighting/pt/ptlightingengine.cpp
```diff
@@ -343,6 +341,11 @@ namespace
         class PathVisitorBase
         {
           public:
+            void on_first_diffuse_bounce(const PathVertex& vertex)
+            {
+                m_path_radiance.m_albedo = vertex.m_albedo;
+            }
+
```
usakhelo: This is the function.

src/appleseed/renderer/kernel/aov/aovaccumulator.h
```diff
@@ -48,6 +48,7 @@ namespace foundation    { class Image; }
 namespace foundation    { class Tile; }
 namespace renderer      { class Frame; }
 namespace renderer      { class PixelContext; }
+namespace renderer      { class AOVComponents; }
```
est77: Order alphabetically.

src/appleseed/renderer/kernel/lighting/null/nulllightingengine.h
```diff
@@ -39,6 +39,7 @@
 
 // Forward declarations.
 namespace renderer  { class PixelContext; }
+namespace renderer  { class AOVComponents; }
```
est77: Order alphabetically

src/appleseed/renderer/kernel/shading/aovcomponents.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include "aovcomponents.h"
+
+namespace renderer
+{
+
+//
+// AOVComponents class implementation.
+//
+
+    AOVComponents::AOVComponents()
```
est77: Fix indentation

src/appleseed/CMakeLists.txt
```diff
@@ -1238,6 +1238,8 @@ set (renderer_kernel_shading_sources
     renderer/kernel/shading/closures.h
     renderer/kernel/shading/directshadingcomponents.cpp
     renderer/kernel/shading/directshadingcomponents.h
+    renderer/kernel/shading/aovcomponents.cpp
+    renderer/kernel/shading/aovcomponents.h
```
est77: Would it make sense to move these two files to renderer/kernel/aovs/ ?
What do you think?

est77:
Thanks for the PR!
A few small changes and I think it is good to merge.

usakhelo:
Thank you @est77
For some reason standard_surface base color stopped geting into albedo.
It's oren_nayar closure, and it's not getting color... Ok, I need to investigate because before it worked.

src/appleseed/renderer/modeling/bsdf/oslbsdf.cpp
```diff
@@ -209,6 +209,7 @@ namespace
                     modes,
                     sample);
             sample.m_value *= c->get_closure_weight(closure_index);
+            sample.m_aov_components.m_albedo *= c->get_closure_weight(closure_index);
```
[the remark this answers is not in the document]
est77: Looks correct

PR #2054, opened 2018-05-23
Add Clamp Roughness test scene

usakhelo:
Hi,
I wasn't sure where to put the scene, so I put it into `/bsdf/clamp roughness/` folder.
Also it looks too big to me, hdri image can be reduced in size.


dictoon:
Superseded by #2055.

PR #2056, opened 2018-05-24
Replace Color diagnostic mode with Albedo

usakhelo:
Hi,
This is my attempt to make Albedo diagnostic override. I just copied compute_light call from physical shader. Looks like it's working but it takes longer than other diagnostic modes.

### DIFF
--- a/src/appleseed/renderer/modeling/surfaceshader/diagnosticsurfaceshader.cpp
+++ b/src/appleseed/renderer/modeling/surfaceshader/diagnosticsurfaceshader.cpp
@@ -233,25 +233,43 @@ void DiagnosticSurfaceShader::evaluate(
         {
             shading_result.set_main_to_opaque_pink();
 
+            const ShadingRay& ray = shading_point.get_ray();
+            if (!ray.m_has_differentials)
+                break;
+
             const Material* material = shading_point.get_material();
             if (material)
             {
                 const Material::RenderData& material_data = material->get_render_data();
 
+                // Execute the OSL shader if there is one.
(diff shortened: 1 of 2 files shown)


src/appleseed/renderer/modeling/surfaceshader/diagnosticsurfaceshader.cpp
```diff
@@ -235,30 +238,20 @@ void DiagnosticSurfaceShader::evaluate(
...
-                        shading_point);
-                }
-
                 if (material_data.m_bsdf)
                 {
-                    const Vector3f direction = -normalize(Vector3f(shading_point.get_ray().m_dir));
+                    // Compute lighting.
+                    AOVComponents components;
+                    ShadingComponents radiance;
+                    shading_context.get_lighting_engine()->compute_lighting(
```
est77: One possible alternative, would be to call BSDF::sample and fetch the resulting albedo from the BSDFSample.
We'd have albedo for camera rays only, but for a diagnostic mode it could be enough.

usakhelo:
Hi @est77 ,
I took version with BSDFSample from `case RaySpread:`. It seems working but I'm not sure if we need to keep two break conditions. I just don't know what are they checking. 
One is this:
```
if (!ray.m_has_differentials)
   break;
```
And other one is this:
```
if (!sample.m_incoming.has_derivatives())
   break;
```

est77:
The RaySpread mode tries to visualize the size of the ray differentials. 
The checks are there to test if the differentials exists or not.

You don't need the checks for the albedo mode.

PR #2060, opened 2018-05-26
Using CMake generator expressions (win-vs.txt)

boberfly:
Using CMake generator expressions so that it doesn't break when using alternative CMake generators (like NMake/JOM/Ninja) 

https://cmake.org/cmake/help/v3.3/manual/cmake-generator-expressions.7.html#informational-expressions

Hoping this doesn't break things...

### DIFF
--- a/cmake/config/win-vs.txt
+++ b/cmake/config/win-vs.txt
@@ -383,8 +383,8 @@ macro (add_copy_target_exe_to_sandbox_command target)
     # Copy the .exe and the PDB file (except in Ship configuration).
     add_custom_command (TARGET ${target} POST_BUILD
         COMMAND if not exist "${bin_path}" mkdir "${bin_path}"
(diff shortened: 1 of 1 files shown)


boberfly:
Hey @dictoon it looks like 2.8.12 does support them here, they do get referenced in the documentation:
https://cmake.org/cmake/help/v2.8.12/cmake.html

However I don't see the .pdb specific one, but I guess we could work around that.

dictoon:
Thanks! Did you check that all files get copied properly? If yes then we can merge and see how it works. Thanks!

boberfly:
It does with NMake Makefiles JOM but I can't test MSBuild here right now as I'm doing a weird experiment with getting gafferDependencies to build in a linux docker container via wine. I can see if I can get something to work with my Win10 machine though.

PR #2061, opened 2018-05-27
OSL related improvements

est77:
- Exported OSL ShaderCompiler wrapper and bind it to python.
- Compile OSL to bytecode memory strings.
- Extend ShaderQuery to work with memory compiled shaders.


dictoon:
I'm adding your PR description to the commit's description.

Issue #2065, opened 2018-05-30
Add an option to control the precision of displayed statistics


oktomus:
Allow to set the number of displayed digits when adding a statistic in a StatisticsVector.

oktomus:
On it.

PR #2067, opened 2018-05-30
Implement precision settings for stats

oktomus:
#2065

### DIFF
--- a/src/appleseed/foundation/utility/statistics.h
+++ b/src/appleseed/foundation/utility/statistics.h
@@ -155,12 +155,14 @@ class Statistics
     struct PopulationEntry
       : public Entry
     {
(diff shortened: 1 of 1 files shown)


src/appleseed/foundation/utility/statistics.h
```diff
@@ -194,7 +196,14 @@ class Statistics
     void insert(
         const std::string&              name,
         const Population<T>&            value,
-        const std::string&              unit);
+        const std::streamsize           precision);
```
dictoon: Why no default value here?

oktomus: To fallback on the default insert method which doesn't know the type

dictoon:
Thanks!

PR #2069, opened 2018-05-31
Add hint for unsample pixels

oktomus:
Highlight in red unsampled areas 

Pixels where `on_pixel_end` wasn't called. Usefull when implementing a new renderer

### DIFF
--- a/src/appleseed/renderer/kernel/rendering/pixelrendererbase.cpp
+++ b/src/appleseed/renderer/kernel/rendering/pixelrendererbase.cpp
@@ -41,6 +41,7 @@
 #include "foundation/image/colorspace.h"
 #include "foundation/image/tile.h"
 #include "foundation/platform/types.h"
+#include "foundation/utility/otherwise.h"
 
 using namespace foundation;
 
@@ -51,8 +52,9 @@ namespace renderer
 // PixelRendererBase class implementation.
 //
 
-const uint8 InvalidSampleHint = 1;
-const uint8 CorrectSampleHint = 2;
+const uint8 NoState = 0;
+const uint8 InvalidSample = 1;
+const uint8 CorrectSample = 2;
 
 PixelRendererBase::PixelRendererBase(
     const Frame&        frame,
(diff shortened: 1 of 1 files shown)


src/appleseed/renderer/kernel/rendering/pixelrendererbase.cpp
```diff
@@ -51,6 +51,7 @@ namespace renderer
 // PixelRendererBase class implementation.
 //
 
+const uint8 NoSampleHint = 0;
 const uint8 InvalidSampleHint = 1;
 const uint8 CorrectSampleHint = 2;
```
dictoon: Maybe it would be more logical that `CorrectSampleHint` is 0 and all problematic cases get non-zero values?

Also, can we find a better term than "hint"? These are not really hints, more like error codes...

src/appleseed/renderer/kernel/rendering/pixelrendererbase.cpp
```diff
@@ -116,6 +117,10 @@ void PixelRendererBase::on_tile_end(
                     tile.get_pixel(x, y, color);
                     color.rgb().set(0.2f * luminance(color.rgb()));     // 20% of luminance
                 }
+                else if (sample_state[0] == NoSampleHint)
```
dictoon: Simplify the logic, maybe with a `switch` where `color` gets assigned once in each case, instead of having a default value that later gets overwritten or altered.

dictoon:
Nice!

src/appleseed/renderer/kernel/rendering/pixelrendererbase.cpp
oktomus: 0 is because the tile is filled with 0 by default. This avoid to do extra process.

src/appleseed/renderer/kernel/rendering/pixelrendererbase.cpp
oktomus: Yes

src/appleseed/renderer/kernel/rendering/pixelrendererbase.cpp
dictoon: Ok, makes sense.

The comment about "hint" stands though.

src/appleseed/renderer/kernel/rendering/pixelrendererbase.cpp
```diff
@@ -51,8 +52,9 @@ namespace renderer
 // PixelRendererBase class implementation.
 //
 
-const uint8 InvalidSampleHint = 1;
-const uint8 CorrectSampleHint = 2;
+const uint8 NoState = 0;
```
dictoon: NoState? Did you mean NoSample?

oktomus: The pixel can be both unsampled or unfinished. In my case, the pixel was sampled but pixel_end was never called.

PR #2075, opened 2018-06-02
Dynamic link fix for Boost.Python

boberfly:
This helps when building Appleseed against Gaffer Windows

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -193,13 +193,6 @@ set (Boost_MULTITHREADED TRUE)
 
 if (USE_STATIC_BOOST)
     set (Boost_USE_STATIC_LIBS TRUE)
-    if (CMAKE_SYSTEM_NAME STREQUAL "Windows")
-        add_definitions (-DBOOST_PYTHON_STATIC_LIB) # link statically against Boost.Python
-    endif ()
-else ()
-    if (CMAKE_SYSTEM_NAME STREQUAL "Windows")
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -193,6 +193,9 @@ set (Boost_MULTITHREADED TRUE)
 
 if (USE_STATIC_BOOST)
     set (Boost_USE_STATIC_LIBS TRUE)
+    add_definitions(-DBOOST_PYTHON_STATIC_LIB) # link statically against Boost.Python
```
dictoon: That used to be defined on Windows only, now it's going to be defined on all platforms. Not sure what the impact is. @est77?

Please add a space after `add_definitions`.

est77: BOOST_PYTHON_STATIC_LIB doesn't seem to be used on unix, at least in boost 1.61.

dictoon: Ok, thanks. Might still be cleaner to restrict this to MSVC...

CMakeLists.txt
```diff
@@ -193,6 +193,13 @@ set (Boost_MULTITHREADED TRUE)
 
 if (USE_STATIC_BOOST)
     set (Boost_USE_STATIC_LIBS TRUE)
+    if (CMAKE_SYSTEM_NAME STREQUAL "Windows")
```
dictoon: Now it occurs to me: can this be done in `win-vs.txt` instead of here? That's where Windows-specific setup should reside, ideally...

PR #2078, opened 2018-06-04
Implement weighted population

oktomus:


src/appleseed/foundation/math/population.h
```diff
@@ -134,12 +134,12 @@ inline void Population<T>::insert(const ValueType& val)
     if (m_max < val)
         m_max = val;
 
-    // Update the size of the population.
-    ++m_size;
-
     // Compute the residual value.
     const double double_val = static_cast<double>(val);
-    const double residual = double_val - m_mean;
+    const double residual = (double_val - m_mean) * amount;
```
dictoon: I haven't checked the math. Are you sure about this formula?

src/appleseed/foundation/math/population.h
```diff
@@ -57,7 +57,7 @@ class Population
     Population();                           // empty population
 
     // Insert a new value into the population.
-    void insert(const ValueType& val);
+    void insert(const ValueType& val, const size_t amount = 1);
```
dictoon: Rename `amount` to `count` and update the comment ("Insert `count` times the value `val` into the population.")

src/appleseed/foundation/math/population.h
oktomus: I wrote a test or two. Do you want me to write more test ?

src/appleseed/foundation/math/population.h
oktomus: Ok !

src/appleseed/foundation/meta/tests/test_population.cpp
```diff
@@ -106,6 +106,39 @@ TEST_SUITE(Foundation_Math_Population)
...
+        Population<float> pop;
+
+        pop.insert(12.0f, 510);
+        pop.insert(102.0f, 20);
+        pop.insert(-3.0f, 101);
+
+        EXPECT_EQ(631, pop.get_size());
+        EXPECT_FEQ(-3.0f, pop.get_min());
+        EXPECT_FEQ(102.0f, pop.get_max());
+        EXPECT_FEQ_EPS(12.4517, pop.get_mean(), 1.0e-4);
```
dictoon: You're mixing float and double here, it won't compile cleanly with VS.

src/appleseed/foundation/math/population.h
```diff
@@ -56,8 +56,8 @@ class Population
     // Constructor.
     Population();                           // empty population
 
-    // Insert a new value into the population.
-    void insert(const ValueType& val);
+    // Insert count times the value val into the population.
```
dictoon: @gospodnetic started using backticks to reference parameters in comments and since then it has become somewhat of a convention, e.g.

    // Insert `count` times the value `val` into the population.

Helps readability :)

PR #2082, opened 2018-06-06
Small python rendering fixes

est77:


src/appleseed.python/bindaov.cpp
```diff
@@ -36,6 +36,7 @@
 
 // appleseed.foundation headers.
 #include "foundation/platform/python.h"
+#include "foundation/image/image.h"
```
dictoon: Order alphabetically.

src/appleseed.python/bindaov.cpp
```diff
@@ -78,6 +79,18 @@ namespace
     {
         return factory->create(bpy_dict_to_param_array(params));
     }
+
+    bpy::list get_channel_names(const AOV* aov)
+    {
+        bpy::list channels;
+
+        const char **names = aov->get_channel_names();
```
dictoon: `const char** names`

PR #2083, opened 2018-06-08
Implement fisheye lens camera

pjessesco:
Implement fisheye lens camera. 

I worked this file based on `pinholecamera.cpp` copy, so some function still same with it(below line 179). I'll take a look in soon.

reference : 
http://michel.thoby.free.fr/Fisheye_history_short/Projections/Fisheye_projection-models.html
https://de.wikipedia.org/wiki/Fischaugenobjektiv

### DIFF
--- a/src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
+++ b/src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
@@ -194,14 +194,17 @@ namespace
         //   http://michel.thoby.free.fr/Fisheye_history_short/Projections/Fisheye_projection-models.html
         //
 
-        void transform_ray_dir(Vector3d& dir) const
+        Vector3d transform_ray_dir(const Vector3d& dir) const
         {
+            Vector3d new_dir = dir;
+            
             // Transform the direction vector to be in the xz plane.
-            const Transformd rot = Transformd(Matrix4d::make_rotation_z(-atan2 (dir[1], dir[0])));
-            dir = rot.vector_to_parent(dir);
+            // TODO : optimize
+            const Transformd rot = Transformd(Matrix4d::make_rotation_z(-atan2(new_dir[1], new_dir[0])));
+            new_dir = rot.vector_to_parent(new_dir);
 
             // Tangent of angle between vector(on xz plane) and z+ axis.
-            const double tan_theta1 = dir[0] / dir[2];
+            const double tan_theta1 = new_dir[0] / new_dir[2];
             const double theta1 = atan(tan_theta1);
             double theta2 = 0.0;
 
@@ -227,10 +230,13 @@ namespace
                   assert(false);
             }
 
+            // TODO : optimize
             const Transformd rot2 = Transformd(Matrix4d::make_rotation_y(theta2 - theta1));
 
-            dir = rot2.vector_to_parent(dir);
-            dir = rot.vector_to_local(dir);
+            new_dir = rot2.vector_to_parent(new_dir);
+            new_dir = rot.vector_to_local(new_dir);
+            
+            return new_dir;
         }
 
         void spawn_ray(
@@ -250,7 +256,7 @@ namespace
             ray.m_org = transform.get_local_to_parent().extract_translation();
             ray.m_dir = normalize(transform.vector_to_parent(-ndc_to_camera(ndc.get_value())));
 
-            transform_ray_dir(ray.m_dir);
+            ray.m_dir = transform_ray_dir(ray.m_dir);
 
             // Compute ray derivatives.
             if (ndc.has_derivatives())
@@ -292,6 +298,7 @@ namespace
             outgoing = point - transform.get_local_to_parent().extract_translation();
 
             // Compute the emitted importance.
+            // TODO
             const Vector3d film_point = ndc_to_camera(ndc);
             const double square_dist_film_lens = square_norm(film_point);
             const double dist_film_lens = sqrt(square_dist_film_lens);
--- a/src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
+++ b/src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
@@ -194,51 +194,6 @@ namespace
         //   http://michel.thoby.free.fr/Fisheye_history_short/Projections/Fisheye_projection-models.html
         //
 
-        Vector3d transform_ray_dir(const Vector3d& dir) const
-        {
-            Vector3d new_dir = dir;
-            
-            // Transform the direction vector to be in the xz plane.
-            // TODO : optimize
-            const Transformd rot = Transformd(Matrix4d::make_rotation_z(-atan2(new_dir[1], new_dir[0])));
-            new_dir = rot.vector_to_parent(new_dir);
-
-            // Tangent of angle between vector(on xz plane) and z+ axis.
-            const double tan_theta1 = new_dir[0] / new_dir[2];
-            const double theta1 = atan(tan_theta1);
-            double theta2 = 0.0;
-
-            switch (m_projection_type) 
-            {
-                case EquisolidAngle:
-                  theta2 = 2 * asin(tan_theta1 * 0.5);
-                  break;
-                
-                case Equidistant:
-                  theta2 = tan_theta1;
-                  break;
-
-                case Stereographic:
-                  theta2 = 2 * atan(tan_theta1 * 0.5);
-                  break;
-
-                case Thoby:
-                  theta2 = asin(tan_theta1 * 0.68027) * 1.40252;
-                  break;
-
-                default:
-                  assert(false);
-            }
-
-            // TODO : optimize
-            const Transformd rot2 = Transformd(Matrix4d::make_rotation_y(theta2 - theta1));
-
-            new_dir = rot2.vector_to_parent(new_dir);
-            new_dir = rot.vector_to_local(new_dir);
-            
-            return new_dir;
-        }
-
         void spawn_ray(
             SamplingContext&    sampling_context,
             const Dual2d&       ndc,
@@ -256,8 +211,6 @@ namespace
             ray.m_org = transform.get_local_to_parent().extract_translation();
             ray.m_dir = normalize(transform.vector_to_parent(-ndc_to_camera(ndc.get_value())));
 
-            ray.m_dir = transform_ray_dir(ray.m_dir);
-
             // Compute ray derivatives.
             if (ndc.has_derivatives())
             {
@@ -266,9 +219,7 @@ namespace
                 ray.m_rx.m_org = ray.m_org;
                 ray.m_ry.m_org = ray.m_org;
                 ray.m_rx.m_dir = normalize(transform.vector_to_parent(-ndc_to_camera(px)));
-                transform_ray_dir(ray.m_rx.m_dir);
                 ray.m_ry.m_dir = normalize(transform.vector_to_parent(-ndc_to_camera(py)));
-                transform_ray_dir(ray.m_ry.m_dir);
                 ray.m_has_differentials = true;
             }
         }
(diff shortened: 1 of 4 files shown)


src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
```diff
@@ -0,0 +1,446 @@
...
+                "  model                         %s\n"
+                "  film width                    %f\n"
+                "  film height                   %f\n"
+                "  focal length                  %f\n"
+                "  near-z                        %f\n"
+                "  shutter open begin time       %f\n"
+                "  shutter open end time         %f\n"
+                "  shutter close begin time      %f\n"
+                "  shutter close end time        %f\n"
+                "  projection type               %d",
```
pjessesco: How can I log enum type?

src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
```diff
@@ -0,0 +1,446 @@
...
+            {
+                // Tangent of angle between vector(on xz plane) and z+ axis.
+                const double tangent_xz = dir[0]/dir[2];
+                double theta1 = atan(tangent_xz);
+                double theta2 = 2 * asin(tangent_xz / 2.0);
+
+                const Transformd rot2 = Transformd(Matrix4d::make_rotation_y(theta2 - theta1));
+
+                dir = rot2.vector_to_parent(dir);
+                dir = rot.vector_to_local(dir);
```
pjessesco: Maybe there's a better approach, but I have no idea :(

src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
```diff
@@ -0,0 +1,446 @@
...
+                ray.m_ry.m_org = ray.m_org;
+                ray.m_rx.m_dir = normalize(transform.vector_to_parent(-ndc_to_camera(px)));
+                transform_ray_dir(ray.m_rx.m_dir);
+                ray.m_ry.m_dir = normalize(transform.vector_to_parent(-ndc_to_camera(py)));
+                transform_ray_dir(ray.m_ry.m_dir);
+                ray.m_has_differentials = true;
+            }
+        }
+
+        bool connect_vertex(
```
pjessesco: I worked this file based on `pinholecamera.cpp` copy, so it is still same with it. I'll take a look in soon.

src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
rgirish28: Maybe, set a string based on the enum value?

pjessesco:
I kept search, and read the existing code after my exam is over.
There is no big difference with my first commit, but it's hard for me to make progress without reviews.

My approach is just transform ray direction in `spawn_ray()`. It is equisolid fisheye lens for now.
1. Rotate  `ray_dir` to be in the xz plane around z axis. `theta1` is angle between z-axis and transformed vector. 
2. `theta2` is calculated in equisolid equation. Rotate `ray_dir` `theta2 - theta1` around y axis.
3. Perform inverse transform of 1.

pjessesco:
stereographic

stereographic

equal distance

equidistance

equisolid

equisolid

Sorry for noise :(

pjessesco:
Thoby ( super slow :( )

thoby

src/appleseed/renderer/modeling/camera/fisheyelenscamera.h
```diff
@@ -0,0 +1,79 @@
+
+//
+// This source file is part of appleseed.
+// Visit https://appleseedhq.net/ for additional information and resources.
+//
+// This software is released under the MIT license.
+//
+// Copyright (c) 2014-2018 Jino Park, The appleseedhq Organization
```
dictoon: That line should be:

    // Copyright (c) 2018 Jino Park, The appleseedhq Organization

src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
```diff
@@ -0,0 +1,511 @@
+
+//
+// This source file is part of appleseed.
+// Visit https://appleseedhq.net/ for additional information and resources.
+//
+// This software is released under the MIT license.
+//
+// Copyright (c) 2014-2018 Jino Park, The appleseedhq Organization
```
dictoon: Same fix here (see comment in header file).

src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
```diff
@@ -0,0 +1,511 @@
...
+        {
+            return Model;
+        }
+
+        void print_settings() const override
+        {
+            const char* projection_type;
+            switch (m_projection_type)
+            {
+                case EquisolidAngle:
```
dictoon: We use half-indentations for `case` statements:

    switch (m_projection_type)
    {
      case EquisolidAngle:
        projection_type = "equisolid angle";
        break;

      case Equidistant:
        projection_type = "equidistant";
        break;
 
      case Stereographic:
        projection_type = "stereographic";
        break;

      case Thoby:
        projection_type = "thoby";
        break;

      default:
        projection_type = "unknown";
        break;
    }

src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
```diff
@@ -0,0 +1,511 @@
...
+                    projection_type = "stereographic";
+                    break;
+
+                case Thoby:
+                    projection_type = "thoby";
+                    break;
+
+                default:
+                    projection_type = "unknown";
+            }
```
dictoon: Add `break` statement here, for safety reasons.

src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
```diff
@@ -0,0 +1,511 @@
...
+
+            // Precompute pixel area.
+            const size_t pixel_count = project.get_frame()->image().properties().m_pixel_count;
+            m_pixel_area = m_film_dimensions[0] * m_film_dimensions[1] / pixel_count;
+
+            const string projection_type = m_params.get_required<string>("projection_type", "equisolid_angle");
+
+            if (projection_type == "equisolid_angle")
+                m_projection_type = EquisolidAngle;
+
```
dictoon: Remove blank lines between `if` / `else` blocks.

src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
```diff
@@ -0,0 +1,511 @@
...
+                    projection_type.c_str());
+                m_projection_type = EquisolidAngle;
+            }
+
+            return true;
+        }
+
+        //
+        // Transform ray direction in selected projection.
+        // Reference :
```
dictoon: *References*

Remove space before `:`

src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
```diff
@@ -0,0 +1,511 @@
...
+        // Reference : 
+        //   https://wiki.panotools.org/Fisheye_Projection
+        //   https://de.wikipedia.org/wiki/Fischaugenobjektiv
+        //   http://michel.thoby.free.fr/Fisheye_history_short/Projections/Fisheye_projection-models.html
+        //
+
+        void transform_ray_dir(Vector3d& dir) const
+        {
+            // Transform the direction vector to be in the xz plane.
+            const Transformd rot = Transformd(Matrix4d::make_rotation_z(-atan2 (dir[1], dir[0])));
```
dictoon: Remove space between `atan2` and opening parenthesis.

I also think this code should be unrolled, simplified and optimized. Constructing a matrix, then a transform, then transforming is a lot of redundant operations.

Especially since this constructor of `foundation::Transform` is actually inverting a matrix...

src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
```diff
@@ -0,0 +1,511 @@
...
+
+            // Tangent of angle between vector(on xz plane) and z+ axis.
+            const double tan_theta1 = dir[0]/dir[2];
+            const double theta1 = atan(tan_theta1);
+            double theta2 = 0.0;
+
+            switch (m_projection_type) 
+            {
+                case EquisolidAngle:
+                    theta2 = 2 * asin(tan_theta1 / 2.0);
```
dictoon: Prefer multiplying by 0.5 to dividing by 2. That transformation is not done by the compiler with our compilation settings.

src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
```diff
@@ -0,0 +1,511 @@
...
+        //
+
+        void transform_ray_dir(Vector3d& dir) const
+        {
+            // Transform the direction vector to be in the xz plane.
+            const Transformd rot = Transformd(Matrix4d::make_rotation_z(-atan2 (dir[1], dir[0])));
+            dir = rot.vector_to_parent(dir);
+
+            // Tangent of angle between vector(on xz plane) and z+ axis.
+            const double tan_theta1 = dir[0]/dir[2];
```
dictoon: Add spaces around `/`.

src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
```diff
@@ -0,0 +1,511 @@
...
+            const double theta1 = atan(tan_theta1);
+            double theta2 = 0.0;
+
+            switch (m_projection_type) 
+            {
+                case EquisolidAngle:
```
dictoon: Adjust indentation as explained earlier.

src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
```diff
@@ -0,0 +1,511 @@
...
+
+        //
+        // Transform ray direction in selected projection.
+        // Reference : 
+        //   https://wiki.panotools.org/Fisheye_Projection
+        //   https://de.wikipedia.org/wiki/Fischaugenobjektiv
+        //   http://michel.thoby.free.fr/Fisheye_history_short/Projections/Fisheye_projection-models.html
+        //
+
+        void transform_ray_dir(Vector3d& dir) const
```
dictoon: I think it would be cleaner to take a const ref and return the new vector.

src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
```diff
@@ -0,0 +1,511 @@
...
+            dir = rot2.vector_to_parent(dir);
+            dir = rot.vector_to_local(dir);
+        }
+
+        void spawn_ray(
+            SamplingContext&    sampling_context,
+            const Dual2d&       ndc,
+            ShadingRay&         ray) const override
+        {
+
```
dictoon: Remove blank line.

src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
```diff
@@ -0,0 +1,511 @@
...
+                    break;
+
+                default:
+                    RENDERER_LOG_ERROR(
+                        "this message should not be logged"
+                        "using default value \"equisolid_angle\".");
+                    theta2 = 2 * asin(tan_theta1 / 2.0);
+            }
+
+            const Transformd rot2 = Transformd(Matrix4d::make_rotation_y(theta2 - theta1));
```
dictoon: Again, this is very expensive.

src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
```diff
@@ -0,0 +1,511 @@
...
+                case Stereographic:
+                    theta2 = 2 * atan(tan_theta1 / 2.0);
+                    break;
+
+                case Thoby:
+                    theta2 = asin(tan_theta1 / 1.47) / 0.713;
+                    break;
+
+                default:
+                    RENDERER_LOG_ERROR(
```
dictoon: If the only way for this to happen is a programmer error, then use an `assert()` instead. Real error messages are reserved for errors that result from user error.

src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
```diff
@@ -0,0 +1,511 @@
...
+
+            // Retrieve the camera transform.
+            Transformd scratch;
+            const Transformd& transform = m_transform_sequence.evaluate(time, scratch);
+
+            // Compute the outgoing direction vector in world space.
+            outgoing = point - transform.get_local_to_parent().extract_translation();
+
+            // Compute the emitted importance.
+            const Vector3d film_point = ndc_to_camera(ndc);
```
dictoon: You will need to properly compute importance for each type of projection.

src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
```diff
@@ -0,0 +1,511 @@
...
+        double      m_rcp_film_width;       // film width reciprocal in camera space
+        double      m_rcp_film_height;      // film height reciprocal in camera space
+        double      m_pixel_area;           // pixel area in meters, in camera space
+
+        enum Projection
+        {
+            EquisolidAngle, Equidistant, Stereographic, Thoby
+        };
+        Projection m_projection_type;
+
```
dictoon: Remove blank line.

src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
```diff
@@ -0,0 +1,511 @@
...
+        // Precomputed values.
+        double      m_rcp_film_width;       // film width reciprocal in camera space
+        double      m_rcp_film_height;      // film height reciprocal in camera space
+        double      m_pixel_area;           // pixel area in meters, in camera space
+
+        enum Projection
+        {
+            EquisolidAngle, Equidistant, Stereographic, Thoby
+        };
+        Projection m_projection_type;
```
dictoon: Add blank line after enum.

src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
```diff
@@ -0,0 +1,511 @@
...
+        double      m_near_z;               // Z value of the near plane in camera space, in meters
+
+        // Precomputed values.
+        double      m_rcp_film_width;       // film width reciprocal in camera space
+        double      m_rcp_film_height;      // film height reciprocal in camera space
+        double      m_pixel_area;           // pixel area in meters, in camera space
+
+        enum Projection
+        {
+            EquisolidAngle, Equidistant, Stereographic, Thoby
```
dictoon: One enumeration value per line.

src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
pjessesco: Can I handle it like this?

    default:
      assert(false);
      break; // maybe not necessary

dictoon: Yes: if you're sure that it's impossible to reach this code without an earlier programming error, then an assertion is sufficient. I usually leave the `break` as well.

src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
pjessesco: Where can I get some informations about emitted importance? Is it included in PBRT?

src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
pjessesco: Should it be like this? 

https://github.com/appleseedhq/appleseed/blob/adbcd712658c818c7df703ddeece0e997e4c9e00/src/appleseed/foundation/meta/tests/test_transform.cpp#L162

or is it problem of my mathematical approach?

dictoon: I only had a quick look but mathematically your approach seems correct. The problem is that there are a lot of hidden operations behind that single line of code, including a full matrix inversion. You don't need all that.

dictoon: You should look at how `renderer::Transform` works and extracts just what you need from it. You can also probably simplify things a bit since you only have a rotation around the Z axis.

src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
dictoon: There are definitely some good information about this topic in PBRT (look for importance sampling). The "importance" that you must compute is in fact the probability that the world space direction `outgoing` get chosen by your `spawn_ray()` method.

dictoon: Note that `connect_vertex()` is not called by the unidirectional path tracer. To see the problem and test your implementation of `connect_vertex()`, you will need to switch to SPPM. I also suggest using a much simpler scene (such as the Cornell Box) for your tests.

If you render the Cornell Box with SPPM, with your new camera and with the current implementation of `connect_vertex()`, you should notice some brightness issues: maybe some areas will be slightly darker/brighter than they should. It can be subtle, which adds to the difficulty of writing a correct implementation of `connect_vertex()`.

pjessesco: > The "importance" that you must compute is in fact the probability 

Since `transform_ray_dir()` is kind of one-to-one mapping function, isn't it works well?

src/appleseed/renderer/modeling/camera/fisheyelenscamera.cpp
pjessesco: I changed my approach (without using transform), and it seems good to me.

pjessesco:
I'll close this PR and reopen it, since my appleseed version is too old and camera codes have been changed

PR #2084, opened 2018-06-09
Randomwalk diffuse tweaks

est77:
Remove the white edges and the bleaching when mfp is large and the geometry has sharp edges.


est77:
> The only curious thing is the dark borders that appeared in 49 - mfp 002 - randomwalk - ibl - no nee.png and 050 - mfp 002 - randomwalk - ibl - nee.png.

The edges were white before because they were "zero scattering edges". Now we use the reflectance instead of white for them.

dictoon:
Ok. Let's merge this. @biart: would be nice if you could post-review this when you have 5 minutes, it's a one-line change.

PR #2087, opened 2018-06-12
Send extra AOVs to stdout

oktomus:


src/appleseed/renderer/modeling/frame/frame.h
```diff
@@ -107,9 +108,15 @@ class APPLESEED_DLLSYMBOL Frame
     // Access the AOV images.
     ImageStack& aov_images() const;
 
+    // Access extra AOV indexs.
+    std::vector<size_t>& extra_aov_indexs() const;
```
dictoon: There are several problems with this:

1. One plural of index is indexes (not indexs), the other is indices. Both are fine.
2. What are AOV indexes?
3. Why is it non-const?
4. The deal breaker: We cannot use anything from `std` in DLL-exported classes such as `Frame`.

If we really cannot avoid exposing this, then look into the `DECLARE_ARRAY()` macro to declare a new DLL-safe array.

src/appleseed/renderer/modeling/frame/frame.h
```diff
@@ -107,9 +108,15 @@ class APPLESEED_DLLSYMBOL Frame
...
+
     // Create an extra AOV image if it does not exist.
     size_t create_extra_aov_image(const char* name) const;
 
+    // Check if extra AOVs are enabled.
+    bool has_extra_aov() const;
```
dictoon: Is this method really needed? It seems completely redundant with `extra_aov_indexes()`.

src/appleseed.cli/stdouttilecallback.cpp
```diff
@@ -248,6 +270,26 @@ namespace
...
+                // Extra AOVs.
+                if (frame.has_extra_aov())
+                {
+                    const vector<size_t>& diagnostic_index = frame.extra_aov_indexs();
+
+                    for (size_t i = 0, e = diagnostic_index.size(); i < e; ++i)
+                    {
+                        const size_t image_index = diagnostic_index[i];
+                        const Image& image = frame.aov_images().get_image(image_index);
+                        const string aov_name = frame.aov_images().get_name(image_index);
```
dictoon: Looks unused.

src/appleseed.cli/stdouttilecallback.cpp
```diff
@@ -163,6 +167,24 @@ namespace
...
```
dictoon: Take const ref rather than copy.

src/appleseed.cli/stdouttilecallback.cpp
```diff
@@ -141,7 +141,11 @@ namespace
             // This header is sent only once and can contains AOVs and frame informations.
             const bool beauty_only = (m_export_options == StdOutTileCallbackFactory::TileOutputOptions::BeautyOnly);
             const size_t chunk_size = 1 * sizeof(uint32);
-            const size_t plane_count = beauty_only ? 1 : 1 + frame.aovs().size();
+            const size_t plane_count =
+                beauty_only ? 1 : 1
+                + frame.aovs().size()
+                + (frame.has_extra_aov() ? frame.extra_aov_indexs().size() : 0);
```
dictoon: Why not just `+ frame.extra_aov_indexs().size()` (which can be 0 if there is none)?

est77: I prefer that we find an alternative way of doing it. Maybe converting the diagnostic AOVs into normal AOVs or something similar.

Also, note that this won't work in the new blenderseed addon. 
The new addon does not use stdout to pass image data to Blender.

dictoon: Note that `--to-stdout` is not exclusive to blenderseed.

PR #2088, opened 2018-06-12
[SQUASH] Fix template dispatch with icavas::set_pixel

oktomus:


src/appleseed/foundation/image/icanvas.h
```diff
@@ -198,10 +198,12 @@ template <typename T>
 inline void ICanvas::set_pixel(
     const size_t        x,
     const size_t        y,
-    const T             components[])
+    T                   components[])
 {
     const CanvasProperties& props = properties();
 
+    assert(sizeof(T) == props.m_pixel_size);
```
dictoon: This assertion is incorrect in multiple ways:

1. T is the type for one channel, not the whole pixel, hence `sizeof(T)` is typically not equal to `m_pixel_size` (maybe you meant `m_channel_size`?).
2. `set_pixel()` does pixel format conversion using `Pixel::convert_to_format()` so we must not assume that pixel formats match.

src/appleseed/foundation/image/icanvas.h
```diff
@@ -64,43 +65,43 @@ class APPLESEED_DLLSYMBOL ICanvas
...
 
     // Structured read access to a given pixel, with automatic pixel format conversion.
     template <typename Color>
     void get_pixel(
-        const size_t    x,
-        const size_t    y,
-        Color&          color) const;
+        const size_t            x,
+        const size_t            y,
+        Color&                  color) const;
```
dictoon: We should do the same here I guess.

src/appleseed/foundation/image/icanvas.h
```diff
@@ -175,15 +176,15 @@ inline const uint8* ICanvas::pixel(
 #define FOUNDATION_CHECK_PIXEL_SIZE(color) \
```
dictoon: Will this macro still be required if you modify `get_pixel()`? If not, please remove it.

PR #2094, opened 2018-06-18
Update packaging script

jdent02:
This allows the building of an unzipped directory formatted for release in a preset location.  Mainly for workflow simplifications for blenderseed and mayaseed building scripts.

### DIFF
--- a/scripts/appleseed.package.py
+++ b/scripts/appleseed.package.py
@@ -262,10 +262,9 @@ class PackageInfo:
 
 class PackageBuilder:
 
-    def __init__(self, settings, package_info, no_zip):
+    def __init__(self, settings, package_info):
         self.settings = settings
         self.package_info = package_info
-        self.no_zip = no_zip
 
     def build_package(self):
         print("Building package:")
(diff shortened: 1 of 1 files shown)


scripts/appleseed.package.py
```diff
@@ -786,17 +807,17 @@ def main():
...
     settings = Settings()
-    package_info = PackageInfo(settings)
+    package_info = PackageInfo(settings, no_zip)
 
     settings.load()
     package_info.load()
 
     if os.name == "nt":
-        package_builder = WindowsPackageBuilder(settings, package_info)
+        package_builder = WindowsPackageBuilder(settings, package_info, no_zip)
```
dictoon: Do we really need to pass `no_zip` to all package builders if it's already available via `PackageInfo`?

scripts/appleseed.package.py
```diff
@@ -437,6 +446,10 @@ def create_preserve_file(self, path):
     def alter_stage(self):
         return
 
+    def build_final_package(self):
```
dictoon: Maybe more accurate: `deploy_stage_to_package_directory`

scripts/appleseed.package.py
```diff
@@ -437,6 +446,10 @@ def create_preserve_file(self, path):
...
+        progress("Building final output directory from staging directory")
```
dictoon: Consequently: `Deploying staging directory to package directory`

Issue #170, opened 2012-08-13
Allow to pause rendering in appleseed.studio


dictoon:

dictoon:
Worker threads, and thus rendering, can now be paused. What's left is exposing this functionality in appleseed.studio's user interface.

dictoon:
Done.

PR #2096, opened 2018-06-19
Expose pause and resume in studio

oktomus:
- #170 

**The new stopwatch model:**

It contains following values:
- *m_start* is the time when the user call *start* or *resume*
- *m_pause_time* is the time when the user call *pause*
- *m_elapsed* is the spent time of the current run + previous runs
- *m_committed* is the spent time of previous runs

Actions:
- Pressing `start` will set the starting time
- Pressing `start` again will act as a *restart*, it will set the starting time again and `clear` the watch. `start` will also clear the *paused* state
- Pressing `pause` change the state of the stopwatch. Call not allowed if the watch is already *paused*. This also stores the current time in `m_pause_time` for future measures.
- Pressing `resume` change the state of the stopwatch and sets the starting time. Call not allowed if the watch isn't in the *paused* state. It also update `start` time and `committed` spent time
         *committed time = committed_time + (now - start) - (now - pause_time)*
- Pressing `clear` will set all *spent times* to 0 but the watch will keep running if it's not paused
- When calling `measure`, the watch stores the *elapsed time*. The whole spend time is computed as following
         *whole spent time = committed_time + (now - start) - (now - pause_time)*
- `get_ticks` returns the whole spent time (*elapsed*) adjusted by the overhead


src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
```diff
@@ -217,7 +227,8 @@ void RenderingManager::start_rendering(
 
 bool RenderingManager::is_rendering() const
 {
-    return m_master_renderer.get() != nullptr;
+    return m_master_renderer.get() != nullptr
```
dictoon: This is going to be problematic.

When a render is paused, we're still rendering in the sense that a lot of operations (such as editing the scene) are still forbidden.

This probably needs to be reverted.

src/appleseed/foundation/utility/stopwatch.h
```diff
@@ -95,6 +103,10 @@ class Stopwatch
 
 template <typename Timer>
 Stopwatch<Timer>::Stopwatch(const size_t overhead_measures)
+  : m_paused(false)
```
dictoon: Initialization order **must** follow declaration order.

src/appleseed/foundation/utility/stopwatch.h
```diff
@@ -83,6 +89,8 @@ class Stopwatch
     uint64  m_overhead;     // measured overhead of calling start() + measure()
     uint64  m_start;        // timer value when start() is called
     uint64  m_elapsed;      // elapsed time when measure() is called, adjusted for overhead
+    uint64  m_stock;        // elapsed time until pause() is called, adjusted for overhead
```
dictoon: What do you mean (in French maybe) by `stock`?

src/appleseed/foundation/utility/stopwatch.h
```diff
@@ -155,7 +183,10 @@ inline Stopwatch<Timer>& Stopwatch<Timer>::measure()
 template <typename Timer>
 inline uint64 Stopwatch<Timer>::get_ticks() const
 {
-    return m_elapsed;
+    if (m_paused)
```
dictoon: I suggest writing a bunch of unit tests to make sure that pausing a stopwatch works in all cases. You can use a custom timer that returns predetermined values to make your test deterministic and robust.

src/appleseed/renderer/kernel/rendering/defaultrenderercontroller.h
```diff
@@ -59,6 +59,9 @@ class APPLESEED_DLLSYMBOL DefaultRendererController
     // This method is called after rendering was aborted.
     void on_rendering_abort() override;
 
+    void on_rendering_paused() override;
```
dictoon: Add docs.

src/appleseed.studio/mainwindow/rendering/qtrenderercontroller.h
```diff
@@ -62,6 +62,9 @@ class QtRendererController
     // This method is called after rendering was aborted.
     void on_rendering_abort() override;
 
+    void on_rendering_paused() override;
```
dictoon: Add docs.

src/appleseed/foundation/utility/stopwatch.h
oktomus: All the time spent before pressing calling `pause`

dictoon: Ok, we'll need to find a better, more explicit name. I could help with naming but I need to fully understand how it works.

Can we call `pause()` multiple times?

oktomus: Now we can't, I added an assert. But once you resume, you can call `pause` again

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -1630,6 +1638,14 @@ void MainWindow::slot_start_rendering_once(const QString& filepath, const QStrin
     }
 }
 
+void MainWindow::slot_pause_or_resume_rendering()
+{
+    if (m_rendering_manager.is_rendering() && !m_rendering_manager.is_rendering_paused())
```
oktomus: I need an advice for this. I suppose the slot will be called only while rendering (otherwise the widget is disabled). So the first condition `is_rendering` can be removed. What do you think ?

src/appleseed/foundation/utility/stopwatch.h
dictoon: Ok, that's what I meant. It's fine to assert if `pause()` is called while already in pause, but of course we should allow calling `pause()` and `resume()` multiple times. If that's what we do then great.

So, here's an idea:

1. In `measure()`, you we _accumulate_ (add) elapsed time to `m_elapsed`.
2. We move the overhead subtraction to `get_ticks()`.
3. We get rid of `m_stock`.

src/appleseed.studio/mainwindow/mainwindow.cpp
dictoon: I think so, yes. You can check `is_rendering()` in an assert.

src/appleseed/foundation/utility/stopwatch.h
oktomus: Thanks for the idea, I haven't tried to use the overhead in `get_ticks` which is why I had problems.

oktomus: The only issue with that, is that we won't be able to call `measure` if the timer is paused. Otherwise `ticks` would be wrong

dictoon: Unless `measure()` doesn't do anything if paused?

oktomus: That's what I implemented :) pushing tomorrow

src/appleseed.studio/mainwindow/mainwindow.ui
```diff
@@ -404,6 +405,14 @@
     <string>F5</string>
    </property>
   </action>
+  <action name="action_rendering_pause_rendering">
+   <property name="text">
+    <string>Pause &amp;Rendering</string>
```
dictoon: That conflicts with `&Rendering Settings`. I suggest putting the & on the P of Pause.

src/appleseed/foundation/utility/stopwatch.h
```diff
@@ -129,12 +143,37 @@ template <typename Timer>
 inline void Stopwatch<Timer>::clear()
 {
     m_elapsed = 0;
+    m_commited = 0;
 }
 
 template <typename Timer>
 inline void Stopwatch<Timer>::start()
 {
+    m_paused = false;
```
dictoon: From your description, `start()` cannot be called in pause state. This should be an assert.

src/appleseed/foundation/utility/stopwatch.h
```diff
@@ -129,12 +143,37 @@ template <typename Timer>
...
+    m_pause_time = m_timer.read_end();
+    m_paused = true;
+}
+
+template <typename Timer>
+inline void Stopwatch<Timer>::resume()
+{
+    assert(m_paused);
+
+    // Update comitted time.
```
dictoon: Typo: *committed*

src/appleseed/renderer/kernel/rendering/defaultrenderercontroller.h
```diff
@@ -59,6 +59,12 @@ class APPLESEED_DLLSYMBOL DefaultRendererController
     // This method is called after rendering was aborted.
     void on_rendering_abort() override;
 
+    // This method is called after rendering was paused.
+    void on_rendering_paused() override;
```
dictoon: For consistency with the other slots, should be `on_rendering_pause()` and `on_rendering_resume()`.

src/appleseed.studio/mainwindow/mainwindow.ui
oktomus: by the way, what's the purpose of &amp; ?

src/appleseed/foundation/utility/stopwatch.h
oktomus: Actually, I would prefer updating my description because it's really convenient to call `start` to restart everything

dictoon: Ok.

src/appleseed.studio/mainwindow/mainwindow.ui
dictoon: It creates a keyboard shortcut: the character following the & gets underlined:

image

dictoon: That way you can do Alt+F followed by N to create a new project.

PR #2098, opened 2018-06-20
Add Transparency Alpha AOV

usakhelo:
Hi,
This is an attempt to implement "transparency affects alpha" feature using separate AOV.

Basic principle is that AOV is filled with white samples and on `on_miss()` function call it's changed to whatever glossy level was in the last sample.

### DIFF
--- a/src/appleseed/renderer/kernel/aov/aovcomponents.h
+++ b/src/appleseed/renderer/kernel/aov/aovcomponents.h
@@ -38,8 +38,8 @@ namespace renderer
 class AOVComponents
 {
   public:
-      Spectrum  m_albedo;
-      Spectrum  m_refraction_value;
+    Spectrum  m_albedo;
+    Spectrum  m_refraction_value;
 
     // Constructor. Clears all components to 0.
     AOVComponents();
--- a/src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.cpp
+++ b/src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.cpp
@@ -675,6 +675,10 @@ namespace
             {
             }
 
+            void store_aov_components(const PathVertex& vertex)
+            {
+            }
+
             bool accept_scattering(
                 const ScatteringMode::Mode  prev_mode,
                 const ScatteringMode::Mode  next_mode) const
--- a/src/appleseed/renderer/kernel/lighting/pathvertex.h
+++ b/src/appleseed/renderer/kernel/lighting/pathvertex.h
@@ -93,6 +93,9 @@ class PathVertex
     ScatteringMode::Mode        m_aov_mode;
     Spectrum                    m_albedo;
     bool                        m_albedo_saved;
+    Spectrum                    m_sample_refraction_value;
+    Spectrum                    m_refraction_alpha_value;
+    bool                        m_has_non_refraction_sample;
 
     // Constructor.
     explicit PathVertex(SamplingContext& sampling_context);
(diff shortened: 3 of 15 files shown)


src/appleseed/renderer/kernel/aov/aovcomponents.h
```diff
@@ -38,7 +38,8 @@ namespace renderer
 class AOVComponents
 {
   public:
-    Spectrum m_albedo;
+      Spectrum  m_albedo;
+      Spectrum  m_refraction_value;
```
oktomus: Align with the method below :)

usakhelo:
Thanks @oktomus, fixed that. And hopefully travis build as well..

dictoon:
@est77 Why is this on hold?

src/appleseed/renderer/kernel/aov/aovcomponents.h
```diff
@@ -38,7 +38,8 @@ namespace renderer
 class AOVComponents
 {
   public:
-    Spectrum m_albedo;
+    Spectrum  m_albedo;
+    Spectrum  m_refraction_value;
```
dictoon: Maybe just `m_refraction`?

src/appleseed/renderer/modeling/bsdf/bsdfsample.h
```diff
@@ -61,6 +61,8 @@ class BSDFSample
     // Roughness.
     float                           m_max_roughness;        // BSDF roughness
 
+    bool                            m_is_refraction_sample;
```
dictoon: Is this an input or an output? I suspect an output. Move it to the right section.

src/appleseed/renderer/modeling/bsdf/bsdfsample.h
```diff
@@ -61,6 +61,8 @@ class BSDFSample
     // Roughness.
     float                           m_max_roughness;        // BSDF roughness
```
dictoon: Unrelated to your PR, but this should be moved to the right section as well. Maybe you can do it in this PR, in a separate commit?

src/appleseed/renderer/modeling/bsdf/bsdfsample.h
```diff
@@ -99,6 +101,7 @@ inline BSDFSample::BSDFSample(
   , m_shading_basis(shading_point->get_shading_basis())
   , m_outgoing(outgoing)
   , m_mode(ScatteringMode::None)
+  , m_is_refraction_sample(false)
```
dictoon: It is really important to initialize members in the same order as they are declared. Here it has no consequences but in other places it's critical to do so, so best keep the habit.

src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.cpp
```diff
@@ -675,6 +675,10 @@ namespace
             {
             }
 
+            void store_aov_components(const PathVertex& vertex)
```
dictoon: Maybe move after `on_scatter()` since this is typically called last? Please check and confirm. The general idea is to order methods logically, i.e. following some kind of logical progression.

src/appleseed/renderer/kernel/lighting/pathvertex.h
```diff
@@ -93,6 +93,9 @@ class PathVertex
     ScatteringMode::Mode        m_aov_mode;
     Spectrum                    m_albedo;
     bool                        m_albedo_saved;
+    Spectrum                    m_sample_refraction_value;
```
dictoon: We need to be very careful about the amount of data we store in this structure since we may keep large numbers of path vertices in memory in bidirectional tracing methods. Why do we need these two?

src/appleseed/renderer/kernel/lighting/pathvertex.h
```diff
@@ -93,6 +93,9 @@ class PathVertex
...
+    Spectrum                    m_refraction_alpha_value;
```
dictoon: Alpha value as a spectrum?

usakhelo:
Hi @dictoon ,
I'll close this PR because I completely rewrote it and will post new version soon.

Thanks,
Sergo.

PR #2099, opened 2018-06-22
Implement and test reading and writting exr attributes

oktomus:
Reading is easy, you just need to know the type so that `to_string` works correctly.
For writing, the fallback is to write the attribute as a `string`, otherwise we try to convert it. And specific cases (chromatics for instance) have to be handled before.

### DIFF
--- a/src/appleseed/foundation/image/exrutils.cpp
+++ b/src/appleseed/foundation/image/exrutils.cpp
@@ -67,6 +67,40 @@ namespace
                 static_cast<int>(System::get_logical_cpu_core_count()));
         }
     };
+
+    void add_attribute(Header& header, const string& attr_name, const string& attr_value)
+    {
+        // Try to guess the type of the value represented by the string.
+
+        try // Vector3
+        {
+            const Vector3d v = from_string<Vector3d>(attr_value);
+            header.insert(attr_name.c_str(), V3dAttribute(Imath::V3d(v[0], v[1], v[2])));
+            return;
+        }
+        catch (ExceptionStringConversionError&) {}
+
+        try // Vector2
+        {
+            const Vector2d v = from_string<Vector2d>(attr_value);
+            header.insert(attr_name.c_str(), V2dAttribute(Imath::V2d(v[0], v[1])));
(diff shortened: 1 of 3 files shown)


src/appleseed/foundation/image/exrutils.cpp
```diff
@@ -67,6 +67,43 @@ namespace
...
+
+        try // Vector3
+        {
+            const Vector3d v = from_string<Vector3d>(attr_value);
+            header.insert(attr_name.c_str(), V3dAttribute(Imath::V3d(v[0], v[1], v[2])));
+            ok = true;
+        }
+        catch (ExceptionStringConversionError&) {}
+
+        try // Vector2
```
dictoon: You probably want to wrap this in `if (!ok)`.

src/appleseed/foundation/image/genericprogressiveimagefilereader.cpp
```diff
@@ -197,8 +199,40 @@ void GenericProgressiveImageFileReader::read_image_attributes(
     ImageAttributes&    attrs)
 {
     assert(is_open());
+    assert(attrs.size() == 0);
```
dictoon: Better: `assert(attrs.empty());`

src/appleseed/foundation/image/genericprogressiveimagefilereader.cpp
```diff
@@ -197,8 +199,40 @@ void GenericProgressiveImageFileReader::read_image_attributes(
...
+    const OIIO::ImageSpec& spec = impl->m_input->spec();
+
+    for (size_t i = 0; i < spec.extra_attribs.size(); ++i)
+    {
+        const OIIO::ParamValue& p(spec.extra_attribs[i]);
+
+        switch (p.type().basetype)
+        {
+          case OIIO::TypeDesc::STRING:
+            attrs.insert(p.name().c_str(), *(const char **) p.data());
```
dictoon: Use `static_cast<>` or `reinterpret_cast<>` instead of C-style casts.

PR #2102, opened 2018-06-25
Add tick marks to color map legend

dictoon:


src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -1933,8 +1934,9 @@ void MainWindow::slot_render_widget_context_menu(const QPoint& point)
         return;
 
     QMenu* menu = new QMenu(this);
-    menu->addAction("Save Frame...", this, SLOT(slot_save_frame()));
-    menu->addAction("Save All AOVs...", this, SLOT(slot_save_all_aovs()));
+    menu->addAction("Save Post-Processed Frame...", this, SLOT(slot_save_post_processed_frame()));
+    menu->addAction("Save Raw Frame...", this, SLOT(slot_save_raw_frame()));
```
oktomus: It's not Raw if you save it as png ?

dictoon: That's right, however you would typically save to OpenEXR.

Issue #2103, opened 2018-06-26
Allow to merge percent statistics


oktomus:
At the moment, percent stats aren't merged because they are stored as strings.

PR #2107, opened 2018-06-28
Allow to merge percent statistics

oktomus:
#2103 
When merging percent entries, numerators and denominators are simply summed. Be sure to use the same scale when adding percent values.

### DIFF
--- a/src/appleseed/foundation/meta/tests/test_statistics.cpp
+++ b/src/appleseed/foundation/meta/tests/test_statistics.cpp
@@ -186,7 +186,7 @@ TEST_SUITE(Foundation_Utility_Statistics)
         EXPECT_EQ("  existing value                17,042", stats.to_string());
     }
 
-    TEST_CASE(Merge_GivenExistingPercentStatistic_InsertsIt)
+    TEST_CASE(Merge_GivenNewPercentStatistic_InsertsIt)
     {
         Statistics stats;
         stats.insert_percent("existing value", 5, 10, 1);
--- a/src/appleseed/foundation/meta/tests/test_statistics.cpp
+++ b/src/appleseed/foundation/meta/tests/test_statistics.cpp
@@ -185,6 +185,32 @@ TEST_SUITE(Foundation_Utility_Statistics)
 
         EXPECT_EQ("  existing value                17,042", stats.to_string());
     }
+
+    TEST_CASE(Merge_GivenExistingPercentStatistic_InsertsIt)
+    {
+        Statistics stats;
+        stats.insert_percent("existing value", 5, 10, 1);
+
+        Statistics other_stats;
+        other_stats.insert_percent("new value", 14, 20, 1);
+
+        stats.merge(other_stats);
+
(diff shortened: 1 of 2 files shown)


src/appleseed/foundation/utility/statistics.h
```diff
@@ -136,6 +136,25 @@ class Statistics
...
+    {
+        T m_numerator;
+        T m_denominator;
+        std::streamsize     m_precision;
+
+        PercentEntry(
+            const std::string&          name,
+            const T&                    numerator,
+            const T&                    denominator,
+            const std::streamsize       m_precision);
```
dictoon: No `m_` prefix on arguments.

src/appleseed/foundation/utility/statistics.h
```diff
@@ -136,6 +136,25 @@ class Statistics
         std::string to_string() const override;
     };
 
+    template <typename T>
+    struct PercentEntry
+      : public Entry
+    {
+        T m_numerator;
+        T m_denominator;
+        std::streamsize     m_precision;
```
dictoon: Vertically align data members.

src/appleseed/foundation/utility/statistics.h
```diff
@@ -431,6 +452,46 @@ const T* Statistics::Entry::cast(const Entry* entry)
...
+{
+    return std::unique_ptr<Statistics::Entry>(new PercentEntry(*this));
+}
+
+template <typename T>
+void Statistics::PercentEntry<T>::merge(const Entry* other)
+{
+    const PercentEntry* pother = cast<PercentEntry>(other);
+
+    m_numerator = m_numerator * pother->m_denominator + pother->m_numerator * m_denominator;
```
dictoon: The problem of this approach is that each time entries are merged, numerator and denominator are getting larger. Ideally we would simplify the fraction afterward (divide by greatest common divisors).

Whether it'll work reliably as is depends on the width of type `T`. What `T` do we have in practice?

oktomus: I'm using size_t in the adaptive sampler, but I should take a look on other uses. Also, I could use the final ratio in only 1 variable and call `pretty_ratio`. But i'm not sure how to add values if we don't keep the denominator and numerator

dictoon: Well you need to keep the numerator and denominator obviously, you just need to simplify the fraction after every merge. I don't think it's necessary in practice since we're not merging millions of values together.

src/appleseed/foundation/meta/tests/test_statistics.cpp
```diff
@@ -185,6 +185,32 @@ TEST_SUITE(Foundation_Utility_Statistics)
 
         EXPECT_EQ("  existing value                17,042", stats.to_string());
     }
+
+    TEST_CASE(Merge_GivenExistingPercentStatistic_InsertsIt)
```
dictoon: The title is incorrect here: it should be `Merge_GivenNewPercentStatistic_InsertsIt`

Issue #2073, opened 2018-06-01
File type ignored when saving an image


oktomus:
Even if png is specified but the extension isn't written by the user, the image is exported as exr
![2018-06-01-10 35 23](https://user-images.githubusercontent.com/4656466/40830875-ca43aba6-6587-11e8-818b-8dfab88c6818.png)
![2018-06-01-10 35 37](https://user-images.githubusercontent.com/4656466/40830881-ce137eb4-6587-11e8-98ea-657048086feb.png)

dictoon:
Definitely working for me.

dictoon:
Only AOVs are always written as OpenEXR files.

oktomus:
Yes I know, but here in the screenshots it's the final image. Even when having the same configuration as in the screenshots your image is being saved as png ?

dictoon:
> Even when having the same configuration as in the screenshots your image is being saved as png

Yes, but I'm on Windows. You should ask someone on Linux.

est77:
It seems the problem is here:
https://github.com/appleseedhq/appleseed/blob/master/src/appleseed.studio/mainwindow/mainwindow.cpp#L1961

I am wondering how this code works on windows? It doesn't look like we use the selected filter to set the extension anywhere.

dictoon:
It looks like by default it picks the first extension from the filter.

dictoon:
> It doesn't look like we use the selected filter to set the extension anywhere.

We certainly don't need to do anything special on Windows. Can you investigate what's the expected behavior on Linux, and whether we need to do something special there?

oktomus:
@est77 or myself ?

dictoon:
Whoever has time :)

oktomus:
It seems that the automatic extension is only working on windows.
https://stackoverflow.com/questions/7234381/specify-default-extension-in-qfiledialoggetsavefilename#15904546
It would be difficult to get the final selected filter to set the extension

oktomus:
Actually, it won't be difficult, I can try :)

PR #2108, opened 2018-06-28
Add an extension based on user selected filter when requesting a saving path

oktomus:
Fix for #2073

### DIFF
--- a/src/appleseed.studio/utility/miscellaneous.cpp
+++ b/src/appleseed.studio/utility/miscellaneous.cpp
@@ -305,7 +305,7 @@ QString get_save_filename(
     const QString dir = get_value(settings, settings_key + SETTINGS_LAST_DIRECTORY);
     QString selected_filter = get_value(settings, settings_key + SETTINGS_SELECTED_FILTER);
 
-    const QString filepath =
+    QString filepath =
         QFileDialog::getSaveFileName(
             parent,
             caption,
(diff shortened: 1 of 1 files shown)


src/appleseed.studio/utility/miscellaneous.cpp
```diff
@@ -324,6 +324,21 @@ QString get_save_filename(
             QDir::toNativeSeparators(QFileInfo(filepath).path()));
     }
 
+    QFileInfo finfo(filepath);
+
+    // Check if there is any extension.
+    // Use the selected filter as the extension only if the filter has one type.
+    if (finfo.suffix().isEmpty() && selected_filter.count('*') == 1)
+    {
+        int start = selected_filter.indexOf("*") + 1;
```
dictoon: begin/end is more common.

src/appleseed.studio/utility/miscellaneous.cpp
```diff
@@ -324,6 +324,21 @@ QString get_save_filename(
...
+        int end = selected_filter.indexOf(")", start);
+
+        if (start != 0 && end > start)
```
dictoon: I think `start` cannot be 0 since it is guaranteed that there is one `*`.

dictoon:
Never mind that last comment, on second though there's nothing special about this case.

src/appleseed.studio/utility/miscellaneous.cpp
oktomus: Let me use an insert then to be sure that the filter is correct

PR #2112, opened 2018-07-06
Allow the camera controller to use the project's camera or a given one

oktomus:
This fixes crashes occurring in interactive mode while editing the
camera.


oktomus:
But what if we set a camera and the pointer is no longer valid ? The controller will use the project's camera. I don't know if that could happen or not... What do you think ?

PR #2113, opened 2018-07-07
Minor tweak to packaging script

jdent02:


scripts/appleseed.package.py
```diff
@@ -446,8 +446,11 @@ def alter_stage(self):
         return
 
     def deploy_stage_to_package_directory(self):
+    	staging_directory = os.path.join(self.settings.package_output_path, "appleseed")
```
dictoon: That should be called `package_directory`.

dictoon: Indentation needs fixing. I'm surprised that Python didn't complain.

jdent02:
Updates in

scripts/appleseed.package.py
```diff
@@ -29,8 +29,7 @@
 #
 
 from __future__ import print_function
-from distutils import archive_util, dir_util
-from xml.etree.ElementTree import ElementTree
+
```
dictoon: What's the rationale behind moving these imports around?

jdent02: No rationale. I did a code format in PyCharm and it moved it

dictoon: Doesn't make much sense to separate them like that :) Let's revert this. Feel free to merge the PR afterward (have you got the rights to do so?)

jdent02: I’m pretty sure I do

PR #2114, opened 2018-07-07
BinaryMesh file format revision 4 (single precision geometry)

dictoon:


src/appleseed/foundation/mesh/binarymeshfilewriter.cpp
```diff
@@ -48,6 +48,12 @@ namespace foundation
 // BinaryMeshFileWriter class implementation.
 //
 
+namespace
+{
+    // Version of the BinaryMesh file format being written by this code.
+    const uint16 Version = 4;
+}
+
```
rgirish28: Not entirely sure why the Version was shifted here, except for readability as no use of Version except in write_version()

dictoon: I moved it at the code here just to make it stand out.

rgirish28: Cool 👍

dictoon:
Thanks for the review!

PR #2115, opened 2018-07-08
Linux linker visibility map fixes

est77:
Hide symbols from libtiff, libpng and libjpeg in libappleseed.so
Fixes possible errors and crashes when using libappleseed.so in host applications 
that use different versions of the above libraries.

### DIFF
--- a/src/appleseed/libappleseed.map
+++ b/src/appleseed/libappleseed.map
@@ -1,9 +1,4 @@
 {
-global: *;
-local:
-    extern "C++"
-    {
-        OSL*;
-        boost*;
-    };
+    global: *;
+    local: TIFF*; _TIFF*; _tiff*; png*; jpeg*;
 };


src/appleseed/libappleseed.map
```diff
@@ -1,9 +1,4 @@
 {
-global: *;
-local:
-    extern "C++"
-    {
-        OSL*;
-        boost*;
-    };
+    global: *;
+    local: TIFF*; _TIFF*; _tiff*; png*; jpeg*;
```
dictoon: No longer hiding `OSL*` and `boost*`?

est77: By default we don't use this file in builds. We were not hidding (and still don't) any symbol.
Hiding boost and OSL was not working. They are not a problem in any case, their versions are fixed in the vfx platform.

I tested that with the changes here and enabling the USE_VISIBILITY_MAP cmake option 
symbols from C image IO libs are correctly hidden.

dictoon: Should `USE_VISIBILITY_MAP` be ON by default? Should the `experimental` status from the option's description be removed?

est77: I was planning to do it after testing a few days locally.

PR #2118, opened 2018-07-10
[WIP] Allow to resume renders with cli

oktomus:


src/appleseed/renderer/kernel/rendering/generic/genericframerenderer.cpp
```diff
@@ -320,14 +331,30 @@ namespace
...
                     if (m_abort_switch.is_aborted())
                     {
                         m_is_rendering = false;
                         return;
                     }
+                    else if(pass > m_pass_start)
+                    {
+                        // Create a resumable pass renderer if required.
+                        RENDERER_LOG_INFO("creating a resumable render starting at pass %s", pretty_uint(pass + 1).c_str());
+                        bf::path bf_file_path("/tmp/resumable.exr");
```
dictoon: Won't work on Windows.

src/appleseed.studio/mainwindow/mainwindow.ui
```diff
@@ -304,6 +305,11 @@
     <string>&amp;Open Project...</string>
    </property>
   </action>
+  <action name="action_file_open_temporary_render">
+   <property name="text">
+    <string>&amp;Open Temporary Render...</string>
```
dictoon: The `O` shortcut is already taken.

src/appleseed/renderer/kernel/rendering/generic/tilejob.h
```diff
@@ -73,6 +73,9 @@ class TileJob
     // Execute the job.
     void execute(const size_t thread_index) override;
 
+    // Check if the render wasn't aborted during this tile.
+    bool is_render_successfull() const;
```
dictoon: Successful (one L)

src/appleseed.studio/utility/miscellaneous.cpp
```diff
@@ -170,6 +170,17 @@ QString get_project_files_filter(const int filters)
     return filter_list.join(";;");
 }
 
+QString get_resumable_render_files_filter()
```
dictoon: Unless this is needed in several places, move it to where it's needed. You can even get rid rid of the function and simply pass the filter as a string.

PR #2120, opened 2018-07-12
Expose checkpoint option and save/restore renders in cli

oktomus:
- [x] Correctly write buffer weights in the checkpoint EXR
- [x] Correctly load weights from the checkpoint
- [x] Correctly save filtered AOVs
- [x] Correctly save un-filtered AOVs
- [x] Correctly load filtered AOVs
- [x] Correctly load un-filtered AOVs
- [x] Allow only `.exr` files for checkpoints
- [x] Add custom message when a checkpoint is invalid
- [x] Compatibility: Check for layer count 
- [x] Color AOVs: don't recreate the same image multiple times
- [x] Support for denoiser AOV
- [x] Reading: use a map containing for each subimage the layer name
- [x] After render finished, rename previously created checkpoint file to not override it
- [x] Restore and save from the frame 

#### Future Work
- Expose option in studio
- Save denoiser checkpoint in the same exr file
- Allow to resume progressive rendering
- Suggested by @est77 : Add checkpoint paths in the project file
- Handle interruption happening while writing a checkpoint file (a solution could be to stop Ctrl-C during that moment). 

#2355

### DIFF
--- a/src/appleseed.cli/main.cpp
+++ b/src/appleseed.cli/main.cpp
@@ -267,7 +267,7 @@ namespace
         if (g_cl.m_passes.is_set())
         {
             params.insert_path(
-                "generic_frame_renderer.passes",
+                "passes",
                 g_cl.m_passes.values()[0]);
         }
 
@@ -296,7 +296,7 @@ namespace
         }
     }
 
-    void apply_frame_command_line_options(Project& project)
+    bool apply_frame_command_line_options(Project& project)
     {
         const Frame* frame = project.get_frame();
         assert(frame != nullptr);
@@ -327,6 +327,53 @@ namespace
             params.insert("noise_seed", noise_seed);
         }
 
+        if (g_cl.m_output.is_set())
+        {
+            const char* file_path = g_cl.m_output.value().c_str();
+            params.insert("output_path", file_path);
+        }
+
+        if (g_cl.m_checkpoint_create.is_set())
+        {
+            params.insert("checkpoint_create", "on");
+
+            if (g_cl.m_checkpoint_create.values().empty() && !g_cl.m_output.is_set())
+            {
+                LOG_ERROR(g_logger, "ouput path must be specified if no path is given for --checkpoint-create");
+                return false;
+            }
+
+            params.insert(
+                "checkpoint_create_path",
+                !g_cl.m_checkpoint_create.values().empty()
+                    ? g_cl.m_checkpoint_create.value().c_str()
+                    : "");
+        }
+
+        if (g_cl.m_checkpoint_resume.is_set())
+        {
+            params.insert("checkpoint_resume", "on");
+
+            if (g_cl.m_checkpoint_resume.values().empty() && !g_cl.m_output.is_set())
+            {
+                LOG_ERROR(g_logger, "ouput path must be specified if no path is given for --checkpoint-resume");
+                return false;
+            }
+
+            params.insert(
+                "checkpoint_resume_path",
+                !g_cl.m_checkpoint_resume.values().empty()
+                    ? g_cl.m_checkpoint_resume.value().c_str()
+                    : "");
+        }
+
+        if (g_cl.m_passes.is_set())
+        {
+            params.insert_path(
+                "passes",
+                g_cl.m_passes.values()[0]);
+        }
+
         auto_release_ptr<Frame> new_frame(
             FrameFactory::create(
                 frame->get_name(),
@@ -334,6 +381,8 @@ namespace
                 frame->aovs()));
 
         project.set_frame(new_frame);
+
+        return true;
     }
 
     void apply_visibility_command_line_options_recursive(
@@ -374,15 +423,17 @@ namespace
         }
     }
 
-    void apply_command_line_options(Project& project, ParamArray& params)
+    bool apply_command_line_options(Project& project, ParamArray& params)
     {
         // Apply command line options that alter renderer settings.
         apply_rendering_settings_command_line_options(params);
         apply_custom_parameter_command_line_options(params);
 
         // Apply command line options that alter the project.
-        apply_frame_command_line_options(project);
+        if (!apply_frame_command_line_options(project)) return false;
         apply_visibility_command_line_options(project);
+
+        return true;
     }
 
 #if defined __APPLE__ || defined _WIN32
@@ -450,9 +501,7 @@ namespace
         params.merge(configuration->get_parameters());
 
         // Apply the command line options.
-        apply_command_line_options(project, params);
-
-        return true;
+        return apply_command_line_options(project, params);
     }
 
     bool is_progressive_render(const ParamArray& params)
@@ -714,3 +763,4 @@ int main(int argc, char* argv[])
 
     return success ? 0 : 1;
 }
+
--- a/src/appleseed.cli/main.cpp
+++ b/src/appleseed.cli/main.cpp
@@ -339,7 +339,10 @@ namespace
 
             if (g_cl.m_checkpoint_create.values().empty() && !g_cl.m_output.is_set())
             {
-                LOG_ERROR(g_logger, "output path must be specified if no path is given for --checkpoint-create");
+                LOG_ERROR(
+                    g_logger,
+                    "output path must be specified if no path is given for %s",
+                    g_cl.m_checkpoint_create.get_name().c_str());
                 return false;
             }
 
@@ -356,7 +359,10 @@ namespace
 
             if (g_cl.m_checkpoint_resume.values().empty() && !g_cl.m_output.is_set())
             {
-                LOG_ERROR(g_logger, "output path must be specified if no path is given for --checkpoint-resume");
+                LOG_ERROR(
+                    g_logger,
+                    "output path must be specified if no path is given for %s",
+                    g_cl.m_checkpoint_resume.get_name().c_str());
                 return false;
             }
 
--- a/src/appleseed.cli/main.cpp
+++ b/src/appleseed.cli/main.cpp
@@ -335,44 +335,40 @@ namespace
 
         if (g_cl.m_checkpoint_create.is_set())
         {
-            params.insert("checkpoint_create", "on");
+            params.insert("checkpoint_create", true);
 
             if (g_cl.m_checkpoint_create.values().empty() && !g_cl.m_output.is_set())
             {
-                LOG_ERROR(g_logger, "ouput path must be specified if no path is given for --checkpoint-create");
+                LOG_ERROR(g_logger, "output path must be specified if no path is given for --checkpoint-create");
                 return false;
             }
 
             params.insert(
                 "checkpoint_create_path",
                 !g_cl.m_checkpoint_create.values().empty()
-                    ? g_cl.m_checkpoint_create.value().c_str()
+                    ? g_cl.m_checkpoint_create.value()
                     : "");
         }
 
         if (g_cl.m_checkpoint_resume.is_set())
         {
-            params.insert("checkpoint_resume", "on");
+            params.insert("checkpoint_resume", true);
 
             if (g_cl.m_checkpoint_resume.values().empty() && !g_cl.m_output.is_set())
             {
-                LOG_ERROR(g_logger, "ouput path must be specified if no path is given for --checkpoint-resume");
+                LOG_ERROR(g_logger, "output path must be specified if no path is given for --checkpoint-resume");
                 return false;
             }
 
             params.insert(
                 "checkpoint_resume_path",
                 !g_cl.m_checkpoint_resume.values().empty()
-                    ? g_cl.m_checkpoint_resume.value().c_str()
+                    ? g_cl.m_checkpoint_resume.value()
                     : "");
         }
 
         if (g_cl.m_passes.is_set())
-        {
-            params.insert_path(
-                "passes",
-                g_cl.m_passes.values()[0]);
-        }
+            params.insert_path("passes", g_cl.m_passes.values()[0]);
 
         auto_release_ptr<Frame> new_frame(
             FrameFactory::create(
@@ -430,7 +426,8 @@ namespace
         apply_custom_parameter_command_line_options(params);
 
         // Apply command line options that alter the project.
-        if (!apply_frame_command_line_options(project)) return false;
+        if (!apply_frame_command_line_options(project))
+            return false;
         apply_visibility_command_line_options(project);
 
         return true;
@@ -763,4 +760,3 @@ int main(int argc, char* argv[])
 
     return success ? 0 : 1;
 }
-
--- a/src/appleseed/foundation/image/genericimagefilewriter.cpp
+++ b/src/appleseed/foundation/image/genericimagefilewriter.cpp
@@ -331,7 +331,6 @@ void GenericImageFileWriter::write_tiles(const size_t image_index)
     // Retrieve image spec.
     assert(image_index < impl->m_spec.size());
     const OIIO::ImageSpec& spec = impl->m_spec[image_index];
-
     assert(spec.nchannels == spec.channelnames.size());
     assert(spec.nchannels == props.m_channel_count);
 
--- a/src/appleseed/foundation/image/genericimagefilewriter.cpp
+++ b/src/appleseed/foundation/image/genericimagefilewriter.cpp
@@ -144,6 +144,7 @@ void GenericImageFileWriter::set_image_channels(
     OIIO::ImageSpec& spec = impl->m_spec.back();
 
     spec.nchannels = static_cast<int>(channel_count);
+    spec.channelnames.clear();
 
     for (size_t i = 0; i < channel_count; i++)
     {
@@ -192,8 +193,12 @@ void GenericImageFileWriter::set_image_spec()
     }
 
     // Channel names.
-    const char* channel_names[] = { "R", "G", "B", "A" };
-    set_image_channels(props.m_channel_count, channel_names);
+    // Needs to be manually set if the number of channels is higher than 4.
+    if (props.m_channel_count <= 4)
+    {
+        const char* channel_names[] = { "R", "G", "B", "A" };
+        set_image_channels(props.m_channel_count, channel_names);
+    }
 
     // Format of the pixel data.
     const boost::filesystem::path filepath(impl->m_filename);
@@ -327,6 +332,9 @@ void GenericImageFileWriter::write_tiles(const size_t image_index)
     assert(image_index < impl->m_spec.size());
     const OIIO::ImageSpec& spec = impl->m_spec[image_index];
 
+    assert(spec.nchannels == spec.channelnames.size());
+    assert(spec.nchannels == props.m_channel_count);
+
     // Compute the tiles' xstride offset in bytes.
     size_t xstride = props.m_pixel_size;
 
@@ -343,7 +351,7 @@ void GenericImageFileWriter::write_tiles(const size_t image_index)
             assert(tile_offset_y <= props.m_canvas_height);
 
             // Compute the tile's ystride offset in bytes.
-            const size_t ystride = xstride * std::min(static_cast<size_t>(spec.width + spec.x - tile_offset_x), 
+            const size_t ystride = xstride * std::min(static_cast<size_t>(spec.width + spec.x - tile_offset_x),
                                                       static_cast<size_t>(spec.tile_width));
 
             // Retrieve the (tile_x, tile_y) tile.
@@ -417,9 +425,9 @@ void GenericImageFileWriter::write_scanlines(const size_t image_index)
 
         // Write scanline into the file.
         if (!impl->m_writer->write_scanlines(
-                static_cast<int>(y_begin), 
-                static_cast<int>(y_end), 
-                0, 
+                static_cast<int>(y_begin),
+                static_cast<int>(y_end),
+                0,
                 convert_pixel_format(props.m_pixel_format),
                 buffer_ptr))
         {
@@ -465,7 +473,7 @@ void GenericImageFileWriter::write_multi_images()
         const std::string msg = impl->m_writer->geterror();
         throw ExceptionIOError(msg.c_str());
     }
-    
+
     for (size_t i = 0, e = get_image_count(); i < e; ++i)
     {
         if (i > 0)
@@ -480,7 +488,7 @@ void GenericImageFileWriter::write_multi_images()
 
         write(i);
     }
-    
+
     close_file();
 }
 
--- a/src/appleseed/foundation/image/genericprogressiveimagefilereader.cpp
+++ b/src/appleseed/foundation/image/genericprogressiveimagefilereader.cpp
@@ -314,9 +314,9 @@ void GenericProgressiveImageFileReader::read_tile(
         const size_t tile_width = min(impl->m_props.m_tile_width, impl->m_props.m_canvas_width - origin_x);
         const size_t tile_height = min(impl->m_props.m_tile_height, impl->m_props.m_canvas_height - origin_y);
 
-        // The tile fits perfectly into the canvas.
         if (tile_width == impl->m_props.m_tile_width && tile_height == impl->m_props.m_tile_height)
         {
+            // The tile fits perfectly into the canvas.
             assert(output_tile->get_width() == impl->m_props.m_tile_width);
             assert(output_tile->get_height() == impl->m_props.m_tile_height);
 
--- a/src/appleseed/foundation/image/genericprogressiveimagefilereader.cpp
+++ b/src/appleseed/foundation/image/genericprogressiveimagefilereader.cpp
@@ -262,89 +262,25 @@ Tile* GenericProgressiveImageFileReader::read_tile(
 {
     assert(is_open());
 
-    if (impl->m_is_tiled)
-    {
-        //
-        // Tiled image.
-        //
-        // In appleseed, for images whose width or height are not multiples
-        // of the tile's width or height, border tiles are actually smaller.
-        // In OpenImageIO, border tiles have the same size as other tiles,
-        // and the image's pixel data window defines which pixels of those
-        // tiles actually belong to the image.
-        //
-        // Since in appleseed we don't propagate pixel data windows through
-        // the whole image pipeline, we make sure here to return tiles of
-        // the correct dimensions, at some expenses.
-        //
+    // Compute tile's dimensions.
+    const size_t tile_width = impl->m_is_tiled
+        ? min(impl->m_props.m_tile_width, impl->m_props.m_canvas_width - (tile_x * impl->m_props.m_tile_width))
+        : impl->m_props.m_canvas_width;
+    const size_t tile_height = impl->m_is_tiled
+        ? min(impl->m_props.m_tile_height, impl->m_props.m_canvas_height - (tile_y * impl->m_props.m_tile_height))
+        : impl->m_props.m_canvas_height;
 
-        unique_ptr<Tile> source_tile(
-            new Tile(
-                impl->m_props.m_tile_width,
-                impl->m_props.m_tile_height,
-                impl->m_props.m_channel_count,
-                impl->m_props.m_pixel_format));
+    // Create the tile.
+    Tile* output_tile = new Tile(
+        tile_width,
+        tile_height,
+        impl->m_props.m_channel_count,
+        impl->m_props.m_pixel_format);
 
-        const size_t origin_x = tile_x * impl->m_props.m_tile_width;
-        const size_t origin_y = tile_y * impl->m_props.m_tile_height;
+    // Read the tile from the file.
+    read_tile(tile_x, tile_y, output_tile);
 
-        if (!impl->m_input->read_tile(
-                static_cast<int>(origin_x),
-                static_cast<int>(origin_y),
-                0, // z
-                impl->m_input->spec().format,
-                source_tile->get_storage()))
-            throw ExceptionIOError(impl->m_input->geterror().c_str());
-
-        const size_t tile_width = min(impl->m_props.m_tile_width, impl->m_props.m_canvas_width - origin_x);
-        const size_t tile_height = min(impl->m_props.m_tile_height, impl->m_props.m_canvas_height - origin_y);
-
-        if (tile_width == impl->m_props.m_tile_width && tile_height == impl->m_props.m_tile_height)
-            return source_tile.release();
-
-        unique_ptr<Tile> shrunk_tile(
-            new Tile(
-                tile_width,
-                tile_height,
-                impl->m_props.m_channel_count,
-                impl->m_props.m_pixel_format));
-
-        for (size_t y = 0; y < tile_height; ++y)
-        {
-            memcpy(
-                shrunk_tile->pixel(0, y),
-                source_tile->pixel(0, y),
-                tile_width * impl->m_props.m_pixel_size);
-        }
-
-        return shrunk_tile.release();
-    }
-    else
-    {
-        //
-        // Scanline image.
-        //
-
-        unique_ptr<Tile> tile(
-            new Tile(
-                impl->m_props.m_canvas_width,
-                impl->m_props.m_canvas_height,
-                impl->m_props.m_channel_count,
-                impl->m_props.m_pixel_format));
-
-        if (!impl->m_supports_random_access)
-        {
-            close();
-            impl->open();
-        }
-
-        if (!impl->m_input->read_image(
-                impl->m_input->spec().format,
-                tile->get_storage()))
-            throw ExceptionIOError(impl->m_input->geterror().c_str());
-
-        return tile.release();
-    }
+    return output_tile;
 }
 
 void GenericProgressiveImageFileReader::read_tile(
--- a/src/appleseed/foundation/image/genericprogressiveimagefilereader.cpp
+++ b/src/appleseed/foundation/image/genericprogressiveimagefilereader.cpp
@@ -85,9 +85,11 @@ struct GenericProgressiveImageFileReader::Impl
             throw ExceptionIOError(OIIO::geterror().c_str());
 
         m_supports_random_access = m_input->supports("random_access") != 0;
+        read_spec(m_input->spec());
+    }
 
-        const OIIO::ImageSpec& spec = m_input->spec();
-
+    void read_spec(const OIIO::ImageSpec& spec)
+    {
         m_is_tiled = spec.tile_width > 0 && spec.tile_height > 0 && spec.tile_depth > 0;
 
         size_t tile_width, tile_height;
@@ -240,6 +242,20 @@ void GenericProgressiveImageFileReader::read_image_attributes(
     }
 }
 
+bool GenericProgressiveImageFileReader::choose_subimage(const size_t subimage) const
+{
+    OIIO::ImageSpec spec;
+    const bool success = impl->m_input->seek_subimage(
+        static_cast<int>(subimage),
+        0,
+        spec);
+
+    if (success)
+        impl->read_spec(spec);
+
+    return success;
+}
+
 Tile* GenericProgressiveImageFileReader::read_tile(
     const size_t        tile_x,
     const size_t        tile_y)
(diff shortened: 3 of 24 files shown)


dictoon:
Will need to be rebased.

src/appleseed/foundation/image/genericprogressiveimagefilereader.h
```diff
@@ -85,6 +85,14 @@ class APPLESEED_DLLSYMBOL GenericProgressiveImageFileReader
         const size_t        tile_x,
         const size_t        tile_y) override;
 
+    // Read an image tile. Outputs the content in the given tile.
+    void read_tile(
+        const size_t        tile_x,
+        const size_t        tile_y,
+        Tile*               output_tile);
+
+    bool choose_subimage(const size_t subimage) const;
```
dictoon: Comment. Move right above the two `read_tile()` methods.

src/appleseed/foundation/image/genericprogressiveimagefilereader.cpp
```diff
@@ -139,6 +141,7 @@ struct GenericProgressiveImageFileReader::Impl
             static_cast<size_t>(spec.nchannels),
             pixel_format);
     }
+
```
dictoon: Revert.

dictoon:
@est77 Could you review the AOV part of this PR please?

src/appleseed/renderer/modeling/aov/aov.cpp
```diff
@@ -110,12 +93,12 @@ ColorAOV::ColorAOV(const char* name, const ParamArray& params)
 
 size_t ColorAOV::get_channel_count() const
 {
-    return 3;
+    return 4;
```
est77: I am not sure about this change. Most of our color AOVs are 3 channel.
Why did you change it?

src/appleseed/renderer/utility/bbox.h
```diff
@@ -116,6 +129,46 @@ BBox interpolate(const Iterator begin, const Iterator end, const double time)
...
 }
 
+template <typename BBox, typename Tile, typename Int>
+BBox compute_tile_image_space_bbox(
+    const Tile&     tile,
+    const Int       tile_x,
+    const Int       tile_y)
+{
+    BBox tile_bbox;
+    const Int tile_origin_x = tile_x * static_cast<Int>(tile.get_width());
```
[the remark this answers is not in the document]
oktomus: IIRC the box is then intersected with the frame. I think i tested that but I will check again.

src/appleseed/renderer/modeling/project/configuration.cpp
```diff
@@ -289,8 +289,6 @@ auto_release_ptr<Configuration> BaseConfigurationFactory::create_base_interactiv
     parameters.insert("spectrum_mode", "rgb");
     parameters.insert("sampling_mode", "qmc");
 
-    parameters.insert("passes", 1);
```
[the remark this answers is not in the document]
oktomus: in Frame::get_starting_pass

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -453,28 +476,501 @@ void Frame::denoise(
...
+    return true;
+}
+
+void Frame::save_checkpoint(
+    IShadingResultFrameBufferFactory&   buffer_factory,
+    const size_t                        pass) const
+{
+    if (!impl->m_checkpoint) return;
+
+    create_parent_directories(impl->m_checkpoint_path.c_str());
```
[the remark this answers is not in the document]
oktomus: Fatal error message ? Or simple error message and continue regular rendering ?

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -453,28 +476,501 @@ void Frame::denoise(
...
+            // Compute the image space bounding box of the tile.
+            const int tile_origin_x = static_cast<int>(
+                m_frame.image().properties().m_tile_width * tile_x);
+            const int tile_origin_y = static_cast<int>(
+                m_frame.image().properties().m_tile_height * tile_y);
+
+            const Tile& frame_tile = m_frame.image().tile(tile_x, tile_y);
+
+            // Compute the tile space bounding box of the pixels to render.
+            return compute_tile_space_bbox(
```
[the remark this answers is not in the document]
oktomus: yes I think so

src/appleseed/renderer/modeling/aov/aov.cpp
oktomus: Because the SadingBuffer is doing 4 * aov_count. So wether we change the shading buffer to use the exact amount of channels which can be difficult or we set all AOV to 4 channels if they are used in the shading buffer.

src/appleseed/renderer/utility/filesystem.cpp
```diff
@@ -0,0 +1,64 @@
...
+    }
+}
+
+void create_parent_directories(const char* file_path)
+{
+    create_parent_directories(bf::path(file_path));
+}
+
+}       // namespace renderer
+
```
dictoon: Remove redundant blank line. The file should simply end with a `\n`.

src/appleseed/renderer/modeling/frame/frame.h
```diff
@@ -62,6 +62,7 @@ namespace foundation    { class Tile; }
 namespace renderer      { class BaseGroup; }
 namespace renderer      { class DenoiserAOV; }
 namespace renderer      { class ImageStack; }
+namespace renderer      { class PermanentShadingResultFrameBufferFactory; }
```
dictoon: Order alphabetically.

src/appleseed/renderer/kernel/rendering/renderercomponents.h
```diff
@@ -86,28 +86,30 @@ class RendererComponents
...
+    ShadingEngine                                                   m_shading_engine;
+    TextureStore&                                                   m_texture_store;
+    OIIOTextureSystem&                                              m_texture_system;
+    OSLShadingSystem&                                               m_shading_system;
+
+    std::unique_ptr<ILightingEngineFactory>                         m_lighting_engine_factory;
+    std::unique_ptr<ISampleRendererFactory>                         m_sample_renderer_factory;
+    std::unique_ptr<ISampleGeneratorFactory>                        m_sample_generator_factory;
+    std::unique_ptr<IPixelRendererFactory>                          m_pixel_renderer_factory;
+    foundation::auto_release_ptr<IShadingResultFrameBufferFactory>  m_shading_result_framebuffer_factory;
```
dictoon: Why this change?

src/appleseed/renderer/kernel/rendering/masterrenderer.cpp
```diff
@@ -461,13 +462,24 @@ struct MasterRenderer::Impl
         // Updating the trace context causes ray tracing acceleration structures to be updated or rebuilt.
         m_project.update_trace_context();
 
+        // Load the checkpoint if any.
+        Frame& frame = *m_project.get_frame();
+
```
dictoon: Remove this blank line and the one below so that the comment applies to the whole block.

src/appleseed/renderer/kernel/rendering/masterrenderer.cpp
```diff
@@ -461,13 +462,24 @@ struct MasterRenderer::Impl
...
+        PermanentShadingResultFrameBufferFactory* buffer_factory =
```
dictoon: What guarantees that the shading result framebuffer is a permanent one here?

src/appleseed/renderer/kernel/rendering/generic/genericframerenderer.cpp
```diff
@@ -37,6 +37,8 @@
 #include "renderer/kernel/rendering/generic/tilejobfactory.h"
 #include "renderer/kernel/rendering/iframerenderer.h"
 #include "renderer/kernel/rendering/ipasscallback.h"
+#include "renderer/kernel/rendering/ishadingresultframebufferfactory.h"
+#include "renderer/kernel/rendering/permanentshadingresultframebufferfactory.h"
```
dictoon: Order alphabetically.

src/appleseed/renderer/kernel/rendering/generic/genericframerenderer.cpp
```diff
@@ -71,6 +73,7 @@ namespace renderer
 
 namespace
 {
+
```
dictoon: Revert.

src/appleseed/foundation/image/genericprogressiveimagefilereader.cpp
```diff
@@ -326,4 +342,100 @@ Tile* GenericProgressiveImageFileReader::read_tile(
     }
 }
 
+void GenericProgressiveImageFileReader::read_tile(
+    const size_t        tile_x,
+    const size_t        tile_y,
+    Tile*               output_tile)
+{
+    assert(is_open());
```
dictoon: Don't duplicate the whole function. I suspect that one function can easily used the other one.

src/appleseed/foundation/image/genericimagefilewriter.cpp
```diff
@@ -331,6 +336,9 @@ void GenericImageFileWriter::write_tiles(const size_t image_index)
     assert(image_index < impl->m_spec.size());
     const OIIO::ImageSpec& spec = impl->m_spec[image_index];
```
dictoon: Remove blank line.

src/appleseed.cli/main.cpp
```diff
@@ -722,3 +771,4 @@ int main(int argc, const char* argv[])
 
     return success ? 0 : 1;
 }
+
```
dictoon: Revert.

src/appleseed.cli/main.cpp
```diff
@@ -267,7 +267,7 @@ namespace
         if (g_cl.m_passes.is_set())
         {
             params.insert_path(
-                "generic_frame_renderer.passes",
+                "passes",
```
dictoon: There was a bug here, no?

src/appleseed.cli/main.cpp
```diff
@@ -374,15 +423,17 @@ namespace
...
-    void apply_command_line_options(Project& project, ParamArray& params)
+    bool apply_command_line_options(Project& project, ParamArray& params)
     {
         // Apply command line options that alter renderer settings.
         apply_rendering_settings_command_line_options(params);
         apply_custom_parameter_command_line_options(params);
 
         // Apply command line options that alter the project.
-        apply_frame_command_line_options(project);
+        if (!apply_frame_command_line_options(project)) return false;
```
dictoon: Move `if` body to its own line.

src/appleseed.cli/main.cpp
```diff
@@ -327,13 +327,62 @@ namespace
...
+            params.insert("output_path", file_path);
+        }
+
+        if (g_cl.m_checkpoint_create.is_set())
+        {
+            params.insert("checkpoint_create", "on");
+
+            if (g_cl.m_checkpoint_create.values().empty() && !g_cl.m_output.is_set())
+            {
+                LOG_ERROR(g_logger, "ouput path must be specified if no path is given for --checkpoint-create");
```
dictoon: Typo: *output*

src/appleseed.cli/main.cpp
```diff
@@ -327,13 +327,62 @@ namespace
...
 
+        if (g_cl.m_output.is_set())
+        {
+            const char* file_path = g_cl.m_output.value().c_str();
+            params.insert("output_path", file_path);
+        }
+
+        if (g_cl.m_checkpoint_create.is_set())
+        {
+            params.insert("checkpoint_create", "on");
```
dictoon: We usually prefer letting the system convert the value to a string, hence the following is preferred:

    params.insert("checkpoint_create", true);

src/appleseed.cli/main.cpp
```diff
@@ -327,13 +327,62 @@ namespace
...
+            if (g_cl.m_checkpoint_create.values().empty() && !g_cl.m_output.is_set())
+            {
+                LOG_ERROR(g_logger, "ouput path must be specified if no path is given for --checkpoint-create");
+                return false;
+            }
+
+            params.insert(
+                "checkpoint_create_path",
+                !g_cl.m_checkpoint_create.values().empty()
+                    ? g_cl.m_checkpoint_create.value().c_str()
```
dictoon: I think you can remove `.c_str()`.

src/appleseed.cli/main.cpp
```diff
@@ -327,13 +327,62 @@ namespace
...
+                    : "");
+        }
+
+        if (g_cl.m_checkpoint_resume.is_set())
+        {
+            params.insert("checkpoint_resume", "on");
```
dictoon: Same thing here.

src/appleseed.cli/main.cpp
```diff
@@ -327,13 +327,62 @@ namespace
...
+
+            if (g_cl.m_checkpoint_resume.values().empty() && !g_cl.m_output.is_set())
+            {
+                LOG_ERROR(g_logger, "ouput path must be specified if no path is given for --checkpoint-resume");
```
dictoon: Typo here too.

src/appleseed.cli/main.cpp
dictoon: Can you retrieve the actual option name instead of repeating it in the error message?

src/appleseed.cli/main.cpp
dictoon: Use actual option name.

src/appleseed.cli/main.cpp
```diff
@@ -327,13 +327,62 @@ namespace
...
+                return false;
+            }
+
+            params.insert(
+                "checkpoint_resume_path",
+                !g_cl.m_checkpoint_resume.values().empty()
+                    ? g_cl.m_checkpoint_resume.value().c_str()
```
dictoon: Same thing here.

src/appleseed.cli/main.cpp
```diff
@@ -327,13 +327,62 @@ namespace
...
+                    : "");
+        }
+
+        if (g_cl.m_passes.is_set())
+        {
+            params.insert_path(
```
dictoon: Fits on a single line (and then the `if` body doesn't need curly braces).

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -462,28 +490,465 @@ void Frame::denoise(
...
+        const string second_layer_name = get<0>(checkpoint_props[1]);
+        if (second_layer_name != "appleseed:RenderingBuffer")
         {
-            bsys::error_code ec;
-            if (!bf::create_directories(parent_path, ec))
+            RENDERER_LOG_ERROR("incorrect checkpoint: rendering buffer layer is missing.");
+            return false;
+        }
+
+        // Check that weight layer has correct amount of channel
```
dictoon: End with period.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -462,28 +490,465 @@ void Frame::denoise(
...
+            frame_props.m_tile_width != beauty_props.m_tile_width ||
+            frame_props.m_tile_height != beauty_props.m_tile_height ||
+            frame_props.m_channel_count != beauty_props.m_channel_count ||
+            frame_props.m_pixel_format != beauty_props.m_pixel_format)
+        {
+            RENDERER_LOG_ERROR("incorrect checkpoint: the beauty layer properties doesn't match the renderer properties.");
+            return false;
+        }
+
+        // Check if aovs are here.
```
dictoon: *AOVs*

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -462,28 +490,465 @@ void Frame::denoise(
...
+        Deepimf& histograms_image = denoiser_aov->histograms_image();
+        Deepimf& covariance_image = denoiser_aov->covariance_image();
+        Deepimf& sum_image = denoiser_aov->sum_image();
+
+        string hist_file_path, cov_file_path, sum_file_path;
+        get_denoiser_checkpoint_paths(checkpoint_path, hist_file_path, cov_file_path, sum_file_path);
+
+        // Load histograms.
+        bool result = ImageIO::loadMultiChannelsEXR(histograms_image, hist_file_path.c_str());
+        // Load covariance accumulator.
```
dictoon: Add a blank line before a full line comment.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -462,28 +490,465 @@ void Frame::denoise(
...
+        const Deepimf& histograms_image = denoiser_aov->histograms_image();
+        const Deepimf& covariance_image = denoiser_aov->covariance_image();
+        const Deepimf& sum_image = denoiser_aov->sum_image();
+
+        string hist_file_path, cov_file_path, sum_file_path;
+        get_denoiser_checkpoint_paths(checkpoint_path, hist_file_path, cov_file_path, sum_file_path);
+
+        // Add histograms layer.
+        bool result = ImageIO::writeMultiChannelsEXR(histograms_image, hist_file_path.c_str());
+        // Add covariances layer.
```
dictoon: Add a blank line before a full line comment.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -462,28 +490,465 @@ void Frame::denoise(
...
+{
+    if  (!impl->m_checkpoint_resume)
+        return true;
+
+    bf::path bf_path(impl->m_checkpoint_resume_path.c_str());
+
+    // Check if the file exists.
+    if (!bf::exists(bf_path))
+    {
+        RENDERER_LOG_WARNING("no checkpoint found, starting a new render");
```
dictoon: End message with a period.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -462,28 +490,465 @@ void Frame::denoise(
...
+
+        reader.read_canvas_properties(layer_canvas_props);
+        reader.read_image_attributes(layer_attributes);
+
+        const string layer_name =
+            layer_attributes.exist("name")
+                ? layer_attributes.get<string>("name")
+                : "undefined";
+
+        checkpoint_props.push_back(
```
dictoon: You can use `emplace_back()` here.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -462,28 +490,465 @@ void Frame::denoise(
...
+        return true;
+    }
+
+    // Open the file.
+    GenericProgressiveImageFileReader reader;
+    reader.open(impl->m_checkpoint_resume_path.c_str());
+
+    // First, read layers name and properties.
+    CheckpointProperties checkpoint_props;
+
```
dictoon: Remove blank line so that the comment applies to the whole block.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -462,28 +490,465 @@ void Frame::denoise(
...
+                layer_canvas_props,
+                layer_attributes));
+
+        ++layer_index;
+    }
+
+    // Check checkpoint's compatibility.
+    if (!is_checkpoint_compatible(impl->m_checkpoint_resume_path, *this, checkpoint_props))
+    {
+        reader.close();
```
dictoon: No need to close the reader manually, it'll close itself on destruction.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -462,28 +490,465 @@ void Frame::denoise(
...
+        return false;
+    }
+
+    const size_t start_pass = get<2>(checkpoint_props[0]).get<size_t>("appleseed:LastPass") + 1;
+
+    // Check if passes have already been rendered.
+    if (impl->m_pass_count <= start_pass)
+    {
+        RENDERER_LOG_WARNING("passes have previously been rendered and saved into the checkpoint");
```
dictoon: End with a period.

dictoon: In an attempt to make the message clearer, may I suggest:

    the requested passes have already been rendered and saved into the checkpoint.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -462,28 +490,465 @@ void Frame::denoise(
...
+        {
+            // Read shading buffer.
+            reader.choose_subimage(1);
+            Tile& shading_tile = shading_canvas.tile(tile_x, tile_y);
+            reader.read_tile(tile_x, tile_y, &shading_tile);
+
+            // No need to read beauty and filtered AOVs because they
+            // are in the shading buffer.
+
+            // Read Unfiltered AOV layers.
```
dictoon: *unfiltered*

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -462,28 +490,465 @@ void Frame::denoise(
...
+    reader.close();
+
+    // Load internal AOVs (from external files).
+    for (size_t i = 0, e = internal_aovs().size(); i < e; ++i)
+    {
+        AOV* aov = internal_aovs().get_by_index(i);
+
+        DenoiserAOV* denoiser_aov = dynamic_cast<DenoiserAOV*>(aov);
+
+        // Save denoiser checkpoint.
```
dictoon: *Load*

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -462,28 +490,465 @@ void Frame::denoise(
...
+        if (denoiser_aov != nullptr)
+        {
+            if (!load_denoiser_checkpoint(impl->m_checkpoint_resume_path, denoiser_aov))
+                return false;
+        }
+    }
+
+    RENDERER_LOG_INFO("opened checkpoint resuming at pass %s",
```
dictoon: Statement fits on a single line.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -462,28 +490,465 @@ void Frame::denoise(
...
+    // Add internal AOVs layers (in external files).
+    for (size_t i = 0, e = internal_aovs().size(); i < e; ++i)
+    {
+        const AOV* aov = internal_aovs().get_by_index(i);
+
+        const DenoiserAOV* denoiser_aov = dynamic_cast<const DenoiserAOV*>(aov);
+
+        // Save denoiser checkpoint.
+        if (denoiser_aov != nullptr)
+        {
```
dictoon: Remove braces around single line statement.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -462,28 +490,465 @@ void Frame::denoise(
...
+            save_denoiser_checkpoint(impl->m_checkpoint_create_path, denoiser_aov);
+        }
+    }
+
+    RENDERER_LOG_INFO("wrote checkpoint for pass %s.", pretty_uint(pass + 1).c_str());
```
dictoon: Let's mention the checkpoint file path.

    wrote checkpoint file %s for pass %s.

src/appleseed/renderer/modeling/frame/frame.cpp
dictoon: End message with a period.

Let's mention the checkpoint file path:

    read checkpoint file %s, resuming rendering at pass %s.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -462,28 +490,465 @@ void Frame::denoise(
...
+    // Interface used to save the rendering buffer in checkpoints.
+    //
+
+    class ShadingBufferCanvas
+      : public ICanvas
+    {
+      public:
+        ShadingBufferCanvas(
+            const Frame&                                frame,
+            PermanentShadingResultFrameBufferFactory*   buffer_factory)
```
dictoon: Do you really need a `PermanentShadingResultFrameBufferFactory`? What for?

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -462,28 +490,465 @@ void Frame::denoise(
...
+
+        AABB2i get_tile_bbox(
+            const size_t            tile_x,
+            const size_t            tile_y) const
+        {
+            const Image& image = m_frame.image();
+            const CanvasProperties& props = image.properties();
+
+            // Compute the image space bounding box of the tile.
+            const int tile_origin_x = static_cast<int>(
```
dictoon: Single line.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -462,28 +490,465 @@ void Frame::denoise(
...
+                props.m_tile_width * tile_x);
+            const int tile_origin_y = static_cast<int>(
```
dictoon: Single line.

src/appleseed/renderer/kernel/rendering/renderercomponents.h
oktomus: Because the ptr is not unique anymore.

oktomus: The master renderer use it.

src/appleseed/renderer/kernel/rendering/masterrenderer.cpp
oktomus: The code that generate the renderer and the buffer :D You think I should dynamic cast and test the result ?

src/appleseed.cli/main.cpp
oktomus: No, because I need the frame to have access to this setting.

src/appleseed.cli/main.cpp
oktomus: I could.

src/appleseed/renderer/modeling/frame/frame.cpp
oktomus: Yes because `EphemeralShadingResultFrameBufferFactory` always create a new tile. But here I want the tiles that was already created and filled.

oktomus:
Thanks for the review @dictoon, really appreciated.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -462,28 +490,462 @@ void Frame::denoise(
...
+    image_attributes.insert("appleseed:LastPass", pass);
+
+    // Create layers.
+    // Buffer containing pixels' weight.
+    ShadingBufferCanvas weight_canvas(*this, buffer_factory);
+
+    // Create channel names.
+    const size_t shading_channel_count = weight_canvas.properties().m_channel_count;
+    vector<string> shading_channel_names; // for shading buffer
+    vector<const char *> shading_channel_names_cr;
```
dictoon: What does the `_cr` suffix mean here? What about `_cstr`?

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -462,28 +490,462 @@ void Frame::denoise(
...
+
+    assert(shading_channel_names.size() == shading_channel_count);
+
+    // Add layers in the file.
+    writer.append_image(&image());
+    image_attributes.insert("image_name", "beauty");
+    writer.set_image_attributes(image_attributes);
+
+    writer.append_image(&weight_canvas);
+    image_attributes.insert("image_name", "appleseed:RenderingBuffer");
```
dictoon: Reusing `image_attributes` is confusing and potentially dangerous: you might end up writing attributes you didn't intend to. For instance, right now the main image and all AOV images have a `appleseed:LastPass` attribute, is that intended?

If possible, I would suggest create a new `ImageAttributes` object every time. If necessary, you can wrap each block in braces (`{}`) in order to isolate them from each other.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -462,28 +490,462 @@ void Frame::denoise(
...
+    writer.set_image_channels(shading_channel_count, shading_channel_names_cr.data());
+    writer.set_image_attributes(image_attributes);
+
+    // Add AOV layers.
+    for (size_t i = 0, e = aovs().size(); i < e; ++i)
```
dictoon: Whenever possible, we should use range-based `for` loops:

    for (const AOV& aov : aovs())
    {
        ...
    }

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -462,28 +490,462 @@ void Frame::denoise(
...
+
+    // Write the file.
+    writer.write();
+
+    // Add internal AOVs layers (in external files).
+    for (size_t i = 0, e = internal_aovs().size(); i < e; ++i)
+    {
+        const AOV* aov = internal_aovs().get_by_index(i);
+
+        const DenoiserAOV* denoiser_aov = dynamic_cast<const DenoiserAOV*>(aov);
```
dictoon: This line belongs to the block below:

    // Save denoiser checkpoint.
    const DenoiserAOV* denoiser_aov = dynamic_cast<const DenoiserAOV*>(aov);
    if (denoiser_aov != nullptr)
        save_denoiser_checkpoint(impl->m_checkpoint_create_path, denoiser_aov);

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -462,28 +490,462 @@ void Frame::denoise(
...
+
+    create_parent_directories(impl->m_checkpoint_create_path.c_str());
+
+    GenericImageFileWriter writer(impl->m_checkpoint_create_path.c_str());
+
+    ImageAttributes image_attributes = ImageAttributes::create_default_attributes();
+    image_attributes.insert("appleseed:LastPass", pass);
+
+    // Create layers.
+    // Buffer containing pixels' weight.
```
dictoon: This buffer only contains weights? If not, I suggest renaming the variable `weight_canvas`, the current name is misleading.

src/appleseed/renderer/kernel/rendering/renderercomponents.h
dictoon: That's not a valid argument :) `foundation::auto_release_ptr<>` is like the now-deprecated `std::auto_ptr<>`, which is the ancestor of `std::unique_ptr<>`. All three are "unique pointers".

The only relevant difference between `foundation::auto_release_ptr<>` and `std::unique_ptr<>` is that the former calls a `release()` method to destroy the object while the later calls the destructor.

In this context, either points would work. I don't see any reason to change the type in this PR.

src/appleseed/renderer/modeling/frame/frame.cpp
dictoon: If the `ShadingBufferCanvas` would work with any `IShadingResultFrameBufferFactory`, then you wouldn't need casts or any special cases anywhere.

src/appleseed/renderer/kernel/rendering/generic/genericframerenderer.cpp
```diff
@@ -316,11 +323,17 @@ namespace
             {
                 set_current_thread_name("pass_manager");
 
+                const size_t start_pass = m_frame.get_initial_pass();
+
+                PermanentShadingResultFrameBufferFactory* buffer_factory =
```
dictoon: See my other comment in `frame.cpp`. We want to avoid those potentially dangerous casts.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -462,28 +490,462 @@ void Frame::denoise(
...
+            const Frame&                                frame,
+            PermanentShadingResultFrameBufferFactory*   buffer_factory)
+          : m_frame(frame)
+          , m_buffer_factory(buffer_factory)
+          , m_props(
+                m_frame.image().properties().m_canvas_width,
+                m_frame.image().properties().m_canvas_height,
+                m_frame.image().properties().m_tile_width,
+                m_frame.image().properties().m_tile_height,
+                (1 + m_frame.aov_images().size()) * 4 + 1,
```
dictoon: Where does that expression come from? Is it the responsability of `ShadingBufferCanvas` to compute that number of channels, are you trying to match the number of channels computed elsewhere?

src/appleseed/renderer/modeling/frame/frame.cpp
oktomus: It was for `char`. but `_cstr` is better indeed

src/appleseed/renderer/modeling/frame/frame.cpp
oktomus: Not it wasn't intended, thanks for pointing it out

src/appleseed/renderer/modeling/frame/frame.cpp
oktomus: Not, it actually contains the full shading buffer.

src/appleseed/renderer/kernel/rendering/renderercomponents.h
oktomus: I didn't knew that :o The names aren't really helping.

src/appleseed/renderer/kernel/rendering/generic/genericframerenderer.cpp
oktomus: It will work without casting but the buffer factory needs to be permanent. Can I point that out in a comment and forget about the cast ?

dictoon: Why does it need to be permanent? What happens if it's not (and assuming you don't cast anymore)?

src/appleseed/renderer/modeling/frame/frame.cpp
dictoon: Why the additional `+ 1`?
Can you instead query the number of channels in the ShadingResultFrameBuffer instead?

src/appleseed/renderer/kernel/rendering/generic/genericframerenderer.cpp
oktomus: If it's not permanent, it means it's ephemeral and it will create an empty tile. So there no point in saving this.

src/appleseed/renderer/modeling/frame/frame.cpp
oktomus: +1 Because i save the beauty image, just in case you want to correctly visualize the checkpoint

src/appleseed/renderer/kernel/rendering/generic/genericframerenderer.cpp
dictoon: I understand, but wouldn't be a correct behavior? If an ephemeral buffer is used, then you get empty checkpoints, no problem, no special case. It's kind of the same as saying, if there are no lights in the scene, the render will be black, so we should abort instead of rendering. We don't, there's no special case for a scene without any light.

oktomus: Agreed

src/appleseed/foundation/image/genericprogressiveimagefilereader.cpp
```diff
@@ -262,59 +309,62 @@ Tile* GenericProgressiveImageFileReader::read_tile(
...
-                static_cast<int>(origin_y),
-                0, // z
-                impl->m_input->spec().format,
-                source_tile->get_storage()))
-            throw ExceptionIOError(impl->m_input->geterror().c_str());
-
         const size_t tile_width = min(impl->m_props.m_tile_width, impl->m_props.m_canvas_width - origin_x);
         const size_t tile_height = min(impl->m_props.m_tile_height, impl->m_props.m_canvas_height - origin_y);
 
+        // The tile fits perfectly into the canvas.
```
dictoon: Nice to have thought of that!

You should move the comment inside the `if`, it would be clearer.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -464,35 +492,478 @@ void Frame::denoise(
...
 {
-    void add_chromaticities_attributes(ImageAttributes& image_attributes)
+    inline size_t get_checkpoint_total_channel_count(const size_t aov_count)
     {
-        // Scene-linear sRGB / Rec. 709 chromaticities.
-        image_attributes.insert("white_xy_chromaticity", Vector2f(0.3127f, 0.3290f));
-        image_attributes.insert("red_xy_chromaticity", Vector2f(0.64f, 0.33f));
-        image_attributes.insert("green_xy_chromaticity", Vector2f(0.30f, 0.60f));
-        image_attributes.insert("blue_xy_chromaticity",  Vector2f(0.15f, 0.06f));
+        // The beauty image plus the shadding result framebuffer.
```
dictoon: *shading*

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -947,6 +1418,84 @@ void Frame::extract_parameters()
...
+            if (use_default_path) path = m_params.get_required<string>("output_path");
+
+            bf::path bf_path(path.c_str());
+
+            const string extension = lower_case(bf_path.extension().string());
+
+            if (extension != ".exr")
+            {
+                // Only .exr files are allowed.
+                RENDERER_LOG_ERROR("checkpoint file must be an \".exr\" file, disabling checkpoint creation");
```
dictoon: End message with a period (`.`).

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -947,6 +1418,84 @@ void Frame::extract_parameters()
...
+            if (use_default_path) path = m_params.get_required<string>("output_path");
+
+            bf::path bf_path(path.c_str());
+
+            const string extension = lower_case(bf_path.extension().string());
+
+            if (extension != ".exr")
+            {
+                // Only .exr files are allowed.
+                RENDERER_LOG_ERROR("checkpoint file must be an \".exr\" file, disabling checkpoint resuming");
```
dictoon: End message with a period (`.`).

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -947,6 +1418,84 @@ void Frame::extract_parameters()
...
+        // Create option.
+        impl->m_checkpoint_create = m_params.get_optional<bool>("checkpoint_create", false);
+        impl->m_checkpoint_create_path = "";
+
+        // Check if the checkpoint create path is valid.
+        if (impl->m_checkpoint_create)
+        {
+            string path = m_params.get_optional<string>("checkpoint_create_path", "");
+            const bool use_default_path = path.empty();
+            if (use_default_path) path = m_params.get_required<string>("output_path");
```
dictoon: Move `if` body to separate line.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -947,6 +1418,84 @@ void Frame::extract_parameters()
...
+        // Resume option.
+        impl->m_checkpoint_resume = m_params.get_optional<bool>("checkpoint_resume", false);
+        impl->m_checkpoint_resume_path = "";
+
+        // Check if the checkpoint resume path is valid.
+        if (impl->m_checkpoint_resume)
+        {
+            string path = m_params.get_optional<string>("checkpoint_resume_path", "");
+            const bool use_default_path = path == "";
+            if (use_default_path) path = m_params.get_required<string>("output_path");
```
dictoon: Move `if` body to separate line.

src/appleseed/foundation/image/genericprogressiveimagefilereader.cpp
```diff
@@ -262,59 +309,62 @@ Tile* GenericProgressiveImageFileReader::read_tile(
...
-
-        if (!impl->m_input->read_tile(
-                static_cast<int>(origin_x),
-                static_cast<int>(origin_y),
-                0, // z
-                impl->m_input->spec().format,
-                source_tile->get_storage()))
-            throw ExceptionIOError(impl->m_input->geterror().c_str());
-
         const size_t tile_width = min(impl->m_props.m_tile_width, impl->m_props.m_canvas_width - origin_x);
```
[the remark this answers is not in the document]
oktomus: No because we need to check if it's a border tile

PR #2123, opened 2018-07-17
Fix FP<float>::construct() bug

fedormatantsev:
Fix a ```FP<float>::construct()``` bug.
Bit offset was previously wrong, so the function produced invalid floats.


dictoon:
There are already a bunch of them in `foundation/meta/tests/test_fp.cpp` that beg to be completed to cover this case.

fedormatantsev:
Yes, I thought about a bunch of uts, but I didn't know the boilerplate is here already. Makes life easier, will write them in a moment

src/appleseed/foundation/meta/tests/test_fp.cpp
```diff
@@ -146,6 +146,44 @@ TEST_SUITE(Foundation_Math_FP_Float)
     {
         EXPECT_EQ(FP<float>::neg_inf(), shift(FP<float>::neg_inf(), -1));
     }
+
+    TEST_CASE(Sign_GivenNegativeAndPositiveFloats_ReturnsOneAndZero)
+    {
+        const float negative = -0.33;
```
dictoon: Won't compile on Windows, the constant needs to be single precision: `-0.33f`.

dictoon: Constants are spelled in UpperCase.

src/appleseed/foundation/meta/tests/test_fp.cpp
```diff
@@ -146,6 +146,44 @@ TEST_SUITE(Foundation_Math_FP_Float)
...
+        const float positive = +0.33;
+        EXPECT_EQ(FP<float>::sign(negative), 1);
```
dictoon: Expected values go first (unlike with Google's unit test framework that doesn't care about the order).

src/appleseed/foundation/meta/tests/test_fp.cpp
```diff
@@ -146,6 +146,44 @@ TEST_SUITE(Foundation_Math_FP_Float)
     {
         EXPECT_EQ(FP<float>::neg_inf(), shift(FP<float>::neg_inf(), -1));
     }
+
+    TEST_CASE(Sign_GivenNegativeAndPositiveFloats_ReturnsOneAndZero)
+    {
+        const float Negative = -0.33f;
+        const float Positive = +0.33f;
+        EXPECT_EQ(FP<float>::sign(Negative), 1);
```
dictoon: Swap args.

src/appleseed/foundation/meta/tests/test_fp.cpp
```diff
@@ -195,4 +233,38 @@ TEST_SUITE(Foundation_Math_FP_Double)
     {
         EXPECT_TRUE(FP<double>::is_qnan(FP<double>::qnan()));
     }
+
+    TEST_CASE(Sign_GivenNegativeAndPositiveFloats_ReturnsOneAndZero)
+    {
+        const double Negative = -0.33;
+        const double Positive = +0.33;
+        EXPECT_EQ(FP<double>::sign(Negative), 1);
```
dictoon: Swap args.

src/appleseed/foundation/meta/tests/test_fp.cpp
fedormatantsev: Ah. Missed those...

PR #2129, opened 2018-07-21
Misc fixes

est77:
- Fix bug in blender progressive tile callback.
- Allow Travis OSX build to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,15 +44,21 @@ matrix:
           compiler: gcc
           env: GCC_VERSION=6
 
+    allow_failures:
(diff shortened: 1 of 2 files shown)


est77:
> Out of curiosity, what are the Travis changes about?

Allow failures for OSX means that travis will not consider the PR failing if the OSX build fails.
Right now, OSX builds fail randomly from time to time.

est77:
> And the other thing, fast finish?

If I read the docs correctly, it marks the build as passing or failing without waiting for the OSX build to complete.

PR #2130, opened 2018-07-22
Add world space position AOV

est77:


src/appleseed/renderer/modeling/aov/positionaov.cpp
```diff
@@ -0,0 +1,202 @@
...
+    return Model;
+}
+
+Dictionary PositionAOVFactory::get_model_metadata() const
+{
+    return
+        Dictionary()
+            .insert("name", Model)
+            .insert("label", "Position")
+            .insert("default_model", "false");
```
dictoon: You can remove this key/value as `false` is already the default.

PR #2134, opened 2018-07-28
Fix invalid noise values in empty areas

oktomus:
- Lower the convergence warning threshold
- Adjust the convergence warning message
- Rename min_samples to batch_size
- Store block's noise into the pixel_variation AOV
- Change the rendering settings window layout

### DIFF
--- a/src/appleseed/foundation/image/filteredtile.cpp
+++ b/src/appleseed/foundation/image/filteredtile.cpp
@@ -179,14 +179,19 @@ float FilteredTile::compute_weighted_pixel_variance(
     const float rcp_second_weight = second_weight == 0.0f ? 0.0f : 1.0f / second_weight;
 
     // Get colors and assign weights.
-    Color4f main_color(abs(main[0]), abs(main[1]), abs(main[2]), abs(main[3]));
+    Color4f main_color(main[0], main[1], main[2], main[3]);
     main_color *= rcp_main_weight;
 
-    Color4f second_color(abs(second[0]), abs(second[1]), abs(second[2]), abs(second[3]));
+    Color4f second_color(second[0], second[1], second[2], second[3]);
     second_color *= rcp_second_weight;
 
+    const float rgb = abs(main_color.r) + abs(main_color.g) + abs(main_color.b);
+
+    if (rgb == 0.0f)
+        return 0.0f;
+
     // Compute variance.
-    return fast_rcp_sqrt(main_color.r + main_color.g + main_color.b) * (
+    return fast_rcp_sqrt(rgb) * (
         abs(main_color.r - second_color.r) +
         abs(main_color.g - second_color.g) +
         abs(main_color.b - second_color.b));
--- a/src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
+++ b/src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
@@ -96,7 +96,7 @@ namespace
     // Minimum allowed size for a block of pixel before splitting.
     const size_t BlockSplittingThreshold = BlockMinAllowedSize * 2;
     // Threshold used to warn the user if blocks don't converge.
-    const int BlockConvergenceWarningThreshold = 70;
+    const int BlockConvergenceWarningThreshold = 50;
 
 
     //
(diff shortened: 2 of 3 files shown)


src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
```diff
@@ -442,8 +444,21 @@ namespace
                         // Retrieve the coordinates of the pixel in the padded tile.
                         const Vector2u pt(x, y);
 
-                        Color3f value(static_cast<float>(pb.m_spp), 0.0f, 0.0f);
-                        m_sample_aov_tile->set_pixel(pt.x, pt.y, value);
+                        Color3f value(0.0f);
+
+                        if (m_sample_aov_tile != nullptr)
+                        {
+                            m_sample_aov_tile->get_pixel(pt.x, pt.y, value);
```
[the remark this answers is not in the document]
oktomus: Yes 3.

src/appleseed/foundation/image/filteredtile.cpp
```diff
@@ -185,8 +185,13 @@ float FilteredTile::compute_weighted_pixel_variance(
...
     second_color *= rcp_second_weight;
 
+    const float rgb = main_color.r + main_color.g + main_color.b;
+
+    if (rgb == 0.0f)
+        return 0.0f;
+
     // Compute variance.
-    return fast_rcp_sqrt(main_color.r + main_color.g + main_color.b) * (
+    return fast_rcp_sqrt(rgb) * (
```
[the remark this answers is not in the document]
oktomus: Because you can't compute the `sqrt` of `rgb` if it's negative. I could trick it by making only the variable `rgb` *absolute*.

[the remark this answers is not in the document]
dictoon: Of course you would still use `abs(main_color.r) + abq(main_color.g) + abs(main_color.b)` when computing `rgb`, but that's not what I'm talking about here...

src/appleseed/foundation/image/filteredtile.cpp
```diff
@@ -179,14 +179,19 @@ float FilteredTile::compute_weighted_pixel_variance(
...
     // Get colors and assign weights.
-    Color4f main_color(abs(main[0]), abs(main[1]), abs(main[2]), abs(main[3]));
+    Color4f main_color(main[0], main[1], main[2], main[3]);
     main_color *= rcp_main_weight;
 
-    Color4f second_color(abs(second[0]), abs(second[1]), abs(second[2]), abs(second[3]));
+    Color4f second_color(second[0], second[1], second[2], second[3]);
     second_color *= rcp_second_weight;
 
+    const float rgb = abs(main_color.r + main_color.g + main_color.b);
```
dictoon: You just changed the semantics and I'm not sure this is correct anymore. What do you _really_ want to do?

oktomus: Negative values occurs because of filtering, but we can't compute `sqrt` of invalid value and we can't just ignore those values.

src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
```diff
@@ -454,10 +469,9 @@ namespace
             // Warn the user if adaptive sampling wasn't efficient on this tile.
             if (static_cast<float>(tile_converged_pixel) / static_cast<float>(pixel_count) < BlockConvergenceWarningThreshold / 100.0f)
             {
-                RENDERER_LOG_WARNING(
-                    "%s of pixels have converged, be sure to increase the maximum number of samples "
-                    "or decrease the noise threshold for better performance",
-                    pretty_percent(tile_converged_pixel, pixel_count, 1).c_str());
+                RENDERER_LOG_WARNING("%s of tile's pixels have converged, average sample/pixel: %s.",
+                    pretty_percent(tile_converged_pixel, pixel_count, 1).c_str(),
+                    pretty_uint(static_cast<size_t>(m_spp.get_mean())).c_str());
```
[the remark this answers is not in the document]
oktomus: Because it's sample/pixel, i used `pretty_scalar` with 1 digit but i think it makes more sense to have integer numbers. Per-pixel, we always sample a with an integer amount.

src/appleseed/foundation/image/filteredtile.cpp
dictoon: That's not what I meant. Just before my review you were adding absolute values, now you're taking the absolute value of the sum.

oktomus: oh yes it's not the same, sorry :(

oktomus: updated.

PR #2141, opened 2018-08-01
Miscellaneous fixes, improvements and cleanups

dictoon:


src/appleseed/foundation/utility/benchmark/benchmarksuite.cpp
```diff
@@ -257,51 +257,50 @@ void BenchmarkSuite::run(
...
+                *benchmark.get(),
+                __FILE__,
+                __LINE__,
+                timing_result);
+
 #ifdef GENERATE_BENCHMARK_PLOTS
             vector<Vector2d> points;
 
-            for (size_t j = 0; j < 100; ++j)
+            const size_t PointCount = 100;
```
Blackhart: To be consistent with other variable naming, do we rename ```PointCount``` into ```point_count``` ?

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -124,20 +124,29 @@ Camera::~Camera()
...
-        m_shutter_close_end_time);
+    m_shutter_open_begin_time =
+        m_params.get_optional<float>(
+            "shutter_open_begin_time",
+            ShutterOpenBeginTimeDefaultValue);
+
+    m_shutter_open_end_time =
+        m_params.get_optional<float>(
+            "shutter_open_end_time",
+            m_shutter_open_begin_time);
```
Blackhart: It's inverted. You name it ```end_time``` but you pass ```begin_time``` as argument.
Just to be sure it's the right behaviour

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -124,20 +124,29 @@ Camera::~Camera()
...
+
+    m_shutter_close_end_time =
+        m_params.get_optional<float>(
+            "shutter_close_end_time",
+            ShutterCloseEndTimeDefaultValue);
+
+    m_shutter_close_begin_time =
+        m_params.get_optional<float>(
+            "shutter_close_begin_time",
+            m_shutter_close_end_time);
```
Blackhart: It's inverted. You name it ```begin_time``` but you pass ```end_time``` as argument.
Just to be sure it's the right behaviour

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -161,131 +170,13 @@ bool Camera::on_render_begin(
...
 
         if (impl->m_use_bezier_shutter_curve)
-        {
             initialize_shutter_curve_bezier();
-        }
-        else
-        {
-            initialize_shutter_curve_linear();
-        }
+        else initialize_shutter_curve_linear();
```
Blackhart: Line break ?

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -452,6 +343,115 @@ void Camera::check_shutter_times_for_consistency() const
...
+        m_shutter_time_interval = m_shutter_close_end_time - m_shutter_open_begin_time;
+
+        cbt_minus_oet_plus_one =
+            impl->m_normalized_close_begin_time - impl->m_normalized_open_end_time + 1.0f;
+    }
+
+    impl->m_shutter_curve_linear_open_multiplier =
+        cbt_minus_oet_plus_one * impl->m_normalized_open_end_time;
+
+    impl->m_shutter_curve_linear_fully_opened_multiplier = cbt_minus_oet_plus_one / 2.0f;
```
Blackhart: Better to multiply by 0.5 than do a quotient

src/appleseed/renderer/modeling/entity/entity.h
```diff
@@ -223,6 +246,26 @@ void invoke_update_asset_paths(
...
+    OnRenderBeginRecorder&              recorder,
+    foundation::IAbortSwitch*           abort_switch)
+{
+    for (auto& entity : entities)
+    {
+        if (foundation::is_aborted(abort_switch))
+            return false;
+
+        if (!entity.on_render_begin(project, parent, recorder, abort_switch))
+            return false;
```
Blackhart: Is that better ?

```cpp
if (foundation::is_aborted(abort_switch) ||
    !entity.on_render_begin(project, parent, recorder, abort_switch)
    return false;
```
Just one branch instruction, less cpu prediction

src/appleseed/foundation/utility/benchmark/benchmarksuite.cpp
dictoon: We spell true constants (i.e. with a constant value, not just constant variables) in CamelCase.

src/appleseed/renderer/modeling/camera/camera.cpp
dictoon: I haven't touched that code, just reformatted it. That said, it's correct. The second argument is the default value. When no open end time is specified, we use the open begin time.

src/appleseed/renderer/modeling/camera/camera.cpp
dictoon: That's how we usually format it...

src/appleseed/renderer/modeling/camera/camera.cpp
dictoon: True. Note that this is called once before rendering starts, not during rendering.

src/appleseed/renderer/modeling/entity/entity.h
dictoon: I'm pretty sure the compiler will compile `||` as a series of branches, so no gains there I'm afraid... (Again, this is done before rendering so not as performance-critical as code that happens during rendering.)

src/appleseed/foundation/utility/benchmark/benchmarksuite.cpp
Blackhart: Okay

src/appleseed/renderer/modeling/camera/camera.cpp
Blackhart: Yep, it's what I was thinking.
Just to be sure.

src/appleseed/renderer/modeling/camera/camera.cpp
Blackhart: Okay, I learn some things.

src/appleseed/renderer/modeling/entity/entity.h
Blackhart: Ok, so it's nano-optimization...
Not important

dictoon:
@Blackhart Thanks for the review!

PR #2147, opened 2018-08-03
Improve adaptive tile renderer settings and behavior

oktomus:
- Change adaptiveness to min_samples. This allow to show a fixed number
of uniform samples. Like this we don't need to change adaptiveness when
we change max_samples.
- Allow to set max_samples to 0. This will stop the render only when
pixels have converged. Though you should use a max_samples when caustics
are enabled or when difficul feature may never converge or provide high
variance.
- When computing the noise threshold of a block, use the maximum instead
of the sum. It makes the renderer much more predictable and
user-friendly. Values are in a more coherent range.

### DIFF
--- a/src/appleseed/foundation/image/filteredtile.cpp
+++ b/src/appleseed/foundation/image/filteredtile.cpp
@@ -39,6 +39,7 @@
 #include "foundation/platform/atomic.h"
 
 // Standard headers.
+#include <algorithm>
 #include <cmath>
 
 using namespace std;
@@ -216,7 +217,7 @@ float FilteredTile::compute_tile_variance(
             const float* main_ptr = main->pixel(x, y);
             const float* second_ptr = second->pixel(x, y);
 
-            error += compute_weighted_pixel_variance(main_ptr, second_ptr);
+            error = max(error, compute_weighted_pixel_variance(main_ptr, second_ptr));
         }
     }
 
--- a/src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
+++ b/src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
@@ -175,11 +175,6 @@ namespace
 
             m_sample_aov_index = frame.aovs().get_index("pixel_sample_count");
             m_variation_aov_index = frame.aovs().get_index("pixel_variation");
-
-            if (m_params.m_adaptiveness == 0.0f && thread_index == 0)
-            {
-                RENDERER_LOG_WARNING("adaptiveness is set to 0, use the uniform renderer for better performance");
-            }
         }
 
         void release() override
@@ -192,13 +187,13 @@ namespace
             RENDERER_LOG_INFO(
                 "adaptive tile renderer settings:\n"
                 "  batch size                    %s\n"
+                "  min samples                   %s\n"
                 "  max samples                   %s\n"
-                "  noise threshold               %f\n"
-                "  adaptiveness                  %f",
+                "  noise threshold               %f",
                 pretty_uint(m_params.m_batch_size).c_str(),
-                pretty_uint(m_params.m_max_samples).c_str(),
-                m_params.m_noise_threshold,
-                m_params.m_adaptiveness);
+                pretty_uint(m_params.m_min_samples).c_str(),
+                m_params.m_max_samples > 0 ? pretty_uint(m_params.m_max_samples).c_str() : "unlimited",
+                m_params.m_noise_threshold);
 
             RENDERER_LOG_DEBUG("adaptive tile renderer splitting threshold: %f",
                 m_params.m_splitting_threshold);
@@ -270,23 +265,20 @@ namespace
             assert(framebuffer);
 
             // Create the buffer into which we will accumulate every second samples.
-            ShadingResultFrameBuffer* second_framebuffer = nullptr;
-
             // If rendering multiple passes, the permanent buffer factory will return
             // the same buffer so we must create a new one.
-            second_framebuffer = new ShadingResultFrameBuffer(
-                tile.get_width(),
-                tile.get_height(),
-                frame.aov_images().size(),
-                tile_bbox,
-                frame.get_filter());
+            unique_ptr<ShadingResultFrameBuffer> second_framebuffer(
+                new ShadingResultFrameBuffer(
+                    tile.get_width(),
+                    tile.get_height(),
+                    frame.aov_images().size(),
+                    tile_bbox,
+                    frame.get_filter()));
 
             if (m_params.m_passes > 1)
                 second_framebuffer->copy_from(*framebuffer);
             else
                 second_framebuffer->clear();
-
-            assert(second_framebuffer);
 
             const size_t pixel_count = framebuffer->get_width() * framebuffer->get_height();
 
@@ -298,16 +290,20 @@ namespace
             create_rendering_blocks(rendering_blocks, padded_tile_bbox, framebuffer->get_crop_window());
 
             // First uniform pass based on adaptiveness parameter.
-            if (m_params.m_adaptiveness < 1.0f)
+            if (m_params.m_min_samples > 0)
             {
-                for (size_t i = 0; i < rendering_blocks.size(); ++i)
-                {
-                    PixelBlock& pb = rendering_blocks.front();
-                    rendering_blocks.pop_front();
+                const size_t batch_size =
+                    m_params.m_max_samples > 0
+                        ? min(m_params.m_min_samples, m_params.m_max_samples)
+                        : m_params.m_min_samples;
 
-                    // First batch contains `max_samples` * `1 - adaptiveness`
-                    const size_t batch_size = static_cast<size_t>(
-                        m_params.m_max_samples * (1.0f - m_params.m_adaptiveness));
+                deque<PixelBlock> blocks = rendering_blocks;
+                rendering_blocks.clear();
+
+                while (!blocks.empty())
+                {
+                    PixelBlock pb = blocks.front();
+                    blocks.pop_front();
 
                     // Draw samples.
                     sample_pixel_block(
@@ -315,7 +311,7 @@ namespace
                             abort_switch,
                             batch_size,
                             framebuffer,
-                            second_framebuffer,
+                            second_framebuffer.get(),
                             tile_origin_x,
                             tile_origin_y,
                             frame,
@@ -330,22 +326,28 @@ namespace
             while (true)
             {
                 // Check if image is converged or rendering was aborted.
-                if (abort_switch.is_aborted() || rendering_blocks.size() < 1)
+                if (abort_switch.is_aborted() || rendering_blocks.empty())
                 {
                     finished_blocks.insert(finished_blocks.end(), rendering_blocks.begin(), rendering_blocks.end());
-                    break;
+
+                    if (rendering_blocks.empty())
+                        break;
+                    else
+                        return;
                 }
 
                 // Sample the block in front of the deque.
-                PixelBlock& pb = rendering_blocks.front();
+                PixelBlock pb = rendering_blocks.front();
                 rendering_blocks.pop_front();
 
                 // Each batch contains 'min' samples.
-                assert(m_params.m_max_samples > pb.m_spp);
-                const size_t remaining_samples = m_params.m_max_samples - pb.m_spp;
-                const size_t batch_size = min(m_params.m_batch_size, remaining_samples);
+                assert(pb.m_spp < m_params.m_max_samples || m_params.m_max_samples == 0);
+                const size_t batch_size =
+                    m_params.m_max_samples > 0
+                        ? min(m_params.m_batch_size, m_params.m_max_samples - pb.m_spp)
+                        : m_params.m_batch_size;
 
-                if (remaining_samples < 1)
+                if (batch_size == 0)
                 {
                     finished_blocks.push_back(pb);
                     continue;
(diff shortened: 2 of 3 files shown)


src/appleseed/foundation/image/filteredtile.cpp
```diff
@@ -216,7 +216,7 @@ float FilteredTile::compute_tile_variance(
             const float* main_ptr = main->pixel(x, y);
             const float* second_ptr = second->pixel(x, y);
 
-            error += compute_weighted_pixel_variance(main_ptr, second_ptr);
+            error = max(error, compute_weighted_pixel_variance(main_ptr, second_ptr));
```
dictoon: Make sure to include `<algorithm>`.

src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
```diff
@@ -270,23 +265,22 @@ namespace
...
+                    tile_bbox,
+                    frame.get_filter()));
 
             if (m_params.m_passes > 1)
                 second_framebuffer->copy_from(*framebuffer);
             else
                 second_framebuffer->clear();
 
-            assert(second_framebuffer);
+            assert(second_framebuffer.get());
```
dictoon: That assertion looks redundant: I don't see how `second_framebuffer` can be null, and you use it before the assertion anyway. I would suggest to remove it.

src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
```diff
@@ -298,24 +292,23 @@ namespace
             create_rendering_blocks(rendering_blocks, padded_tile_bbox, framebuffer->get_crop_window());
 
             // First uniform pass based on adaptiveness parameter.
-            if (m_params.m_adaptiveness < 1.0f)
+            if (m_params.m_min_samples > 0)
             {
+                const size_t batch_size = m_params.m_max_samples > 0 ?
```
dictoon: More readable IMHO:

    const size_t batch_size =
        m_params.m_max_samples > 0
            ? min(m_params.m_min_samples, m_params.m_max_samples)
            : m_params.m_min_samples;

src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
```diff
@@ -298,24 +292,23 @@ namespace
...
+                    min(m_params.m_min_samples, m_params.m_max_samples) : m_params.m_min_samples;
+
                 for (size_t i = 0; i < rendering_blocks.size(); ++i)
```
dictoon: That loop looks suspicious. I think it works as expected but it would be a lot clearer if you replaced it by something like

    while (!rendering_blocks.empty())
    {
        ...
    }

src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
```diff
@@ -298,24 +292,23 @@ namespace
...
                 {
                     PixelBlock& pb = rendering_blocks.front();
                     rendering_blocks.pop_front();
```
dictoon: I'm not sure this is correct. Is `pb` guaranteed to still be valid once you called `pop_front()`? Intuitively I would say no.

src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
```diff
@@ -333,19 +326,23 @@ namespace
                 if (abort_switch.is_aborted() || rendering_blocks.size() < 1)
                 {
                     finished_blocks.insert(finished_blocks.end(), rendering_blocks.begin(), rendering_blocks.end());
-                    break;
+
+                    if (rendering_blocks.size() < 1)
```
dictoon: Maybe slightly clearer:

    if (rendering_blocks.empty())

src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
```diff
@@ -333,19 +326,23 @@ namespace
                 if (abort_switch.is_aborted() || rendering_blocks.size() < 1)
```
dictoon: Same here.

src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
```diff
@@ -333,19 +326,23 @@ namespace
...
                 // Sample the block in front of the deque.
                 PixelBlock& pb = rendering_blocks.front();
                 rendering_blocks.pop_front();
 
                 // Each batch contains 'min' samples.
-                assert(m_params.m_max_samples > pb.m_spp);
-                const size_t remaining_samples = m_params.m_max_samples - pb.m_spp;
-                const size_t batch_size = min(m_params.m_batch_size, remaining_samples);
+                assert(pb.m_spp < m_params.m_max_samples || m_params.m_max_samples == 0);
+                const size_t batch_size = m_params.m_max_samples > 0 ?
```
dictoon: Same formatting improvement here.

src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
```diff
@@ -357,66 +354,64 @@ namespace
...
                     tile_origin_x,
                     tile_origin_y,
                     frame,
                     frame_width,
                     pass_hash,
                     aov_count);
 
-                // Check if this was the last batch.
-                if (remaining_samples - batch_size < 1)
+                const AABB2u& block_image_bb = AABB2i::intersect(framebuffer->get_crop_window(), pb.m_surface);
```
dictoon: `AABB2i::intersect()` returns a bounding box by copy. I find it best to acknowledge that fact by not keeping a const reference to the result:

    const AABB2u block_image_bb = AABB2i::intersect(framebuffer->get_crop_window(), pb.m_surface);

src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
```diff
@@ -357,66 +354,64 @@ namespace
...
+                    // The block needs to be splitted.
+                    if (pb.m_main_axis == PixelBlock::Axis::Horizontal
+                        && block_image_bb.extent(0) >= BlockSplittingThreshold)
                     {
-                        pb.m_converged = true;
-                        finished_blocks.push_back(pb);
+                        split_pixel_block(
+                            pb,
+                            rendering_blocks,
+                            static_cast<int>(block_image_bb.min.x)
```
dictoon: What is the `- 0.5f` part for?

If all you want is the middle of the box, you can use `block_image_bb.center(0)` (but please make sure it does exactly what you need).

src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
```diff
@@ -903,34 +895,34 @@ Dictionary AdaptiveTileRendererFactory::get_params_metadata()
         "batch_size",
         Dictionary()
             .insert("type", "int")
-            .insert("default", "8")
+            .insert("default", "16")
             .insert("label", "Batch Size")
             .insert("help", "How many samples to render before each noise calculation"));
```
dictoon: What about "before each convergence estimation" instead of "before each noise calculation"?

src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
```diff
@@ -903,34 +895,34 @@ Dictionary AdaptiveTileRendererFactory::get_params_metadata()
...
 
+    metadata.dictionaries().insert(
+        "min_samples",
+        Dictionary()
+            .insert("type", "int")
+            .insert("default", "0")
+            .insert("label", "Min Samples")
+            .insert("help", "Amount of uniform samples to render before adaptive samples"));
```
dictoon: Replace "Amount of" by "Number of".
Replace "before adaptive samples" by "before adaptive sampling".

src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
```diff
@@ -903,34 +895,34 @@ Dictionary AdaptiveTileRendererFactory::get_params_metadata()
...
+            .insert("help", "Maximum number of anti-aliasing samples (0 for unlimited)"));
 
     metadata.dictionaries().insert(
         "noise_threshold",
         Dictionary()
             .insert("type", "float")
-            .insert("default", "5.0")
+            .insert("default", "1.0")
             .insert("label", "Noise Threshold")
             .insert("help", "Rendering stop threshold"));
```
dictoon: Can you think of a better help message here?

src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
oktomus: What do you mean by valid?

src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
oktomus: "Convergance rate. This is the maximum amount of noise allowed for areas of the image." ?

src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
dictoon: It seems that it's potentially a reference to a destroyed object.

src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
dictoon: Something like that yes, maybe:

    "Maximum amount of noise allowed in the image"

?

src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
oktomus: Ah I see, I should copy it ! I still make this mistake a lot. Thanks

src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
oktomus: Yes I like it

src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
oktomus: I don't remember why I did it this way. I think it's because I wanted to truncate the value

src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
dictoon: To be clear:

Your code was valid because C++ extends the lifetime of temporaries to match the lifetime of const references to them:

https://blog.galowicz.de/2016/03/23/const_reference_to_temporary_object/

src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
dictoon: Yeah this is clearly a bug, or a potential serious bug:

https://stackoverflow.com/questions/6244931/reference-front-and-pop-front

dictoon: Demo of the problem with VC++:
http://rextester.com/FPUA29279

Same problem with clang. However, interestingly it works fine with gcc. Obviously we're deep into undefined territory here so compilers are free to do whatever they want.

src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
dictoon: Maybe you mean "round the value" instead of "truncate the value"? Because that's what happens when you truncate `x + 0.5f`: you round `x` to the nearest integer.

In any case that code should be checked and clarified.

src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
oktomus: Wow thanks, it's weird i didn't noticed any issue so far

src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
oktomus: good to know, thanks

src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
oktomus: There isn't any `center` for integral bboxes, which make sense

src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
```diff
@@ -357,66 +359,62 @@ namespace
...
+                    // The block needs to be splitted.
+                    if (pb.m_main_axis == PixelBlock::Axis::Horizontal
+                        && block_image_bb.extent(0) >= BlockSplittingThreshold)
                     {
-                        pb.m_converged = true;
-                        finished_blocks.push_back(pb);
+                        split_pixel_block(
+                            pb,
+                            rendering_blocks,
+                            static_cast<int>((block_image_bb.min.x + block_image_bb.max.x) * 0.5f));
```
dictoon: I believe that

    static_cast<int>((block_image_bb.min.y + block_image_bb.max.y) * 0.5f)

is equivalent to

    (block_image_bb.min.y + block_image_bb.max.y) / 2

dictoon:
Thank you!

Issue #569, opened 2014-09-16
Implement camera shift


dictoon:

est77:
Done in 19ac25a4e329b58837ed0a46ead5653a432cab5b

PR #2155, opened 2018-08-17
Implement camera shift. Misc work.

est77:
- Implemented camera shift. Issue #569
- Added optional fresnel_weight param to as_glossy OSL closure.

### DIFF
--- a/src/appleseed.studio/mainwindow/rendering/lightpathswidget.cpp
+++ b/src/appleseed.studio/mainwindow/rendering/lightpathswidget.cpp
@@ -51,12 +51,8 @@
 
 // Standard headers.
 #include <algorithm>
+#include <cmath>
 #include <string>
-
-// Platform headers.
-#ifdef __APPLE__
-#include <GLKit/GLKMatrix4.h>
-#endif
 
 using namespace foundation;
 using namespace renderer;
@@ -213,7 +209,7 @@ namespace
         void end_object() override
         {
             glEnd();
-        };
+        }
 
         void rasterize(const Triangle& triangle) override
         {
@@ -288,21 +284,22 @@ void LightPathsWidget::paintGL()
 
     glMatrixMode(GL_PROJECTION);
     glLoadIdentity();
-#ifdef __APPLE__
-    glMultMatrixf(
-        GLKMatrix4MakePerspective(
-            rc.m_hfov / rc.m_aspect_ratio,
-            rc.m_aspect_ratio,
-            0.01,
-            1000.0
-        ).m);
-#else
-    gluPerspective(
-        rad_to_deg(rc.m_hfov) / rc.m_aspect_ratio,
-        rc.m_aspect_ratio,
-        0.01,
-        1000.0);
-#endif
+
+    const double ZNear = 0.01;
+    const double ZFar = 1000.0;
+
+    const double fy = tan(rc.m_hfov / rc.m_aspect_ratio * 0.5) * ZNear;
+    const double fx = fy * rc.m_aspect_ratio;
+
+    const double shift_x = rc.m_shift_x * 2.0 * fx;
+    const double shift_y = rc.m_shift_y * 2.0 * fy;
+
+    const double left   = -fx + shift_x;
+    const double right  =  fx + shift_x;
+    const double top    = -fy + shift_y;
+    const double bottom =  fy + shift_y;
+
+    glFrustum(left, right, top, bottom, ZNear, ZFar);
 
     glMatrixMode(GL_MODELVIEW);
     glLoadIdentity();
--- a/src/appleseed/renderer/kernel/rasterization/rasterizationcamera.h
+++ b/src/appleseed/renderer/kernel/rasterization/rasterizationcamera.h
@@ -40,6 +40,8 @@ class APPLESEED_DLLSYMBOL RasterizationCamera
   public:
     double  m_aspect_ratio;     // aspect ratio of the film (width / height)
     double  m_hfov;             // horizontal field of view in radians
+    double  m_shift_x;          // horizontal camera shift
+    double  m_shift_y;          // vertical camera shift
 };
 
 }       // namespace renderer
--- a/src/appleseed/renderer/kernel/shading/closures.cpp
+++ b/src/appleseed/renderer/kernel/shading/closures.cpp
@@ -658,6 +658,7 @@ namespace
             float           anisotropy;
             float           ior;
             float           energy_compensation;
+            float           fresnel_weight;
         };
 
         static const char* name()
@@ -683,6 +684,7 @@ namespace
             // Initialize keyword parameter defaults.
             Params* params = new (data) Params();
             params->energy_compensation = 0.0f;
+            params->fresnel_weight = 1.0f;
         }
 
         static void register_closure(OSLShadingSystem& shading_system)
@@ -697,6 +699,7 @@ namespace
                 CLOSURE_FLOAT_PARAM(Params, anisotropy),
                 CLOSURE_FLOAT_PARAM(Params, ior),
                 CLOSURE_FLOAT_KEYPARAM(Params, energy_compensation, "energy_compensation"),
+                CLOSURE_FLOAT_KEYPARAM(Params, fresnel_weight, "fresnel_weight"),
                 CLOSURE_FINISH_PARAM(Params)
             };
 
@@ -723,6 +726,8 @@ namespace
             ClosureID cid;
 
             const float roughness = saturate(p->roughness);
+            const float fresnel_weight = saturate(p->fresnel_weight);
+
             const float highlight_falloff = saturate(p->highlight_falloff);
             const float ior = max(p->ior, 0.001f);
             float w = luminance(weight);
@@ -730,17 +735,17 @@ namespace
             if (p->dist == g_ggx_str)
             {
                 cid = GlossyGGXID;
-                w *= sample_weight<GGXMDF>(roughness, ior);
+                w *= sample_weight<GGXMDF>(roughness, ior, fresnel_weight);
             }
             else if (p->dist == g_beckmann_str)
             {
                 cid = GlossyBeckmannID;
-                w *= sample_weight<BeckmannMDF>(roughness, ior);
+                w *= sample_weight<BeckmannMDF>(roughness, ior, fresnel_weight);
             }
             else if (p->dist == g_std_str)
             {
                 cid = GlossySTDID;
-                w *= sample_weight_std(roughness, highlight_falloff, ior);
+                w *= sample_weight_std(roughness, highlight_falloff, ior, fresnel_weight);
             }
             else
             {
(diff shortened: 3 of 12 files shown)


src/appleseed.studio/mainwindow/rendering/lightpathswidget.cpp
```diff
@@ -213,7 +209,7 @@ namespace
         void end_object() override
         {
             glEnd();
-        };
+        }
```
dictoon: Wow, that compiled?!

src/appleseed.studio/mainwindow/rendering/lightpathswidget.cpp
```diff
@@ -288,21 +284,22 @@ void LightPathsWidget::paintGL()
...
-        ).m);
-#else
-    gluPerspective(
-        rad_to_deg(rc.m_hfov) / rc.m_aspect_ratio,
-        rc.m_aspect_ratio,
-        0.01,
-        1000.0);
-#endif
+
+    const double znear = 0.01;
```
dictoon: Capitalize true constants: `ZNear`, `ZFar`.

src/appleseed/renderer/modeling/camera/camera.h
```diff
@@ -170,10 +170,11 @@ class APPLESEED_DLLSYMBOL Camera
     // Return a camera representation suitable for rasterization.
     virtual RasterizationCamera get_rasterization_camera() const = 0;
 
-  protected:
+  private:
     struct Impl;
     Impl* impl;
 
+  protected:
```
dictoon: Move protected section ahead of private one so there's a logical progression (public -> protected -> private).

src/appleseed/renderer/modeling/camera/perspectivecamera.h
```diff
@@ -0,0 +1,113 @@
...
+//
+// Perspective camera base class.
+//
+
+class APPLESEED_DLLSYMBOL PerspectiveCamera
+  : public Camera
+{
+  public:
+    PerspectiveCamera(
+        const char*             name,
```
dictoon: Align arguments.

src/appleseed/renderer/modeling/camera/perspectivecamera.h
```diff
@@ -0,0 +1,113 @@
...
+
+  protected:
+    // Parameters.
+    foundation::Vector2d    m_film_dimensions;      // film dimensions in camera space, in meters
+    double                  m_focal_length;         // focal length in camera space, in meters
+    double                  m_near_z;               // Z value of the near plane in camera space, in meters
+    foundation::Vector2d    m_shift;                // camera shift in camera space, in meters
+
+    // Precomputed values.
+    double      m_rcp_film_width;       // film width reciprocal in camera space
```
dictoon: Align data members.

src/appleseed/renderer/modeling/camera/thinlenscamera.cpp
```diff
@@ -883,6 +794,22 @@ DictionaryArray ThinLensCameraFactory::get_input_metadata() const
             .insert("use", "optional")
             .insert("default", "-0.001"));
 
+    metadata.push_back(
+        Dictionary()
+            .insert("name", "shift_x")
+            .insert("label", "Shift X")
+            .insert("type", "text")
```
dictoon: Should be `numeric` with proper min/max bounds (probably soft ones); `text` is really meant for strings.

src/appleseed/renderer/modeling/camera/thinlenscamera.cpp
```diff
@@ -883,6 +794,22 @@ DictionaryArray ThinLensCameraFactory::get_input_metadata() const
...
+            .insert("use", "optional")
+            .insert("default", "0.0"));
+
+    metadata.push_back(
+        Dictionary()
+            .insert("name", "shift_y")
+            .insert("label", "Shift Y")
+            .insert("type", "text")
```
dictoon: Same thing.

dictoon:
Very welcome refactoring!

src/appleseed/renderer/modeling/camera/pinholecamera.cpp
```diff
@@ -416,6 +308,22 @@ DictionaryArray PinholeCameraFactory::get_input_metadata() const
             .insert("use", "optional")
             .insert("default", "-0.001"));
 
+    metadata.push_back(
```
dictoon: Maybe we should factorize perspective camera metadata like we already factorize common camera metadata (`CameraFactory::get_input_metadata()`).

src/appleseed/renderer/api/camera.h
```diff
@@ -36,6 +36,7 @@
 #include "renderer/modeling/camera/cameratraits.h"
 #include "renderer/modeling/camera/icamerafactory.h"
 #include "renderer/modeling/camera/orthographiccamera.h"
+#include "renderer/modeling/camera/perspectivecamera.h"
```
dictoon: Not sure it's relevant to include that header file here, it's not really part of the API since `PerspectiveCamera` is a private implementation detail, no?

src/appleseed.studio/mainwindow/rendering/lightpathswidget.cpp
est77: clang emits a warning. All other compilers accept it happily.

src/appleseed/renderer/kernel/shading/closures.cpp
```diff
@@ -658,6 +658,7 @@ namespace
             float           anisotropy;
             float           ior;
             float           energy_compensation;
+            float           fresnel_weight;
```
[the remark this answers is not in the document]
est77: It is a special case.
The glossy BRDF is the only one that supports disabling Fresnel, originally to render furnace test scenes.
Disabling Fresnel and replacing it by something more artistic can be useful for NPR shaders and that is why I am 
exposing it to OSL.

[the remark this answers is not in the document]
est77: > Is there a reason why we're not prefixing members with m_ here?

Not really. The OSL closure params members have always been unprefixed.
There is no good reason to prefix them either. We only cast pointers from OSL to these structs to access 
the closure params and copy them to our BSDF inputs.

src/appleseed/renderer/kernel/shading/closures.cpp
```diff
@@ -765,27 +770,37 @@ namespace
...
         template <typename MDF>
-        static float sample_weight(const float roughness, const float ior)
+        static float sample_weight(
+            const float roughness,
+            const float ior,
+            const float fresnel_weight)
         {
             const float eavg = get_average_albedo(MDF(), roughness);
-            const float favg = average_fresnel_reflectance_dielectric(ior);
+            const float favg = lerp(
```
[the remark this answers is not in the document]
est77: It is correct. fresnel weight == 0 -> F = 1.

src/appleseed/renderer/modeling/camera/camera.h
est77: I prefer the private implementation to be the first member of the class. I think it would be surprising if it wasn't.
Maybe we can make an exception in cases like this?
Alternatively, I can make it protected again. I like it a bit less, but I still prefer impl to be first.

src/appleseed/renderer/modeling/camera/pinholecamera.cpp
est77: Done

src/appleseed/renderer/kernel/shading/closures.cpp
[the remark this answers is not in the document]
est77: Fixed

src/appleseed/renderer/kernel/rasterization/rasterizationcamera.h
```diff
@@ -40,6 +40,8 @@ class APPLESEED_DLLSYMBOL RasterizationCamera
   public:
     double  m_aspect_ratio;     // aspect ratio of the film (width / height)
     double  m_hfov;             // horizontal field of view in radians
+    double  m_shift_x;          // horizontal camera shift
+    double  m_shift_y;          // verical camera shift
```
dictoon: Typo

src/appleseed/renderer/modeling/camera/camera.cpp
```diff
@@ -788,4 +746,90 @@ DictionaryArray CameraFactory::get_input_metadata()
...
+            .insert("default", "-0.001"));
+}
+
+void CameraFactory::add_shift_metadata(DictionaryArray& metadata)
+{
+    metadata.push_back(
+        Dictionary()
+            .insert("name", "shift_x")
+            .insert("label", "Shift X")
+            .insert("type", "text")
```
dictoon: Must be numeric.

est77: I tried. Numerical inputs require a min value. They crash if one is not present (a bug?).
The valid range of shifts is -inf to +inf and the values depend on the camera film sizes.
I don't have a way to assign min / max values to shifts.

src/appleseed/renderer/kernel/rasterization/rasterizationcamera.h
est77: Fixed

src/appleseed/renderer/modeling/camera/camera.cpp
dictoon: It's not a bug. You must assign min and max values, but they can be soft limits if you mark them as so. The idea is to provide a reasonable starting range for the slider to work correctly; the user is still able to go outside that range and the slider will adapt. Check how it's done elsewhere.

est77: Ok, but I still don't have meaninful values for min and max, soft or hard.
It is probably the same reason focal length and film width and heights are also text inputs?

dictoon: Shifting is done in multiples of the film dimensions no? In that case it's unit-less and relatively consistent across scenes, so it should be possible to infer soft min and max values.

Eventually we must turn focal length and film width/height into numeric inputs as well. Ranges aren't hard to come up with, i.e. 1-100 mm (0.001-0.100 m) for film width/height, etc.

est77: No. Shifting is done in meters (or the units you use for the film dimensions).

I checked what the apps we support do. Absolute shift is the result you get from camera callibration software and 3d tracking apps, camera databases and what Maya uses. 
Relative shift is used in Max and Blender. In the end I decided to use absolute for our camera.

dictoon: Ok, in that case it's the same question as the other absolute values like film dimensions, focal length, etc. Maybe we should convert them all to numeric in one go, in a subsequent PR.

PR #2168, opened 2018-08-24
Various cleanups in adaptive tile renderer

dictoon:


src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
```diff
@@ -790,41 +795,40 @@ namespace
                     static_cast<float>(pt.y + s.y),
                     shading_result);
 
-                if (second)
+                // Only half the samples go into the second scratch framebuffer.
+                if (i & 1)
```
oktomus: I like this :D

dictoon:
> I'm sorry I had forgot so many things..

No worries at all, code is never perfect, improving it is a never-ending process...

PR #2170, opened 2018-08-25
Misc work

est77:
Added BaseGroup and Assembly clear method: 
- Useful to implement interactive updates of instancers in plugins.

Embree cleanups:
- Remove extra ray origin and dir copy.
- Remove unnecesary impl classes and levels of indirection to access some member variables.
- Move APPLESEED_WITH_EMBREE checks where they are needed, like other optional features.

### DIFF
--- a/src/appleseed/renderer/kernel/intersection/assemblytree.h
+++ b/src/appleseed/renderer/kernel/intersection/assemblytree.h
@@ -32,7 +32,9 @@
 
 // appleseed.renderer headers.
 #include "renderer/kernel/intersection/curvetree.h"
+#ifdef APPLESEED_WITH_EMBREE
 #include "renderer/kernel/intersection/embreescene.h"
+#endif
 #include "renderer/kernel/intersection/probevisitorbase.h"
 #include "renderer/kernel/intersection/regiontree.h"
 #include "renderer/kernel/intersection/treerepository.h"
--- a/src/appleseed/renderer/kernel/intersection/embreescene.cpp
+++ b/src/appleseed/renderer/kernel/intersection/embreescene.cpp
@@ -53,12 +53,6 @@
 #include "foundation/utility/statistics.h"
 #include "foundation/utility/stopwatch.h"
 
-// Embree headers.
-#include <embree3/rtcore.h>
-
-// Standard headers.
-#include <memory>
-
 using namespace foundation;
 using namespace renderer;
 using namespace std;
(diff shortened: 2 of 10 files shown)


src/appleseed/renderer/kernel/intersection/assemblytree.h
```diff
@@ -32,7 +32,9 @@
 
 // appleseed.renderer headers.
 #include "renderer/kernel/intersection/curvetree.h"
+#ifdef APPLESEED_WITH_EMBREE
```
fedormatantsev: I don't know if it's necessary here. The intent was to hide this check inside the header to keep outer code cleaner.

src/appleseed/renderer/kernel/intersection/embreescene.cpp
```diff
@@ -250,17 +221,20 @@ namespace
         const ShadingRay&   shading_ray,
         RTCRay&             embree_ray)
     {
-        const RayBuffer buffer(shading_ray.m_org, shading_ray.m_dir);
+        embree_ray.org_x = static_cast<float>(shading_ray.m_org.x);
```
fedormatantsev: Can you explain this revert?

src/appleseed/renderer/kernel/intersection/embreescene.h
```diff
@@ -42,6 +41,9 @@
 #include "foundation/utility/poolallocator.h"
 #include "foundation/utility/uid.h"
 
+// Embree headers.
+#include <embree3/rtcore.h>
```
fedormatantsev: If this move is made, no need in EmbreeDevice object at all. It was introduce to hide embree guts from outer code.

src/appleseed/renderer/kernel/intersection/intersector.h
```diff
@@ -32,7 +32,9 @@
 
 // appleseed.renderer headers.
 #include "renderer/kernel/intersection/curvetree.h"
+#ifdef APPLESEED_WITH_EMBREE
 #include "renderer/kernel/intersection/embreescene.h"
```
fedormatantsev: Again, I vote for no checks in outer code, check is done inside EmbreeScene.h

src/appleseed/renderer/kernel/intersection/assemblytree.h
est77: We do the check outside for all our others optional dependencies.
I did this change for consistency.

src/appleseed/renderer/kernel/intersection/embreescene.cpp
est77: You were casting and copying manually the org and dir from shading_ray to RayBuffer only to copy them again to the embree ray, this time using mm_store_ps.

You can copy them directly to the embree ray and avoid the store_ps intrinsics.

src/appleseed/renderer/kernel/intersection/embreescene.h
est77: EmbreeScene is not part of the public API and it is not included by any public header.
Removing the private implementation saves a memory allocation and one indirection when accessing the class members.

I still think that EmbreeDevice can be useful as a wrapper, if we want to set options in the embree device, so I didn't remove it.

src/appleseed/renderer/kernel/intersection/intersector.h
est77: I prefer them outside myself, it is more readable for me. But it is not a big preference.
What I don't like is that we check outside for some optional dependencies and inside for embree.

src/appleseed/renderer/kernel/intersection/embreescene.cpp
fedormatantsev: Ah, okay. This was done based on embree's performance tips. They are in documentation pdf located somewhere in embree distribution folder in built appleseed deps

src/appleseed/renderer/kernel/intersection/embreescene.h
fedormatantsev: Sounds reasonable, nice!

src/appleseed/renderer/kernel/intersection/intersector.h
fedormatantsev: Okay, then you're safe to remove this check inside the header itself

src/appleseed/renderer/modeling/scene/basegroup.cpp
```diff
@@ -98,6 +98,16 @@ ShaderGroupContainer& BaseGroup::shader_groups() const
     return impl->m_shader_groups;
 }
 
+void BaseGroup::clear()
+{
+    impl->m_colors.clear();
+    impl->m_textures.clear();
+    impl->m_texture_instances.clear();
+    impl->m_shader_groups.clear();
+    impl->m_assemblies.clear();
```
dictoon: Do we acknowledge here that we're not calling `clear()` on sub-assemblies?

src/appleseed/renderer/kernel/intersection/embreescene.h
```diff
@@ -96,6 +97,9 @@ class EmbreeScene
   private:
     struct Impl;
     Impl* impl;
+
```
dictoon: Is the private implementation still useful/reasonable at this point?

src/appleseed/renderer/modeling/scene/basegroup.cpp
est77: We are simply deleting sub-assemblies.

src/appleseed/renderer/kernel/intersection/embreescene.h
est77: Removed

src/appleseed/renderer/kernel/intersection/embreescene.cpp
```diff
@@ -53,86 +53,54 @@
...
-
 using namespace foundation;
 using namespace renderer;
 using namespace std;
 
 namespace renderer
 {
 
-namespace
+class EmbreeGeometryData
```
dictoon: Should be inside an anonymous namespace.

dictoon: Ah no sorry, disregard.

PR #2182, opened 2018-08-30
Add --libraries command line option

dictoon:


src/appleseed.shared/application/commandlinehandlerbase.cpp
```diff
@@ -42,6 +42,38 @@
...
+// IlmBase headers.
+#include <OpenEXR/IlmBaseConfig.h>
+
+// libpng headers.
+#include <png.h>
+
+// OpenColorIO headers.
+#include <OpenColorIO/OpenColorABI.h>
+
+// OpenEXR headers.
```
est77: Missing #include "foundation/platform/_beginexrheaders.h"?

src/appleseed.shared/application/commandlinehandlerbase.cpp
```diff
@@ -42,6 +42,38 @@
...
+#include <OpenEXR/OpenEXRConfig.h>
+
+// OpenImageIO headers.
```
est77: Missing #include "foundation/platform/_beginoiioheaders.h"?

src/appleseed/renderer/kernel/intersection/embreescene.h
```diff
@@ -43,7 +43,9 @@
 #include "foundation/utility/uid.h"
 
 // Embree headers.
+#ifdef APPLESEED_WITH_EMBREE
```
est77: We already do the test before including this header. I don't think we need to repeat it here.

src/appleseed.shared/application/commandlinehandlerbase.cpp
dictoon: Good point.

src/appleseed.shared/application/commandlinehandlerbase.cpp
dictoon: Good point.

PR #2185, opened 2018-08-30
Fix double travis OSX build

est77:


.travis.yml
```diff
@@ -44,9 +44,6 @@ matrix:
           compiler: gcc
           env: GCC_VERSION=6
 
-        - os: osx
```
dictoon: I'm not sure to understand what's going on. Were we building the macOS version with clang twice?

PR #2186, opened 2018-08-31
Add third party libraries versions to appleseed.studio's About dialog

dictoon:


src/appleseed.studio/help/about/aboutwindow.cpp
```diff
@@ -88,9 +98,35 @@ void AboutWindow::set_version_string()
             .arg(Appleseed::get_lib_compilation_date())
             .arg(Appleseed::get_lib_compilation_time())
             .arg(Compiler::get_compiler_name())
-            .arg(Compiler::get_compiler_version());
-
-    m_ui->label_version_string->setText(version_string);
+            .arg(Compiler::get_compiler_version()));
+
+    LibraryVersionArray versions = ThirdParties::get_versions();
+    versions.push_back(APIStringPair("Qt", QT_VERSION_STR));
```
Blackhart: Why didn't you add Qt's version in the `get_versions()` method ?

src/appleseed.studio/help/about/aboutwindow.cpp
```diff
@@ -88,9 +98,35 @@ void AboutWindow::set_version_string()
...
+    vector<size_t> versions_indices(versions.size());
+    for (size_t i = 0, e = versions.size(); i < e; ++i)
+        versions_indices[i] = i;
+    sort(
+        versions_indices.begin(),
+        versions_indices.end(),
+        [&versions](const size_t lhs, const size_t rhs)
+        {
+            return strcmp(versions[lhs].m_first.c_str(), versions[rhs].m_first.c_str()) < 0;
+        });
```
Blackhart: I undertsand the process, but we know all the dependencies we need.
Is it better to sort theses dependencies inside the `get_version()` method ?

src/appleseed.studio/help/about/aboutwindow.cpp
dictoon: Because `ThirdParties::get_versions()` is in appleseed.dll which is built without Qt support. Only appleseed.studio is built with Qt support.

src/appleseed.studio/help/about/aboutwindow.cpp
dictoon: `ThirdParties::get_versions()` returns a sorted list, but since I then append Qt to that list, the list needs to be sorted. I don't see any other option at the moment, except if one knows how to call `std::sort()` on a custom container such as `foundation::APIArray` which only provides index-based access.

Issue #1993, opened 2018-04-21
Don't store items with 0 probability in CDF?


dictoon:

dictoon:
We're storing items with probability = 0 in the `foundation::CDF` class. Since these items cannot be picked, we should not store them at all. Should this be done by _users_ of `foundation::CDF`? What's the impact of not storing items with zero probability?

pjessesco:
How can I test it if something happens when I make some changes?

pjessesco:
Done.

dictoon:
PR #2196 reverted via #2314 since it introduces regressions, in particular in `test scenes/environment/20 - sparse latlong environment map with horizontal shift.appleseed`.

PR #2196, opened 2018-09-13
Do not store zero weight in `Foundation::cdf`

pjessesco:
#1993 
> Should this be done by users of `foundation::CDF`?

I thought it may force to make pair `if clause(if weight > 0)` with `cdf::insert()` in future usage. It doesn't seem good to me, so I chose to change `cdf::insert()` rather than change _users_ of `cdf`. 

Feedback appreciated :)

### DIFF
--- a/src/appleseed/foundation/math/cdf.h
+++ b/src/appleseed/foundation/math/cdf.h
@@ -181,7 +181,8 @@ template <typename Item, typename Weight>
 inline void CDF<Item, Weight>::insert(const Item& item, const Weight weight)
 {
     assert(weight >= Weight(0.0));
-    if(weight > Weight(0.0)){
+    if (weight > Weight(0.0))
+    {
         m_items.push_back(std::make_pair(item, weight));
         m_weight_sum += weight;
     }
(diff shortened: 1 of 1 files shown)


src/appleseed/foundation/math/cdf.h
```diff
@@ -181,8 +181,10 @@ template <typename Item, typename Weight>
 inline void CDF<Item, Weight>::insert(const Item& item, const Weight weight)
 {
     assert(weight >= Weight(0.0));
-    m_items.push_back(std::make_pair(item, weight));
-    m_weight_sum += weight;
+    if(weight > Weight(0.0)){
```
est77: Add space after if and move { to the next line.

```
if (condition)
{
    ....
}
```

est77:
Thanks for the PR. I think it is correct. Only some very small formatting issues and we can merge.

src/appleseed/foundation/math/cdf.h
pjessesco: Oops, thanks for review.

dictoon:
Looks good. Do unit tests pass?

pjessesco:
Yes it does.

PR #2199, opened 2018-09-16
Fix motion blur embree bug

fedormatantsev:
Previously we used vertices from the first motion segment to construct supporting plane for the hit no matter what. After the fix vertices from two nearest motion steps are getting interpolated based on the ray time.

### Before:
bug

### After:
fix

### DIFF
--- a/src/appleseed/renderer/kernel/intersection/embreescene.cpp
+++ b/src/appleseed/renderer/kernel/intersection/embreescene.cpp
@@ -448,12 +448,46 @@ void EmbreeScene::intersect(ShadingPoint& shading_point) const
         const uint32 v1_idx = geometry_data->m_primitives[rayhit.hit.primID * 3 + 1];
         const uint32 v2_idx = geometry_data->m_primitives[rayhit.hit.primID * 3 + 2];
 
-        const TriangleType triangle(
-            Vector3d(geometry_data->m_vertices[v0_idx]),
-            Vector3d(geometry_data->m_vertices[v1_idx]),
-            Vector3d(geometry_data->m_vertices[v2_idx]));
+        if (geometry_data->m_motion_steps_count > 1)
+        {
+            const uint32 last_motion_step_idx = geometry_data->m_motion_steps_count - 1;
+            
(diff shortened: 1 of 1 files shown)


src/appleseed/renderer/kernel/intersection/embreescene.cpp
```diff
@@ -448,12 +448,47 @@ void EmbreeScene::intersect(ShadingPoint& shading_point) const
...
         const uint32 v2_idx = geometry_data->m_primitives[rayhit.hit.primID * 3 + 2];
 
-        const TriangleType triangle(
-            Vector3d(geometry_data->m_vertices[v0_idx]),
-            Vector3d(geometry_data->m_vertices[v1_idx]),
-            Vector3d(geometry_data->m_vertices[v2_idx]));
+        if (geometry_data->m_motion_steps_count > 1)
+        {
+            // N is the last motion step index.
+            const uint32 N = geometry_data->m_motion_steps_count - 1;
```
dictoon: Spell variables in lower case. May I suggest `last_motion_step_index` instead?

src/appleseed/renderer/kernel/intersection/embreescene.cpp
```diff
@@ -448,12 +448,47 @@ void EmbreeScene::intersect(ShadingPoint& shading_point) const
...
+            const uint32 motion_step_end_idx = motion_step_begin_idx + 1;
+
+            const uint32 motion_step_begin_offset = motion_step_begin_idx * geometry_data->m_vertices_count;
+            const uint32 motion_step_end_offset = motion_step_end_idx * geometry_data->m_vertices_count;
+
+            const float motion_step_begin_time = static_cast<float>(motion_step_begin_idx) / N;
+            
+            // Linear interpolation coefficients.
+            const float p = (rayhit.ray.time - motion_step_begin_time) * N;
+            const float q = (1.0f - p);
```
dictoon: Remove redundant parentheses.

src/appleseed/renderer/kernel/intersection/embreescene.cpp
fedormatantsev: Ah, no prob. It was just a variable name from my paper scribbling.

PR #2202, opened 2018-09-18
Refactor project points out of tests and expose it to Python

est77:


src/appleseed/renderer/meta/tests/test_pinholecamera.cpp
```diff
@@ -50,41 +45,20 @@ TEST_SUITE(Renderer_Modeling_Camera_PinholeCamera)
...
-        auto_release_ptr<Scene> scene(SceneFactory::create());
-        scene->cameras().insert(
-            PinholeCameraFactory().create(
-                "camera",
-                ParamArray()
-                    .insert("film_width", "0.025")
-                    .insert("film_height", "0.025")
-                    .insert("focal_length", "0.035")));
+        auto_release_ptr<Camera> camera(
+                    PinholeCameraFactory().create(
```
dictoon: Fix indentation.

src/appleseed/renderer/meta/tests/test_pinholecamera.cpp
```diff
@@ -50,41 +45,20 @@ TEST_SUITE(Renderer_Modeling_Camera_PinholeCamera)
...
-        OnFrameBeginRecorder frame_begin_recorder;
-        success = project->get_scene()->on_frame_begin(project.ref(), nullptr, frame_begin_recorder);
-        ASSERT_TRUE(success);
-
-        const Camera* camera = project->get_scene()->get_active_camera();
+        const ProjectPoints proj(camera, Vector2u(512, 512));
 
         Vector2d projected;
-        success = camera->project_point(0.0f, Vector3d(0.0, 0.0, -1.0), projected);
+        bool success = proj.project_point(0.0f, Vector3d(0.0, 0.0, -1.0), projected);
```
dictoon: Make const.

src/appleseed/renderer/modeling/camera/projectpoints.h
```diff
@@ -0,0 +1,84 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+#ifndef APPLESEED_RENDERER_MODELING_CAMERA_PROJECTPOINTS_H
+#define APPLESEED_RENDERER_MODELING_CAMERA_PROJECTPOINTS_H
+
+// appleseed.renderer headers.
```
dictoon: Remove empty section.

src/appleseed/renderer/modeling/camera/projectpoints.cpp
```diff
@@ -0,0 +1,142 @@
...
+{
+
+struct ProjectPoints::Impl
+{
+    Impl()
+      : m_camera(nullptr)
+    {
+    }
+
+    Camera*                     m_camera;
```
dictoon: Move members before methods.

src/appleseed/renderer/modeling/camera/projectpoints.cpp
```diff
@@ -0,0 +1,142 @@
...
+};
+
+ProjectPoints::ProjectPoints(
+    auto_release_ptr<Camera> camera,
+    const Vector2u&          resolution)
+  : impl(new Impl())
+{
+    auto_release_ptr<Scene> scene = SceneFactory::create();
+
+    const string camera_name = camera->get_name();
```
dictoon: Get rid of `camera_name` variable, use `camera->get_name()` directly when creating the frame.

src/appleseed/renderer/modeling/camera/projectpoints.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+    bool success =
+        impl->m_project->get_scene()->on_render_begin(
+            impl->m_project.ref(),
+            nullptr,
+            impl->m_render_begin_recorder);
+
+    if (!success)
+    {
+        RENDERER_LOG_ERROR("Error initializing ProjectPoints: on_render_begin failed.");
```
dictoon: I would prefer if this component didn't emit messages (apart from those emitted by cameras). 

What about setting a `m_is_initialized` member to true if construction went well, and add an `is_initialized()` method that the caller should use to check that construction went well? The caller is then free to emit a message or throw an exception if it needs to.

It's a pretty standard pattern; it's even used in the C++ standard library.

src/appleseed/renderer/modeling/camera/projectpoints.cpp
```diff
@@ -0,0 +1,142 @@
...
+// appleseed.renderer headers.
+#include "renderer/modeling/entity/onframebeginrecorder.h"
+#include "renderer/modeling/entity/onrenderbeginrecorder.h"
+#include "renderer/modeling/frame/frame.h"
+#include "renderer/modeling/project/project.h"
+#include "renderer/modeling/scene/scene.h"
+#include "renderer/utility/paramarray.h"
+
+// appleseed.foundation headers.
+#include "foundation/utility/iostreamop.h"
```
dictoon: I don't think you need that?

src/appleseed/renderer/modeling/camera/projectpoints.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+// Standard headers.
+#include <string>
```
dictoon: You won't need that anymore.

src/appleseed/renderer/modeling/camera/projectpoints.cpp
est77: camera is an auto_release_ptr and it is gone after I insert it into the scene.
I need to save the name before inserting it into the scene.

src/appleseed/renderer/modeling/camera/projectpoints.cpp
est77: I need it to use Vectors with ParamArray.

est77:
Fixed and rebased

src/appleseed/renderer/modeling/camera/projectpoints.cpp
dictoon: Ok.

src/appleseed/renderer/modeling/camera/projectpoints.cpp
dictoon: Ok.

src/appleseed/renderer/modeling/camera/projectpoints.cpp
```diff
@@ -0,0 +1,141 @@
...
+{
+    return impl->m_camera != nullptr;
+}
+
+bool ProjectPoints::project_point(
+    const float              time,
+    const Vector3d&          point,
+    Vector2d&                ndc) const
+{
+    if (is_initialized())
```
dictoon: It would make more sense to assert. One isn't supposed to use `ProjectPoints` if it couldn't be properly initialized.

src/appleseed/renderer/modeling/camera/projectpoints.cpp
```diff
@@ -0,0 +1,141 @@
...
+{
+    impl->m_frame_begin_recorder.on_frame_end(impl->m_project.ref());
+    impl->m_render_begin_recorder.on_render_end(impl->m_project.ref());
+
+    delete impl;
+}
+
+bool ProjectPoints::is_initialized() const
+{
+    return impl->m_camera != nullptr;
```
dictoon: This works but it's slightly hacky to rely on the camera being set _last_ in the constructor. An explicit flag would be cleaner.

est77:
Updated and rebased

est77:
> One last thing: I think projectpoints.{h,cpp} really belongs to the appleseed.python project

I want to keep it accessible from both C++ and Python. You never know when are you going to need it and the cost is almost nothing.

> I'm not too comfortable keeping this in the middle of the cameras.

I agree. I moved ProjectPoints to renderer/utility.

src/appleseed/CMakeLists.txt
```diff
@@ -2019,6 +2019,8 @@ set (renderer_utility_sources
     renderer/utility/messagecontext.h
     renderer/utility/paramarray.cpp
     renderer/utility/paramarray.h
+    renderer/utility/projectpoints.cpp
```

PR #2204, opened 2018-09-18
Fix OSL texture wrap modes

est77:


dictoon:
Worth checking if we have that bug anywhere else (don't we have other similar texture shaders at the moment?).

PR #2229, opened 2018-09-30
Ignore libraries whose path contains `@rpath` in macOS

pjessesco:
Ignore libraries whose path contains `@rpath`


pjessesco:
Is it make sense that I add commits for bundle python in this PR?

PR #2232, opened 2018-10-08
Add MatteSurface closure and shader, change samplerenderer to support it

usakhelo:
Hi,
This is my attempt to implement matte surface shader. There's a new closure and the shader, getting two parameters. These parameters are passed to `genericsamplerenderer` and it modifies rgb and alpha for the matte sample.
Hopefully closure and parameter passing are implemented correctly.
Shader has 3ds max class id, so it's ready to be tested in max.

image

image

image

Thanks for review,
Sergo.


dictoon:
Amazing!

Small suggestion: maybe enlarge the Input Surface button to take advantage of the free space on its left?

image

src/appleseed.shaders/CMakeLists.txt
```diff
@@ -303,6 +303,7 @@ set (src_appleseed_sources
      src/appleseed/as_vary_color.osl
      src/appleseed/as_voronoi2d.osl
      src/appleseed/as_voronoi3d.osl
+     src/appleseed/as_matte_surface.osl
```

src/appleseed.shaders/src/appleseed/as_matte_surface.osl
```diff
@@ -0,0 +1,85 @@
+
```
dictoon: Add copyright notice.

src/appleseed.shaders/src/appleseed/as_matte_surface.osl
```diff
@@ -0,0 +1,85 @@
...
+#include "appleseed/math/as_math_helpers.h"
+
+shader as_matte_surface
+[[
+    string as_maya_node_name = "asMatteSurface",
+    string as_maya_classification = "drawdb/shader/surface:rendernode/appleseed/surface:shader/surface:swatch/AppleseedMatteSurface",
+    string as_blender_node_name = "asMatteSurface",
+    string as_blender_category = "shader",
+    string help = "Matte Surface",
+    string icon = "",
```
dictoon: Maybe @luisbarrancos can provide an icon? Could you ask him please?

src/appleseed/renderer/kernel/shading/shadingcontext.h
```diff
@@ -132,6 +132,12 @@ class ShadingContext
         const ShaderGroup&  shader_group,
         const ShadingPoint& shading_point) const;
 
+    void execute_osl_matte(
+        const ShaderGroup&      shader_group,
```
dictoon: Align arguments vertically. While you're at it, maybe you can do the same for `execute_osl_npr()`'s arguments.

src/appleseed/renderer/modeling/shadergroup/shadergroup.h
```diff
@@ -122,6 +122,9 @@ class APPLESEED_DLLSYMBOL ShaderGroup
     // Return true if the shader group contains at least one NPR closure.
     bool has_npr() const;
 
+    // Return true if the shader group contains at least one Matteshadow closure.
```
dictoon: *Matteshadow closure*?

src/appleseed/renderer/modeling/shadergroup/shadergroup.cpp
```diff
@@ -227,6 +229,7 @@ bool ShaderGroup::create_optimized_osl_shader_group(
         report_has_closure(g_holdout_str.c_str(), HasHoldout);
         report_has_closure(g_debug_str.c_str(), HasDebug);
 
+        report_has_closure(g_mattesurface_str.c_str(), HasMatte);
```
dictoon: Move after the next line to respect the order you've followed everywhere (matte after NPR).

src/appleseed/renderer/kernel/rendering/generic/genericsamplerenderer.cpp
```diff
@@ -253,6 +254,22 @@ namespace
                 if (!shading_point_ptr->hit_surface())
                     break;
 
+                const Material* material = shading_point_ptr->get_material();
+                if (material &&
```
dictoon: We now try to be explicit in pointer comparisons:

    if (material != nullptr &&

src/appleseed/renderer/kernel/lighting/pathtracer.h
```diff
@@ -355,6 +355,14 @@ size_t PathTracer<PathVisitor, VolumeVisitor, Adjoint>::trace(
             continue;
         }
 
+        // Apply OSL holdout.
+        if (material_data.m_shader_group &&
+            material_data.m_shader_group->has_matte() &&
+            vertex.m_path_length == 1)
```
dictoon: Make `vertex.m_path_length == 1` the first condition since it's going to evaluate to false for all bounces but the first, hence short-circuiting the evaluation of the remaining conditions (it's a micro-optimization).

src/appleseed/renderer/kernel/lighting/pathtracer.h
```diff
@@ -355,6 +355,14 @@ size_t PathTracer<PathVisitor, VolumeVisitor, Adjoint>::trace(
             continue;
         }
 
+        // Apply OSL holdout.
```
dictoon: Maybe "Handle" rather than "Apply"?

src/appleseed/renderer/kernel/lighting/pathtracer.h
```diff
@@ -355,6 +355,14 @@ size_t PathTracer<PathVisitor, VolumeVisitor, Adjoint>::trace(
...
+        if (material_data.m_shader_group &&
```
dictoon: Explicitly compare against `nullptr`.

src/appleseed/renderer/kernel/rendering/generic/genericsamplerenderer.cpp
```diff
@@ -253,6 +254,22 @@ namespace
                 if (!shading_point_ptr->hit_surface())
                     break;
 
+                const Material* material = shading_point_ptr->get_material();
```
dictoon: Add a short comment summarizing what this code block does.

My main question here is: is this code where it ought to be (that is, between checking whether we hit the environment and whether we've reached full opacity)?

src/appleseed/renderer/kernel/lighting/pathtracer.h
dictoon: I'm not entirely sure to understand why holdout must be handled both in the Generic Sample Renderer and here.

src/appleseed/renderer/kernel/shading/closures.h
```diff
@@ -122,6 +122,9 @@ enum ClosureID
     NPRShadingID,
     NPRContourID,
 
+    // Matte closures.
+    MatteSurfaceID,
```
dictoon: I think that should go in the "Special closures" section.

src/appleseed/renderer/kernel/rendering/generic/genericsamplerenderer.cpp
usakhelo: I thought it's first thing to check - if it's matte surface or not, that's why I put it right after surface hit check.

src/appleseed/renderer/kernel/lighting/pathtracer.h
usakhelo: Ok, I checked and yes, even with current version this section is not required, but I think it speeds things up in case of the matte surface.

src/appleseed/renderer/kernel/shading/closures.cpp
```diff
@@ -2367,6 +2409,20 @@ namespace
                 }
                 break;
 
+              case MatteSurfaceID:
```
est77: We can remove this block. I don't think we need it.

src/appleseed/renderer/modeling/shadergroup/shadergroup.cpp
```diff
@@ -346,6 +349,8 @@ void ShaderGroup::get_shadergroup_closures_info(OSLShadingSystem& shading_system
                 m_flags |= HasNPR;
             else if (closures[i] == g_npr_contour_str)
                 m_flags |= HasNPR;
+            else if (closures[i] == g_mattesurface_str)
```
est77: Here, test for holdout too. Remove the previous holdout test.

src/appleseed/renderer/modeling/shadergroup/shadergroup.h
```diff
@@ -151,17 +154,19 @@ class APPLESEED_DLLSYMBOL ShaderGroup
         HasHoldout      = 1 << 4,
```
est77: Since holdout is a special case of Matte, lets remove HasHoldout and treat holdouts as matte.

src/appleseed/renderer/kernel/shading/closures.cpp
```diff
@@ -2400,6 +2456,21 @@ Color3f process_background_tree(const OSL::ClosureColor* ci)
     return do_process_closure_id_tree(ci, BackgroundID);
 }
 
+void process_matte_tree(const OSL::ClosureColor* ci, foundation::Color3f& matte_color, Alpha& matte_alpha)
+{
+    if (ci && ci->id == MatteSurfaceID)
```
est77: It is not guaranteed that there is only one closure and that it is the one we are looking for.
We need to traverse the tree (handle OSL::ClosureColor::MUL and OSL::ClosureColor::ADD and use recursion) 
like in the other process_closure_tree functions.

I am not sure what should we do if we get a shader like: 

0.75 * as_matte(colA, alphaA) + 0.1 * as_matte(colB, alphaB)

Maybe as_matte(0.75 * colA + 0.1 * colB, 0.75 * alphaA + 0.1 * alphaB) ?

src/appleseed/renderer/kernel/rendering/generic/genericsamplerenderer.cpp
usakhelo: I think I should have done compound the same way as `CompositeEmissionClosure::process_closure_tree`. I'll fix it tomorrow.

usakhelo:
I think I should have done compound the same way as `CompositeEmissionClosure::process_closure_tree`. I'll fix it tomorrow.

src/appleseed/renderer/kernel/shading/closures.cpp
```diff
@@ -2456,17 +2442,47 @@ Color3f process_background_tree(const OSL::ClosureColor* ci)
...
+            {
+                const OSL::ClosureComponent* c = reinterpret_cast<const OSL::ClosureComponent*>(closure);
+
+                if (c->id == MatteSurfaceID)
+                {
+                    const MatteClosure::Params* p = static_cast<const MatteClosure::Params*>(c->data());
+                    matte_alpha.set(p->matte_alpha);
+                    matte_color = p->matte_color;
+                }
+            }
```
est77: We need to also check for HoldoutID here and do matte_alpha = 0, matte_color = 0 if found.

usakhelo: Hi @est77 , I've made this change and then I changed so any shader except Matte will return zero color and zero alpha. It made blending matte with normal shaders to not fail. Now blending of two mattes works correctly, but blending with normal shader looks as if matte is blended with 0.

usakhelo:
I also renamed MatteSurface to Matte as @luisbarrancos suggested.
One other interesting thing with blending - when I put matte into first blend input and make multiplier <1.0 it makes matte darker, but also normal material darker in reflection/refraction. It should probably somehow skip any processing if it's not camera ray. Or probably blend should be matte-aware.

src/appleseed/renderer/kernel/shading/closures.cpp
est77: We don't need the else part. Just ignore and skip anything that is not a MatteID.

est77:
We are getting close!

I was experimenting with it too, trying to avoid executing the shader once more.
What do you think of something like this:

`
void ShadingEngine::shade_hit_point(
    SamplingContext&            sampling_context,
    const PixelContext&         pixel_context,
    const ShadingContext&       shading_context,
    const ShadingPoint&         shading_point,
    AOVAccumulatorContainer&    aov_accumulators,
    ShadingResult&              shading_result) const
{
    // Compute the alpha channel of the main output.
    shading_result.m_main.a = shading_point.get_alpha()[0];

    // Retrieve the material of the intersected surface.
    const Material* material = shading_point.get_material();

    // Apply OSL transparency and matte if needed.
    if (material && material->get_render_data().m_shader_group)
    {
        const ShaderGroup* sg = material->get_render_data().m_shader_group;

        if (sg->has_transparency() || sg->has_matte())
        {
            // This is similar to other functions but
            // does not call process_xxx_tree().
            // Folding the two cases together, we save one shading group exec.
            shading_context.execute_osl_transparency_and_matte(
                *material->get_render_data().m_shader_group,
                shading_point);

            if (sg->has_matte())
            {
                Color3f matte_color;
                Alpha matte_alpha;
                const bool any_matte_closure = process_matte_tree(
                    shading_point.get_osl_shader_globals().Ci,
                    matte_color,
                    matte_alpha);

                if (any_matte_closure)
                {
                    shading_result.m_main = (matte_color, matte_alpha);
                    return;
                }
            }

            if (sg->has_transparency())
            {
                Alpha alpha;
                process_transparency_tree(...);
                shading_result.m_main.a *= alpha[0];
            }
        }
    }
`

usakhelo:
I really like it @est77 . Does it mean matte happens before all the tracing? That's cool. genericsamplerenderer.cpp will not need to be changed, cool.
I had plans to add other closures like refraction transparency, which probably can also go here.

usakhelo:
I'll fix it in coming days. Thanks @est77 !

usakhelo:
Hi @est77 , I've tried this version and `shading_result.apply_alpha_premult();` makes any color black if alpha 0.0. Also I still had to break the loop in samplerenderer, otherwise matte objects look as they are invisible, not cut-out.
Is it possible to move `alpha_premult()` after the matte check?

dictoon:
@usakhelo What's the status of this changeset? Can you rebase it on top of master and squash it please?

usakhelo:
Hi @dictoon , I've made some changes but I'm not sure if @est77 reviewed them. Rebased and squashed commits.

src/appleseed/renderer/kernel/shading/shadingengine.cpp
```diff
@@ -97,25 +98,52 @@ void ShadingEngine::shade_hit_point(
...
+                Color3f matte_rgb;
+                Alpha matte_alpha;
+                const bool any_matte_closure = process_matte_tree(
+                    shading_point.get_osl_shader_globals().Ci,
+                    matte_rgb,
+                    matte_alpha);
+
+                if (any_matte_closure)
+                {
+                    matte_color.rgb() = matte_rgb;
```
est77: can we write matte_color directly to shading_result?

usakhelo: I moved matte check to sample renderer to break the loop somehow. Otherwise it renders everything beneath the matte surface as if it's not visible.

src/appleseed/renderer/kernel/shading/closures.cpp
```diff
@@ -2392,14 +2434,57 @@ void process_transparency_tree(const OSL::ClosureColor* ci, Alpha& alpha)
...
 {
-    return do_process_closure_id_tree(ci, BackgroundID);
+    if (closure)
+    {
+        switch (closure->id)
+        {
+          case OSL::ClosureColor::MUL:
+            {
+                const OSL::ClosureMul* c = reinterpret_cast<const OSL::ClosureMul*>(closure);
+                process_matte_tree(c->closure, matte_color, matte_alpha);
```
est77: We are loosing the result of the process_matte_tree here.
return process_matte_tree(...)

src/appleseed/renderer/kernel/shading/closures.cpp
```diff
@@ -2392,14 +2434,57 @@ void process_transparency_tree(const OSL::ClosureColor* ci, Alpha& alpha)
...
+                matte_alpha.set(saturate(luminance(Color3f((c->weight) * matte_alpha[0]))));
+            }
+            break;
+
+          case OSL::ClosureColor::ADD:
+            {
+                const OSL::ClosureAdd* c = reinterpret_cast<const OSL::ClosureAdd*>(closure);
+                foundation::Color3f color_a(0.0f), color_b(0.0f);
+                Alpha alpha_a(0.0f), alpha_b(0.0f);
+                process_matte_tree(c->closureA, color_a, alpha_a);
```
est77: return process_matte_tree(c->closureA, color_a, alpha_a) || process_matte_tree(c->closureB, color_b, alpha_b);

src/appleseed/renderer/kernel/shading/shadingengine.cpp
est77: I see. Maybe we can set matte color and matte alpha here and return any_matte_closure?
In the genericsamplerenderer if shade returns true, we can stop.

src/appleseed/renderer/kernel/rendering/generic/genericsamplerenderer.cpp
```diff
@@ -253,6 +258,17 @@ namespace
                 if (!shading_point_ptr->hit_surface())
                     break;
 
+                // Check if matte shader is enabled and apply alpha and color if it is.
+                const Material* material = shading_point_ptr->get_material();
+                if (material != nullptr &&
+                    material->get_render_data().m_shader_group != nullptr &&
+                    material->get_render_data().m_shader_group->has_matte())
```
est77: This test is not correct. has_matte() means that the result of executing the shadergroup may contain matte closures, 
not that it always contains matte closures. We need to get that information from the shading engine after executing the shader and knowing for sure that as_matte() is present.

src/appleseed/renderer/kernel/shading/closures.cpp
```diff
@@ -2462,10 +2463,11 @@ bool process_matte_tree(
                 const OSL::ClosureAdd* c = reinterpret_cast<const OSL::ClosureAdd*>(closure);
                 foundation::Color3f color_a(0.0f), color_b(0.0f);
                 Alpha alpha_a(0.0f), alpha_b(0.0f);
-                process_matte_tree(c->closureA, color_a, alpha_a);
-                process_matte_tree(c->closureB, color_b, alpha_b);
+                bool is_matte_a = process_matte_tree(c->closureA, color_a, alpha_a);
+                bool is_matte_b = process_matte_tree(c->closureB, color_b, alpha_b);
                 matte_color = color_a + color_b;
                 matte_alpha.set(alpha_a[0] + alpha_b[0]);
+                return is_matte_a || is_matte_b;
```
usakhelo: I made this part `return process_matte_tree(c->closureA, color_a, alpha_a) || process_matte_tree(c->closureB, color_b, alpha_b)`, as you suggested but then I thought maybe we need to combine matte colors? I'm not sure if it's needed, but it's possible to blend mattes with blend shader.

usakhelo:
Hi @est77, I made requested changes but there are couple of notes that I added to the source code.

usakhelo:
Hi @est77 , I've rebased the branch, hope that makes is it easier to check.

PR #2240, opened 2018-10-17
Allow skipping shadow rays for environment EDFs and non-physical lights

KarthikRIyer:
#140


src/appleseed/renderer/modeling/environmentedf/environmentedf.h
```diff
@@ -113,8 +113,13 @@ class APPLESEED_DLLSYMBOL EnvironmentEDF
     virtual float evaluate_pdf(
         const foundation::Vector3f& outgoing) const = 0;        // world space emission direction, unit-length
 
+    virtual bool get_cast_shadows() const;
```
dictoon: I don't think any of these methods need to be virtual.

src/appleseed/renderer/modeling/environmentedf/environmentedf.cpp
```diff
@@ -102,4 +102,18 @@ bool EnvironmentEDF::on_frame_begin(
     return true;
 }
 
+bool EnvironmentEDF::get_cast_shadows() const
+{
+
```
dictoon: Remove redundant blank lines.

src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
```diff
@@ -402,14 +402,20 @@ void DirectLightingIntegrator::add_emitting_triangle_sample_contribution(
 
     // Compute the transmission factor between the light sample and the shading point.
     Spectrum transmission;
-    m_material_sampler.trace_between(
-        m_shading_context,
-        sample.m_point,
-        transmission);
+    if(material->get_cast_shadows()){
```
dictoon: Please read and follow our coding conventions: https://github.com/appleseedhq/appleseed/wiki/Coding-Philosophy-and-Guidelines

src/appleseed/renderer/modeling/environmentedf/environmentedf.h
```diff
@@ -113,8 +113,13 @@ class APPLESEED_DLLSYMBOL EnvironmentEDF
...
+
+    virtual void set_cast_shadows(const bool cast_shadows);
+
   protected:
     TransformSequence m_transform_sequence;
+    bool m_env_cast_shadows = true;
```
dictoon: We can't initialize members directly because we still need to support VS 2012 which doesn't support this feature.

src/appleseed/renderer/modeling/material/material.cpp
```diff
@@ -270,4 +270,20 @@ IBasisModifier* Material::create_basis_modifier(const MessageContext& context) c
...
+bool Material::get_cast_shadows() const
+{
+
+  return m_cast_shadows;
+
+}
+
+void Material::set_cast_shadows(const bool cast_shadows)
+{
+    if(has_emission())
```
dictoon: I don't think it makes sense to test if the material has emission.

dictoon:
The VS 2012 build did fail on AppVeyor.

KarthikRIyer:
Right now, for the test scenes I have to turn off shadows by making the boolean false in the constructor and build again. So when I export the .appleseed file from Blender, the boolean flag isn't saved. Is that fine?

est77:
You need to add the new option (for lights) here:

https://github.com/appleseedhq/appleseed/blob/master/src/appleseed/renderer/modeling/light/ilightfactory.cpp#L44

Then, do the same thing we do for the "cast_indirect_light" option. 

For environments, it is similar.

You can remove all the setter methods()

src/appleseed/renderer/modeling/material/material.cpp
```diff
@@ -87,6 +87,8 @@ Material::Material(
     set_name(name);
 
     m_inputs.declare("surface_shader", InputFormatEntity, "");
+
+    m_cast_shadows = true;
```
est77: I would remove the option to disable cast shadows for materials. 
The option should be in the EDF class, but this can be hard to do in some cases, like OSL shaders with emission.
We can think about how can we do it and leave this part for later

KarthikRIyer: Ok. I'll make the changes.

src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
```diff
@@ -402,7 +402,8 @@ void DirectLightingIntegrator::add_emitting_triangle_sample_contribution(
 
     // Compute the transmission factor between the light sample and the shading point.
     Spectrum transmission;
-    if(material->get_cast_shadows()){
+    if(material->get_cast_shadows())
```
dictoon: Insert space between `if` and opening parenthesis.

src/appleseed/renderer/modeling/environmentedf/environmentedf.cpp
```diff
@@ -62,6 +62,7 @@ EnvironmentEDF::EnvironmentEDF(
   : ConnectableEntity(g_class_uid, params)
 {
     set_name(name);
+    m_env_cast_shadows = true;
```
dictoon: Move to initializer list.

src/appleseed/renderer/modeling/material/material.cpp
```diff
@@ -272,18 +274,12 @@ IBasisModifier* Material::create_basis_modifier(const MessageContext& context) c
...
-
   return m_cast_shadows;
-
 }
 
 void Material::set_cast_shadows(const bool cast_shadows)
 {
-    if(has_emission())
-    {
         m_cast_shadows = cast_shadows;
```
dictoon: Fix indentation

KarthikRIyer:
I added options for lights and environment. It works properly initially for gradient environment. If I switch to some other kind of environment it doesn't seem to work properly. Even if I switch back to gradient, the environment colour changes and Even after switching off shadow, there is still a faint shadow visible.
image

image

dictoon:
I don't know about the other issues you mentioned, but regarding this one:

> [...] there is still a faint shadow visible

It's to be expected. Rays that bounce off the ground plane might still be occluded by the cube and thus won't reach the environment, causing a shadow.

I think it would be expensive to handle this properly (we would need to check if the ray would have it one of the area light that doesn't cast shadows). For the time being, we should see the "Cast Shadows" flag as an optimization (i.e. don't bother casting shadow rays to small lights), not as an artistic control that totally removes shadows.

@est77 @Narann @luisbarrancos Opinions?

Narann:
From the artistic sense, the more you give, the best it is. :D

But it have to be consistent with how the raytracer work as a whole and I'm not sure it is a good idea here.

From a technical standpoint, I wonder how a light "can not" cast shadow. I mean, objects cast shadows and you choose to make object to don't cast shadow so they are virtually "invisible" to shadow rays.

To provide another "mind representation". Without light, the scene is all black, shadow only, The more light you put, the more the scene will be revealed. So shadow is always here, lights actually remove shadows from unoccluded areas.

You can't "add" shadow, so you can't "disable" shadow from a light. So technically nothing "cast shadow", objects simply "occlude" lights. What you can "disable" is an object to occlude a light (and, indirectly, "cast shadow").

My 2 cts.

Narann:
Thinking about it, it could be a `never_occluded` parameter on the light. This mean the light is sampled without doing any shadow ray intersection. This would result as a "video game light effect" (no shadow). But why not.

KarthikRIyer:
So, what changes should I make? Should I change the name of the flag variable?

src/appleseed/renderer/modeling/environmentedf/environmentedf.cpp
```diff
@@ -99,7 +99,17 @@ bool EnvironmentEDF::on_frame_begin(
...
+        m_env_cast_shadows = true;
+    else
+        m_env_cast_shadows = false;
+
     return true;
 }
 
+bool EnvironmentEDF::get_cast_shadows() const
+{
+  return m_env_cast_shadows;
```
est77: Use 4 spaces for indentation.

src/appleseed/renderer/modeling/light/light.cpp
```diff
@@ -145,4 +150,11 @@ void Light::sample(
             probability);
 }
 
+bool Light::get_cast_shadows() const
+{
+
```
est77: Remove empty lines

sandbox/tests/test scenes/lights/21 - non physical light shadow off.appleseed
```diff
@@ -0,0 +1,246 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- File generated by appleseed.so version 2.0.0-beta (SSE4.2). -->
+<project format_revision="27">
+    <search_paths>
+        <search_path>
+            /home/karthik/appleseed-git/appleseed/sandbox/shaders/appleseed
```
dictoon: Since the test scenes must run everywhere, they must not contain absolute paths.

src/appleseed/renderer/modeling/light/light.h
```diff
@@ -137,12 +137,15 @@ class APPLESEED_DLLSYMBOL Light
...
 
+    bool get_cast_shadows() const;
+
   protected:
     int m_flags;
 
   private:
     struct Impl;
     Impl* impl;
+    bool m_cast_shadows;
```
dictoon: The idea of the [PImpl idiom](https://en.cppreference.com/w/cpp/language/pimpl) is to hide implementation details in the .cpp file, so in theory `m_cast_shadows` should go in the PImpl, not here.

However, for performance reasons, we make some exceptions to this rule, as you can see with `m_flags` just above. I suggest moving `m_cast_shadows` next to it.

src/appleseed/renderer/modeling/light/light.cpp
```diff
@@ -113,6 +113,11 @@ bool Light::on_frame_begin(
     else
         m_flags &= ~CastIndirectLight;
 
+    if (m_params.get_optional<bool>("cast_shadows", true))
```
dictoon: Simpler:

    m_cast_shadows = m_params.get_optional<bool>("cast_shadows", true);

src/appleseed/renderer/modeling/environmentedf/mirrorballmapenvironmentedf.cpp
```diff
@@ -282,6 +282,15 @@ DictionaryArray MirrorBallMapEnvironmentEDFFactory::get_input_metadata() const
             .insert("use", "optional")
             .insert("help", "Environment exposure multiplier"));
 
+    metadata.push_back(
+        Dictionary()
+            .insert("name", "env_cast_shadows")
```
dictoon: I suggest simply calling it `cast_shadows` (same everywhere).

src/appleseed/renderer/modeling/environmentedf/environmentedf.cpp
```diff
@@ -99,7 +99,17 @@ bool EnvironmentEDF::on_frame_begin(
                 Matrix4d::make_rotation(rotation)));
     }
 
+    if (m_params.get_optional<bool>("env_cast_shadows", true))
```
dictoon: Simplify.

src/appleseed/renderer/modeling/environmentedf/environmentedf.cpp
```diff
@@ -99,7 +99,17 @@ bool EnvironmentEDF::on_frame_begin(
...
+        m_env_cast_shadows = true;
+    else
+        m_env_cast_shadows = false;
+
     return true;
 }
 
+bool EnvironmentEDF::get_cast_shadows() const
+{
+    return m_env_cast_shadows;
```
dictoon: `m_cast_shadows`

src/appleseed/renderer/modeling/environmentedf/latlongmapenvironmentedf.cpp
```diff
@@ -595,6 +595,15 @@ DictionaryArray LatLongMapEnvironmentEDFFactory::get_input_metadata() const
             .insert("use", "optional")
             .insert("help", "Environment texture vertical shift in degrees"));
 
+    metadata.push_back(
```
dictoon: I would suggest moving common metadata to a separate function in the base factory class (e.g. `add_common_sky_input_metadata()`), like it's done for some other entity types (sky environment EDFs, and maybe lights, IIRC).

src/appleseed/renderer/kernel/lighting/imagebasedlighting.cpp
```diff
@@ -207,16 +207,23 @@ void compute_ibl_environment_sampling(
 
         // Discard occluded samples.
         Spectrum transmission;
-        const ShadingPoint& shading_point = material_sampler.trace(
-            shading_context,
-            incoming,
-            transmission);
+        if(environment_edf.get_cast_shadows())
```
dictoon: Add space between `if` and parenthesis.

src/appleseed/renderer/kernel/lighting/imagebasedlighting.cpp
```diff
@@ -207,16 +207,23 @@ void compute_ibl_environment_sampling(
...
+                shading_context,
+                incoming,
+                transmission);
 
-        if (shading_point.hit_surface())
-            continue;
-        
-        if (max_value(transmission) == 0.0f)
-            continue;
+            if (shading_point.hit_surface()){}
```
dictoon: You have a very serious bug here (the `{}` at the end).

Did you test your changes?

src/appleseed/renderer/kernel/lighting/directlightingintegrator.cpp
```diff
@@ -506,14 +506,21 @@ void DirectLightingIntegrator::add_non_physical_light_sample_contribution(
 
     // Compute the transmission factor between the light sample and the shading point.
     Spectrum transmission;
-    m_material_sampler.trace_between(
-        m_shading_context,
-        emission_position,
-        transmission);
+    if(light->get_cast_shadows())
```
dictoon: Space after `if`.

sandbox/tests/test scenes/lights/ 23  - non physical light shadow on.appleseed
```diff
@@ -0,0 +1,239 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dictoon: Space at the beginning of the filename?

src/appleseed/renderer/modeling/environmentedf/environmentedf.cpp
KarthikRIyer: The variable defined is m_env_cast shadows. Do I need to rename it?

src/appleseed/renderer/kernel/lighting/imagebasedlighting.cpp
```diff
@@ -204,9 +204,23 @@ void compute_ibl_environment_sampling(
 
         // Discard occluded samples.
         Spectrum transmission;
-        material_sampler.trace_simple(shading_context, incoming, transmission);
-        if (is_zero(transmission))
-            continue;
+        if (environment_edf.get_cast_shadows())
+        {
+            const ShadingPoint& shading_point = material_sampler.trace_full(
```
dictoon: Why did you change the logic here?

src/appleseed/renderer/modeling/material/material.h
```diff
@@ -161,7 +161,8 @@ class APPLESEED_DLLSYMBOL Material
     const RenderData& get_render_data() const;
 
   protected:
-    RenderData m_render_data;
+    bool        m_has_render_data;
```
dictoon: Revert.

src/appleseed/renderer/kernel/lighting/imagebasedlighting.cpp
KarthikRIyer: Do you mean why I used trace_full instead of trace_simple? It's because trace_simple doesn't return the shading point.

dictoon: Why would you need the shading point? Why changing the logic (with a significant performance impact) when all is needed is wrapping the test in a "if cast shadows"?

KarthikRIyer: I'm sorry. I got confused. I was thinking about it a bit wrong. I'll update the PR.

src/appleseed/renderer/kernel/lighting/imagebasedlighting.cpp
```diff
@@ -204,9 +204,20 @@ void compute_ibl_environment_sampling(
 
         // Discard occluded samples.
         Spectrum transmission;
-        material_sampler.trace_simple(shading_context, incoming, transmission);
-        if (is_zero(transmission))
-            continue;
+        if (environment_edf.get_cast_shadows())
+        {
+            material_sampler.trace_simple(
```
dictoon: There's another `trace_simple()` call in that file that should obey the cast shadow flag, in `compute_ibl_material_sampling()`.

Make sure to re-render the reference images for your test scenes after you've done this change.

sandbox/tests/test scenes/lights/21 - non physical light shadow off.appleseed
```diff
@@ -0,0 +1,239 @@
...
+                <parameter name="next_event_estimation" value="true" />
+                <parameter name="optimize_for_lights_outside_volumes" value="True" />
+                <parameter name="record_light_paths" value="false" />
+                <parameter name="rr_min_path_length" value="6" />
+                <parameter name="volume_distance_samples" value="2" />
+            </parameters>
+            <parameters name="uniform_pixel_renderer">
+                <parameter name="decorrelate_pixels" value="true" />
+                <parameter name="force_antialiasing" value="true" />
+                <parameter name="samples" value="64" />
```
dictoon: That's quite a high sampling rate for a test scene. What's the render time like?

We try to have test scenes render in seconds rather than minutes, because there are so many of them (and sometimes we render the full test suite with a debug build of appleseed, that takes 18-24 hours already).

KarthikRIyer: With a release build it takes 5.856 seconds. But that's with 16 threads. With 4 threads it's around 14.376 seconds. Should 16 samples be enough?

dictoon: Oh that's quite fast, it's fine with 64 spp then. Thanks!

dictoon: Ah I see that you changed it to 16 spp, that's fine too.

dictoon:
@KarthikRIyer Upon closer inspection a number of issues remained, I fixed them in the commit above, the details are in the commit message.

I'll merge this when CI passes.

Thanks!

KarthikRIyer:
@dictoon thanks for pointing them out. I see that I paid less attention to some things. Now I've seen about some things I should be careful about while working on such stuff in the future.

PR #2258, opened 2018-10-21
Remove mesh regions

est77:


src/appleseed/renderer/kernel/intersection/embreescene.cpp
```diff
@@ -110,21 +109,16 @@ namespace
         // Retrieve the object.
         Object& object = object_instance.get_object();
 
-        // TODO: remove regions.
-        // Retrieve the region kit of the object.
-        const Access<RegionKit> region_kit(&object.get_region_kit());
+        const MeshObject& mesh = static_cast<const MeshObject&>(object);
```
dictoon: Is this safe?

src/appleseed/renderer/kernel/intersection/intersectionfilter.cpp
```diff
@@ -58,19 +56,9 @@ namespace
...
-        for (const_each<RegionKit> i = *region_kit; i; ++i)
-        {
-            const IRegion* region = *i;
-            Access<StaticTriangleTess> tess(&region->get_static_triangle_tess());
-
-            triangle_count += tess->m_primitives.size();
-        }
-
-        return triangle_count;
+        const MeshObject& mesh = static_cast<const MeshObject&>(object);
```
dictoon: Again, is this safe? Not all objects are mesh objects...

src/appleseed/renderer/kernel/intersection/intersectionfilter.h
```diff
@@ -158,8 +158,6 @@ inline bool IntersectionFilter::accept(
     const double                u,
     const double                v) const
 {
-    assert(triangle_key.get_region_index() == 0);
```
dictoon: Remove `<cassert>` (and include it in the .cpp file if necessary).

src/appleseed/renderer/kernel/intersection/triangletree.cpp
```diff
@@ -1101,11 +1084,23 @@ namespace
 
     // Collect the set of object instances referenced by a set of regions.
```
dictoon: Fix comment.

src/appleseed/renderer/modeling/scene/assembly.cpp
```diff
@@ -347,16 +345,6 @@ Dictionary AssemblyFactory::get_model_metadata() const
 DictionaryArray AssemblyFactory::get_input_metadata() const
 {
     DictionaryArray metadata;
-
-    metadata.push_back(
```
dictoon: We should provide a project migration step that removes this flag from assemblies.

est77: Done

sandbox/tests/unit tests/inputs/test_projectfilereader_configurationblocks.appleseed
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project format_revision="27">
+<project format_revision="28">
```
dictoon: It would be nice to find a better solution than having to update this data file every time :)

PR #2263, opened 2018-10-22
Create replace_header_guard script

jdent02:
Replaces #ifndef header guards with #pragma once and removes associated #define and #endif statements


dictoon:
Very nice, thanks!

How did you test the script? On many header files?

jdent02:
I tested src\appleseed\renderer\kernel\rendering and it picked up all the files in the subfolders as well when '-r' was used.

PR #2268, opened 2018-10-27
Hide node input in node editor (maya, blender, gaffer)

luisbarrancos:
Metal had a null widget for energy conservation since this was not be exposed to the user, but the node editor input socket was still visible if you expanded the attributes in Maya.
Added the token to disable and hide this input socket in Maya, Gaffer, Blender.
Max probably needs an similar metadata token to hide an existing input from the nodes in its graph editor.

### DIFF
--- a/src/appleseed.shaders/src/appleseed/as_fresnel.osl
+++ b/src/appleseed.shaders/src/appleseed/as_fresnel.osl
@@ -172,7 +172,7 @@ shader as_fresnel
 
         if (exterior_medium_ior != 1.0)
         {
-            float costheta = dot(normalize(N), -normalize(I));
+            float costheta = dot(normalize(N), -I);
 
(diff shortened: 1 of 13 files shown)


src/appleseed.shaders/src/appleseed/as_fresnel.osl
```diff
@@ -170,7 +170,7 @@ shader as_fresnel
 
         if (exterior_medium_ior != 1.0)
         {
-            float costheta = dot(normalize(N), -normalize(I));
+            float costheta = dot(normalize(N), -I); // I is unit lenght in as
```
dictoon: Typo. That said I would simply remove those comments...

dictoon:
(On hold until after the release.)

dictoon:
No longer on hold, this needs to be fixed and rebased.

PR #2273, opened 2018-10-29
Fix travis build on macOS

pjessesco:


dictoon:
Thanks for looking into this @pjessesco, much appreciated... What a mess right? :)

pjessesco:
I fixed script after I checked it doesn't work. Finally it seems working, I'll squash them. Sorry for annoying.

dictoon:
> Sorry for annoying.

Not at all! You proceeded exactly as we discussed. Awesome work man, thank you!

PR #2288, opened 2018-11-03
Add max ray intensity for SPPM

### LINKED ISSUE #2266: Enable ray intensity clamping for photon mapping
Add a `Max Ray Intensity` function to SPPM similar to the one for path tracing.

oktomus:
This close #2266.


dictoon:
Thanks!

I'll review the PR, but looking at the screenshot, I notice that the Advanced section doesn't have the same layout as the other sections (which are centered on the `:`).

src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
```diff
@@ -1275,6 +1283,27 @@ namespace
...
+
+            QVBoxLayout* layout = new QVBoxLayout();
+            groupbox->setLayout(layout);
+
+            create_advanced_max_ray_intensity_settings(layout);
+        }
+
+        void create_advanced_max_ray_intensity_settings(QVBoxLayout* parent)
+        {
+            QDoubleSpinBox* max_ray_intensity = create_double_input("advanced.max_ray_intensity", 0.0, 1.0e9, 3, 0.1);
```
dictoon: 1.0e9? Is that really the max value for MRI in the unidirectional path tracer?

Also precision of 3 digits after the decimal place seems excessive, aren't we using 1 in UPT?

src/appleseed/renderer/kernel/lighting/sppm/sppmparameters.cpp
```diff
@@ -109,6 +109,8 @@ SPPMParameters::SPPMParameters(const ParamArray& params)
   , m_photon_tracing_rr_min_path_length(fixup_path_length(params.get_optional<size_t>("photon_tracing_rr_min_path_length", 6)))
   , m_path_tracing_max_bounces(fixup_bounces(params.get_optional<int>("path_tracing_max_bounces", -1)))
   , m_path_tracing_rr_min_path_length(fixup_path_length(params.get_optional<size_t>("path_tracing_rr_min_path_length", 6)))
+  , m_path_tracing_max_ray_intensity(params.get_optional<float>("path_tracing_max_ray_intensity", 0.0f))
+  , m_path_tracing_has_max_ray_intensity(params.get_optional<float>("path_tracing_max_ray_intensity", 0.0f) != 0.0f)
```
dictoon: Different logic than in UPT where we simply check if the `max_ray_intensity` is set or not. We must use the same logic everywhere, for our own sanity. The advantage of the logic in UPT is that one can set MRI to 0 (it's not useful in practice but there's no reason to not support this corner case if we can).

src/appleseed/renderer/utility/spectrumclamp.h
```diff
@@ -0,0 +1,78 @@
...
+//
+// Clamp the contribution of a given spectrum.
+//
+
+template <typename T>
+inline void clamp_contribution(Spectrum& radiance, const T high)
+{
+    const float avg = foundation::average_value(radiance);
+
+    assert(avg != T(0));
```
dictoon: This assertion is invalid: `avg` may very well be 0.

src/appleseed/renderer/utility/spectrumclamp.h
```diff
@@ -0,0 +1,78 @@
...
+
+namespace renderer
+{
+
+//
+// Clamp the contribution of a given spectrum.
+//
+
+template <typename T>
+inline void clamp_contribution(Spectrum& radiance, const T high)
```
dictoon: This doesn't need to be templatized, the only type that makes sense here is `float`.

src/appleseed/renderer/utility/spectrumclamp.h
```diff
@@ -0,0 +1,78 @@
...
+
+#ifndef APPLESEED_RENDERER_UTILITY_SPECTRUMCLAMP_H
+#define APPLESEED_RENDERER_UTILITY_SPECTRUMCLAMP_H
+
+// appleseed.renderer headers.
+#include "renderer/global/globaltypes.h"
+#include "renderer/kernel/shading/directshadingcomponents.h"
+
+// appleseed.foundation headers.
+#include "foundation/image/color.h"
```
dictoon: I don't think we need this header?

src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
oktomus: I copied the line from the UPT and haven't modified anything

src/appleseed/renderer/kernel/lighting/sppm/sppmparameters.cpp
oktomus: ok

src/appleseed/renderer/utility/spectrumclamp.h
oktomus: Indeed, my bad

src/appleseed/renderer/utility/spectrumclamp.h
oktomus: In fact. I orginaly wanted to template the clamp function using Color<T, N> but that wasn't a good idea

src/appleseed/renderer/utility/spectrumclamp.h
```diff
@@ -0,0 +1,76 @@
...
+
+//
+// Clamp the contribution of a given spectrum.
+//
+
+inline void clamp_contribution(Spectrum& radiance, const float high)
+{
+    const float avg = foundation::average_value(radiance);
+
+    assert(high != 0.0f);
```
dictoon: Still there?

src/appleseed/renderer/utility/spectrumclamp.h
```diff
@@ -0,0 +1,76 @@
...
+
+// appleseed.renderer headers.
+#include "renderer/global/globaltypes.h"
+#include "renderer/kernel/shading/directshadingcomponents.h"
+
+// appleseed.foundation headers.
+#include "foundation/image/color.h"
+
+// Standard headers.
+#include <cassert>
```
dictoon: Remove.

src/appleseed/renderer/utility/spectrumclamp.h
dictoon: Remove.

Issue #2276, opened 2018-10-29
Unify DCC specific OSL metadata in shaders when possible


est77:
We have some per DCC duplicated metadata entries, for example as_blender_input_socket, as_maya_connectable, ...
When possible, we should have instead only an as_connectable entry.

luisbarrancos:
I think that'll be maya_attribute_hidden, for the node editor input sockets, which matches the blender_input_socket, and gaffer's gafferNoduleLayoutVisible.

PR #2289, opened 2018-11-04
Gafferseed work

est77:
Part of #2276

- Remove duplicated OSL shader.
- Unify top level shaders metadata:
      as_maya_node_name , as_blender_node_name -> as_node_name
      as_blender_category -> as_category (now also used in Gafferseed)

- Small code improvements and tweaks.


src/appleseed/foundation/platform/compiler.h
```diff
@@ -223,6 +223,18 @@ namespace foundation
 #endif
 
 
+//
+//  noexcept support.
+//
+
+// Visual C++ 2012 and earlier.
+#if !defined(_MSC_VER) || _MSC_VER >= 1800
+    #define APPLESEED_NOEXCEPT
```
dictoon: Incorrect, it's the other way around.

Here's a correct implementation, you can copy/paste it:

    // Visual C++ 2012 and earlier: ignore.
    #if defined _MSC_VER && _MSC_VER < 1800
        #define APPLESEED_NOEXCEPT

    // Other compilers.
    #else
        #define APPLESEED_NOEXCEPT noexcept
    #endif

src/appleseed.shaders/CMakeLists.txt
```diff
@@ -145,7 +145,6 @@ source_group ("include\\appleseed\\transform" FILES
 
 set (src_blenderseed_sources
      src/blenderseed/as_blender_areaLight.osl
-     src/blenderseed/as_closure2surface.osl
```
dictoon: Is this shader currently used by blenderseed? If yes, we can't simply remove it, that would break all scenes created until the next release of appleseed/blenderseed.

est77: We have an exact copy of it in the appleseed folder. In my tests, we pick the appleseed one without issues.

src/appleseed/foundation/platform/compiler.h
est77: Fixed

src/appleseed.shaders/CMakeLists.txt
dictoon: Ok, cool.

PR #2297, opened 2018-11-12
Revert "Fix shader name capitalization"

luisbarrancos:
Reverts appleseedhq/appleseed#2296
The shader name change will break backwards compat.


luisbarrancos:
Actually, i'll make a new PR.

PR #2300, opened 2018-11-15
Fix time on render stamp when pausing rendering

### LINKED ISSUE #2290: Render Stamp shows incorrect render time when Pause/Resume is used
When a render is paused and resumed the render stamp shows the time inclusive the pause where the render is suspended instead of the active render time (see screenshot, bottom left corner is the correct render time).

capture

oktomus:
This fix #2290


dictoon:
Awesome, thanks!

PR #2301, opened 2018-11-15
Add triangulation stage to C++ blenderseed exporter

jdent02:


src/appleseed.python/bindblendermeshconverter.cpp
```diff
@@ -130,45 +124,90 @@ void convert_bl_mesh(
...
+            blender_mesh->push_triangle(
+                Triangle(
+                    face.v[2],
+                    face.v[3],
+                    face.v[0],
+                    face.mat_nr));
+        }
     }
 
+    size_t bl_face_index = 0;
```
oktomus: Can't it be initialized directly in the for loop ? Like this

oktomus: `         for (size_t tri_index = 0, bl_face_index = 0; tri_index < triangle_length; ++tri_index, ++bl_face_index) `

src/appleseed.python/bindblendermeshconverter.cpp
```diff
@@ -130,45 +124,90 @@ void convert_bl_mesh(
...
+            {
+                Triangle& tri_b = blender_mesh->get_triangle(++bl_face_index);
+                tri_b.m_n0 = face.v[2];
+                tri_b.m_n1 = face.v[3];
+                tri_b.m_n2 = face.v[0];
+            }
         }
     }
 
+    size_t bl_uv_index = 0;
```
oktomus: Same as bl_face_index

src/appleseed.python/bindblendermeshconverter.cpp
```diff
@@ -130,45 +124,88 @@ void convert_bl_mesh(
...
+
     // Tie vertex normals to mesh faces
     if (export_normals == true)
     {
-        for (size_t tri_index = 0; tri_index < triangle_length; ++tri_index)
+        for (size_t tri_index = 0; tri_index < triangle_length; ++tri_index, ++bl_face_index)
         {
             MeshFace &face = triangles[tri_index];
-            Triangle& tri = blender_mesh->get_triangle(tri_index);
+            Triangle& tri = blender_mesh->get_triangle(bl_face_index);
```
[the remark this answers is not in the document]
jdent02: Yes but the issue is there may be two triangles for every blender UV face, so both will have to be retrieved in the loop.  You can see the second tri is actually called tri_b

src/appleseed.python/bindblendermeshconverter.cpp
```diff
@@ -130,45 +124,88 @@ void convert_bl_mesh(
...
             blender_mesh->push_tex_coords(
                 GVector2(
                     tex_face.uv[2][0],
                     tex_face.uv[2][1]));
-            tri.m_a2 = uv_index++;
+            tri.m_a2 = uv_vertex_index++;
+
+            if (tex_face.uv[3][0] != 0)
+            {
+                Triangle& tri_b = blender_mesh->get_triangle(++bl_uv_index);
```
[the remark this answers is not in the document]
jdent02: See above

src/appleseed.python/bindblendermeshconverter.cpp
```diff
@@ -130,45 +124,88 @@ void convert_bl_mesh(
...
+            tri.m_a1 = uv_vertex_index++;
 
             blender_mesh->push_tex_coords(
                 GVector2(
                     tex_face.uv[2][0],
                     tex_face.uv[2][1]));
-            tri.m_a2 = uv_index++;
+            tri.m_a2 = uv_vertex_index++;
+
+            if (tex_face.uv[3][0] != 0)
```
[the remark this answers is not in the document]
jdent02: I'll check what Blender shows for a UV face with only tris but there might not be another way.

jdent02:
Would it make more sense maybe to have the triangulation stage store it's results in an intermediate struct?  That way when it comes time to create vertices and faces that intermediate list can be iterated through in one go.

src/appleseed.python/bindblendermeshconverter.cpp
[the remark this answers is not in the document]
dictoon: So using `bl_uv_index` (UV?) as a triangle index is intentional here? Is `bl_uv_index` the right name then?

dictoon:
> Would it make more sense maybe to have the triangulation stage store it's results in an intermediate struct? That way when it comes time to create vertices and faces that intermediate list can be iterated through in one go.

The triangulation for a while object you mean? It will lead to higher memory consumption and I'm not sure what it's buying us.

src/appleseed.python/bindblendermeshconverter.cpp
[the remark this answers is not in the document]
jdent02: Yes.  If the blender face is dynamically triangulated the tex coord assignment has to have a way of offsetting the face index when it goes to pull the correct appleseed triangle(s).

No, it's probably not the best name.

src/appleseed.python/bindblendermeshconverter.cpp
```diff
@@ -35,81 +35,76 @@
...
-    short mode, tile, unwrap;
-};
+    struct MeshTexFace
+    {
+        float uv[4][2];
+        void *tpage;
+        char flag, transp;
+        short mode, tile, unwrap;
+    };
+}  // namespace
```
dictoon: We don't usually put those kind of markers, I would suggest to remove it.

src/appleseed.python/bindblendermeshconverter.cpp
```diff
@@ -35,81 +35,76 @@
...
 
 
 // The following function takes a series of pointers to Blender mesh data and modifies the appleseed MeshObject entity
 //      blender_object = the appleseed MeshObject created earlier in the export process
-//      vertices_length = the number of vertices in the mesh
+//      vertices_size = the number of vertices in the mesh
```
dictoon: Not sure what those are referring to, function parameters have different names?

src/appleseed.python/bindblendermeshconverter.cpp
```diff
@@ -35,81 +35,76 @@
...
-    const size_t                vertices_length,
-    const size_t                vertex_pointer,
-    const size_t                triangle_length,
-    const size_t                triangle_pointer,
-    const size_t                uv_layer_pointer,
-    const bool                  export_normals,
-    const bool                  export_uvs)
+    MeshObject*                     blender_mesh,
+    const size_t                    bl_vert_size,
+    const uintptr_t                 bl_vert_ptr,
```
dictoon: Make sure to include `<cstdint>` for `uintptr_t`.

src/appleseed.python/bindblendermeshconverter.cpp
```diff
@@ -35,81 +35,76 @@
...
-    MeshObject*                 blender_mesh,
-    const size_t                vertices_length,
-    const size_t                vertex_pointer,
-    const size_t                triangle_length,
-    const size_t                triangle_pointer,
-    const size_t                uv_layer_pointer,
-    const bool                  export_normals,
-    const bool                  export_uvs)
+    MeshObject*                     blender_mesh,
+    const size_t                    bl_vert_size,
```
dictoon: May I suggest `bl_vert_count`?

src/appleseed.python/bindblendermeshconverter.cpp
```diff
@@ -118,78 +113,119 @@ void convert_bl_mesh(
...
     }
 
-    int uv_index = 0;
-
     // Push triangles
-    for (size_t tri_index = 0; tri_index < triangle_length; ++tri_index)
+    for (size_t tri_index = 0; tri_index < bl_faces_size; ++tri_index)
     {
-        MeshFace &face = triangles[tri_index];
+        const MeshFace &face = bl_faces[tri_index];
```
dictoon: Move ampersand to type's side.

src/appleseed.python/bindblendermeshconverter.cpp
```diff
@@ -118,78 +113,119 @@ void convert_bl_mesh(
...
-    if (export_normals == true)
+    if (export_normals)
     {
-        for (size_t tri_index = 0; tri_index < triangle_length; ++tri_index)
+        size_t bl_face_index = 0;
+        for (size_t tri_index = 0; tri_index < bl_faces_size; ++tri_index, ++bl_face_index)
         {
-            MeshFace &face = triangles[tri_index];
-            Triangle& tri = blender_mesh->get_triangle(tri_index);
+            const MeshFace& face = bl_faces[tri_index];
```
dictoon: Is it intentional that you're using `tri_index` to index into _faces_, and `bl_face_index` to index into triangles?

src/appleseed.python/bindblendermeshconverter.cpp
jdent02: Done

src/appleseed.python/bindblendermeshconverter.cpp
jdent02: Done

jdent02:
PR updated

src/appleseed.python/bindblendermeshconverter.cpp
```diff
@@ -32,84 +32,83 @@
...
-    if (export_normals == true)
+    // Push normals.
+    if (export_normals)
     {
-        for (size_t vertex_index = 0; vertex_index < vertices_length; ++vertex_index)
+        blender_mesh->reserve_vertex_normals(bl_vert_count);
+        for (size_t vertex_index = 0; vertex_index < bl_vert_count; ++vertex_index)
         {
-            MeshVert &vert = vertices[vertex_index];
+            const MeshVert &vert = bl_vertices[vertex_index];
```
dictoon: Move ampersand.

src/appleseed.python/bindblendermeshconverter.cpp
```diff
@@ -118,12 +117,10 @@ void convert_bl_mesh(
         }
     }
 
-    int uv_index = 0;
-
     // Push triangles
```
dictoon: End with period (fix everywhere).

src/appleseed.python/bindblendermeshconverter.cpp
```diff
@@ -118,12 +117,10 @@ void convert_bl_mesh(
...
-    for (size_t tri_index = 0; tri_index < triangle_length; ++tri_index)
+    for (size_t tri_index = 0; tri_index < bl_faces_count; ++tri_index)
```
dictoon: For consistency's sake, rename `tri_index` to `face_index` (like you've done in other places).

src/appleseed.python/bindblendermeshconverter.cpp
```diff
@@ -198,12 +194,12 @@ void convert_bl_vertex_pose(
                 vert.co[1],
                 vert.co[2]));
     }
-    // Push normals
-    if (export_normals == true)
+    // Push normals.
```
dictoon: Insert blank line before full line comment.

PR #2304, opened 2018-11-17
clang-format config

est77:
- Added clang-format config, trying to match appleseed's style.
- Added a basic script to disable formatting of forward declaration blocks.
- Added a commit with some sample results (to be deleted before merging the PR)

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,118 @@
+---
+Language:        Cpp
+# BasedOnStyle:  LLVM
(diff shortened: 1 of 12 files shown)


oktomus:
What's the status of this ? I like the idea of having something automated but IIRC @FrancoisBeaune do not agree that ?

dictoon:
We'll probably switch to clang-format after the upcoming release.

PR #2309, opened 2018-11-20
Expose library info and third party version info to Python bindings

jdent02:


src/appleseed.python/bindappleseedinfo.cpp
```diff
@@ -0,0 +1,99 @@
...
+using namespace foundation;
+
+namespace
+{
+    // Assembles version information on all third party dependencies
+    bpy::dict get_deps_lib_versions()
+    {
+        bpy::dict output_dict;
+        const LibraryVersionArray& info_array = ThirdParties::get_versions();
+        for (size_t index = 0; index < info_array.size(); index++)
```
dictoon: Include `<cstddef>` for `size_t` (which technically should be `std::size_t`... We need to fix this in the entire codebase. Until then let's stick to unqualified `size_t`).

src/appleseed.python/bindappleseedinfo.cpp
```diff
@@ -0,0 +1,99 @@
...
+namespace bpy = boost::python;
+using namespace foundation;
+
+namespace
+{
+    // Assembles version information on all third party dependencies
+    bpy::dict get_deps_lib_versions()
+    {
+        bpy::dict output_dict;
+        const LibraryVersionArray& info_array = ThirdParties::get_versions();
```
dictoon: `ThirdParties::get_versions()` returns a value, not a reference, hence it's cleaner in my opinion to acknowledge that fact and make `info_array` a value too instead of a reference (even if in practices it doesn't change anything since the value referenced by `info_array` would live long enough for the reference not to dangle).

src/appleseed.python/bindappleseedinfo.cpp
```diff
@@ -0,0 +1,99 @@
...
+#include "foundation/core/appleseed.h"
+#include "foundation/core/thirdparties.h"
+#include "foundation/platform/python.h"
+
+namespace bpy = boost::python;
+using namespace foundation;
+
+namespace
+{
+    // Assembles version information on all third party dependencies
```
dictoon: End full line comments with a period.

src/appleseed.python/bindappleseedinfo.cpp
```diff
@@ -0,0 +1,99 @@
...
+    bpy::dict get_deps_lib_versions()
```
dictoon: Let's use the same vocabulary: `get_third_parties_versions()`

src/appleseed.python/module.cpp
```diff
@@ -35,6 +35,7 @@
 
 // Forward declarations.
 void bind_aov();
+void bind_appleseed_info();
```
dictoon: Maybe that should be called `bind_appleseed()` since it's exposing the `Appleseed` class (more or less) to Python...

PR #2317, opened 2018-11-23
Fix call to foundation::SearchPaths in sandbox/samples/cpp/basic.cpp

Mango-3:


dictoon:
Thanks!

Are other C++ samples OK?

PR #2318, opened 2018-11-24
Fix anisotropy when shading samples is greater than 1

### LINKED ISSUE #2315: Unexpected results with increasing number of shading samples
When rendering a network pluggin a radial (greyscale, linear) ramp, tiled 4 times along U, V, into the anisotropy angle of the metal shader, we have the expect results with 1 shading sample.
However if we increase shading samples, to 4, and to 9 samples, instead of having the same visual appearance with smoother results, we have these extra radial bands in the resulting images.
I also noticed some visual change with SSS as shading samples change but attributed it perhaps incorrectly, to some accidental change of parameters on my part.

shading samples 1

shading samples 4

shading samples 9

The shading graph layout follows for clarity
aniso_simple_node

I couldn't find the shading samples in the appleseedz file, so the maya archive is also included. No textures required.

[aniso_simplenode.appleseedz.zip](https://github.com/appleseedhq/appleseed/files/2609892/aniso_simplenode.appleseedz.zip)

[anisotest.mb.zip](https://github.com/appleseedhq/appleseed/files/2609888/anisotest.mb.zip)

Something similar which might be related was reported earlier in https://github.com/appleseedhq/appleseed/issues/1539

est77:
Fixes #2315


src/appleseed/renderer/modeling/surfaceshader/physicalsurfaceshader.cpp
```diff
@@ -113,18 +113,32 @@ namespace
...
+                shading_context,
+                shading_point,
+                radiance,
+                components);
+
             if (m_lighting_samples > 1)
+            {
+                for (size_t i = 1, e = m_lighting_samples; i < e; ++i)
+                {
+                    shading_point.set_shading_basis(basis);
```
dictoon: Maybe add a short comment explaining that the shading basis stored in the shading point gets modified under certain circumstances?

PR #2321, opened 2018-11-26
Fix insert search path in sandbox/samples/cpp/heightfield.cpp

Mango-3:
The other C++ samples are not affected by the updated insert-searchpath call.


dictoon:
Thanks!

Issue #2322, opened 2018-11-27
OSL shaders such as as_disney_material shouldn't limit bounces by default


dictoon:
The global bounce limits should prevail by default.

luisbarrancos:
@dictoon we always provide a override to the maximum number of bounces in the material shaders since the user might want to for instance, limit the number of bounces allowed for a refractive sphere in a particular setup, while having other spheres with lower/higher bounces. That was a conscious choice. Do you want to remove this feature from the materials? If so, they need to be hidden and not removed in order not to break backwards compat. If the problem is having a default number of bounces, perhaps we can instead increase the default value to a more convenient number. Suggestions?

dictoon:
I'm totally fine with providing overrides for the maximum number of bounces in materials. The point is that by default, that override should be _disabled_. Setting it to a large enough value (100) seems fine.

dictoon:
This applies to all materials and to all overrides for global settings.

luisbarrancos:
Addressed in https://github.com/appleseedhq/appleseed/pull/2348

PR #2348, opened 2018-12-06
Increase default number of bounces

luisbarrancos:
To address issue #2322 
64 bounces ought to be enough for anybody.... though we settled for the maximum allowed by default, which is 100. Metadata for default sliders soft maximum was increased from 8 to 16.

### DIFF
--- a/src/appleseed.shaders/src/appleseed/as_subsurface.osl
+++ b/src/appleseed.shaders/src/appleseed/as_subsurface.osl
@@ -49,7 +49,7 @@ shader as_subsurface
         string as_maya_attribute_name = "sssProfile",
         string as_maya_attribute_short_name = "ssp",
         string widget = "mapper",
-        string options = "Better Dipole:0|Directional Dipole:1|Gaussian:2|Normalized Diffusion:3|Standard Dipole:4|Random Walk:5",
+        string options = "Better Dipole:0|Directional Dipole:1|Gaussian:2|Normalized Diffusion:3|Standard Dipole:4|Diffuse Random Walk:5",
         string label = "Subsurface Profile",
         string page = "Subsurface",
         int as_maya_attribute_connectable = 0,
@@ -92,15 +92,16 @@ shader as_subsurface
         string label = "Depth Scale",
         string page = "Subsurface"
     ]],
-    int in_sss_maximum_ray_depth = 2
+    int in_sss_maximum_ray_depth = 8
     [[
         string as_maya_attribute_name = "sssMaximumRayDepth",
         string as_maya_attribute_short_name = "ssd",
         int min = 1,
-        int max = 16,
-        int softmax = 8,
+        int max = 100,
+        int softmax = 16,
         string label = "Subsurface Ray Depth",
         string page = "Subsurface.Advanced",
+        string help = "Maximum number of bounces allowed",
         int as_maya_attribute_connectable = 0,
         int as_maya_attribute_keyable = 0,
         int as_maya_attribute_hidden = 1,
@@ -260,13 +261,15 @@ shader as_subsurface
     [[
         string as_maya_attribute_name = "outColor",
         string as_maya_attribute_short_name = "oc",
-        string widget = "null"
+        string widget = "null",
+        string label = "Output Color"
     ]],
     output closure color out_outMatteOpacity = 0
     [[
         string as_maya_attribute_name = "outMatteOpacity",
         string as_maya_attribute_short_name = "om",
         string widget = "null",
+        string label = "Output Matte",
         int as_maya_attribute_hidden = 1,
         int as_blender_input_socket = 0,
         int gafferNoduleLayoutVisible = 0
@@ -285,6 +288,18 @@ shader as_subsurface
 
         out_outColor += out_outMatteOpacity;
     }
+
+    int ray_depth;
+    int status = getattribute("path:ray_depth", ray_depth);
+
+    if (!status || ray_depth > in_sss_maximum_ray_depth)
+    {
+        return;
+    }
+
+    normal Nn = normalize(in_bump_normal);
+
+    color albedo = clamp(in_sss_amount * in_color, 0.01, 0.99);
 
     string sss_profile;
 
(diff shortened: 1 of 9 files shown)


src/appleseed.shaders/src/appleseed/as_subsurface.osl
```diff
@@ -92,13 +92,13 @@ shader as_subsurface
         string label = "Depth Scale",
         string page = "Subsurface"
     ]],
-    int in_sss_maximum_ray_depth = 2
+    int in_sss_maximum_ray_depth = 100
     [[
         string as_maya_attribute_name = "sssMaximumRayDepth",
         string as_maya_attribute_short_name = "ssd",
         int min = 1,
```
dictoon: min = 1 is intentional here I presume?

luisbarrancos: I'm not entirely sure if to allow a minimum of 0 since that disables SSS completely and this is an exclusively SSS shader (with a specular on top). What do you think? Allow 0 or always keep at least 1 bounce for the SSS to produce anything?

dictoon: It actually looks like this parameter isn't used at all?

luisbarrancos: You're right,  i added it in stdsurface but didn't in this PR. I have it in the diffuse+glass randomwalk. Hold this for a bit while i add these changes.

src/appleseed.shaders/src/appleseed/as_toon.osl
```diff
@@ -460,13 +460,13 @@ shader as_toon
         string page = "Specular.Toon",
         string help = "Attenuate facing incidence reflection, similar to Fresnel, retaining reflection intensity at the grazing angles."
     ]],
-    int in_specular_ray_depth = 2
+    int in_specular_ray_depth = 100
     [[
         string as_maya_attribute_name = "specularRayDepth",
         string as_maya_attribute_short_name = "srd",
         int min = 0,
         int max = 16,
```
dictoon: I think `max` should be 100 here.

src/appleseed.shaders/src/appleseed/as_subsurface.osl
```diff
@@ -92,15 +92,33 @@ shader as_subsurface
...
+        float min = -1.0,
+        float max = 1.0,
+        int as_maya_attribute_connectable = 0,
+        int as_maya_attribute_keyable = 0,
+        int as_maya_attribute_hidden = 1,
+        int as_blender_input_socket = 0,
+        int gafferNoduleLayoutVisible = 0,
+        string label = "Volume Anisotropy",
+        string page = "Subsurface",
+        string help = "Scattering anisotropy, with negative values producing back scattering, 0 isotropic scattering, and positive values producing forward scattering",
```
dictoon: "0 **producing** isotropic scattering" if you want to be consistent.

src/appleseed.shaders/src/appleseed/as_subsurface.osl
```diff
@@ -92,15 +92,33 @@ shader as_subsurface
...
+        int divider = 1
+    ]],
+    int in_sss_maximum_ray_depth = 100
```
dictoon: I'm not sure we want to change this default value, especially since there is no other global control for this in appleseed yet (unlike max diffuse/glossy/specular bounces). I would suggest to leave it at 2.

src/appleseed.shaders/src/appleseed/as_subsurface.osl
```diff
@@ -315,74 +343,89 @@ shader as_subsurface
...
         {
-            out_outColor += as_subsurface(
-                sss_profile,
-                Nn,
-                albedo,
-                in_sss_mfp_scale * in_sss_mfp,
-                in_ior,
-                "fresnel_weight", in_fresnel_weight);
+            out_outColor += (1.0 - in_specular_weight) *
+                as_subsurface(
```
dictoon: Maybe add a comment explaining that you're computing random-walk SSS with a _diffuse_ interface here.

src/appleseed.shaders/src/appleseed/as_subsurface.osl
```diff
@@ -315,74 +343,89 @@ shader as_subsurface
...
+                    Nn,
+                    albedo,
+                    in_sss_mfp_scale * in_sss_mfp,
+                    in_ior,
+                    in_specular_roughness,
+                    "fresnel_weight", 1.0,
+                    "volume_anisotropy", in_volume_anisotropy);
         }
-        else
+        if (in_specular_weight < 1.0)
```
dictoon: I'm not sure to understand the logic. You're (nearly) always computing and adding together glass and diffuse random-walk components? In other words, do you confirm that the absence of `else` is on purpose here?

luisbarrancos: Still WIP but the idea is doing a linear blend between the *diffuse* and *glass* randomwalk profiles via *specular weight*. The GGX term is not added yet, the *glass* random walk is not accounting for reflectance.It would give some measure of control for *randomwalk* while sharing some parameters with the specular term.
Alternatively we could add *diffuse randomwalk* and *glass randomwalk*, i'm open to suggestions.

luisbarrancos: Regarding breaking changes, i know in Maya and Gaffer adding a parameter poses no problem, but in Max adding a parameter needs to be accounted for, correct @usakhelo ?
In blender i'm not sure, @jdent02 ?

src/appleseed.shaders/src/appleseed/as_subsurface.osl
luisbarrancos: Though it's an interesting idea to blend between the diffuse and glassy random walk with the specular term, it will restrict the ability to layer the GGX lobe over the diffuse random walk. Adding a parameter *transmission* to linearly blend from the *diffuse* to *glassy* random walk would be a possibility, but i'm not sure this is desirable at this stage, so we have instead the explicit *diffuse* random walk as previously, and a new *glassy* one. I think this removes the need for the comment here.

src/appleseed.shaders/src/appleseed/as_subsurface.osl
luisbarrancos: This affects the specular term as well and has a considerable impact on the *glassy* random walk - there, a default of 2 is just too low. 8?

luisbarrancos: Done

src/appleseed.shaders/src/appleseed/as_subsurface.osl
Mango-3: Yes, that is unfortunately correct. I tried it out with your new version of asSubsurface  and it breaks scenes (Error loading ParamBlock2) which have the current version of asSubsurface as a material assigned to objects.

jdent02:
Looks okay to me. Soft min and soft max are ui related parameters.  Any bounce settings that have been changed by a user will be retained.

src/appleseed.shaders/src/appleseed/as_subsurface.osl
luisbarrancos: I'll leave the *glassy* random walk profile and phase function control to a v2 version but retain the bounces changes which should pose no problem.

luisbarrancos: Removed the glassy random walk profile.

PR #2349, opened 2018-12-06
Disable blender input socket for metal EC

luisbarrancos:
This addresses https://github.com/appleseedhq/blenderseed/issues/355
EC was inadvertanly exposed in blender, but existing scenes will break
if there are any input connections to this, so the metadata token
 as_blender_input_socket must be commented out for now.


luisbarrancos:
Closing this, fix made on blenderseed.

PR #2351, opened 2018-12-06
Extracted ColorMap class from ColorMapPostProcessingStage and used it for diagnostic AOVs

### LINKED ISSUE #2249: Extract ColorMap class from ColorMapPostProcessingStage and use it for diagnostic AOVs

mayankDhiman:
@dictoon. This should fix #2249.


src/appleseed/renderer/modeling/aov/invalidsamplesaov.cpp
```diff
@@ -191,6 +193,196 @@ namespace
     };
 }
 
+//
+// Color mapping
+//
+
+class ColorMap
```
oktomus: lass ?

PR #2352, opened 2018-12-10
Add support of Cryptomatte AOV

usakhelo:
Hi guys!

This is an attempt to add Cryptomatte AOV. It's done in the same way as denoiser, so in order to save cryptomatte AOV you will need to enable it on beauty output and save main output after rendering. It has two modes - object names and material names. It works fine on my simple test scene where I used depth of field and opacity to test matte selection. I tested it on Nuke and Fusion.

Thanks for review!


CMakeLists.txt
```diff
@@ -283,6 +283,7 @@ endif ()
 include_directories (
     src/thirdparty/bcd
     src/thirdparty/lz4
+    src/thirdparty/murmurhash3
```
est77: We already have murmurhash in foundation/utility, wrapped in a class. It looks like the same version.
Use that one, modifying it if needed, instead of adding another copy of the code.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,736 @@
...
+
+void CryptomatteAOV::clear_image()
+{
+    const int num_channels = (impl->m_num_layers * 2) + 3;
+    for (size_t ry = 0, e = impl->m_image->properties().m_canvas_height; ry < e; ++ry)
+    {
+        for (size_t rx = 0, e = impl->m_image->properties().m_canvas_width; rx < e; ++rx)
+        {
+                vector<float> pixel_values;
+                for (size_t i = 0; i < num_channels; ++i)
```
est77: Move this out of the loop. Use the vector constructor that creates N elements with a value.

CMakeLists.txt
usakhelo: Their specification requires "32 bit variant of MurmurHash3 (Also known as MurmurHash3_x86_32)"
The foundation version is a x64_128 bit version. I tried to use that hash truncating it to 32 bits but it didn't work. Is there any way I can get the same 32 bit value from 128bits?

usakhelo: Maybe I should add x86_32 version to an existing `MurmurHash` class?

est77: I see. I think that we can keep it like it is, as an implementation detail of cryptomatte.

usakhelo: Maybe I should remove all other implementations from external code and leave only x86_32bit?

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -952,6 +970,25 @@ void Frame::extract_parameters()
             impl->m_denoising_mode = DenoisingMode::Off;
         }
     }
+
+    // Retrieve cryptomatte parameters.
+    {
+        const string cryptomatte_mode = m_params.get_optional<string>("cryptomatte", "off");
```
est77: I think we can remove this now

src/appleseed/renderer/modeling/frame/frame.h
```diff
@@ -170,6 +170,12 @@ class APPLESEED_DLLSYMBOL Frame
         const size_t                thread_count,
         foundation::IAbortSwitch*   abort_switch) const;
 
+    enum class CryptomatteMode
```
est77: Remove this enum too.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,752 @@
...
+{
+    return
+        Dictionary()
+        .insert("name", get_model())
+        .insert("label", "Cryptomatte");
+}
+
+DictionaryArray CryptomatteAOVFactory::get_input_metadata() const
+{
+    DictionaryArray metadata;
```
est77: We can add here the option to choose between object names or material names.
We don't have a way to edit them in studio but we'll have it.

Another option would be to have 2 Cryptomatte AOVs. One for object IDs and one for material IDs.
This option might be easier to support in plugins. I'll check that.

usakhelo: I thought about two AOVs too, it will require another line `register_factory(auto_release_ptr<FactoryType>(new CryptomatteAOVFactory()));`
But mode parameter can passed to Factory class maybe?

usakhelo:
Hi @est77 , I've fixed comments and changed Factory so that it can create two different AOVs. Tested in 3ds max and it works. This branch requires rebase I guess.

src/appleseed/renderer/modeling/aov/aovfactoryregistrar.cpp
```diff
@@ -32,6 +32,7 @@
 // appleseed.renderer headers.
 #include "renderer/modeling/aov/aovtraits.h"
 #include "renderer/modeling/aov/albedoaov.h"
+#include "renderer/modeling/aov/cryptomatte.h"
```
dictoon: The files should be called `cryptomatteaov.{h,cpp}`.

src/appleseed/renderer/modeling/aov/cryptomatte.h
```diff
@@ -0,0 +1,133 @@
...
+
+// appleseed.renderer headers.
+#include "renderer/modeling/aov/aov.h"
+#include "renderer/modeling/aov/iaovfactory.h"
+
+// appleseed.foundation headers.
+#include "foundation/utility/autoreleaseptr.h"
+
+// BCD headers.
+#include "bcd/DeepImage.h"
```
dictoon: This doesn't look needed?

src/appleseed/renderer/modeling/aov/cryptomatte.h
```diff
@@ -0,0 +1,133 @@
...
+    foundation::auto_release_ptr<AOVAccumulator> create_accumulator() const override;
+
+    bool write_images(
+        const char*                         file_path,
+        const foundation::ImageAttributes&  image_attributes) const override;
+
+  private:
+    friend class CryptomatteAOVFactory;
+
+    struct                      Impl;
```
dictoon: We only require that data members are aligned with other data members. Since there are no others here, you don't need all these spaces.

src/appleseed/renderer/modeling/aov/cryptomatte.h
```diff
@@ -0,0 +1,133 @@
...
+    const char** get_channel_names() const override;
+
+    bool has_color_data() const override;
+
+    void create_image(
+        const size_t    canvas_width,
+        const size_t    canvas_height,
+        const size_t    tile_width,
+        const size_t    tile_height,
+        ImageStack&     aov_images) override;
```
dictoon: `ImageStack` needs to be forward-declared.

src/appleseed/renderer/modeling/aov/cryptomatte.h
```diff
@@ -0,0 +1,133 @@
...
+  : public AOV
+{
+  public:
+    enum class CryptomatteType
+    {
+        ObjectNames,
+        MaterialNames,
+    };
+
+    CryptomatteAOV(const ParamArray& params);
```
dictoon: Mark constructor as `explicit`.

src/appleseed/renderer/modeling/aov/cryptomatte.h
```diff
@@ -0,0 +1,133 @@
...
+
+// BCD headers.
+#include "bcd/DeepImage.h"
+
+// Standard headers.
+#include <cstddef>
+
+// Forward declarations.
+namespace foundation    { class ImageAttributes; }
+namespace renderer      { class AOV; }
```
dictoon: Not required since you include `aov.h`.

src/appleseed/renderer/modeling/aov/cryptomatte.h
```diff
@@ -0,0 +1,133 @@
...
+    Impl*                       impl;
+};
+
+
+//
+// A factory for cryptomatte AOVs.
+//
+
+class APPLESEED_DLLSYMBOL CryptomatteAOVFactory
+    : public IAOVFactory
```
dictoon: Indent with two spaces.

src/appleseed/renderer/modeling/aov/cryptomatte.h
```diff
@@ -0,0 +1,133 @@
...
+  private:
+    friend class CryptomatteAOVFactory;
+
+    struct                      Impl;
+    Impl*                       impl;
+};
+
+
+//
+// A factory for cryptomatte AOVs.
```
dictoon: *Cryptomatte*

src/appleseed/renderer/modeling/aov/cryptomatte.h
```diff
@@ -0,0 +1,133 @@
...
+      CryptomatteAOVFactory(const CryptomatteAOV::CryptomatteType aov_type);
+
+      // Delete this instance.
+      void release() override;
+
+      // Return a string identifying this AOV model.
+      const char* get_model() const override;
+
+      // Return metadata for this AOV model.
+      foundation::Dictionary get_model_metadata() const override;
```
dictoon: Foward declare `foundation::Dictonary`.

src/appleseed/renderer/modeling/aov/cryptomatte.h
```diff
@@ -0,0 +1,133 @@
...
+
+      // Return metadata for the inputs of this AOV model.
+      foundation::DictionaryArray get_input_metadata() const override;
```
dictoon: Foward declare `foundation::DictonaryArray`.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+// Interface header.
+#include "cryptomatte.h"
+
+// appleseed.renderer headers.
+#include "renderer/kernel/aov/aovaccumulator.h"
+#include "renderer/kernel/rendering/pixelcontext.h"
+#include "renderer/kernel/shading/shadingcomponents.h"
+#include "renderer/kernel/shading/shadingpoint.h"
+#include "renderer/kernel/shading/shadingresult.h"
+#include "renderer/modeling/aov/aov.h"
```
dictoon: Not required since already included in the header.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+
+// Murmurhash3 headers.
+#include "MurmurHash3.h"
+
+// Boost headers.
+#include "boost/filesystem.hpp"
+
+// Standard headers.
+#include <cmath>
+#include <cstddef>
```
dictoon: Not required for the same reason.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+
+using namespace foundation;
+using namespace std;
+namespace bf = boost::filesystem;
+
+namespace
+{
+    class MultiChannelExrFileWriter
+        : public IImageFileWriter
```
dictoon: Indent with two spaces.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+    {
+      public:
+        explicit MultiChannelExrFileWriter(const char* filename, const ICanvas* image)
+            : m_canvas(image)
```
dictoon: Indent with two spaces.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+            if (!m_writer->supports("tiles"))
+                return;
+
+            if (!m_writer->open(m_filename, m_spec))
+            {
+                const string msg = m_writer->geterror();
+                throw ExceptionIOError(msg.c_str());
+            }
+
+            // m_writer->write_image(OIIO::TypeDesc::FLOAT, m_canvas);
```
dictoon: Remove dead code?

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+        {
+            for (const_each<ImageAttributes> i = image_attributes; i; ++i)
+            {
+                // Fetch the name and the value of the attribute.
+                const string attr_name = i->key();
+                const string attr_value = i->value<string>();
+
+                if (attr_name == "software")
+                    m_spec.attribute("Software", attr_value.c_str());
+
```
dictoon: Remove blank lines between `if` / `else if` blocks.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+            {
+                const string msg = OIIO::geterror();
+                throw ExceptionIOError(msg.c_str());
+            }
+
+            m_spec = OIIO::ImageSpec();
+            set_image_spec();
+        }
+
+        ~MultiChannelExrFileWriter()
```
dictoon: Mark all base class methods you override with `override`.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+            }
+
+            // m_writer->write_image(OIIO::TypeDesc::FLOAT, m_canvas);
+            write_tiles();
+            close_file();
+        }
+
+        void set_image_attributes(const ImageAttributes& image_attributes)
+        {
+            for (const_each<ImageAttributes> i = image_attributes; i; ++i)
```
dictoon: Prefer range-based `for` loops in new code.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+                }
+
+                else
+                    m_spec.attribute(attr_name, attr_value.c_str());
+            }
+        }
+
+        void set_image_channels(
+            const size_t            channel_count,
+            const vector<string>&   channel_names)
```
dictoon: Include `<vector>`.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+                    // Compute the tile's ystride offset in bytes.
+                    const size_t ystride = xstride * min(static_cast<size_t>(m_spec.width + m_spec.x - tile_offset_x),
+                        static_cast<size_t>(m_spec.tile_width));
+
+                    // Retrieve the (tile_x, tile_y) tile.
+                    const Tile& tile = m_canvas->tile(tile_x, tile_y);
+
+                    // Write the tile into the file.
+                    if (!m_writer->write_tile(
+                        static_cast<int>(tile_offset_x),
```
dictoon: Indent method arguments once more.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+    {
+      public:
+        explicit MultiChannelExrFileWriter(const char* filename, const ICanvas* image)
+            : m_canvas(image)
+            , m_filename(filename)
+        {
+            m_writer = OIIO::ImageOutput::create(m_filename);
+            if (m_writer == nullptr)
+            {
+                const string msg = OIIO::geterror();
```
dictoon: Include `<string>`.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+        }
+
+    private:
+        const ICanvas*      m_canvas;
+        OIIO::ImageSpec     m_spec;
+        OIIO::ImageOutput*  m_writer;
+        const char*         m_filename;
+    };
+
+    float hash_to_float(uint32_t hash) {
```
dictoon: Move brace to its own line.
Add reference to where you took this code from.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+        // if all exponent bits are 0 (subnormals, +zero, -zero) set exponent to 1
+        // if all exponent bits are 1 (NaNs, +inf, -inf) set exponent to 254
+        uint32_t exponent = hash >> 23 & 255; // extract exponent (8 bits)
+        if (exponent == 0 || exponent == 255)
+            hash ^= 1 << 23; // toggle bit
+        float f;
+        memcpy(&f, &hash, 4);
```
dictoon: Include `<cstring>`.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+    }
+}
+
+namespace renderer
+{
+
+namespace
+{
+
+    typedef map<uint32_t, string> NameMap;
```
dictoon: Include `<map>`.
Use `uint32` from `foundation` and include `foundation/platform/types.h`.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+    typedef map<uint32_t, float> WeightMap;
+
```
dictoon: Two blank lines before block comment.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+      : public AOVAccumulator
+    {
+      public:
+        CryptomatteAOVAccumulator(
+            Image&                              aov_image,
+            vector<WeightMap>&                  pixel_samples,
+            map<pair<int, int>, NameMap>&       tile_name_map,
+            int                                 num_layers,
+            CryptomatteAOV::CryptomatteType     layer_type)
+            : m_aov_image(aov_image)
```
dictoon: Indent only with two spaces.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+    // Cryptomatte AOV accumulator.
+    //
+
+    class CryptomatteAOVAccumulator
+      : public AOVAccumulator
+    {
+      public:
+        CryptomatteAOVAccumulator(
+            Image&                              aov_image,
+            vector<WeightMap>&                  pixel_samples,
```
dictoon: A vector of maps? Is that intended?

If it is, how many maps will we have? One per pixel in the tile?

How many items will each map typically hold? Could a lighter (probably home-grown) data structure be used instead of `std::map<>`?

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+            const size_t                tile_x,
+            const size_t                tile_y) override
+        {
+            for (int ry = m_tile_origin_y; ry <= m_tile_end_y; ++ry)
+            {
+                for (int rx = m_tile_origin_x; rx <= m_tile_end_x; ++rx)
+                {
+                    const WeightMap& weight_map = m_pixel_samples[ry * m_frame_width + rx];
+
+                    if (weight_map.size() > 0)
```
dictoon: `!weight_map.empty()`

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+                    {
+                        float total_weight = 0.0f;
+                        for (const auto& item : weight_map)
+                            total_weight += item.second;
+
+                        if (total_weight == 0.0f)
+                            total_weight = 1.0f;
+
+                        vector<pair<float, uint32_t>> ranked_vector;
```
dictoon: Reserve memory in advance since you know exactly how many items the vector will hold.

Even better would be to have a single vector allocated outside of the loop, and cleared (using `foundation::clear_keep_memory()`) every time. You wouldn't need to reserve memory anymore and you would save a lot of memory allocations and deallocations.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+                        for (const auto& item : weight_map)
+                            ranked_vector.push_back(make_pair(item.second, item.first));
+                        sort(ranked_vector.begin(), ranked_vector.end(),
```
dictoon: Include `<algorithm>`.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+                            [](pair<float, uint32_t> const& a, pair<float, uint32_t> const& b) { return a.first > b.first; });
```
dictoon: Include `<utility>`.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+
+                        uint32_t m3hash = ranked_vector[0].second;
+
+                        vector<float> pixel_values;
+                        pixel_values.push_back(m3hash == 0 ? 0.0f : hash_to_float(m3hash));
+                        pixel_values.push_back(m3hash == 0 ? 0.0f : ((float)((m3hash << 8)) / (float)UINT32_MAX));
```
dictoon: Use `static_cast<>()` instead of C-style casts.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+                        vector<pair<float, uint32_t>> ranked_vector;
+                        for (const auto& item : weight_map)
+                            ranked_vector.push_back(make_pair(item.second, item.first));
+                        sort(ranked_vector.begin(), ranked_vector.end(),
+                            [](pair<float, uint32_t> const& a, pair<float, uint32_t> const& b) { return a.first > b.first; });
+
+                        uint32_t m3hash = ranked_vector[0].second;
+
+                        vector<float> pixel_values;
+                        pixel_values.push_back(m3hash == 0 ? 0.0f : hash_to_float(m3hash));
```
dictoon: Test `m3hash` just once.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+                        pixel_values.push_back(m3hash == 0 ? 0.0f : ((float)((m3hash << 8)) / (float)UINT32_MAX));
+                        pixel_values.push_back(m3hash == 0 ? 0.0f : ((float)((m3hash << 16)) / (float)UINT32_MAX));
+
+                        // Remove background contribution
```
dictoon: End full line comments with a period.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+                        if (ranked_vector.size() > 1 && m3hash == 0)
+                            ranked_vector.erase(ranked_vector.begin());
+
+                        for (size_t i = 0, e = ranked_vector.size(); i < e && i < m_num_layers; ++i)
```
dictoon: Set `e` to `min(ranked_vector.size(), m_num_layers)`?

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+                        {
+                            m3hash = ranked_vector[i].second;
+                            pixel_values.push_back(m3hash == 0 ? 0.0f : hash_to_float(m3hash));
+                            pixel_values.push_back(m3hash == 0 ? 0.0f : ranked_vector[i].first / total_weight);
+                        }
+
+                        const int num_channels = (m_num_layers * 2) + 3;
+                        for (size_t i = ranked_vector.size(); i < num_channels; ++i)
```
dictoon: You're mixing `int` and `size_t` here. Use `size_t`.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+        }
+    };
+
+    const char* CryptomatteObjectAOVModel = "cryptomatte_object_aov";
+    const char* CryptomatteMaterialAOVModel = "cryptomatte_material_aov";
+}
+
+
+//
+// Cryptomatte AOV class implementation.
```
dictoon: The class is called `CryptomatteAOV`.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+    unique_ptr<Image>                   m_image;
+    int                                 m_num_layers;
+    CryptomatteAOV::CryptomatteType     m_layer_type;
+};
+
+CryptomatteAOV::CryptomatteAOV(const ParamArray& params)
+  : AOV("cryptomatte", params)
+  , impl(new Impl())
+{
+    const string cryptomatte_type = params.get_optional<string>("cryptomatte_type", "object_names");
```
dictoon: You can pass a list of valid values to `get_optional()`, that way an error message will be emitted if the value is not valid.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+    const size_t    canvas_height,
+    const size_t    tile_width,
+    const size_t    tile_height,
+    ImageStack&     aov_images)
+{
+    const int w = static_cast<int>(canvas_width);
+    const int h = static_cast<int>(canvas_height);
+
+    // Create underlying images.
+    const int num_channels = (impl->m_num_layers * 2) + 3;
```
dictoon: `size_t`

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+}
+
+void CryptomatteAOV::create_image(
+    const size_t    canvas_width,
+    const size_t    canvas_height,
+    const size_t    tile_width,
+    const size_t    tile_height,
+    ImageStack&     aov_images)
+{
+    const int w = static_cast<int>(canvas_width);
```
dictoon: Why the casts? Those variables don't seem to be useful, just use `canvas_width` and `canvas_height`.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+            tile_height,
+            num_channels,
+            PixelFormatFloat));
+    impl->m_pixel_samples.resize(w * h);
+    clear_image();
+}
+
+void CryptomatteAOV::clear_image()
+{
+    const int num_channels = (impl->m_num_layers * 2) + 3;
```
dictoon: `size_t`

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+    vector<float> pixel_values(num_channels, 0.0f);
+    for (size_t ry = 0, e = impl->m_image->properties().m_canvas_height; ry < e; ++ry)
```
dictoon: Call `impl->m_image->properties()` once.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+    {
+        for (size_t rx = 0, e = impl->m_image->properties().m_canvas_width; rx < e; ++rx)
```
dictoon: You're using `e` in both loops.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+      case CryptomatteType::ObjectNames:
+        layer_name = "AppleseedObjectName";
+        break;
+
+      case CryptomatteType::MaterialNames:
+        layer_name = "AppleseedMaterialName";
+        break;
+
+      default:
+        break;
```
dictoon: Use `assert_otherwise` from `foundation/utility/otherwise.h`.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+{
+    const bf::path boost_file_path(file_path);
+    const bf::path directory = boost_file_path.parent_path();
+    const string base_file_name = boost_file_path.stem().string();
+    const string extension = boost_file_path.extension().string();
+
+    const string exr_file_name = base_file_name + ".cryptomatte" + extension;
+    const string exr_file_path = (directory / exr_file_name).string();
+
+    string layer_name = "AppleseedObjectName";
```
dictoon: I don't think you need to initialize the variable here?

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+    string layer_prefix("cryptomatte/");
+    layer_prefix.append(string(type_name_hex).erase(7));
+    image_attributes_copy.insert(layer_prefix + "/conversion", "uint32_to_float32");
+    image_attributes_copy.insert(layer_prefix + "/hash", "MurmurHash3_32");
+    image_attributes_copy.insert(layer_prefix + "/name", layer_name);
+
+    NameMap name_map;
+    for (auto& pair : impl->m_tile_name_map)
+    {
+        for (auto& hash : pair.second)
```
dictoon: `const`?

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+
+    image_attributes_copy.insert(layer_prefix + "/manifest", manifest_string.c_str());
+
+    vector<string> channel_names;
+    channel_names.push_back("R");
+    channel_names.push_back("G");
+    channel_names.push_back("B");
+
+    int num_exr_layers = truncate<int>(fast_ceil(impl->m_num_layers / 2.0));
+    for(size_t i = 0; i < num_exr_layers; ++i)
```
dictoon: Space after `for`.
Mixing signed/unsigned here. Use `size_t`.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+    {
+        string layer_number = get_numbered_string("##", i);
```
dictoon: Make const.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,784 @@
...
+    return true;
+}
+
+
+//
+// CryptomatteAOVFactory class implementation.
+//
+
+CryptomatteAOVFactory::CryptomatteAOVFactory(const CryptomatteAOV::CryptomatteType aov_type)
+    : m_aov_type(aov_type)
```
dictoon: Indent with two spaces.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
usakhelo: Hi @dictoon ,
Yes it's a map per pixel, it's variable size and holds up to the amount of objects that share the pixel. Typically up to 3-4 in test scenes. I'll check if there is anything lighter in foundation.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
usakhelo: Thanks for this tip.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
usakhelo: Terrible. How it ever worked?!

usakhelo:
Hi @dictoon , I've fixed the comments and replaced per-pixel map with in-house simple object. Maybe it is super naive, I'm not sure.
Also I'm not sure if I should do rebase, because CI complains.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -402,7 +527,7 @@ namespace
             const AOVComponents&        aov_components,
             ShadingResult&              shading_result) override
         {
-            uint32_t m3hash = 0;
+            uint32 m3hash = 0;
```
dictoon: I can't comment further down in this method, but can you mark as `const` all variables that can be const?

dictoon: You can also precompute `1.0 / m_rcp_canvas_width` etc.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,908 @@
...
+void CryptomatteAOV::clear_image()
+{
+    const size_t num_channels = (impl->m_num_layers * 2) + 3;
+    vector<float> pixel_values(num_channels, 0.0f);
+    const auto image_props = impl->m_image->properties();
+    for (size_t ry = 0, ey = image_props.m_canvas_height; ry < ey; ++ry)
+    {
+        for (size_t rx = 0, ex = image_props.m_canvas_width; rx < ex; ++rx)
+        {
+                impl->m_image->set_pixel(rx, ry, pixel_values.data());
```
dictoon: Fix indentation.

dictoon: That's going to be very slow because for each pixel `ICanvas::set_pixel()` needs to find the associated tile.

Here's a faster implementation (untested, please check):

```cpp
for (size_t ty = 0, ey = image_props.m_tile_count_y; ty < ey; ++ty)
{
    for (size_t tx = 0, ex = image_props.m_tile_count_x; tx < ex; ++tx)
    {
        Tile& tile = impl->m_image->tile(tx, ty);
        memset(tile.get_storage(), 0, tile.get_size());
    }
}
```

That works because `0` is the bit pattern for `0.0f` (how convenient!).

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,908 @@
...
+            tile_height,
+            num_channels,
+            PixelFormatFloat));
+    impl->m_pixel_samples.resize(canvas_width * canvas_height, WeightMap(impl->m_num_layers));
+    clear_image();
+}
+
+void CryptomatteAOV::clear_image()
+{
+    const size_t num_channels = (impl->m_num_layers * 2) + 3;
```
dictoon: The expression `(impl->m_num_layers * 2) + 3` is used in multiple places, what about factoring it into a method (of `Impl` for instance) and using it instead?

It would also make sense to explain the expression (why the multiplication by 2? Why adding 3).

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,908 @@
...
+            );
+            
+            filter(
+                static_cast<float>(tile_pixel_x + sample_pos_x),
+                static_cast<float>(tile_pixel_y + sample_pos_y),
+                tile_crop_window,
+                m3hash);
+        }
+
+    private:
```
dictoon: Indent with 2 spaces.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,908 @@
...
+        for (const auto& hash : pair.second)
+        {
+            if (hash.first == 0)
+                continue;
+            name_map[hash.first] = hash.second;
+        }
+    }
+
+    string manifest_str;
+    for (auto& hash : name_map)
```
dictoon: Make const.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,908 @@
...
+    {
+        char hash_hex[9];
+        sprintf(hash_hex, "%08x", hash.first);
+        manifest_str.append(format("\"{0}\":\"{1}\",", hash.second, string(hash_hex)));
+    }
+    if (manifest_str.size() > 0)
```
dictoon: `!empty()`

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,908 @@
...
+            name_map[hash.first] = hash.second;
+        }
+    }
+
+    string manifest_str;
+    for (auto& hash : name_map)
+    {
+        char hash_hex[9];
+        sprintf(hash_hex, "%08x", hash.first);
+        manifest_str.append(format("\"{0}\":\"{1}\",", hash.second, string(hash_hex)));
```
dictoon: This is fairly inefficient, use `std::stringstream` instead.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,908 @@
...
+    }
+    if (manifest_str.size() > 0)
+        manifest_str.pop_back();
+
+    const string manifest_string = format("{{0}}", manifest_str);
```
dictoon: Farily inefficient too, I would suggest appending a `{` at the beginning and a `}` at the end.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,908 @@
...
+            {
+                for (int rx = footprint.min.x; rx <= footprint.max.x; ++rx)
+                {
+                    const float weight = m_renderer_filter->evaluate(rx - dx, ry - dy);
+                    WeightMap& weight_map = m_pixel_samples[(m_tile_origin_y + ry) * m_frame_width + (m_tile_origin_x + rx)];
+                    const float old_value = weight_map.get(m3hash);
+                    weight_map.insert(m3hash, old_value + weight); // possible race conditions writing to pixels on the tile edges
+                }
+            }
+            return;
```
dictoon: Remove.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,908 @@
...
+                return;
+
+            for (int ry = footprint.min.y; ry <= footprint.max.y; ++ry)
+            {
+                for (int rx = footprint.min.x; rx <= footprint.max.x; ++rx)
+                {
+                    const float weight = m_renderer_filter->evaluate(rx - dx, ry - dy);
+                    WeightMap& weight_map = m_pixel_samples[(m_tile_origin_y + ry) * m_frame_width + (m_tile_origin_x + rx)];
+                    const float old_value = weight_map.get(m3hash);
+                    weight_map.insert(m3hash, old_value + weight); // possible race conditions writing to pixels on the tile edges
```
dictoon: That's a big problem!

`FilteredTile` handles it by using `foundation::atomic_add()` to atomically add weights; for that to work here, `WeightMap` needs to expose a mutable refs to values.

You can look at how `FilteredTile::atomic_add()` is implemented.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,908 @@
...
+        double                          m_rcp_canvas_width;
+        double                          m_rcp_canvas_height;
+        AABB2u                          m_crop_window;
+        Image&                          m_aov_image;
+        size_t                          m_num_layers;
+        vector<WeightMap>&              m_pixel_samples;
+        map<pair<int, int>, NameMap>&   m_tile_name_map;
+        CryptomatteAOV::CryptomatteType m_layer_type;
+
+        // Copy of the FilteredTile::add().
```
dictoon: Remove "the" or add "method" at the end.

Maybe "inspired" instead of "copy" since it's not actually a copy?

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,908 @@
...
+
+                MurmurHash3_x86_32(reinterpret_cast<const unsigned char*>(obj_name.c_str()), static_cast<int>(obj_name.size()), 0, &m3hash);
+            }
+
+            const Vector2i pixel_pos = pixel_context.get_pixel_coords();
+            const Vector2d sample_pos = pixel_context.get_sample_position();
+            int tile_pixel_x = pixel_pos.x - m_tile_origin_x;
+            int tile_pixel_y = pixel_pos.y - m_tile_origin_y;
+
+            m_tile_name_map[make_pair(m_tile_origin_x, m_tile_origin_y)][m3hash] = obj_name;    //tile_origin indexing of the map should help avoiding the race condition
```
dictoon: Either it completely avoids the race condition, or it doesn't: _helping_ avoiding the race condition is not enough :)

In any case, it is completely unsafe to modify a `std::map<>` concurrently so it would first need to be populated with enough entries such that the map itself is never modified.

That said, it seems you don't even need a map at all, you just need an array indexed by tile index (i.e. tile_y * tile_count_y + tile_x).

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,908 @@
...
+        manifest_str.append(format("\"{0}\":\"{1}\",", hash.second, string(hash_hex)));
+    }
+    if (manifest_str.size() > 0)
+        manifest_str.pop_back();
+
+    const string manifest_string = format("{{0}}", manifest_str);
+
+    image_attributes_copy.insert(layer_prefix + "/manifest", manifest_string.c_str());
+
+    vector<string> channel_names;
```
dictoon: You can reserve a number of items here since you already know how many you'll push.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,908 @@
...
+    {
+        const size_t num_channels = (impl->m_num_layers * 2) + 3;
+        MultiChannelExrFileWriter writer(exr_file_path.c_str(), &image);
+        writer.set_image_channels(num_channels, channel_names);
+        writer.set_image_attributes(image_attributes_copy);
+        writer.write();
+    }
+    catch (const exception& e)
+    {
+        RENDERER_LOG_ERROR(
```
dictoon: Include `renderer/global/globallogger.h`

dictoon:
Will need to be rebased on top of master.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -243,10 +249,111 @@ namespace
...
+
+        WeightMap(size_t size)
+          : m_size(size)
+          , m_index(0)
+        {
+            assert(size > 0);
+            m_map = new Entry[size];
+        }
+
+        WeightMap(const WeightMap& other)
```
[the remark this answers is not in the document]
usakhelo: I had it, then I tried to compile without it and it compiled, so I decided to remove it. Will const m_size work with assignment operator?

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -326,46 +433,64 @@ namespace
...
-                        pixel_values.push_back(m3hash == 0 ? 0.0f : ((float)((m3hash << 16)) / (float)UINT32_MAX));
+                        clear_keep_memory(pixel_values);
 
-                        // Remove background contribution
+                        // Preview channels (deprecated in recent Cryptomatte specification).
+                        float r(0.0), g(0.0), b(0.0);
+                        if (m3hash != 0)
+                        {
+                            r = hash_to_float(m3hash);
+                            g = (static_cast<float>(m3hash << 8) / static_cast<float>(UINT32_MAX));
```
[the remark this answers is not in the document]
usakhelo: Why do I need to precompute? To multiply each time instead of dividing?

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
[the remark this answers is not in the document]
dictoon: Ah, no, it won't, sorry. Fine as is then.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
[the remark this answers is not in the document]
dictoon: Yes, because the division is a particularly expensive operation.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,908 @@
...
+                  default:
+                    break;
+                }
+
+                MurmurHash3_x86_32(reinterpret_cast<const unsigned char*>(obj_name.c_str()), static_cast<int>(obj_name.size()), 0, &m3hash);
+            }
+
+            const Vector2i pixel_pos = pixel_context.get_pixel_coords();
+            const Vector2d sample_pos = pixel_context.get_sample_position();
+            int tile_pixel_x = pixel_pos.x - m_tile_origin_x;
```
dictoon: Make const.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -0,0 +1,908 @@
...
+            int tile_pixel_y = pixel_pos.y - m_tile_origin_y;
+
+            m_tile_name_map[make_pair(m_tile_origin_x, m_tile_origin_y)][m3hash] = obj_name;    //tile_origin indexing of the map should help avoiding the race condition
+
+            double sample_pos_x = (sample_pos.x / m_rcp_canvas_width) - pixel_pos.x;
```
dictoon: Make const.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -243,10 +249,111 @@ namespace
...
+        }
+
+        Entry* begin() const
+        {
+            return &m_map[0];
+        }
+
+        Entry* end() const
+        {
+            return &m_map[m_index];
```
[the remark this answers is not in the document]
usakhelo: I think begin/end are required for for-range loops, right?

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
usakhelo: Great tip, thanks. I don't know why I used map of maps here.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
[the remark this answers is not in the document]
dictoon: They are, but are you using for-range loops on that data structure? Iterating over a map is not so common...

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -266,11 +266,11 @@ namespace
         };
 
         explicit WeightMap(const size_t size)
-          : m_size(size)
+          : m_size(static_cast<uint32>(size))
           , m_index(0)
         {
             assert(size > 0);
-            m_map = new Entry[size];
+            m_map = new Entry[static_cast<uint32>(size)];
```
dictoon: You can simply use `size` here.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -684,6 +691,8 @@ CryptomatteAOV::CryptomatteAOV(const ParamArray& params)
 
 CryptomatteAOV::~CryptomatteAOV()
 {
+    if (impl->m_tile_name_array != nullptr)
```
dictoon: The check is unnecessary, it's safe to call `delete` on `nullptr`.

src/appleseed/renderer/modeling/aov/cryptomatte.cpp
```diff
@@ -639,8 +661,8 @@ namespace
                 {
                     const float weight = m_renderer_filter->evaluate(rx - dx, ry - dy);
                     WeightMap& weight_map = m_pixel_samples[(m_tile_origin_y + ry) * m_frame_width + (m_tile_origin_x + rx)];
-                    const float old_value = weight_map.get(m3hash);
-                    weight_map.insert(m3hash, old_value + weight); // possible race conditions writing to pixels on the tile edges
+                    float* APPLESEED_RESTRICT ptr = weight_map.get_value_ptr(m3hash);
```
dictoon: `get_value_ptr()` is not thread-safe, is that OK?

usakhelo: I guess it's not OK.
It's probably that it creates new entries in the dictionary, is that what makes it unsafe, right?

dictoon: Yes.

usakhelo: Hi @dictoon, is it possible to make thread safe dictionary without guards/mutexes? I think it should be something quick and simple right? Shouldn't block the thread I think..
Maybe it's ok to have thread collisions on some of the samples at the edge of the tiles?

usakhelo: Forgot to mention - I checked arnold's cryptomatte and it looks like they don't do thread safe writes on their sample buffer. It's possible they have totally different implementation of the tiles but I was just curious.

usakhelo: Hi @dictoon , I thought maybe it should have a sample storage per tile and that will make it thread safe? Sorry it takes me long time to fix properly.

usakhelo: I think I misunderstood and tile doesn't write outside its boundaries, so maybe it was perfectly thread safe before.

dictoon: That's correct.

PR #2360, opened 2018-12-16
Do not allow clamping ray intensity after specular bounces

### LINKED ISSUE #1267: Don't apply Max Ray Intensity after specular bounces

KarthikRIyer:
Fixes #1267


dictoon:
Thanks!

PR #2365, opened 2018-12-20
Add color inversion OSL shader

Mango-3:
This shader is an OSL implementation of RenderMan's PxrInvert node:

https://rmanwiki.pixar.com/display/REN22/PxrInvert

It allows to invert individual or all color channels of an input color with selectable color model. Output is a RGB color.

### DIFF
--- a/src/appleseed.shaders/src/appleseed/as_inv_color.osl
+++ b/src/appleseed.shaders/src/appleseed/as_inv_color.osl
@@ -1,0 +1,159 @@
+
+//
+// This source file is part of appleseed.
+// Visit https://appleseedhq.net/ for additional information and resources.
+//
+// This software is released under the MIT license.
+//
+// Copyright (c) 2018 Herbert Crepaz, The appleseedhq Organization
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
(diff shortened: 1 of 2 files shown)


src/appleseed.shaders/src/appleseed/as_inv_color.osl
```diff
@@ -0,0 +1,159 @@
...
+    string as_category = "color",
+    
+    string as_max_class_id = "1105087319 977155459", 
+    string as_max_plugin_type = "texture",
+    
+    int as_maya_type_id = 0x00127a08,
+    string as_maya_classification = "drawdb/shader:rendernode/appleseed/utility"
+]]
+(
+    color in_color = color(0,0,0)
```
dictoon: Insert spaces after commas.

src/appleseed.shaders/src/appleseed/as_inv_color.osl
```diff
@@ -0,0 +1,159 @@
...
+        string as_maya_attribute_short_name = "oc",
+        string widget = "null",
+        string label = "Color"
+    ]]
+)
+{
+    color in_color_trans = transformc(in_color_model, in_color);
+    color out_color_trans = color(0);
+    
+    if (as_invert_channel_0 == 1 && as_invert_channel_1 == 0 && as_invert_channel_2 == 0)
```
dictoon: The code can be simplified very significantly:

    out_color_trans = in_color_trans;

    if (as_invert_channel_0 == 1)
         out_color_trans[0] = 1 - in_color_trans[0];

    if (as_invert_channel_1 == 1)
         out_color_trans[1] = 1 - in_color_trans[1];

    if (as_invert_channel_2 == 1)
         out_color_trans[2] = 1 - in_color_trans[2];

src/appleseed.shaders/src/appleseed/as_inv_color.osl
```diff
@@ -0,0 +1,159 @@
...
+    {
+        out_color_trans[0] = 1 - in_color_trans[0];
+        out_color_trans[1] = 1 - in_color_trans[1];
+        out_color_trans[2] = 1 - in_color_trans[2];
+    }
+    else
+    {
+        out_color_trans = in_color_trans;
+    }
+    out_outColor =  transformc(in_color_model, "rgb", out_color_trans);
```
dictoon: Remove extra space after `=`.

src/appleseed.shaders/src/appleseed/as_inv_color.osl
Mango-3: ok

src/appleseed.shaders/src/appleseed/as_inv_color.osl
Mango-3: I tried a simplified logic initially. But after testing I saw it didn't work when selecting 2 or more channels as inverted. So each case is explicitly stated, may look overly complicated but it works reliably. OSL is different from C++ and not all things work equally, e.g. there is no bool data type in OSL.

src/appleseed.shaders/src/appleseed/as_inv_color.osl
Mango-3: done

Mango-3:
Thanks for the speedy review!

src/appleseed.shaders/src/appleseed/as_inv_color.osl
dictoon: I'm not sure to understand. OSL and C++ are different but still, valid OSL code will behave as expected, and the code I wrote above looks valid to me, so it should do the expected thing.

> But after testing I saw it didn't work when selecting 2 or more channels as inverted.

What happened? Are you sure your code was correct? Did you try mine?

Mango-3: You are right. I must have had a bug initially when I tested the simplified version the first time. Your simplification works as expected.

PR #2366, opened 2018-12-21
Make AOV::write_image() virtual

usakhelo:
Hi @est77,
This is quick version of the change that you've suggested me to do. I just moved some writing functionality from frame.cpp into AOV class and made the function virtual.
I tested it on max scene and it worked.
If I understand correctly this change will not reduce cryptomatte's write_image code because of restriction of `GenericImageFileWriter` to 4 channels. Cryptomatte will still need its own FileWriter.

### DIFF
--- a/src/appleseed/renderer/modeling/aov/aov.cpp
+++ b/src/appleseed/renderer/modeling/aov/aov.cpp
@@ -37,6 +37,7 @@
 #include "foundation/image/color.h"
 #include "foundation/image/genericimagefilewriter.h"
 #include "foundation/image/image.h"
+#include "foundation/image/imageattributes.h"
 
 // Standard headers.
 #include <exception>
@@ -107,7 +108,7 @@ bool AOV::write_images(const char* file_path) const
 
         writer.append_image(&get_image());
 
-        if(has_color_data())
+        if (has_color_data())
             writer.set_image_output_format(PixelFormatHalf);
 
         writer.set_image_channels(get_channel_count(), get_channel_names());
--- a/src/appleseed/renderer/modeling/aov/aov.cpp
+++ b/src/appleseed/renderer/modeling/aov/aov.cpp
@@ -30,11 +30,16 @@
 #include "aov.h"
 
 // appleseed.renderer headers.
+#include "renderer/global/globallogger.h"
 #include "renderer/kernel/aov/imagestack.h"
 
 // appleseed.foundation headers.
 #include "foundation/image/color.h"
+#include "foundation/image/genericimagefilewriter.h"
 #include "foundation/image/image.h"
+
+// Standard headers.
+#include <exception>
 
 using namespace foundation;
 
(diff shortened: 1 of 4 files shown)


est77:
Yes, this is the idea. The Cryptomatte AOV will override write_image() and use its own writer.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -717,8 +705,8 @@ bool Frame::write_aov_images(const char* file_path) const
         const string aov_file_path = (directory / aov_file_name).string();
 
         // Write AOV image.
-        if (!write_image(aov_file_path.c_str(), aov->get_image(), aov))
-            success = false;
```
dictoon: Why the change of logic here?

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -763,8 +751,9 @@ bool Frame::write_main_and_aov_images() const
                 filepath = new_filepath;
             }
 
-            if (!write_image(filepath.string().c_str(), aov->get_image(), aov))
-                success = false;
+            // Write AOV image.
+            if (!aov->write_images(filepath.string().c_str()))
+                return false;
```
dictoon: And here?

src/appleseed/renderer/modeling/aov/aov.cpp
```diff
@@ -94,6 +99,37 @@ void AOV::create_image(
...
 
+bool AOV::write_images(const char* file_path) const
+{
+    try
+    {
+        GenericImageFileWriter writer(file_path);
+
+        writer.append_image(&get_image());
+
+        if(has_color_data())
```
dictoon: Insert space after `if`.

src/appleseed/renderer/modeling/aov/aov.cpp
```diff
@@ -94,6 +99,37 @@ void AOV::create_image(
...
+            writer.set_image_output_format(PixelFormatHalf);
+
+        writer.set_image_channels(get_channel_count(), get_channel_names());
+
+        ImageAttributes image_attributes = ImageAttributes::create_default_attributes();
```
dictoon: Include `foundation/image/imageattributes.h`.

src/appleseed/renderer/modeling/aov/aov.h
```diff
@@ -92,6 +92,9 @@ class APPLESEED_DLLSYMBOL AOV
     // Apply any post-processing needed to the AOV image.
     virtual void post_process_image(const Frame& frame);
 
+    // Write image to EXR file.
```
dictoon: *OpenEXR*

dictoon: You called the method `write_images()` (plural), is this intended?

src/appleseed/renderer/modeling/frame/frame.cpp
usakhelo: You are asking about `success` variable, right? Sorry I misunderstood. It's because of the for loop. I thought breaking the loop and returning false on first failing aov.

src/appleseed/renderer/modeling/aov/aov.h
usakhelo: I think standalone method in `frame.cpp` was called plural, and I made copy of that.

src/appleseed/renderer/modeling/frame/frame.cpp
dictoon: Yes, the `success` variable. I prefer that we don't change the logic, if possible.

src/appleseed/renderer/modeling/aov/aov.h
dictoon: Ok.

PR #2370, opened 2018-12-23
added geometry for orthographic camera

mayankDhiman:


dictoon:
Forgot to say that it's fine to mix all geometry in `_geometry/`, we don't care much.

PR #2371, opened 2018-12-23
added geometry for orthgraphic camera scene

mayankDhiman:


dictoon:
Thanks! This looks a lot better.

PR #2389, opened 2019-01-23
Initial OIIO 2 support

luisbarrancos:
ImageInput and ImageOutput static create() and open() methods now return
unique_ptr rather than raw pointers, and OpenVDB is now read via texture3d(). It also brings us the gradients needed for bump2roughness. Notes at https://github.com/OpenImageIO/oiio/releases/tag/Release-2.0.3

### DIFF
--- a/src/appleseed/foundation/image/genericimagefilewriter.cpp
+++ b/src/appleseed/foundation/image/genericimagefilewriter.cpp
@@ -62,7 +62,7 @@ struct GenericImageFileWriter::Impl
 #if OIIO_VERSION >= 20000
     std::unique_ptr<OIIO::ImageOutput>  m_writer;
 #else
-    OIIO::ImageOutput* m_writer;
+    OIIO::ImageOutput*                  m_writer;
 #endif
     std::vector<const ICanvas*>         m_canvas;
     std::vector<OIIO::ImageSpec>        m_spec;
--- a/src/appleseed/foundation/image/genericimagefilewriter.cpp
+++ b/src/appleseed/foundation/image/genericimagefilewriter.cpp
@@ -43,6 +43,9 @@
 // Boost headers.
 #include "boost/filesystem/path.hpp"
 
+// OIIO headers.
+#include "OpenImageIO/version.h"
+
 // Standard headers.
 #include <algorithm>
 #include <cstring>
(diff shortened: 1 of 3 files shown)


src/appleseed/foundation/image/genericimagefilewriter.h
```diff
@@ -92,7 +93,12 @@ class APPLESEED_DLLSYMBOL GenericImageFileWriter
     struct Impl;
     Impl* impl;
 
+#if OIIO_VERSION >= 20000
+    std::unique_ptr<OIIO::ImageOutput> m_writer;
```
dictoon: You need to include `<memory>` for the definition of `std::unique_ptr<>`.

dictoon: Also, can we move `m_writer` and `m_filename` to the private impl?

src/appleseed/foundation/image/genericprogressiveimagefilereader.cpp
```diff
@@ -177,7 +184,9 @@ void GenericProgressiveImageFileReader::close()
     // but OpenImageIO 1.5.20 (the version included in appleseed-deps at the time
     // of writing) is too old to have this method. Since on Windows we link to
     // OpenImageIO statically, this should be safe anyway.
+#if OIIO_VERSION < 20000
```
dictoon: Not strictly part of your PR, but... Is the comment above still relevant? If it is and we keep the comment, then it should be inside the `#if ... #endif` block.

luisbarrancos: Not relevant anymore, and OIIO::ImageInput::destroy() can be called.

src/appleseed/foundation/image/genericimagefilewriter.h
luisbarrancos: Done, but still testing.

src/appleseed/foundation/image/genericimagefilewriter.cpp
```diff
@@ -57,6 +60,13 @@ struct GenericImageFileWriter::Impl
 {
     std::vector<const ICanvas*>     m_canvas;
     std::vector<OIIO::ImageSpec>    m_spec;
+
+#ifdef OIIO_VERSION >= 20000
+    std::unique_ptr<OIIO::ImageOutput>  m_writer;
```
dictoon: Align members.

dictoon: I would also suggest ordering them similarly to how they are initialized, which seems to be: filename, writer, canvas, spec.

src/appleseed/foundation/image/genericimagefilewriter.cpp
```diff
@@ -320,6 +331,7 @@ void GenericImageFileWriter::write_tiles(const size_t image_index)
     // Compute the tiles' xstride offset in bytes.
     size_t xstride = props.m_pixel_size;
 
+
```
dictoon: Remove redundant blank line.

src/appleseed/foundation/image/genericimagefilewriter.h
```diff
@@ -38,8 +38,12 @@
 // OpenImageIO headers.
 #include "foundation/platform/_beginoiioheaders.h"
 #include "OpenImageIO/imageio.h"
+#include "OpenImageIO/version.h"
 #include "foundation/platform/_endoiioheaders.h"
 
+// Standard headers.
```
dictoon: Remove dead code.

PR #2395, opened 2019-01-27
Tweaks to ASC CDL help, parameters

luisbarrancos:
- The parameters for the RGB and white primaries is not *colorspace* but RGB primaries and white, so name it accordingly.
 - The parameters for primaries and clamping should not be mapeable and visible in the node graph editor, so hide them accordingly.
 - Parameter names have *in_* or *out_* prefix, but the clamp parameter had an *as_* prefix, which is incorrect. The convention is this prefix being used for appleseed generic shader names (not DCC specific).
 - Add proper references for the implementation (URL, ISBN), keep it consistent with everything else.

### DIFF
--- a/src/appleseed.shaders/src/appleseed/as_asc_cdl.osl
+++ b/src/appleseed.shaders/src/appleseed/as_asc_cdl.osl
@@ -26,18 +26,23 @@
 // THE SOFTWARE.
 //
 
-// American Society of Cinematographers Color Decision List implementation
+//
+// References:
+//
+//  American Society of Cinematographers Color Decision List
+//  https://en.wikipedia.org/wiki/ASC_CDL
+//
 
 #include "appleseed/color/as_color_helpers.h"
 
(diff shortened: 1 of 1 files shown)


src/appleseed.shaders/src/appleseed/as_asc_cdl.osl
```diff
@@ -111,36 +121,40 @@ shader as_asc_cdl
...
-    if (as_clamp_output == 1)
-    {   
-        out_outColor = clamp(unclamped, 0, 1);
-    }
-    else
-    {   
-        out_outColor = unclamped; 
-    }
+
+    out_outColor = (in_clamp_output)
```
dictoon: The parentheses here are redundant.

luisbarrancos: @dictoon it'll be in the next PR

PR #2275, opened 2018-10-29
Rename CMake options USE_EXTERNAL_* to USE_SYSTEM_*

Blackhart:

PR #2406, opened 2019-02-02
Small improvements in CMake scripts

dictoon:
Based on @Blackhart's work done in PRs https://github.com/appleseedhq/appleseed/pull/2274 and https://github.com/appleseedhq/appleseed/pull/2275, which this PR supersedes.


cmake/config/linux-gcc-clang.txt
```diff
@@ -91,16 +91,6 @@ set (c_compiler_flags_common
...
-        CMAKE_BUILD_TYPE STREQUAL "Release")
-        set (c_compiler_flags_common
-            ${c_compiler_flags_common}
-            -Werror                                     # treat Warnings As Errors
-        )
-    endif ()
-endif ()
 set (cxx_compiler_flags_common
     ${cxx_compiler_flags_common}
     -std=c++11                                          # enable C++11
```
est77: We can replace this by:

set (CMAKE_CXX_STANDARD 11)

in the top level cmakefile

Issue #2411, opened 2019-02-09
Remove limitation to writing frames in PNG/OpenEXR formats


dictoon:
`renderer::Frame` is still limited to writing PNG and OpenEXR files. Now that we're piggybacking on OpenImageIO we should be able to write a much larger variety of image file formats.

This requires updating the `write_image()` free function in `frame.cpp`, replacing `g_appleseed_image_files_filter` by `g_qt_image_files_filter` in appleseed.studio and finally getting rid of `g_appleseed_image_files_filter` altogether.

dictoon:
Done.

PR #2415, opened 2019-02-10
 Allow saving other image formats

KarthikRIyer:
#2411
@dictoon I'm unable to save jpg successfully and the bmp images o not appear right. Could you please tell me what changes I might need to make to get it right?


dictoon:
We don't want to explicit handle some image formats, instead we want to automatically support all formats supported by OIIO.

CMakeLists.txt
```diff
@@ -100,8 +100,6 @@ endif ()
 # Compiler features.
 #--------------------------------------------------------------------------------------------------
 
-set (CMAKE_CXX_STANDARD 11)
```
dictoon: Revert changes to this file.

src/appleseed/foundation/image/genericimagefilewriter.cpp
```diff
@@ -297,6 +297,13 @@ void GenericImageFileWriter::set_exr_image_attributes(const ImageAttributes& ima
     }
 }
 
+void GenericImageFileWriter::set_bmp_image_attributes(const ImageAttributes& image_attributes)
+{
+    OIIO::ImageSpec& spec = impl->m_spec.back();
+
+    spec.attribute("ResolutionUnit", "m");
```
dictoon: What does this do? Why is it necessary?

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -550,30 +550,26 @@ namespace
...
+      create_parent_directories(file_path);
 
-        image_attributes.insert("color_space", "sRGB");
-        writer.set_image_attributes(image_attributes);
+      const std::string filename = file_path.string();
+      GenericImageFileWriter writer(filename.c_str());
 
-        writer.write();
+      writer.append_image(&transformed_image);
+      writer.set_image_output_format(PixelFormat::PixelFormatUInt8);
```
dictoon: Just like @slongle PR, as far as I can tell this will break writing high dynamic range formats such as OpenEXR.

src/appleseed/foundation/image/genericimagefilewriter.cpp
KarthikRIyer: bmp_oiio
It says in the oiio docs that for bmp ResolutionUnit is always m.
I think ResolutionUnit matters only when we print the image.
resolutionunit_oiio

CMakeLists.txt
KarthikRIyer: Upto which commit should I revert? If this line is necessary can I just add it in another commit?

src/appleseed/foundation/image/genericimagefilewriter.cpp
dictoon: I'm not sure to understand. What in the docs makes you think you need to set `ResolutionUnit` to `m`? To me it seems that for the BMP format, it's always set to `m` anyway.

CMakeLists.txt
dictoon: You should revert all changes to this file. You need make sure your PR doesn't introduce unnecessary changes, and even more importantly that it doesn't accidentally revert older commits (this is what it's doing right now).

dictoon:
Superseded by #2445.

PR #2418, opened 2019-02-11
Make alembicassembly sample compiles again

Narann:
I've updated the alembic assembly sample to make it compile again with new inheritance.

Any feedback appreciate. :)

### DIFF
--- a/sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
+++ b/sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
@@ -32,6 +32,7 @@
 #include "renderer/api/log.h"
 #include "renderer/api/object.h"
 #include "renderer/api/project.h"
+#include "renderer/api/rasterization.h"
 #include "renderer/api/scene.h"
 #include "renderer/api/utility.h"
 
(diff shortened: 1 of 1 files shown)


sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -1163,6 +1164,86 @@ namespace
...
+            asf::DictionaryArray metadata;
+
+            metadata.push_back(
+                asf::Dictionary()
+                    .insert("name", "file_path")
+                    .insert("label", "File path")
+                    .insert("type", "text")
+                    .insert("use", "required"));
+
+            metadata.push_back(
```
dictoon: Are these supposed to be the metadata for the pinhole camera?

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -1214,8 +1299,10 @@ namespace
         }
 
         bool on_render_begin(
-            const asr::Project&      project,
-            asf::IAbortSwitch*       abort_switch) override
+            const asr::Project&             project,
```
dictoon: Align arguments with neighboring methods.

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
Narann: No, those are assembly parameters use to sample the given alembic file.

For now, if not setted, [it use `0.0f`](https://github.com/appleseedhq/appleseed/blob/01a053de46c70359511dcc2494e59e364edcbfc7/sandbox/samples/cpp/alembicassembly/alembicassembly.cpp#L432).

IIRC there is no way to retrieve the shutter open/close of the current scene, so you need to pass it explicitly to each assembly.

Doc [here](https://github.com/appleseedhq/appleseed/wiki/Alembic-Assembly#shutter_open_time-and-shutter_close_time-float-optional)

The camera doesn't have shutter open/close but  [explicit `time_offset`](https://github.com/appleseedhq/appleseed/wiki/Alembic-Camera#time_offset-float)

Any suggestion ?

Narann:
> What's the status of the importer, functionality-wise?
This PR simply make it compile with new appleseed code. No modification.

I have [a board](https://github.com/appleseedhq/appleseed/projects/10) I try to keep up to date.

For now I will focus on implicit instancing. I will need to have a dedicated method (maybe static) for polymesh extraction because it's a big part of the code and it make the rest of the logic (specially, the code flow for instancing) hard to read. I will try to do hte PR in two times. Split polymesh extraction method, and add implicit instancing.

Have you any feature you would want to see supported ?

sandbox/samples/cpp/alembicassembly/alembicassembly.cpp
dictoon: > IIRC there is no way to retrieve the shutter open/close of the current scene, so you need to pass it explicitly to each assembly.

You _can_ retrieve the shutter times of the camera. In `do_expand_contents()` you've got a `Project` so you first need to retrieve the active camera used for rendering:

    const Camera* camera = project.get_scene()->get_active_camera();

Then you can use `camera->get_shutter_open_begin_time()`, etc. to retrieve shutter times.

Narann:
Yes we can. The next PR would be the implicit instancing system.

PR #2424, opened 2019-02-14
add implicit instancing support in AlembicAssembly

Narann:
As its name states, this commit add implicit instancing in AlembicAssembly using `IObject::getPropertyHash()` on poly mesh.

The weird part of the code review will be that a tiny refacto appear: I created a `do_extract_polymesh()` method that put the content of the Alembic `IPolyMesh` into the appleseed `MeshObject`. This obfuscate the diff but it make the whole logic easier to read.


Narann:
I also think I should provide some test scene but I have no idea where I should put them. `sandbox/samples/cpp/alembicassembly` looks weird to me.

dictoon:
> I also think I should provide some test scene but I have no idea where I should put them. `sandbox/samples/cpp/alembicassembly` looks weird to me.

A `testscenes/` subdirectory in `sandbox/samples/cpp/alembicassembly/` sounds perfectly acceptable to me.

Narann:
Thanks, I try to use relative paths in the appleseed XML file:
```xml
<parameter name="file_path" value="alembics/xform_ogawa.abc" />
```
This path is relative to the scene but I suspect I have to resolve it as I'm in `bin/Ship` when I run `appleseed.studio`.

Any hint, in the code, where a similar problem is solved?

Narann:
```
2019-02-22T20:08:03.968452Z <003>   165 MB info    | time_offset: 0.000000
2019-02-22T20:08:03.968452Z <003>   165 MB info    | verbose: true
2019-02-22T20:08:03.969460Z <003>   167 MB warning | invalid archive: alembics/xform_ogawa.abc
2019-02-22T20:08:03.973803Z <001>   167 MB info    | rendering finished in 0 ms.
```
It's my code logging this: https://github.com/Narann/appleseed/blob/master/sandbox/samples/cpp/alembicassembly/alembicassembly.cpp#L450

So I need to "interpret" a relative file path. Once again, I'm sure you guys already had to deal with that.

Narann:
Reading other scenes, it looks like .obj are always passed relative to the .appleseed scene:
```xml
                <object name="Sphere" model="mesh_object">
                    <parameter name="filename" value="meshes/Sphere.obj" />
                </object>
```
But the obj parser doesn't seems to do anything to convert to absolute path:
https://github.com/appleseedhq/appleseed/blob/9c6daf3dc8667b57d824ff2f7967f046da317dd2/src/appleseed/foundation/mesh/objmeshfilereader.cpp#L574
https://github.com/appleseedhq/appleseed/blob/96c4992476ed9a5c5d49e64ce21259cb09c1cacd/src/appleseed/foundation/mesh/objmeshfilelexer.h#L78
So I suspect the working directory used when rendering test scenes must be the place where scenes are.
Am I good ?

dictoon:
> So I suspect the working directory used when rendering test scenes must be the place where scenes are.

That's correct.

Issue #2345, opened 2018-12-04
Make auto_release_ptr bool-convertible


dictoon:

est77:
Could we use something like this?

```
#include <memory>

struct ReleaseDeleter
{
    template <class T>
    void operator()(T* x) {x->release();}
};

template <class T>
using auto_release_ptr = std::unique_ptr<T, ReleaseDeleter>;
```

dictoon:
Indeed, I've been thinking of doing that for a while now (but using a `typedef` since VS 2012 doesn't support defining types with `using`).

dictoon:
Done.

PR #2425, opened 2019-02-15
Fix issue #2345 : auto_release_ptr is now bool-convertible

Uedaki:
Issue: https://github.com/appleseedhq/appleseed/issues/2345

### DIFF
--- a/src/appleseed/foundation/utility/autoreleaseptr.h
+++ b/src/appleseed/foundation/utility/autoreleaseptr.h
@@ -147,6 +147,7 @@ class auto_release_ptr
         return m_ptr;
     }
 
+    // Implicit bool conversion of the wrapped pointer.
     operator bool() const
     {
         return static_cast<bool>(m_ptr);
--- a/src/appleseed/foundation/utility/autoreleaseptr.h
+++ b/src/appleseed/foundation/utility/autoreleaseptr.h
@@ -148,7 +148,7 @@ class auto_release_ptr
     }
 
     // Implicit bool conversion of the wrapped pointer.
-    explicit operator bool() const
+    operator bool() const
     {
         return static_cast<bool>(m_ptr);
     }
--- a/src/appleseed/foundation/utility/autoreleaseptr.h
+++ b/src/appleseed/foundation/utility/autoreleaseptr.h
@@ -148,7 +148,7 @@ class auto_release_ptr
     }
 
     // Implicit bool conversion of the wrapped pointer.
-    operator bool() const
+    explicit operator bool() const
     {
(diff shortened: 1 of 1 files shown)


src/appleseed/foundation/utility/autoreleaseptr.h
```diff
@@ -147,6 +147,11 @@ class auto_release_ptr
         return m_ptr;
     }
 
+    operator bool() const
```
dictoon: Looks good! Can you add a comment please? Something like

    // Implicit conversion of the wrapped pointer to a bool.

est77: Missing explicit, maybe?

Uedaki: Comment and explicit added in 2 separate commit so if the explicit wasnt necessary, we just have to revert the last commit.
@est77, why do you think explicit is important for this function?

dictoon: @est77 Good catch, I hadn't noticed that the conversion was explicit on `std::unique_ptr<>`.

@Uedaki Ideally we would like the conversion to be explicit to mimic `std::unique_ptr<>` and other standard smart pointers. The reason they're marked as explicit is to avoid things like

    std::unique_ptr<int> x = std::make_unique(42);
    const bool b = x;  // error

while allowing

    std::unique_ptr<int> x = std::make_unique(42);
    if (x)  // OK
    {
        ...
    }

Unfortunately, as you may have noticed, explicit conversion operators aren't supported by VS 2012 which we still need to support. I suggest making explicit conditional to allow catching errors on newer compilers:

        // Bool conversion of the wrapped pointer.
    #if defined _MSC_VER && _MSC_VER >= 1800
        explicit
    #endif
        operator bool() const
        {
            return static_cast<bool>(m_ptr);
        }

Make sure to remove the word "implicit" from the comment as it's no longer the case.

Uedaki:
Thanks for the explanation
I will do the change later this day

src/appleseed/foundation/utility/autoreleaseptr.h
```diff
@@ -147,6 +147,15 @@ class auto_release_ptr
         return m_ptr;
     }
 
+    // Bool conversion of the wrapped pointer.
+#if (defined _MSC_VER && _MSC_VER >= 1800) || defined(__GNUC__)
```
dictoon: I think this is correct, but an even better solution is this:

    #if !defined(_MSC_VER) || _MSC_VER >= 1800

PR #2426, opened 2019-02-15
Add new shaders, Maya matte support, fixes

luisbarrancos:
- Added raySwitch surface shader, for closures. We have one for textures.
- Added blendNormals node to blend **tangent space normal maps** via reoriented normal mapping.
- Added surface luminance node via NPR rays, color space aware.
- Added support for matte opacity in the Maya shaders, but restricted only to matte opacity mode 0 (*Black Hole*). For the rest, piping the output to the [as_matte](https://github.com/appleseedhq/appleseed/blob/master/src/appleseed.shaders/src/appleseed/as_matte.osl) shader via shader passthrough provides equivalent functionality.
- Added minor performance optimizations to the Maya nodes, mostly related to opacity tests and *shadow* and *transparency* rays.
- Added orientation check to *as_fresnel*, to ensure the normal is forward facing.
- Added access to appleseed exclusive global primvars dNdu, dNdv, and others.
- Added access to *vertex_color*, currently a placeholder attribute.
- Chromaticity coordinates can be negative, so ensure the UI metadata reflects this in *as_luminance* node.
- Added the *glass random walk* profile to *as_subsurface*.
- Added flip/switch XY/RG components of the texture sections of the triplanar node, required to blend correctly tangent space normal maps if they follow different conventions (OpenGL vs DX)
- Minor fixes and cleanups.

### DIFF
--- a/src/appleseed.shaders/CMakeLists.txt
+++ b/src/appleseed.shaders/CMakeLists.txt
@@ -267,6 +267,7 @@ set (src_appleseed_sources
      src/appleseed/as_attributes.osl
      src/appleseed/as_blackbody.osl
      src/appleseed/as_blend_color.osl
+     src/appleseed/as_blend_normal.osl
      src/appleseed/as_blend_shader.osl
      src/appleseed/as_bump.osl
      src/appleseed/as_closure2surface.osl
@@ -290,10 +291,12 @@ set (src_appleseed_sources
      src/appleseed/as_noise3d.osl
      src/appleseed/as_plastic.osl
      src/appleseed/as_ray_switch.osl
+     src/appleseed/as_ray_switch_surface.osl
      src/appleseed/as_sbs_pbrmaterial.osl
      src/appleseed/as_space_transform.osl
      src/appleseed/as_standard_surface.osl
      src/appleseed/as_subsurface.osl
+     src/appleseed/as_surface_luminance.osl
      src/appleseed/as_switch_surface.osl
      src/appleseed/as_switch_texture.osl
      src/appleseed/as_swizzle.osl
--- a/src/appleseed.shaders/include/appleseed/math/as_math_helpers.h
+++ b/src/appleseed.shaders/include/appleseed/math/as_math_helpers.h
@@ -81,48 +81,14 @@ float fast_gain(float value, float g)
 //
 // Reference:
 //
-//      Lehmer linear congruential generator (Park-Miller)
-//
-//      Random Number Generators: Good Ones Are Hard To Find,
-//      Stephen K.Park and Keith W.Miller
-//
-//      http://www.firstpr.com.au/dsp/rand31/p1192-park.pdf
-//      https://en.wikipedia.org/wiki/Lehmer_random_number_generator
+//  https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/divmodnote-letter.pdf
 //
 
-int rand_int(int seed)
+int remainder(int x, int y)
 {
-    int a = 16807;
-    int q = 127773;
-    int r = 2836;
-    int m = 2147483647;
-
-    int hi = seed / q;
-    int lo = seed % q;
-    int x = a * lo - r * hi;
-
-    return x > 0 ? x : x + m;
-}
-
-int rand_float(int seed, output float result)
-{
-    int x = rand_int(seed);
-    result = x / 2147483647;
-
-    return x;
-}
-
-float random(float x, float s, float t)
-{
-    vector A = vector(s, t, 0); // s=u, t=v usually
-    vector B = vector(4.213536, 8.34621351, 0);
-
-    return mod(sin(dot(A, B)) * 917853.4917593, 1.0);
-}
-
-float random(float x)
-{
-    return random(x, u, v);
+    int rem = x % y;
+    if ((rem > 0 && y < 0) || (rem < 0 && y > 0)) rem += y;
+    return rem;
 }
 
 float fract(float x)
--- a/src/appleseed.shaders/src/appleseed/as_globals.osl
+++ b/src/appleseed.shaders/src/appleseed/as_globals.osl
@@ -30,7 +30,7 @@ shader as_globals
 [[
     string help = "OSL global variables.",
     string icon = "asGlobals.png",
-    string URL = "https://appleseed.readthedocs.io/projects/appleseed-maya/en/latest/shaders/utilities/as_globals.html#label-as-globals",
+    string URL = "https://appleseed.readthedocs.io/projects/appleseed-maya/en/latest/shaders/utilities/as_globals.html",
 
     string as_node_name = "asGlobals",
     string as_category = "utility",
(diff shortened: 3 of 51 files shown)


luisbarrancos:
This requires the type IDs and other changes currently in progress for appleseed-maya.
Until Max can address adding parameters without breaking backwards compatibility, the following nodes are breaking backwards compatibility:
 * as_bump
 * as_subsurface

Mango-3:
Very nice!
One thing, as_bump is not used in appleseed-Max, it uses it's own bump map implementation which is added to each material node. So the reorientNormal checkbox you added to as_bump.osl would need to be replicated in the appleseed-Max C++ code.

luisbarrancos:
Note to update this with IDs for https://github.com/appleseedhq/appleseed/pull/2537

luisbarrancos:
@usakhelo what changes are requested for Max? Can max already deal gracefully with new parameters?

src/appleseed.shaders/CMakeLists.txt
```diff
@@ -267,6 +267,7 @@ set (src_appleseed_sources
      src/appleseed/as_attributes.osl
      src/appleseed/as_blackbody.osl
      src/appleseed/as_blend_color.osl
+     src/appleseed/as_blend_normals.osl
```
dictoon: We have `as_blend_color` (singular) and `as_blend_shader` (singular), should the new shader be called `as_blend_normal` or is there a reason why it's spelled with a plural?

src/appleseed.shaders/include/appleseed/math/as_math_helpers.h
```diff
@@ -34,7 +34,6 @@
 float sqr(float x) { return x * x; }
 vector sqr(vector x) { return x * x; }
 
-
```
dictoon: Block comments should be preceded by two blank lines so I would suggest to revert these changes.

src/appleseed.shaders/include/appleseed/math/as_math_helpers.h
```diff
@@ -77,52 +75,17 @@ float fast_gain(float value, float g)
         : fast_bias(value * 2 - 1, 1 - g) * 0.5 + 0.5;
 }
 
-
 //
 // Reference:
 //
-//      Lehmer linear congruential generator (Park-Miller)
```
dictoon: Out of curiosity, why did you remove the RNG stuff?

src/appleseed.shaders/src/appleseed/as_surface_luminance.osl
```diff
@@ -0,0 +1,476 @@
...
+        string as_maya_attribute_short_name = "r",
+        string label = "Luminance Scalar"
+    ]],
+    output closure color out_color = 0
+    [[
+        string as_maya_attribute_name = "outColor",
+        string as_maya_attribute_short_name = "oc",
+        string label = "Surface Luminance",
+        string widget = "null"
+    ]],
```
dictoon: Remove trailing comma.

src/appleseed.shaders/src/appleseed/as_switch_texture.osl
```diff
@@ -202,7 +203,7 @@ shader as_switch_texture
         grey_id);
 
     int index = (in_cycle_mode == 0)
-        ? (int) mod(hash_id, len)
+        ? remainder(hash_id, len + 1)
```
dictoon: It would be more logical if `len` was really the length of the array, as you did in another shader.

src/appleseed.shaders/CMakeLists.txt
luisbarrancos: It should be *as_blend_normal* for consistency yes. I'll update the node IDs, if needed as well.

src/appleseed.shaders/src/appleseed/as_switch_texture.osl
luisbarrancos: I agree it can be clearer, editing now. FYI, the idea is to get a cycling [0,7] range index for texture lookups in one mode, and to get a clamped [0,7] range index in the other mode.

src/appleseed.shaders/include/appleseed/math/as_math_helpers.h
luisbarrancos: It was only used previously in the implementation of the maya specific procedural noises, and we replaced that by OSL's native hashnoise(), which allowed us also to discard the big arrays/tables. If it's handy to have the RNG in general (albeit unused atm) let me know, i'll reinstate it.

luisbarrancos: It was only used previously in the implementation of the maya specific procedural noises. There it was replaced by OSL's native **hashnoise()**, which allowed us also to discard the big arrays/tables. If it's handy to have the (pseudo) RNG in general (albeit unused atm) let me know so i can reinstate it.

src/appleseed.shaders/CMakeLists.txt
luisbarrancos: Addressed together with this PR:
https://github.com/appleseedhq/appleseed-maya/pull/210

src/appleseed.shaders/include/appleseed/math/as_math_helpers.h
dictoon: Was just curious, thanks for the explanation!

dictoon:
> Does this still pose a problem for Max?

I think so. We need to make 3ds Max and Blender test scenes that use the shaders (in their initial version) for which params were added/removed in this PR, then locally merge this PR and try to open the test scenes again.

luisbarrancos:
@dictoon don't merge yet since there were some triplanar conflicts that require some testing.

src/appleseed.shaders/src/appleseed/as_triplanar.osl
```diff
@@ -81,7 +81,7 @@ shader as_triplanar
         string label = "Surface Point",
         string page = "Projection",
         int divider = 1,
-        int as_max_param_id = 3
+        int as_max_param_id = 2
```
dictoon: I don't think this is correct, @usakhelo has set the `as_max_param_id` values carefully as to preserve backward compatibility, and these shouldn't be changed.

src/appleseed.shaders/src/appleseed/as_subsurface.osl
```diff
@@ -97,6 +97,23 @@ shader as_subsurface
...
+        float max = 1.0,
+        int as_maya_attribute_connectable = 0,
+        int as_maya_attribute_keyable = 0,
+        int as_maya_attribute_hidden = 1,
+        int as_blender_input_socket = 0,
+        int gafferNoduleLayoutVisible = 0,
+        string label = "Volume Anisotropy",
+        string page = "Subsurface",
+        string help = "Scattering anisotropy, with negative values producing back scattering, 0 producing isotropic scattering, and positive values producing forward scattering",
+        int divider = 1
```
dictoon: You need to set a new, unused `as_max_param_id` value for all new parameters.

usakhelo:
Just wanted to mention that because `as_globals.osl` is an existing shader its output parameters should be assigned specific param ID - `int as_max_param_id = 2`. And it should be the same param ID for each output parameter.
And that rules applies to all shaders - output parameters should have the same ID because they all gathered under one 3ds max UI parameter.

dictoon:
Correction on my [earlier comment](https://github.com/appleseedhq/appleseed/pull/2426#pullrequestreview-326478910):

`as_maya_lambert.osl` does NOT need `as_max_param_id` attributes since it's Maya-only and not used in Max (thanksfully).

src/appleseed.shaders/src/appleseed/as_triplanar.osl
usakhelo: Hi,
I can confirm that we shouldn't change parameter IDs.
@luisbarrancos, I'm wondering if I shared these ID values with you already? They seem to be close to what I have in my branch but not exactly the same... I was wondering if it's something I shared with you at some point and then changed it.. Sorry, I don't remember anything.
Anyway, I have current IDs for all existing shaders in case we need it.

src/appleseed.shaders/src/appleseed/as_globals.osl
```diff
@@ -136,33 +136,73 @@ shader as_globals
...
+    normal dNdu = 0
     [[
         int lockgeom = 0,
+        string widget = "null",
+        int as_maya_attribute_connectable = 0,
+        int as_maya_attribute_keyable = 0,
+        int as_maya_attribute_hidden = 1,
+        int as_blender_input_socket = 0,
+        int gafferNoduleLayoutVisible = 0,
+        string widget = "null"
```
dictoon: You will need to rebase on top of master and add `as_max_param_id` values to all new attributes **in shaders that already use them** (NOT the other shaders).

src/appleseed.shaders/src/appleseed/as_triplanar.osl
dictoon: Revert and use what's in master instead.

luisbarrancos: @usakhelo i think you shared this with me when this PR was opened but it was some time ago and i don't remember it either. In any case, it should be ok now. Merged with the values in master.

Issue #2325, opened 2018-11-28
Warnings in randomwalkbssrdf.cpp


oktomus:
```
In file included from /home/oktomus/dev/appleseedhq/appleseed.code/1.9.shared/src/appleseed/foundation/math/quaternion.h:35:0,                                                                                     
                 from /home/oktomus/dev/appleseedhq/appleseed.code/1.9.shared/src/appleseed/foundation/math/matrix.h:36,                                                                    
                 from /home/oktomus/dev/appleseedhq/appleseed.code/1.9.shared/src/appleseed/foundation/image/color.h:36,                                                                                           
                 from /home/oktomus/dev/appleseedhq/appleseed.code/1.9.shared/src/appleseed/renderer/utility/dynamicspectrum.h:33,                                                                                 
                 from /home/oktomus/dev/appleseedhq/appleseed.code/1.9.shared/src/appleseed/renderer/global/globaltypes.h:34,                                                                                      
                 from /home/oktomus/dev/appleseedhq/appleseed.code/1.9.shared/src/appleseed/renderer/modeling/bssrdf/randomwalkbssrdf.h:33,                                                                        
                 from /home/oktomus/dev/appleseedhq/appleseed.code/1.9.shared/src/appleseed/renderer/modeling/bssrdf/randomwalkbssrdf.cpp:30:                                                                       
/home/oktomus/dev/appleseedhq/appleseed.code/1.9.shared/src/appleseed/foundation/math/vector.h:751:9: warning: ‘slab_normal’ may be used uninitialized in this function [-Wmaybe-uninitialized]                    
         result += lhs[i] * rhs[i];                                                                                                                                                                                
         ^                                                                                                                                                                                                         
/home/oktomus/dev/appleseedhq/appleseed.code/1.9.shared/src/appleseed/renderer/modeling/bssrdf/randomwalkbssrdf.cpp:302:22: note: ‘slab_normal’ was declared here                    
             Vector3f slab_normal;                                                                                                                                                                                 
                      ^                 
```

oktomus:
Can't we do 
```cpp
Vector3f slab_normal(0.0f);
```
?

dictoon:
I wouldn't want to initialize `slab_normal` just to silent a warning without making sure there isn't a deeper problem.

For instance, a cursory look reveals that `trace_zero_scattering_path_glass()` could return `true` without initializing `slab_normal`. That looks like a genuine bug.

dictoon:
Taken care of in #2428.

PR #2428, opened 2019-02-16
Miscellaneous fixes and cleanups

dictoon:
@biart Would you mind reviewing the changes in the Random-Walk BSSRDF please?


biart:
Looks good to me, I approve. Thanks for caring about the code.

Just a single question: are you poisoning these variables to track possible errors? It is related to this, right: #2325 ?

dictoon:
Thanks for the reviews @est77 and @biart.

Yes, the poisoning is related to #2325. As far as I can tell the code is correct (`slab_normal` won't be used if it's not been initialized), but I want to make sure there will be an easy way to detect if we start using an uninitialized variable, now or later.

One thing I realized we need to do is make `poison()` automatically do nothing in release builds. I'll do that in a subsequent PR.

PR #2429, opened 2019-02-17
Disable poisoning everywhere in debug builds

dictoon:


dictoon:
Thanks for the review @biart! Merging now.

Issue #1543, opened 2017-07-19
Allow applying materials via drag & drop


dictoon:

Uedaki:
https://github.com/appleseedhq/appleseed/pull/2431

dictoon:
Done.

PR #2432, opened 2019-02-20
Implementing drag & drop to apply materials

Uedaki:
issue: https://github.com/appleseedhq/appleseed/issues/1543


src/appleseed.studio/mainwindow/droppedmaterialmenu.cpp
```diff
@@ -0,0 +1,73 @@
+
+//
+// This source file is part of appleseed.
+// Visit https://appleseedhq.net/ for additional information and resources.
+//
+// This software is released under the MIT license.
+//
+// Copyright (c) 2010-2013 Francois Beaune, Jupiter Jazz Limited
+// Copyright (c) 2014-2018 Francois Beaune, The appleseedhq Organization
```
pjessesco: You should change to your name!

src/appleseed.studio/mainwindow/droppedmaterialmenu.h
```diff
@@ -0,0 +1,75 @@
+
+//
+// This source file is part of appleseed.
+// Visit https://appleseedhq.net/ for additional information and resources.
+//
+// This software is released under the MIT license.
+//
+// Copyright (c) 2010-2013 Francois Beaune, Jupiter Jazz Limited
```
pjessesco: Same here.

src/appleseed.studio/mainwindow/rendering/renderwidget.h
```diff
@@ -46,14 +46,18 @@ namespace OCIO = OCIO_NAMESPACE;
 #include <QMutex>
 #include <QPainter>
 #include <QWidget>
+#include <QDropEvent>
```
pjessesco: Could you sort these include statement in ascending order?

src/appleseed.studio/mainwindow/rendering/renderwidget.cpp
```diff
@@ -45,6 +46,7 @@
 #include <QColor>
 #include <QMutexLocker>
 #include <Qt>
+#include <QMimeData>
```
pjessesco: Same here

src/appleseed.studio/CMakeLists.txt
```diff
@@ -296,6 +296,8 @@ set (mainwindow_sources
     mainwindow/mainwindow.cpp
     mainwindow/mainwindow.h
     mainwindow/mainwindow.ui
+	mainwindow/droppedmaterialmenu.cpp
```
dictoon: Still appears to use tabs here.

dictoon: Keep the alphabetical ordering of files.

src/appleseed.studio/mainwindow/droppedmaterialmenu.cpp
```diff
@@ -0,0 +1,74 @@
+
+//
+// This source file is part of appleseed.
+// Visit https://appleseedhq.net/ for additional information and resources.
+//
+// This software is released under the MIT license.
+//
+// Copyright (c) 2010-2013 Thibault Vergne, Jupiter Jazz Limited
```
dictoon: The copyright notice for all files created this year should simply be

    Copyright (c) 2019 Thibault Vergne, The appleseedhq Organization

More details here: https://github.com/appleseedhq/appleseed/wiki/Copyright-Notices

src/appleseed.studio/mainwindow/rendering/renderwidget.cpp
```diff
@@ -422,5 +428,29 @@ void RenderWidget::paintEvent(QPaintEvent* event)
     m_painter.end();
 }
 
+void RenderWidget::dragEnterEvent(QDragEnterEvent* event)
+{
+    if (event->mimeData()->hasFormat("text/plain") && m_project.has_trace_context())
```
dictoon: It looks like you only need the project here. I wonder if it wouldn't be simpler to check if the trace context exists elsewhere, that way the only responsability of `RenderWidget` would be to accept drag & drop events and emit the right signal. It sounds simpler to me.

src/appleseed.studio/mainwindow/rendering/renderwidget.cpp
```diff
@@ -422,5 +428,29 @@ void RenderWidget::paintEvent(QPaintEvent* event)
...
+        event->acceptProposedAction();
+}
+
+void RenderWidget::dragMoveEvent(QDragMoveEvent* event)
+{
+    if (event->pos().x() > pos().x() + width() || event->pos().y() > pos().y() + height())
```
dictoon: Three questions:

- What are these checks for exactly?
- Are you sure it shouldn't be `pos().x() + width() - 1`? Or, alternatively, `>=` instead of `>`?
- Isn't checking the other sides (< 0) also necessary?

src/appleseed.studio/mainwindow/rendering/renderwidget.cpp
```diff
@@ -422,5 +428,29 @@ void RenderWidget::paintEvent(QPaintEvent* event)
...
+    else
+        event->accept();
+}
+
+void RenderWidget::dropEvent(QDropEvent* event)
+{
+    const Vector2d drop_pos = Vector2d(
+        static_cast<double>(event->pos().x()) / width(),
+        static_cast<double>(event->pos().y()) / height());
+    std::string material_name = event->mimeData()->text().toUtf8().constData();
```
dictoon: `event->mimeData()->text().toStdString()` seems a little bit simpler.

Also, make the variable const.

src/appleseed.studio/mainwindow/rendering/renderwidget.cpp
```diff
@@ -422,5 +428,29 @@ void RenderWidget::paintEvent(QPaintEvent* event)
...
+{
+    if (event->pos().x() > pos().x() + width() || event->pos().y() > pos().y() + height())
+        event->ignore();
+    else
+        event->accept();
+}
+
+void RenderWidget::dropEvent(QDropEvent* event)
+{
+    const Vector2d drop_pos = Vector2d(
```
dictoon: You actually don't need the `Vector2()` part:

    const Vector2d drop_pos(
        static_cast<double>(event->pos().x()) / width(),
        static_cast<double>(event->pos().y()) / height());

src/appleseed.studio/mainwindow/rendering/renderwidget.h
```diff
@@ -53,7 +55,8 @@ namespace OCIO = OCIO_NAMESPACE;
 
 // Forward declarations.
 namespace foundation    { class CanvasProperties; }
-namespace renderer      { class Frame; }
+namespace renderer      { class Frame; class Project; }
```
dictoon: One class at a time (repeat the namespace).

src/appleseed.studio/mainwindow/rendering/renderwidget.h
```diff
@@ -42,6 +42,8 @@
 namespace OCIO = OCIO_NAMESPACE;
 
 // Qt headers.
+#include <QDragEnterEvent>
```
dictoon: You can forward declare those (and include them in the .cpp file instead).

src/appleseed.studio/mainwindow/rendering/renderwidget.h
```diff
@@ -139,6 +145,14 @@ class RenderWidget
...
 
     void paintEvent(QPaintEvent* event) override;
+    void dragEnterEvent(QDragEnterEvent* event) override;
+    void dragMoveEvent(QDragMoveEvent* event) override;
+    void dropEvent(QDropEvent* event) override;
+
+signals:
+    void signal_material_dropped(
+        const foundation::Vector2d& drop_pos,
+        const std::string& material_name);
```
dictoon: You need to include `<string>` and `foundation/math/vector.h`.

src/appleseed.studio/mainwindow/rendering/renderwidget.h
```diff
@@ -139,6 +145,14 @@ class RenderWidget
         const size_t            tile_y);
 
     void paintEvent(QPaintEvent* event) override;
+    void dragEnterEvent(QDragEnterEvent* event) override;
+    void dragMoveEvent(QDragMoveEvent* event) override;
+    void dropEvent(QDropEvent* event) override;
+
+signals:
```
dictoon: Half-indent (2 spaces), like `public` etc.

dictoon: Signals are part of the public API, move that section before the private section.

src/appleseed.studio/mainwindow/rendering/renderingmanager.h
```diff
@@ -202,6 +202,13 @@ class RenderingManager
     void slot_camera_change_begin();
     void slot_camera_changed();
     void slot_camera_change_end();
+    void slot_material_dropped(
+        const foundation::Vector2d& drop_pos,
+        const std::string& material_name);
```
dictoon: Align arguments vertically (check how it's done throughout the codebase).

dictoon: Add any necessary header file (I haven't checked if you need anything new, but just in case).

src/appleseed.studio/mainwindow/project/projectexplorer.cpp
```diff
@@ -47,6 +47,8 @@
 #include <Qt>
 #include <QTreeWidget>
 #include <QTreeWidgetItem>
+#include <QDrag>
```
dictoon: Order alphabetically.

src/appleseed.studio/mainwindow/project/projectexplorer.cpp
```diff
@@ -247,6 +253,20 @@ void ProjectExplorer::slot_edit_item(QTreeWidgetItem* item, int column)
     static_cast<ItemBase*>(item)->slot_edit();
 }
 
+void ProjectExplorer::slot_drag_item(QTreeWidgetItem* item, int column)
+{
+    if (item && item->flags() & Qt::ItemIsDragEnabled)
+    {
+        QDrag* drag = new QDrag(m_tree_widget);
```
dictoon: The code might be correct, but who owns the allocated memory? When is it released?

src/appleseed.studio/mainwindow/project/projectexplorer.cpp
```diff
@@ -247,6 +253,20 @@ void ProjectExplorer::slot_edit_item(QTreeWidgetItem* item, int column)
...
+        QMimeData* mimeData = new QMimeData;
```
dictoon: Conventionally, we always include the constructor's parentheses here.

src/appleseed.studio/mainwindow/project/itembase.cpp
```diff
@@ -49,6 +49,7 @@ ItemBase::ItemBase(
   , m_allow_edition(true)
   , m_allow_deletion(true)
 {
+    setFlags(Qt::ItemIsEnabled | Qt::ItemIsSelectable | Qt::ItemIsUserCheckable);
```
dictoon: Why is this code necessary?

src/appleseed.studio/mainwindow/droppedmaterialmenu.h
```diff
@@ -0,0 +1,75 @@
+
+//
+// This source file is part of appleseed.
+// Visit https://appleseedhq.net/ for additional information and resources.
+//
+// This software is released under the MIT license.
+//
+// Copyright (c) 2010-2013 Thibault Vergne, Jupiter Jazz Limited
```
dictoon: Same thing.

src/appleseed.studio/mainwindow/droppedmaterialmenu.h
```diff
@@ -0,0 +1,75 @@
...
+
+// Qt headers.
+#include <QMenu>
+#include <QObject>
+
+namespace appleseed {
+namespace studio {
+
+class DroppedMaterialMenu
+    : public QMenu
```
dictoon: Indent with two spaces only.

src/appleseed.studio/mainwindow/droppedmaterialmenu.h
```diff
@@ -0,0 +1,75 @@
...
+{
+    Q_OBJECT
+
+public:
```
dictoon: Indent with two spaces only (check how it's done throughout the codebase).

src/appleseed.studio/mainwindow/droppedmaterialmenu.h
```diff
@@ -0,0 +1,75 @@
...
+    // constructor
+    DroppedMaterialMenu(const foundation::Vector2d& drop_pos, const std::string& material_name);
```
dictoon: Include `<string>`.

src/appleseed.studio/mainwindow/droppedmaterialmenu.h
```diff
@@ -0,0 +1,75 @@
...
+
+    // destructor
+    ~DroppedMaterialMenu() override = default;
+
+signals:
+    void signal_apply_material(
+        const foundation::Vector2d& drop_pos,
+        const std::string& material_name,
+        renderer::ObjectInstance::Side side);
```
dictoon: Make arg const.

src/appleseed.studio/mainwindow/droppedmaterialmenu.h
```diff
@@ -0,0 +1,75 @@
...
+public:
+    // constructor
+    DroppedMaterialMenu(const foundation::Vector2d& drop_pos, const std::string& material_name);
+
+    // destructor
+    ~DroppedMaterialMenu() override = default;
+
+signals:
+    void signal_apply_material(
+        const foundation::Vector2d& drop_pos,
```
dictoon: Align args vertically.

src/appleseed.studio/mainwindow/droppedmaterialmenu.h
```diff
@@ -0,0 +1,75 @@
...
+        const std::string& material_name,
+        renderer::ObjectInstance::Side side);
+
+private:
+    const foundation::Vector2d& m_drop_pos;
```
dictoon: Align members vertically.

dictoon: Are you sure keeping just a (const) ref is enough here? It might be, I'm just asking. Storing a value instead would be perfectly fine.

src/appleseed.studio/mainwindow/droppedmaterialmenu.h
```diff
@@ -0,0 +1,75 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+#pragma once
+
+// appleseed.foundation headers.
+#include "foundation/math/vector.h"
+
+// appleseed.renderer headers.
```
dictoon: Order include files from the most specific ones (appleseed.studio, then appleseed.renderer) to the more generic ones (appleseed.foundation, then Qt headers, then standard headers).

src/appleseed.studio/mainwindow/droppedmaterialmenu.cpp
```diff
@@ -0,0 +1,74 @@
...
+    const foundation::Vector2d& drop_pos,
+    const std::string& material_name)
+    : m_drop_pos(drop_pos)
+    , m_material_name(material_name)
+{
+    QAction* front = addAction("Front side");
+    QAction* back = addAction("Back side");
+    QAction* both = addAction("Both sides");
+
+    connect(front, SIGNAL(triggered()),
```
dictoon: Write on one line. Same for the other two statements. Then you no longer need to separate with blank lines, they really belong together.

src/appleseed.studio/mainwindow/droppedmaterialmenu.cpp
```diff
@@ -0,0 +1,74 @@
...
+namespace appleseed {
+namespace studio {
+
+DroppedMaterialMenu::DroppedMaterialMenu(
+    const foundation::Vector2d& drop_pos,
+    const std::string& material_name)
+    : m_drop_pos(drop_pos)
+    , m_material_name(material_name)
+{
+    QAction* front = addAction("Front side");
```
dictoon: Use title capitalization in UI items (https://capitalizemytitle.com/).

src/appleseed.studio/mainwindow/droppedmaterialmenu.cpp
```diff
@@ -0,0 +1,74 @@
...
+// Interface header.
+#include "mainwindow/droppedmaterialmenu.h"
+
+namespace appleseed {
+namespace studio {
+
+DroppedMaterialMenu::DroppedMaterialMenu(
+    const foundation::Vector2d& drop_pos,
+    const std::string& material_name)
+    : m_drop_pos(drop_pos)
```
dictoon: Half-indent (only two spaces).

src/appleseed.studio/mainwindow/droppedmaterialmenu.cpp
```diff
@@ -0,0 +1,74 @@
...
+//
+
+// Interface header.
+#include "mainwindow/droppedmaterialmenu.h"
+
+namespace appleseed {
+namespace studio {
+
+DroppedMaterialMenu::DroppedMaterialMenu(
+    const foundation::Vector2d& drop_pos,
```
dictoon: Align args vertically.

src/appleseed.studio/mainwindow/droppedmaterialmenu.h
```diff
@@ -0,0 +1,75 @@
...
+namespace appleseed {
+namespace studio {
+
+class DroppedMaterialMenu
+    : public QMenu
+{
+    Q_OBJECT
+
+public:
+    // constructor
```
dictoon: Start full line comments with a capital and end them with a period (`.`). Like a normal sentence :)

src/appleseed.studio/mainwindow/droppedmaterialmenu.h
```diff
@@ -0,0 +1,75 @@
...
+    DroppedMaterialMenu(const foundation::Vector2d& drop_pos, const std::string& material_name);
+
+    // destructor
+    ~DroppedMaterialMenu() override = default;
```
dictoon: Overriding it to use the default implementation is not useful, you can completely remove the destructor here. The base class' one will do the job just fine.

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
```diff
@@ -490,6 +495,105 @@ void RenderingManager::slot_camera_change_end()
 {
 }
 
+void RenderingManager::slot_material_dropped(const foundation::Vector2d& drop_pos, const std::string& material_name)
+{
+    if (!m_project->has_trace_context())
+    {
+        RENDERER_LOG_INFO("the scene must be rendering or must have been rendered at least once for drag&drop to be available.");
```
dictoon: "drag and drop".

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
```diff
@@ -490,6 +495,105 @@ void RenderingManager::slot_camera_change_end()
 {
 }
 
+void RenderingManager::slot_material_dropped(const foundation::Vector2d& drop_pos, const std::string& material_name)
+{
+    if (!m_project->has_trace_context())
```
dictoon: Since you're doing this check here, it seems to me that the other check is really redundant.

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
```diff
@@ -490,6 +495,105 @@ void RenderingManager::slot_camera_change_end()
...
+    {
+        RENDERER_LOG_INFO("the scene must be rendering or must have been rendered at least once for drag&drop to be available.");
+        return;
+    }
+
+    DroppedMaterialMenu* menu = new DroppedMaterialMenu(drop_pos, material_name);
+
+    connect(menu,
```
dictoon: Nice formatting :)

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
```diff
@@ -490,6 +495,105 @@ void RenderingManager::slot_camera_change_end()
...
 
+void RenderingManager::slot_material_dropped(const foundation::Vector2d& drop_pos, const std::string& material_name)
+{
+    if (!m_project->has_trace_context())
+    {
+        RENDERER_LOG_INFO("the scene must be rendering or must have been rendered at least once for drag&drop to be available.");
+        return;
+    }
+
+    DroppedMaterialMenu* menu = new DroppedMaterialMenu(drop_pos, material_name);
```
dictoon: Do you confirm that the menu is destroyed automatically?

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
```diff
@@ -490,6 +495,105 @@ void RenderingManager::slot_camera_change_end()
...
+            renderer::ObjectInstance::Side)));
+
+    menu->exec(QCursor::pos());
+}
+
+namespace
+{
+    struct MaterialSlot
+    {
+        std::string name;
```
dictoon: Prefix class and struct data members with `m_`.

Align members vertically.

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
```diff
@@ -490,6 +495,105 @@ void RenderingManager::slot_camera_change_end()
...
+        ObjectInstance::Side side;
+    };
+
+    class ApplyDroppedMaterial
+        : public RenderingManager::IScheduledAction
```
dictoon: Half-indent etc.

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
```diff
@@ -490,6 +495,105 @@ void RenderingManager::slot_camera_change_end()
...
+            const std::string& material_name)
+            : m_instance(instance),
+            m_material_slots(material_slots),
+            m_material_name(material_name)
+        {
+        }
+
+        void operator()(Project& project) override
+        {
+            for (const_each<std::vector<MaterialSlot>> slot = m_material_slots; slot; ++slot)
```
dictoon: Use modern range-based for loop. `foundation::const_each<>` is deprecated, we're in the process of converting them all to their modern counterparts (you can help too!).

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
```diff
@@ -490,6 +495,105 @@ void RenderingManager::slot_camera_change_end()
...
+        std::string name;
+        ObjectInstance::Side side;
+    };
+
+    class ApplyDroppedMaterial
+        : public RenderingManager::IScheduledAction
+    {
+    public:
+        explicit ApplyDroppedMaterial(
+            ObjectInstance& instance,
```
dictoon: Align vertically.

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
```diff
@@ -490,6 +495,105 @@ void RenderingManager::slot_camera_change_end()
...
+    {
+        std::string name;
+        ObjectInstance::Side side;
+    };
+
+    class ApplyDroppedMaterial
+        : public RenderingManager::IScheduledAction
+    {
+    public:
+        explicit ApplyDroppedMaterial(
```
dictoon: While it's not incorrect, we typically don't mark multi-argument constructors as explicit.

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
```diff
@@ -490,6 +495,105 @@ void RenderingManager::slot_camera_change_end()
...
+}
+
+void RenderingManager::slot_change_material(
+    const foundation::Vector2d& drop_pos,
+    const std::string& material_name,
+    renderer::ObjectInstance::Side side)
+{
+    const ScenePicker scene_picker(*m_project);
+    const ScenePicker::PickingResult result = scene_picker.pick(drop_pos);
+    renderer::ObjectInstance* instance = const_cast<renderer::ObjectInstance*>(result.m_object_instance);
```
dictoon: I know we discussed that earlier, but I would prefer that you make `ScenePicker::PickingResult` members non-const instead.

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
```diff
@@ -490,6 +495,105 @@ void RenderingManager::slot_camera_change_end()
...
+
+    std::vector<MaterialSlot> material_slots;
+    if (side == ObjectInstance::Side::FrontSide || side == ObjectInstance::Side::BothSides)
+    {
+        for (auto item = instance->get_front_material_mappings().begin();
+            item != instance->get_front_material_mappings().end();
+            ++item)
+        {
+            material_slots.push_back({ item.key(), ObjectInstance::Side::FrontSide });
```
dictoon: Surprised that this compiles with VS 2012, good news in any case.

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
```diff
@@ -490,6 +495,105 @@ void RenderingManager::slot_camera_change_end()
...
+    {
+        for (auto item = instance->get_back_material_mappings().begin();
+            item != instance->get_back_material_mappings().end();
+            ++item)
+        {
+            material_slots.push_back({ item.key(), ObjectInstance::Side::BackSide });
+        }
+    }
+
+    schedule_or_execute(unique_ptr<RenderingManager::IScheduledAction>(
```
dictoon: Our formatting style tends to be more... staged:

    schedule_or_execute(
        unique_ptr<RenderingManager::IScheduledAction>(
            ...));

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
```diff
@@ -490,6 +495,105 @@ void RenderingManager::slot_camera_change_end()
...
+        new ApplyDroppedMaterial(
+            *instance,
+            material_slots,
+            material_name)));
+
+    if (is_rendering())
```
dictoon: Redundant I think, this is already done by `schedule_or_execute()`.

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
```diff
@@ -490,6 +495,105 @@ void RenderingManager::slot_camera_change_end()
 {
 }
 
+void RenderingManager::slot_material_dropped(const foundation::Vector2d& drop_pos, const std::string& material_name)
```
dictoon: It doesn't seem like this whole code is in the right place in `RenderingManager`. Not sure exactly what would be the best place yet, can you think or something? What about `ObjectInstanceItem` where other material assignments are done?

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
```diff
@@ -490,6 +495,105 @@ void RenderingManager::slot_camera_change_end()
...
+    renderer::ObjectInstance::Side side)
+{
+    const ScenePicker scene_picker(*m_project);
+    const ScenePicker::PickingResult result = scene_picker.pick(drop_pos);
+    renderer::ObjectInstance* instance = const_cast<renderer::ObjectInstance*>(result.m_object_instance);
+
+    std::vector<MaterialSlot> material_slots;
+    if (side == ObjectInstance::Side::FrontSide || side == ObjectInstance::Side::BothSides)
+    {
+        for (auto item = instance->get_front_material_mappings().begin();
```
dictoon: You're currently assigning the material to all slots of the objects (i.e. to the whole object), that's a good first step but I'm wondering if we could only assign the material to the slot used by the part of the object the material was dropped on.

src/appleseed.studio/mainwindow/rendering/renderwidget.cpp
Uedaki: This checks are to be sure the user can only drop the material inside the scene. The default behavior always start when you enter the area but have not the behaviour wanted when it leaves the area, so this "if" is here to fix this.
I just checked, and yes, it has an undefined behavior when you leaves the area by the top or left side so I will add more checks.

src/appleseed.studio/mainwindow/project/itembase.cpp
Uedaki: I use the QTreeWidgetItem flags to know if I the item can be dragged, but seeing it again make me wonder if a boolean in ItemBase would have been simplier
(this are the default flag without Qt::ItemIsDragEnabled that I had in MaterialItem)

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
Uedaki: ScenePicker doesnt have any data on the slot used by the part the material was dropped on but I can try to look if it is possible to add it
We could also do the same for the side selection if needed

src/appleseed.studio/mainwindow/project/itembase.cpp
dictoon: My point is, if these are the default flags, why are you setting them?

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
dictoon: Agreed. Both improvements can be done in another PR.

src/appleseed.studio/mainwindow/project/itembase.cpp
Uedaki: Default flags:  Qt::ItemIsSelectable | Qt::ItemIsUserCheckable | Qt::ItemIsEnabled | Qt::ItemIsDragEnabled | Qt::ItemIsDropEnabled
I remove Qt::ItemIsDragEnabled and Qt::ItemIsDropEnabled for the ItemBase class

dictoon: Ok.

src/appleseed.studio/mainwindow/rendering/renderwidget.cpp
Uedaki: The reason why I check if the trace exist here, it is because it change the way the cursor is visible.
If no trace exist, the icon will stay as "action non available" and the drag and drop will no work.
If you prefer that we let the user drop the item even if it is not available, I will change it

dictoon: I think it would be cleaner, also it would allow to print a log message explaining why drag and drop is not available (versus just changing the icon right now). What do you think?

Uedaki: I like the idea to have a visual feedback which say that it is not a legal action but it is true that it doesnt explain why it is illegal. I dont really have a opinion on this, it is why both check have beeb done

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
Uedaki: I took for example materialassigmenteditorwindow.cpp L367
`    m_editor_context.m_rendering_manager.schedule_or_execute(
        unique_ptr<RenderingManager::IScheduledAction>(
            new AssignMaterialsAction(
                m_object_instance,
                m_object_instance_item,
                slot_values)));

    if (old_front_mappings != m_object_instance.get_front_material_mappings() ||
        old_back_mappings != m_object_instance.get_back_material_mappings())
        m_editor_context.m_project_builder.slot_notify_project_modification();

    if (m_editor_context.m_rendering_manager.is_rendering())
        m_editor_context.m_rendering_manager.reinitialize_rendering();`
Do you want that I change this one too?

src/appleseed.studio/mainwindow/rendering/renderwidget.cpp
Uedaki: I dont understand what you mean, there is no implicit conversion between QPoint and vector2d

src/appleseed.studio/mainwindow/project/itembase.cpp
Uedaki: So do you prefer a boolean like m_isDragEnabled or we keep using the intern flags of QTreeWidgetItem?

src/appleseed.studio/mainwindow/project/projectexplorer.cpp
Uedaki: I dont know, I followed the example provided by qt but there is nothing provided about the memory management of QDrag
And it seems to be the same for the menu (droppedmaterialmenu)

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
Uedaki: do I need to change something for this one? I am not sure

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
Uedaki: I just made a test project and it is not
we could define menu as a static variable to prevent appleseed from building one each time you drop an item.

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
Uedaki: ScenePicker retreive m_object_instance from shadingpoint.cpp L164
`// Retrieve the object instance.
    m_object_instance = m_assembly->object_instances().get_by_index(m_object_instance_index);
    assert(m_object_instance);`
and I didnt find a way to make this line retreive a non-const object_instance

src/appleseed.studio/mainwindow/rendering/renderwidget.cpp
dictoon: The move _is_ legal, it's just not possible for a technical reason that might disappear one day. It's also not clear from a user's point of view why it's sometimes legal and sometimes illegal. I'm in favor of removing this check, simplifying the code accordingly and printing a clear error message when the move ends up being impossible.

src/appleseed.studio/mainwindow/rendering/renderwidget.cpp
dictoon: What I mean is that you don't need to create a temporary and assign it to your variable, you can simply construct the `Vector2d` directly.

src/appleseed.studio/mainwindow/project/itembase.cpp
dictoon: I'm fine with the internal flags.

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
dictoon: Nope :)

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
dictoon: Maybe it's not destroyed automatically because you're not passing a parent.

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
dictoon: I see. I guess we need to turn `ShadingPoint::m_object_instance` into a mutable pointer then.

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
dictoon: Yeah, that was a mistake, it seems that we need to fix that too. Please make sure it continues to work afterward (i.e. open the material assignment window and assign materials _during_ rendering).

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
Uedaki: one way would be to create the menu in RenderWidget and connect signal_apply_material to a objectInstanceItem slot slot_apply_material

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
Uedaki: I did some test with a dummy project but it keep crashing but it seems to work fine on appleseed
I added this 2 lines right after each signal_apply_material in droppedmaterialmenu.cpp
`destroy();`
`delete this;`
Do you have idea to test the stability of this solution?

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
Uedaki: It is really weird, it seems assign material apply the material 2 times, when you select the material and when you apply it
can you confirm it does the same on your version?

Uedaki: it seems it is the wanted behavior

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
dictoon: I don't think you want to do any of this. Just passing a parent to `DroppedMaterialMenu` and eventually to `QMenu` should fix any ownership issue.

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
dictoon: Yes, it's the desired behavior, the first assignment is caused by this code:

```
void MaterialAssignmentEditorWindow::slot_entity_browser_accept(
    QLineEdit*          line_edit,
    QString             page_name,
    QString             entity_name)
{
    line_edit->setText(entity_name);

	// HERE
    assign_materials(get_slot_values());

    qobject_cast<QWidget*>(sender()->parent())->close();
}
```

src/appleseed.studio/mainwindow/rendering/renderwidget.h
```diff
@@ -50,10 +51,15 @@ namespace OCIO = OCIO_NAMESPACE;
 // Standard headers.
 #include <cstddef>
 #include <memory>
+#include <string>
 
 // Forward declarations.
 namespace foundation    { class CanvasProperties; }
 namespace renderer      { class Frame; }
+namespace renderer      { class Project; }
```
dictoon: You no longer need `Project` anywhere in the render widget, right?

src/appleseed.studio/mainwindow/rendering/renderwidget.h
```diff
@@ -113,10 +120,17 @@ class RenderWidget
     QMutex& mutex();
     QImage& image();
 
+  signals:
+    void signal_material_dropped(
+        const foundation::Vector2d& drop_pos,
```
dictoon: Align arguments vertically with their neighbors.

src/appleseed.studio/mainwindow/rendering/renderwidget.h
```diff
@@ -33,9 +33,10 @@
 #include "mainwindow/rendering/renderclipboardhandler.h"
 
 // appleseed.foundation headers.
-#include "foundation/platform/compiler.h"
 #include "foundation/image/image.h"
 #include "foundation/image/tile.h"
+#include "foundation/math/vector.h"
+#include "foundation/platform/compiler.h"
```
dictoon: I think we can safely get rid of `compiler.h`.

src/appleseed.studio/mainwindow/rendering/renderwidget.cpp
```diff
@@ -422,5 +430,31 @@ void RenderWidget::paintEvent(QPaintEvent* event)
...
 
+void RenderWidget::dragEnterEvent(QDragEnterEvent* event)
+{
+    if (event->mimeData()->hasFormat("text/plain"))
+        event->acceptProposedAction();
+}
+
+void RenderWidget::dragMoveEvent(QDragMoveEvent* event)
+{
+    if (pos().x() <= event->pos().x() && pos().y() <= event->pos().y()
```
dictoon: Formatting proposal for improved readability:

    if (pos().x() <= event->pos().x() &&
        pos().y() <= event->pos().y() &&
        event->pos().x() < pos().x() + width() &&
        event->pos().y() < pos().y() + height())

src/appleseed.studio/mainwindow/rendering/renderwidget.cpp
```diff
@@ -422,5 +430,31 @@ void RenderWidget::paintEvent(QPaintEvent* event)
...
+        && event->pos().x() < pos().x() + width() && event->pos().y() < pos().y() + height())
+    {
```
dictoon: Remove braces for single-line `if` body.

src/appleseed.studio/mainwindow/rendering/renderwidget.cpp
```diff
@@ -32,6 +32,7 @@
 
 // appleseed.renderer headers.
 #include "renderer/api/frame.h"
+#include "renderer/api/project.h"
```
dictoon: Remove.

src/appleseed.studio/mainwindow/rendering/renderwidget.cpp
```diff
@@ -62,11 +66,13 @@ namespace studio {
 //
 
 RenderWidget::RenderWidget(
+    const Project&          project,
```
dictoon: Remove.

src/appleseed.studio/mainwindow/droppedmaterialmenu.cpp
```diff
@@ -0,0 +1,74 @@
...
+    const std::string&          material_name)
+  : m_drop_pos(drop_pos)
+  , m_material_name(material_name)
+{
+    connect(addAction("Front Side"), SIGNAL(triggered()),
```
dictoon: For improved readability, write these statements each on a single line.

src/appleseed.studio/mainwindow/droppedmaterialmenu.h
```diff
@@ -0,0 +1,75 @@
+
+//
+// This source file is part of appleseed.
+// Visit https://appleseedhq.net/ for additional information and resources.
+//
+// This software is released under the MIT license.
+//
+// Copyright (c) 2019 Thibault Vergne, The appleseedhq Organization
+// Copyright (c) 2014-2018 Thibault Vergne, The appleseedhq Organization
```
dictoon: Remove that second copyright line.

src/appleseed.studio/mainwindow/droppedmaterialmenu.cpp
```diff
@@ -0,0 +1,74 @@
+
+//
+// This source file is part of appleseed.
+// Visit https://appleseedhq.net/ for additional information and resources.
+//
+// This software is released under the MIT license.
+//
+// Copyright (c) 2019 Thibault Vergne, The appleseedhq Organization
+// Copyright (c) 2014-2018 Thibault Vergne, The appleseedhq Organization
```
dictoon: Remove that second copyright line.

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
Uedaki: I tried with passing the mainwindow as a parent, but it doesnt clean the menu
If you pass a parent, it will only clean the widget when the parent will be destroy:

> Although a popup menu is always a top-level widget, if a parent is passed the popup menu will be deleted when that parent is destroyed

src/appleseed.studio/mainwindow/rendering/renderwidget.h
Uedaki: It s still here because I am still thinking for a way to remove the assign_material from renderingmanager and I will need project to implement it in RenderWidget

src/appleseed.studio/mainwindow/rendering/renderwidget.h
Uedaki: ok, I just moved it so it is order alphabetical order

src/appleseed.studio/mainwindow/rendering/renderwidget.cpp
Uedaki: this part is not finish, if I leave the scene by the left side, the icon change to show that drop is not available but if you continue it come back to the drop-available icon.
I need to understand why and fix it before closing this request

Uedaki: to give a little idea, inside the scene x change between 0 and 700, and when I live from the left (so x was 0), x become 700 again and decrement when you go futher to the left
I dont really see a way to fix this...

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
dictoon: Some hints about possible solutions here:
https://stackoverflow.com/questions/39259862/when-and-how-to-properly-destroy-qmenu-context-menu

Make sure to check out both solutions and read all comments of the first solution.

I kind of like the second solution.

src/appleseed.studio/mainwindow/rendering/renderwidget.h
dictoon: We don't want the code for material assignment in `RenderWidget` either. Render widget is (mostly) responsible for show render results, and that should be pretty much it. We don't want to add anything more to it. It's the Single Responsibility Principle: https://www.google.com/search?q=Single+Responsability+Principle&oq=Single+Responsability+Principle

src/appleseed.studio/mainwindow/rendering/renderwidget.cpp
dictoon: That's weird. There must be a way. Did you check the docs carefully? What about posting a question on Stack Overflow or on Qt Centre?

src/appleseed.studio/mainwindow/droppedmaterialmenu.cpp
```diff
@@ -0,0 +1,70 @@
...
+{
+    connect(addAction("Front Side"), SIGNAL(triggered()), SLOT(slot_apply_to_front()));
+    connect(addAction("Back Side"), SIGNAL(triggered()), SLOT(slot_apply_to_back()));
+    connect(addAction("Both Sides"), SIGNAL(triggered()), SLOT(slot_apply_to_both()));
+}
+
+void DroppedMaterialMenu::slot_apply_to_front()
+{
+    emit signal_apply_material(m_drop_pos, m_material_name, renderer::ObjectInstance::Side::FrontSide);
+    destroy();
```
dictoon: A cleaner solution would have been to call `deleteLater()` (to be checked if that's really a proper solution). Unfortunately it doesn't cover the case where the user decides not to click on any menu item.

src/appleseed.studio/mainwindow/rendering/renderwidget.cpp
dictoon: Also, did you check carefully how `QDragMoveEvent` works? https://doc.qt.io/archives/qt-4.8/qdragmoveevent.html#details

src/appleseed.studio/mainwindow/droppedmaterialmenu.h
```diff
@@ -0,0 +1,74 @@
...
+
+  private:
+    const foundation::Vector2d  m_drop_pos;
+    const std::string           m_material_name;
+
+  private slots:
+    void slot_apply_to_front();
+    void slot_apply_to_back();
+    void slot_apply_to_both();
+
```
dictoon: Remove blank line.

src/appleseed.studio/mainwindow/droppedmaterialmenu.cpp
Uedaki: you can still use the signal aboutToHide in case the user didnt click on a menu item

dictoon: Indeed, as suggested on that Stack Overflow post. Maybe `aboutToHide()` + `deleteLater()` is a good enough solution?

src/appleseed.studio/mainwindow/rendering/renderwidget.cpp
Uedaki: I tried accept(const QRect & rectangle) but it is worth
I dont understand why the drag and drop behave like this

src/appleseed.studio/mainwindow/project/projectexplorer.cpp
```diff
@@ -285,5 +287,10 @@ void ProjectExplorer::slot_delete_items()
     items.first()->delete_multiple(items);
 }
 
+const MaterialDropHandler* ProjectExplorer::get_material_drop_handler() const
+{
+    return (&*m_material_drop_handler);
```
dictoon: Or simpler:

    return m_material_drop_handler.get();

src/appleseed.studio/mainwindow/project/projectexplorer.h
```diff
@@ -84,17 +85,20 @@ class ProjectExplorer
...
-    ItemRegistry                m_item_registry;
-    EntityEditorContext         m_editor_context;
-    std::unique_ptr<QShortcut>  m_delete_shortcut;
+    QTreeWidget*                            m_tree_widget;
+    AttributeEditor*                        m_attribute_editor;
+    ProjectBuilder                          m_project_builder;
+    ItemRegistry                            m_item_registry;
+    EntityEditorContext                     m_editor_context;
+    std::unique_ptr<QShortcut>              m_delete_shortcut;
+    std::shared_ptr<MaterialDropHandler>    m_material_drop_handler;
```
dictoon: Do we need a `std::shared_ptr<>` here? Or would a `std::unique_ptr<>` be enough?

src/appleseed.studio/mainwindow/project/projectexplorer.cpp
```diff
@@ -115,6 +115,8 @@ ProjectExplorer::ProjectExplorer(
     connect(
         &m_project_builder, SIGNAL(signal_frame_modified()),
         SIGNAL(signal_frame_modified()));
+
+    m_material_drop_handler.reset(new MaterialDropHandler(project, rendering_manager));
```
dictoon: In my idea, `MaterialDropHandler` was created by `RenderTab`, just like all other handlers. That would simplify things even more.

If that means you'll need to pass the `RenderingManager` to `RenderTab`, then fine.

src/appleseed.studio/mainwindow/rendering/materialdrophandler.cpp
```diff
@@ -0,0 +1,159 @@
+
+//
+// This source file is part of appleseed.
+// Visit https://appleseedhq.net/ for additional information and resources.
+//
+// This software is released under the MIT license.
+//
+// Copyright (c) 2010-2013 Francois Beaune, Jupiter Jazz Limited
```
dictoon: Just like in the header file, replace these two lines with

    // Copyright (c) 2019 Thibault Vergne, The appleseedhq Organization

src/appleseed.studio/CMakeLists.txt
```diff
@@ -288,6 +290,8 @@ set (mainwindow_sources
     mainwindow/configurationmanagerwindow.cpp
     mainwindow/configurationmanagerwindow.h
     mainwindow/configurationmanagerwindow.ui
+    mainwindow/droppedmaterialmenu.cpp
```
dictoon: Maybe we should rename these files to `MaterialDropMenu` to match `MaterialDropHandler` and make it clearer that these components are related. What do you think?

src/appleseed.studio/mainwindow/rendering/renderwidget.h
```diff
@@ -113,6 +117,11 @@ class RenderWidget
     QMutex& mutex();
     QImage& image();
 
+  signals:
+    void signal_material_dropped(
+        const foundation::Vector2d& drop_pos,
+        const std::string&          material_name);
```
dictoon: Actually it's probably better to use `QString` in slots and signals.

src/appleseed.studio/mainwindow/rendering/renderingmanager.h
```diff
@@ -41,6 +41,7 @@
 
 // appleseed.foundation headers.
 #include "foundation/math/transform.h"
+#include "foundation/math/vector.h"
```
dictoon: Probably not needed anymore.

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
```diff
@@ -31,6 +31,7 @@
 #include "renderingmanager.h"
 
 // appleseed.studio headers.
+#include "mainwindow/droppedmaterialmenu.h"
```
dictoon: Probably not needed anymore.

src/appleseed.studio/mainwindow/rendering/materialdrophandler.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+void MaterialDropHandler::assign_material(const renderer::ObjectInstance::Side side)
+{
+    const renderer::ScenePicker scene_picker(m_project);
+    const renderer::ScenePicker::PickingResult result = scene_picker.pick(m_drop_pos);
+    renderer::ObjectInstance* instance = const_cast<renderer::ObjectInstance*>(result.m_object_instance);
+
+    std::vector<MaterialSlot> material_slots;
+    if (side == renderer::ObjectInstance::Side::FrontSide
+        || side == renderer::ObjectInstance::Side::BothSides)
```
dictoon: You can improve readability by moving the `||` at the end of the previous line:

    if (side == renderer::ObjectInstance::Side::FrontSide ||
        side == renderer::ObjectInstance::Side::BothSides)

However, you'll notice that `renderer::ObjectInstance::Side` is actually a bitmask, so you can simply do something like

    if (side & renderer::ObjectInstance::Side::FrontSide)

And the same thing for the backside.

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
Uedaki: yeah, was just waiting for confirmation before removing completly droppedmaterialmenu

src/appleseed.studio/CMakeLists.txt
Uedaki: DroppedMaterialMenu is no longer use, it was created to keep the mouse position and the material name when the menu is open but with the new handler, I no longer need it as the information are stocked in the handler

src/appleseed.studio/mainwindow/project/projectexplorer.h
```diff
@@ -89,12 +89,12 @@ class ProjectExplorer
     void signal_frame_modified() const;
 
   private:
-    QTreeWidget*                m_tree_widget;
-    AttributeEditor*            m_attribute_editor;
-    ProjectBuilder              m_project_builder;
-    ItemRegistry                m_item_registry;
-    EntityEditorContext         m_editor_context;
-    std::unique_ptr<QShortcut>  m_delete_shortcut;
+    QTreeWidget*                            m_tree_widget;
```
dictoon: Revert these changes now that you're not adding anything anymore.

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
```diff
@@ -51,7 +51,6 @@
 #include "foundation/image/image.h"
 #include "foundation/platform/defaulttimers.h"
 #include "foundation/platform/types.h"
-#include "foundation/utility/foreach.h"
```
dictoon: Revert, still needed.

src/appleseed.studio/mainwindow/rendering/rendertab.h
```diff
@@ -86,6 +87,8 @@ class RenderTab
     CameraController* get_camera_controller() const;
     ScenePickingHandler* get_scene_picking_handler() const;
 
+    void set_material_drop_handler(MaterialDropHandler *material_drop_handler);
```
dictoon: Ownership of the handler is transfered here but it's not clear from the API. Take a `std::unique_ptr<>` instead.

src/appleseed.studio/mainwindow/rendering/rendertab.cpp
```diff
@@ -32,6 +32,7 @@
 
 // appleseed.studio headers.
 #include "mainwindow/project/projectexplorer.h"
+#include "mainwindow/rendering/materialdrophandler.h"
```
dictoon: Remove, already included in `rendertab.h`.

src/appleseed.studio/mainwindow/rendering/rendertab.cpp
```diff
@@ -111,6 +112,21 @@ ScenePickingHandler* RenderTab::get_scene_picking_handler() const
     return m_scene_picking_handler.get();
 }
 
+void RenderTab::set_material_drop_handler(MaterialDropHandler *handler)
+{
+    m_material_drop_handler.reset(handler);
+    if (handler && m_render_widget)
```
dictoon: Add braces for multiline `if` body (even if it's a single statement).

src/appleseed.studio/mainwindow/rendering/renderwidget.h
```diff
@@ -50,10 +50,14 @@ namespace OCIO = OCIO_NAMESPACE;
 // Standard headers.
 #include <cstddef>
 #include <memory>
+#include <string>
 
 // Forward declarations.
 namespace foundation    { class CanvasProperties; }
 namespace renderer      { class Frame; }
+
```
dictoon: Remove blank line.

dictoon: You're missing a fwd decl for `QDragMoveEvent`.

src/appleseed/renderer/kernel/shading/shadingpoint.h
```diff
@@ -221,6 +221,7 @@ class ShadingPoint
     const Assembly& get_assembly() const;
 
     // Return the object instance that was hit.
+    ObjectInstance& get_object_instance();
```
dictoon: Can't you just change the existing `get_object_instance()` method to return a non-const ref but keeping the method itself const?

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -979,6 +980,12 @@ void MainWindow::add_render_tab(const QString& label)
             *m_project_manager.get_project(),
             m_ocio_config);
 
+    // Set material drop handler
+    render_tab->set_material_drop_handler(
```
dictoon: It would be cleaner if this was done inside `RenderTab`'s constructor, like it's done for all other handlers. Pass the rendering manager if needed.

Uedaki:
I posted it but I dont think I will have a answer, from what I have tested, the problem come from the fact MainWindow is drop enabled. When your leave RenderWidget, Qt still think you are inside RenderWidget but QDragMoveEvent is now relative to MainWindow, which is why my check does not fully work.
This bug will not break the software but it would be nice to fix it.

dictoon:
Alright, let's merge this PR as is. We'll think of a fix.

dictoon:
Thanks for your contribution!

PR #2443, opened 2019-02-26
Add an alpha channel to the DepthAOV

achalpandeyy:
When we don't hit an object, we set the depth to 0 and alpha to 0. This allows us to use depth = 0 instead of numeric_limits::max() which some applications don't like.

### DIFF
--- a/src/appleseed/renderer/modeling/aov/depthaov.cpp
+++ b/src/appleseed/renderer/modeling/aov/depthaov.cpp
@@ -94,8 +94,7 @@ namespace
             if (shading_point.hit_surface())
             {
                 shading_result.m_aovs[0].a = 1.0f;
-                const float depth = static_cast<float>(shading_point.get_distance());
-                *out = depth;
+                *out = static_cast<float>(shading_point.get_distance());
             }
             else
             {
@@ -144,7 +143,7 @@ namespace
 
         void clear_image() override
         {
-            m_image->clear(Color<float, 1>(numeric_limits<float>::max()));
+            m_image->clear(Color<float, 1>(0.0f));
         }
 
       protected:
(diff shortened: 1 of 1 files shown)


src/appleseed/renderer/modeling/aov/depthaov.cpp
```diff
@@ -91,12 +91,17 @@ namespace
...
-            const float depth =
-                shading_point.hit_surface()
-                    ? static_cast<float>(shading_point.get_distance())
-                    : numeric_limits<float>::max();
-
-            *out = depth;
+            if (shading_point.hit_surface())
+            {
+                shading_result.m_aovs[0].a = 1.0f;
+                const float depth = static_cast<float>(shading_point.get_distance());
```
dictoon: You can probably get rid of the `depth` variable here.

est77:
Thanks for the PR. It looks pretty good. 
There is one thing missing: you need to modify DepthAOV::clear_image().
After that, we can merge.

src/appleseed/renderer/modeling/aov/depthaov.cpp
```diff
@@ -128,18 +132,18 @@ namespace
...
         {
-            static const char* ChannelNames[] = { "Z" };
+            static const char* ChannelNames[] = { "Z", "A" };
             return ChannelNames;
         }
 
         void clear_image() override
         {
-            m_image->clear(Color<float, 1>(numeric_limits<float>::max()));
+            m_image->clear(Color<float, 1>(0.0f));
```
dictoon: You need to clear with a two-channel color here.

dictoon: `Color<float, 2>(0.0f)` should work.

achalpandeyy: what should be the default values for each of them?

src/appleseed/renderer/modeling/aov/depthaov.cpp
```diff
@@ -144,7 +143,7 @@ namespace
...
```
est77: You need to use Color<float, 2> since the image has two channels, and pass two values to it.

achalpandeyy:
Thank you for the merge!

dictoon:
Thanks for your contribution.

PR #2445, opened 2019-03-01
Support write all formats images using OIIO

slongle:
https://github.com/appleseedhq/appleseed/issues/2411
Also I fix the bug at GenericImageFileWriter::write_scanlines, so I think maybe some bug caused by invoking this wrong function originally will be fixed.


sandbox/settings/appleseed.studio.xml
```diff
@@ -10,6 +10,12 @@
         <parameter name="tile_ordering" value="spiral" />
     </parameters>
     <parameters name="ui">
+        <parameters name="file_dialog">
```
dictoon: Don't commit changes to this file.

src/appleseed.studio/main/main.cpp
```diff
@@ -76,7 +76,7 @@ using namespace foundation;
 using namespace std;
 namespace bf = boost::filesystem;
 
-namespace
+namespace
```
dictoon: Revert.

src/appleseed/foundation/meta/tests/test_stlallocatortestbed.cpp
```diff
@@ -1,1395 +1,1395 @@
...
-
-    SET_DEFAULT_CONSTRUCTABLE_OFF(AlignedAllocator);
-
-    TEST_CASE(AlignedAllocator)
-    {
-        AlignedAllocator<int> allocator(32);
-        TestAlloc(allocator);
-    }
-}
+﻿
```
dictoon: Revert changes to this file, it's a third party source file.

src/appleseed/foundation/image/genericimagefilewriter.cpp
```diff
@@ -419,7 +419,7 @@ void GenericImageFileWriter::write_scanlines(const size_t image_index)
         }
 
         // Compute y dimensional scanline border.
-        const size_t y_begin = tile_y;
+        const size_t y_begin = tile_y * props.m_tile_height;
```
dictoon: That looked like a serious bug. This function wasn't used anywhere until now?

dictoon: I can't comment there, but there appears to be another bug on line 388:

    std::unique_ptr<uint8> buffer(new uint8[props.m_canvas_width * props.m_tile_height * props.m_pixel_size]);

should be

    std::unique_ptr<uint8[]> buffer(new uint8[props.m_canvas_width * props.m_tile_height * props.m_pixel_size]);

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -1071,29 +1027,31 @@ namespace
...
-            {
-                write_exr_image(
-                    bf_file_path,
-                    image,
-                    image_attributes);
+            Image transformed_image(image);
+
+            // todo: we may want to be more specific here.
+            if (image.properties().m_channel_count == 4)
+                transform_to_srgb(transformed_image);
```
dictoon: That's wrong, we don't want to convert images to sRGB when writing OpenEXR (and other high dynamic range) files.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -1071,29 +1027,31 @@ namespace
...
+            create_parent_directories(bf_file_path);
+
+            const std::string filename = bf_file_path.string();
+            GenericImageFileWriter writer(filename.c_str());
+
+            writer.append_image(&transformed_image);
+
+            writer.set_image_output_format(PixelFormat::PixelFormatUInt8);
+
+            if (extension == ".exr") {
```
dictoon: We must account for all HDR file formats here.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -1120,7 +1078,7 @@ namespace
 
 bool Frame::write_main_image(const char* file_path) const
 {
-    assert(file_path);
+    assert(file_path);
```
dictoon: Revert.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -1071,29 +1027,31 @@ namespace
...
+                transform_to_srgb(transformed_image);
+
+            create_parent_directories(bf_file_path);
+
+            const std::string filename = bf_file_path.string();
+            GenericImageFileWriter writer(filename.c_str());
+
+            writer.append_image(&transformed_image);
+
+            writer.set_image_output_format(PixelFormat::PixelFormatUInt8);
```
dictoon: Again, that's incorrect for all HDR formats.

slongle: set the exported bit depth as
OpenEXR     .exr                                 32-bit
TIFF             .tiff/.tif                             32-bit
RGBE           .hdr                                 32-bit
BMP            .bmp                                 8-bit
JPEG            .jpg/.jpe/.jpeg/.jif/.jfif/.jfi  8-bit

slongle:
Now we can support 
HDR format:
OpenEXR 32-bit
TIFF 32-bit
RGBE 32-bit
Others:
BMP 8-bit
JPEG 8-bit
PNG 8-bit
But we can't support FPM format now, because OIIO don't support it.
HDR format is set in linear, and others are set in sRGB.

sandbox/settings/appleseed.studio.xml
dictoon: You still need to revert all your changes to this file.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -1010,43 +992,28 @@ namespace
         }
     }
 
-    void transform_to_srgb(Image& image)
+    void transform_to_srgb(Image *image)
```
dictoon: Why these changes? Please revert.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -1071,29 +1038,48 @@ namespace
...
         try
         {
-            if (extension == ".exr")
-            {
-                write_exr_image(
-                    bf_file_path,
-                    image,
-                    image_attributes);
+            bool high_dynamic_range_format = false;
+            if (extension == ".exr" || extension == ".tiff" || extension == ".tif" || extension == ".hdr")
```
dictoon: Even simpler:

    const bool high_dynamic_range_format =
        extension == ".exr" ||
        extension == ".tiff" ||
        extension == ".tif" ||
        extension == ".hdr";

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -1071,29 +1038,48 @@ namespace
...
+                high_dynamic_range_format = true;            
+
+            create_parent_directories(bf_file_path);
+
+            const std::string filename = bf_file_path.string();
+            GenericImageFileWriter writer(filename.c_str());
+
+            // Add image to writer
+            Image *transformed_image;
```
dictoon: You have a memory leak since you're never deleting the transformed image. Please use `std::unique_ptr<>` to avoid this problem.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -1071,29 +1038,48 @@ namespace
...
+            if (!high_dynamic_range_format && image.properties().m_channel_count == 4) {
```
dictoon: Move braces to their own lines.

Make sure to check our coding conventions: https://github.com/appleseedhq/appleseed/wiki/Coding-Philosophy-and-Guidelines

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -1071,29 +1038,48 @@ namespace
...
+                transformed_image = new Image(image);
+                writer.append_image(transformed_image);
+            }
+            else {
                 RENDERER_LOG_ERROR(
-                    "failed to write image file %s for frame \"%s\": unsupported image format.",
-                    bf_file_path.string().c_str(),
-                    frame.get_path().c_str());
-
+                    "failed to write image file %s: unsupport image format.",
```
dictoon: Why did you change this? Please revert.

src/appleseed/foundation/image/image.h
```diff
@@ -80,6 +80,11 @@ class APPLESEED_DLLSYMBOL Image
         const size_t        tile_height,        // tile height, in pixels
         const PixelFormat   pixel_format);
 
+    // Copy image data but allow to tweak the channel_count    
+    Image(
```
dictoon: I understand the need here but this API is too limited, it only allows to _reduce_ the number of channels. In fact your implementation is even more limited since it asserts that the source image has exactly four channels and the destination one has exactly three channels, which is not reflected in any comment here.

`renderer::Tile` offers a more powerful mechanism with "shuffle tables", I would suggest exposing this mechanism in `renderer::Image` as well.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -1010,43 +992,28 @@ namespace
...
-
-        image_attributes.insert("color_space", "sRGB");
-        writer.set_image_attributes(image_attributes);
-
-        writer.write();
-    }
 
+    /*
+     * Export bit depth
+     * OpenEXR .exr        32-bit
```
dictoon: Are you sure we're writing 32-bit OpenEXR files? I'm pretty sure we aren't, we're almost certainly writing 16-bit files instead.

src/appleseed/renderer/modeling/frame/frame.cpp
slongle: Thanks, I'll change it.

src/appleseed/renderer/modeling/frame/frame.cpp
slongle: Sorry about that, I've read it now.
I'll fix it.

src/appleseed/renderer/modeling/frame/frame.cpp
slongle: I change this because you want to automatically handle the different formats using OIIO.
I just put the function `write_png_image` and `write_exr_image` in `'write_image`.

src/appleseed/renderer/modeling/frame/frame.cpp
slongle: Maybe I didn't figure out the bit.
I checked the origin Appleseed's output .exr file, opened it in Photoshop, and I saw
.
And it showed the same picture after I changed the code.

slongle: For the bit, I handle it implicitly, which means OIIO will set the bit for different format automatically.

sandbox/settings/appleseed.studio.xml
slongle: Sorry for that, every time after I run the Appleseed, this file is modified. I will check it carefully.

src/appleseed/renderer/modeling/frame/frame.cpp
slongle: For the function `write_exr_image`, I move it into `write_image`.

src/appleseed/renderer/modeling/frame/frame.cpp
slongle: Oh, I misunderstand the origin code. I will fix this serious problem.

src/appleseed/foundation/image/image.h
slongle: Wow, I get it. I will use that function to do it.

slongle:
The output image's bit problem is still.
I even don't know how to detect any format's bit number now.

slongle:
Now, for HDR output, it can support output 16-bit (half) OpenEXR file, 32-bit (float) TIFF file and 32-bit (float) RGBE file.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -1010,40 +992,43 @@ namespace
...
 
-    void write_png_image(
-        const bf::path&         file_path,
-        const Image&            image,
-        ImageAttributes         image_attributes)
+    /*
+     * Default Export Format
+     * OpenEXR   .exr          4-channel   16-bit (half)    Linear  
+     * TIFF      .tiff/.tif    4-channel   32-bit (float)   Linear
+     * RGBE      .hdr          3-channel   32-bit (float)   Linear
```
dictoon: Incorrect, RGBE is 8-bit integer + 8-bit shared integer exponent.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -1010,40 +992,43 @@ namespace
...
+     * BMP       .bmp          4-channel    8-bit (uint8)     sRGB
+     * PNG       .png          4-channel    8-bit (uint8)     sRGB    
+     * JPEG      .jpg/.jpe/    3-channel    8-bit (uint8)     sRGB
+     *           .jpeg/.jif/          
+     *           .jfif/.jfi           
+     */
+    void write_image(
+        const std::string&         filename,
+        std::unique_ptr<Image>&    image,
```
dictoon: Pass a `const Image&` like it was initially, it makes more sense.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -1010,40 +992,43 @@ namespace
...
+        ImageAttributes            image_attributes,
+        PixelFormat                image_output_format = PixelFormat::PixelFormatDefault)
```
dictoon: In fact, as far as I can tell, you're never using that last argument so `PixelFormat::PixelFormatDefault` is truly not needed.

slongle: No, my expression is not clear.
I want to use it when users can choose what pixelformat they want.

src/appleseed/foundation/image/image.h
```diff
@@ -80,6 +80,14 @@ class APPLESEED_DLLSYMBOL Image
         const size_t        tile_height,        // tile height, in pixels
         const PixelFormat   pixel_format);
 
+    // Construct a image by converting an existing image to a given pixel format,
```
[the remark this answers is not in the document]
slongle: Sorry for typo

src/appleseed/renderer/modeling/frame/frame.cpp
slongle: If you don't want this, I'll change them in origin function.

src/appleseed/foundation/image/pixel.h
```diff
@@ -60,7 +60,8 @@ enum PixelFormat
     PixelFormatUInt32,
     PixelFormatHalf,
     PixelFormatFloat,
-    PixelFormatDouble
+    PixelFormatDouble,
+    PixelFormatDefault
```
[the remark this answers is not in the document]
slongle: I want to explain it.
This default is for OIIO, which means without setting PixelFormat OIIO will handle it default.
But I indeed didn't think about this problem, I'll check them.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -1010,40 +992,43 @@ namespace
         }
     }
 
-    void transform_to_srgb(Image& image)
+    void transform_to_srgb(Image* image)
```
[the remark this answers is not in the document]
slongle: Ok, revert.

src/appleseed/renderer/modeling/frame/frame.cpp
slongle: Thanks, changed.

src/appleseed/renderer/modeling/frame/frame.cpp
slongle: Reverted.

src/appleseed/renderer/modeling/frame/frame.cpp
dictoon: You're still introducing a typo here: *unsupport* -> *unsupported*

slongle: Thanks, done.

slongle:
Thanks for your every review very much, François!
I'll do better on other issues.

dictoon:
Thanks for the fix.

Could you rebase your PR on top of master? GitHub cannot merge it right now due to potential conflicts. Thanks.

Issue #2438, opened 2019-02-22
Print final render time estimations


dictoon:
From user Dark_z on Discord:

> It would be useful if, in the renderer's info, the renderer made an attempt to estimate the time it could take to render the whole frame based on the rolling average time it's already taken to render the current tiles. This will definitely be a rough estimation for sure, but getting an idea of how long a frame will take to render without rendering out the entire frame first is something that would be very beneficial.

achalpandeyy:
Is anybody working on this? I'd like to try it.

dictoon:
Go ahead.

achalpandeyy:
How often should the renderer print estimated render time in the logs?

dictoon:
That's a good question. Maybe every time there's a progress update? What do you think?

achalpandeyy:
What exactly do you mean by a progress update?

How about every time we finish rendering a tile?

achalpandeyy:
One way of approaching this issue-

1. add a member variable to `tilecallbackbase.h` to keep track of how much time it takes for a tile to get rendered

2. start the timer in `on_tile_begin()` and end it on `on_tile_end()` callback functions and store the time elapsed in the member variable

3. take the average of this member variable for all the already rendered tiles in the frame

4. since all the tiles are being rendered in parallel, we can return this average as the estimated time

what do you think?

dictoon:
Done.

PR #2454, opened 2019-03-04
Add support for estimating render time in appleseed.cli

achalpandeyy:
As the first step towards issue #2438.

I initially submitted these changes in #2453  but realized that the indentation was off so I tried deleting the PR a bunch of times which made a big mess.
Also, the travis-ci build failed most probably due to some network issues which only added to the mess.
So I decided to close that PR and open a new one, hope you don't mind!

### DIFF
--- a/src/appleseed.cli/progresstilecallback.cpp
+++ b/src/appleseed.cli/progresstilecallback.cpp
@@ -39,6 +39,7 @@
 #include "foundation/image/image.h"
 #include "foundation/image/tile.h"
 #include "foundation/platform/thread.h"
+#include "foundation/utility/stopwatch.h"
 #include "foundation/utility/string.h"
 
 // Standard headers.
@@ -65,6 +66,7 @@ namespace
         explicit ProgressTileCallback(Logger& logger)
           : m_logger(logger)
           , m_rendered_pixels(0)
+          , m_rendered_tiles(0)
         {
         }
 
@@ -72,6 +74,16 @@ namespace
         {
             // The factory always return the same tile callback instance.
             // Prevent this instance from being destroyed by doing nothing here.
+        }
+
+        void on_tiled_frame_begin(const Frame* frame) override
+        {
+            m_stopwatch.start();
+        }
+
+        void on_tiled_frame_end(const Frame* frame) override
+        {
+            m_stopwatch.clear();
         }
 
         void on_tile_end(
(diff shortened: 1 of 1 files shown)


src/appleseed.cli/progresstilecallback.cpp
```diff
@@ -90,12 +102,28 @@ namespace
...
+            // Keep track of the total number of rendered tiles.
+            m_rendered_tiles++;
+
+            // Retrieve the total number of tiles in the frame.
+            const size_t total_tiles = frame->image().properties().m_tile_count;
+
+            // Estimate remaining render time.
+            m_stopwatch.measure();
+            const double elapsed_time = m_stopwatch.get_seconds();
+            const double estimated_time = (elapsed_time / m_rendered_tiles) * (total_tiles - m_rendered_tiles);
```
dictoon: Maybe `remaining_time` is clearer?

src/appleseed.cli/progresstilecallback.cpp
```diff
@@ -90,12 +102,28 @@ namespace
...
+
+            // Print estimated time.
+            LOG_INFO(m_logger, "remaining time %.3lf seconds", estimated_time);
```
dictoon: `%.3f` would have been enough here. That said, you should use `pretty_time()` which will print time in a more friendly manner than just seconds. Finally, I don't think we need subsecond precision for time estimations (they are very approximate anyway).

I would also prefer that both the progress message and the remaining time are printed on the same line, something like

    rendering, 53% done; about 5 hours, 30 minutes and 21 seconds remaining...

`pretty_time()` will take care of the latter part.

Btw, make sure to call `c_str()` on the string returned by `pretty_time()`.

src/appleseed.cli/progresstilecallback.cpp
```diff
@@ -90,12 +102,28 @@ namespace
...
         }
 
       private:
         Logger&             m_logger;
         boost::mutex        m_mutex;
         size_t              m_rendered_pixels;
+        size_t              m_rendered_tiles;
+        Stopwatch<DefaultWallclockTimer>    m_stopwatch;
```
dictoon: Align members vertically.

You need to include `foundation/platform/defaulttimers.h` for `DefaultWallclockTimer`.

src/appleseed.cli/progresstilecallback.cpp
achalpandeyy: done :)

src/appleseed.cli/progresstilecallback.cpp
achalpandeyy: sorry didn't see it! gonna include it in next commit

achalpandeyy: done

src/appleseed.cli/progresstilecallback.cpp
```diff
@@ -89,13 +102,33 @@ namespace
             const size_t total_pixels = frame->image().properties().m_pixel_count;
 
             // Print a progress message.
-            LOG_INFO(m_logger, "rendering, %s done", pretty_percent(m_rendered_pixels, total_pixels).c_str());
+            //LOG_INFO(m_logger, "rendering, %s done", pretty_percent(m_rendered_pixels, total_pixels).c_str());
```
dictoon: Don't forget to remove dead code.

achalpandeyy: I keep missing things for some reason! This time it is perfect.

achalpandeyy:
Thanks for the merge! :)

Issue #2452, opened 2019-03-03
libSM and libuuid error when building


kushWithoutWax:
Using gcc 4.8.5 on Ubuntu 16.04.5, I faced some issues while building appleseed studio.

 Linking CXX executable appleseed.studio
```bash
**//usr/lib/x86_64-linux-gnu/libSM.so.6: undefined reference to `uuid_generate@UUID_1.0'
//usr/lib/x86_64-linux-gnu/libSM.so.6: undefined reference to `uuid_unparse_lower@UUID_1.0'**
collect2: error: ld returned 1 exit status
src/appleseed.studio/CMakeFiles/appleseed.studio.dir/build.make:8119: recipe for target 'src/appleseed.studio/appleseed.studio' failed
make[2]: *** [src/appleseed.studio/appleseed.studio] Error 1
CMakeFiles/Makefile2:527: recipe for target 'src/appleseed.studio/CMakeFiles/appleseed.studio.dir/all' failed
make[1]: *** [src/appleseed.studio/CMakeFiles/appleseed.studio.dir/all] Error 2
```

I used the prebuilt dependencies package and used cmake with the flags in https://github.com/appleseedhq/appleseed/wiki/Building-appleseed-on-Linux

est77:
This is caused most likely because of using precompiled libs built on Fedora on Ubuntu.
The correct fix is to add -DAPPLESEED_LINK_EXTRA_LIBRARIES=-luuid to the cmake command line 
(or rebuild the dependencies on Ubuntu).

PR #2455, opened 2019-03-05
fixed build error due to libSM and libuuid 

kushWithoutWax:
added a uuid flag to link libraries to fix issue #2452

### DIFF
--- a/src/appleseed.studio/CMakeLists.txt
+++ b/src/appleseed.studio/CMakeLists.txt
@@ -644,6 +644,7 @@ target_link_libraries (appleseed.studio
     ${QT_LIBRARIES}
     ${OPENGL_LIBRARY}
     ${PYTHON_LIBRARIES}
+    uuid
 )
 
 if (WITH_DISNEY_MATERIAL)


src/appleseed.studio/CMakeLists.txt
```diff
@@ -644,6 +644,7 @@ target_link_libraries (appleseed.studio
     ${QT_LIBRARIES}
     ${OPENGL_LIBRARY}
     ${PYTHON_LIBRARIES}
+    uuid
```
est77: This will not work correctly on Windows, and maybe on OSX (I don't know if there is an uuid lib on OSX)

est77:
Thanks for the PR and for taking the time to look into this.
I don't think there is a problem in the build system. The problem is actually using precompiled libs built on Fedora on Ubuntu.

I think that the fix is correct for your specific case, but we should not do this change for everyone.
Instead, we have a cmake variable that you can use: APPLESEED_LINK_EXTRA_LIBRARIES.

You can add -DAPPLESEED_LINK_EXTRA_LIBRARIES=-luuid in your cmake command line and that should probably fix the issue for your build.

Issue #2403, opened 2019-02-01
Use proper color space when computing variations in Adaptive Tile Sampler


dictoon:
We're computing variations in linear RGB right now. Using sRGB would already be an improvement I think.

raviteja469:
@dictoon  I would like to contribute to this. I am new to this repository as well as open source. It would be great if you could help me get started.

wilimitis:
Hi @raviteja469, I'm curious if there's been any progress on this, otherwise I was considering getting started on it now, thanks!

Issue #2224, opened 2018-09-28
Allow increasing sampling quality in dark areas


dictoon:
It would make sense to artificially increase the variance estimation in darker areas of the image. It could be controlled with a setting ("Increase sampling quality in dark areas") that would allow adjusting the level of sampling quality boost in dark areas.

PR #2465, opened 2019-03-10
Compute adaptive tiling variance in sRGB

wilimitis:
A minimally invasive attempt at exploring https://github.com/appleseedhq/appleseed/issues/2403 in order to initiate discussion.

In analyzing the curve of [linear_rgb_to_srgb](https://github.com/appleseedhq/appleseed/blob/master/src/appleseed/foundation/image/colorspace.h#L662) (simplified to omit the `x < 0.003` ramp-up) we can begin to understand how computing variance in sRGB will affect the rendering process.
Screen Shot 2019-03-09 at 6 46 18 PM

Linear rgb values (especially those closer to 0) will be modified in a way such that the variance between them increases. For instance, using the existing variance computation from the reference

| | RGB | sRGB |
| --- | --- | --- |
| C1 | (0.01, 0.01, 0.01) | (0.1, 0.1, 0.1) |
| C2 | (0.25, 0.25, 0.25) | (0.53, 0.53, 0.53) |
| Variance(C1, C2) | ~0.42 | ~0.75 |

To see how this manifests in test renders, consider the following test case: [adaptive-variance-srgb.appleseedz.zip](https://github.com/appleseedhq/appleseed/files/2948999/adaptive-variance-srgb.appleseedz.zip) (starter Cornell box with adaptive tiling at a noise threshold of 0.1).

| RGB Variance | sRGB Variance |
| --- | --- |
| adaptive_rgb3 | adaptive_srgb_naive3 |
| Screen Shot 2019-03-09 at 6 52 12 PM | Screen Shot 2019-03-09 at 6 51 44 PM |
| Screen Shot 2019-03-09 at 6 53 30 PM | Screen Shot 2019-03-09 at 6 54 10 PM |

Thus, in a given render pass, we have less noise, but we are also performing more work (in relation to the parabolic nature of the transformation) to do so, since there are potentially more samples required to reduce pixel variance below the threshold.

Since there will be higher variance in comparisons involving darker colors of the spectrum, we can perhaps consider wiring this up as a toggle.


luisbarrancos:
How do we proceed when the target isn't using the [sRGB OETF](https://www.colour-science.org/posts/the-importance-of-terminology-and-srgb-uncertainty/) for example? When it's using [Rec.709](https://www.itu.int/dms_pubrec/itu-r/rec/bt/R-REC-BT.709-6-201506-I!!PDF-E.pdf) (closer to gamma 2.4)? Or when it's using gamma 2.6, for digital cinema, [Hybrid Log Gamma](https://en.wikipedia.org/wiki/Hybrid_Log-Gamma) or [DolbyPQ](https://en.wikipedia.org/wiki/Rec._2100) for HDR television? Since we're now tying the sampling to a perceptual domain, will we be able to override the OETF, CCTF or OOTF?

dictoon:
Very nice analysis, thanks!

In your images, both render times and variances vary, making it difficult if computing variances in sRGB is a net improvement. What about trying to do equal-time comparisons? The process is a bit annoying, but it can be done by iteratively adjusting the noise threshold until render times match.

> I would like to find a scientific article backing this, but there seems to be loose consensus that the human eye is more sensitive to variations in dark shades than light shades. Since there will be higher variance involving darker colors of the spectrum, this approach seems intuitive, and we can perhaps consider wiring this up as a toggle.

That's a great observation, and we even have an issue covering this optimization: https://github.com/appleseedhq/appleseed/issues/2224

dictoon:
@luisbarrancos That's indeed a good question. My thinking is that doing variance computations in sRGB is _anyway_ better than doing them in linear RGB, and will be closer to the "correct thing" even when using non-sRGB OETFs.

Finding a way to use the actual OETF would be the next step.

wilimitis:
Using the same test project linked above, following @dictoon's suggested procedure yields

| RGB Noise 0.08 | RGB Noise 0.3 ~19.5s | sRGB Noise 0.4 ~19.5s |
| --- | --- | --- |
| Untitled | adaptive_srgb_naive3 | adaptive_rgb3 |
| Screen Shot 2019-03-10 at 10 30 30 PM | Screen Shot 2019-03-10 at 10 29 48 PM | Screen Shot 2019-03-10 at 10 29 55 PM |
| Screen Shot 2019-03-10 at 10 32 46 PM | Screen Shot 2019-03-10 at 10 32 23 PM | Screen Shot 2019-03-10 at 10 32 35 PM |
| Screen Shot 2019-03-10 at 10 43 38 PM | Screen Shot 2019-03-10 at 10 43 15 PM | Screen Shot 2019-03-10 at 10 43 26 PM |

There seems to still be an improvement overall when considering equivalent render times (though, there is runtime variance on my machine of about ~200ms for identical renders), however it is of course less obvious and even the last example seems questionable.

dictoon:
Thanks!

Strangely, the noise on the ceiling is significantly reduced, but there is slightly more noise in the shadow of the block.

What adaptive tile renderer settings are you using, beside the noise threshold?

wilimitis:
Yeah, I noticed that as well, and I just double-checked the output images again to confirm.

I'm using the default adaptive tile settings, namely
```
Passes: 1
Batch Size: 16
Min Samples: 0
Max Samples: 256
```

dictoon:
Might be interesting to try with smaller batches: maybe what happens is batches of 16 samples cause oversampling in some areas such as the ceiling, for instance if the noise there is just slightly above the threshold. Maybe batches of 4 samples maybe?

It's also worth experimenting on a slightly more realistic scene such as the Cookies scene that was used to test the new Adaptive Tile Renderer:

https://www.dropbox.com/s/b2kumvc93vc86xq/cookies.appleseedz?dl=0

wilimitis:
Looks like GitHub creates distinct events when force pushing to a branch that tags a PR, sorry for the noise! (😉)

**_More Realistic Scene_**
Circling back on this, I've lowered the batch size from 16 to 8, and since there were assets missing from cookies, I found a more realistic scene by crawling the forums https://forum.appleseedhq.net/t/indoor-scene-with-chairs-around-a-table/422.

| Ref* | RGB | sRGB |
| --- | --- | --- |
| ref | rgb | srgb3 11 |

**Ref**: 16 passes, 8 per batch, 0 min, 256 max, 0.1 noise
**rgb**: 4 passes, 8 per batch, 0 min, 256 max, 0.3 noise, 3m21s, 0.040569 rmse**
**srgb**: 4 passes, 8 per batch, 0 min, 256 max, 0.27 noise, 3m17s, 0.040632 rmse**

I've dug into each of the rgb/srgb examples and found cases (like above) where one is some non-zero number of pixels less noisy than the other, but it is still not clear whether this is an overall improvement.

_*The right two renders are cropped due to the firefly in the ref spiking the max error value._
_**The rmse might seem a bit lower since we're dividing by the sum under the root operator by 3, per our discord discussion, and I haven't found other sources yet which compute it the same way._

**_Cornell Box_**
Decreasing the batch size from 16 to 12 and then 8 still caused slightly more variance in sRGB's lower-right box-shadow.

**_Next Steps_**
We could add an option to compute sample variance in sRGB, and even provide context (via a tooltip?) as to how darker areas will be sampled more than lighter areas.

Creating a threshold seems like it would be interesting to play with, and might require accepting two metrics
1. What exactly is a "dark" area? Avg luminance for the tile below a given value range seems functional.
2. How exactly is increased sampling performed? Decreasing the acceptable noise threshold by a given value seems functional as well.

Do you think either of these are worth pursuing? I'm hesitant to sink more time into generating renders without adding more tools for analysis.

dictoon:
> Looks like GitHub creates distinct events when force pushing to a branch that tags a PR, sorry for the noise! 

No worries.

Regarding your comparisons: I'm not sure computing the reference image with adaptive sampling is a good idea, I would instead suggest computing it with uniform sampling and at a very high quality level (i.e. noise-free).

That said, I agree that the visual improvement is not obvious at all (these are equal-time renders right?).

Another question: the renders you posted are expressed in sRGB right?

It's not clear to me yet why there isn't any clear improvement. Could you visualize and compare the number of samples (using the right AOV) in both linear and sRGB cases?

wilimitis:
>Regarding your comparisons: I'm not sure computing the reference image with adaptive sampling is a good idea, I would instead suggest computing it with uniform sampling and at a very high quality level

This makes sense to me.

>these are equal-time renders right?

Yeah, equal in the sense of using the method you discussed above, the ones I posted were consistently within 4 seconds of each other. Getting closer than that on my machine seems tricky without implementing more robust timing termination logic.

>the renders you posted are expressed in sRGB right?

They're the output of the frame to exr then converted directly to png (which from reading `write_exr_image`, leads me to believe is linear rgb). Thank you for pointing out this distinction though, I've posted new renders of equal timing in both linear rgb/srgb, where the refs are essentially uniformly sampled for 30 min.

>(using the right AOV)

Sorry, before I go and produce _N_ renders again, what exactly do you mean by this? As in, is there a specific AOV you think is relevant to analyze here, like the sample time AOV?

dictoon:
Note that you can also save PNG directly from appleseed.studio and appleseed.cli, if needed.

Comparing sRGB images is absolutely the right thing to do, it wouldn't make sense to compare linear images.

wilimitis:
I've incorporated your feedback into a new set of renders. I will post an identical table for the cookies scene soon after the "noise-free" (i.e. long running uniform sampling) render has completed. In all renders it was assured my machine had full access to the cpu (nothing else consuming > 1% cpu) and they were run several times to confirm the render times are consistent.

| ref | rgb var short | srgb var short | rgb var long | srgb var long |
|---|---|---|---|---|
| png (sRGB)<br/>uniform<br/>16 passes<br/>32 spp | png (sRGB)<br/>adaptive<br/>1.0 noise<br/>6.1 sec<br/>0.03725 rmsd | png (sRGB)<br/>adaptive<br/>0.815 noise<br/>6.1 sec<br/>0.039309 rmsd | png (sRGB)<br/>adaptive<br/>0.1 noise<br/>1 min 2 sec<br/>0.013664 rmsd | png (sRGB)<br/>adaptive<br/>0.173 noise<br/>1 min 2 sec<br/>0.012934 rmsd |
| ref | rgb-1n-6s-037250 | srgb-0815n-6s-039309 | rgb-01n-1m2s-013664 | srgb-0173n-1m2s-012934 |
| sample count aov | rgb-1n-6s-037250 pixel_sample_count | srgb-0815n-6s-039309 pixel_sample_count | rgb-01n-1m2s-013664 pixel_sample_count | srgb-0173n-1m2s-012934 pixel_sample_count |

It is immediately noticeable that the long rgb render has a much cleaner ceiling, whereas the long srgb render has a much cleaner back and right wall. Since the noises level had to vary to acquire matching renders, and yet the srgb render had an overall lower rmsd with an overall less noisy appearance besides near the light, I'm tempted to confirm that it is better overall in cases where lights are not dominant features in the scene. However, the rmsd point may not be valid since rmsd is computed in linear space.

The cookies scene may reveal more.

dictoon:
Thanks for the detailed comparisons, that's very interesting. I'll wait for the Cookies scene before making any conclusion, but these preliminary results on the Cornell Box seem to indicate that computing the variance in linear RGB is overall the best option as it puts more samples on more noisy areas such as the ceiling and the small cube's front face.

I'm just very surprised. I wouldn't have expected sRGB variance to be so difference from its linear counterpart, and I would have expected it to be a better metric. Moreover, I'm surprised that the light itself is so oversampled in the "srgb var long" render. That's so surprising that I'm kind of suspecting a bug really. Any idea why it's doing that?

src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
```diff
@@ -164,10 +164,12 @@ namespace
         const float rcp_second_weight = second_weight == 0.0f ? 0.0f : 1.0f / second_weight;
 
         // Get colors and assign weights.
-        Color4f main_color(main[0], main[1], main[2], main[3]);
+        Color3f main_color(main[0], main[1], main[2]);
+        main_color = linear_rgb_to_srgb(main_color);
```
dictoon: Wait a minute, shouldn't the color space conversion happen _after_ the division by the weight? I'm pretty sure it should.

wilimitis: Nice catch, yes!

wilimitis:
Having reran the srgb renders with the transformation applied after weighting:

| ref | rgb var short | srgb var short | rgb var long | srgb var long |
|---|---|---|---|---|
| png (sRGB)<br/>uniform<br/>16 passes<br/>32 spp | png (sRGB)<br/>adaptive<br/>1.0 noise<br/>6.1 sec<br/>0.03725 rmsd | png (sRGB)<br/>adaptive<br/>0.91 noise<br/>6.1 sec<br/>0.05201 rmsd | png (sRGB)<br/>adaptive<br/>0.1 noise<br/>1 min 2 sec<br/>0.013664 rmsd | png (sRGB)<br/>adaptive<br/>0.091 noise<br/>1 min 2 sec<br/>0.015275 rmsd |
| ref | rgb-1n-6s-037250 | rgb-091n-6s-05201 | rgb-01n-1m2s-013664 | srgb-0091-1m2s-015275 |
| sample count aov | rgb-1n-6s-037250 pixel_sample_count | rgb-091n-6s-05201 pixel_sample_count | rgb-01n-1m2s-013664 pixel_sample_count | srgb-0091-1m2s-015275 pixel_sample_count |
| error aov<br/>(light cropped out since it dominates error viz) | rgb1 pixel_error | quicksave pixel_error | rgb2 pixel_error | quicksave pixel_error |

The expected outcome - more time spent reducing noise in darker areas - is verified by the aovs. Though overall, in this render at least, it isn't obvious imo that the srgb variance render is less noisy overall, since the noise on the back wall is immediately noticeable.

wilimitis:
Cookies scene ran with latest commit:

| ref | rgb var short | srgb var short | rgb var long | srgb var long |
|---|---|---|---|---|
| png (sRGB)<br/>uniform<br/>16 passes<br/>32 spp | png (sRGB)<br/>adaptive<br/>1.0 noise<br/>3 min 16 sec<br/>0.035291 rmsd | png (sRGB)<br/>adaptive<br/>0.89 noise<br/>3 min 16 sec<br/>0.040898 rmsd | png (sRGB)<br/>adaptive<br/>0.2 noise<br/>15 min 40 sec<br/>0.020112 rmsd | png (sRGB)<br/>adaptive<br/>0.265 noise<br/>15 min 30 sec<br/>0.024131 rmsd |
| ref | rgb-1n-3m16s-035291 | srgb-089n-3m15s-040898 | rgb-02n-15m43s-020112 | srgb-0265n-15m30s-024131 |
| sample count aov | rgb-1n-3m16s-035291 pixel_sample_count | srgb-089n-3m15s-040898 pixel_sample_count | rgb-02n-15m43s-020112 pixel_sample_count | srgb-0265n-15m30s-024131 pixel_sample_count |
| error aov | rgb-1n-3m16s-035291 pixel_error | srgb-089n-3m15s-040898 pixel_error | rgb-02n-15m43s-020112 pixel_error | srgb-0265n-15m30s-024131 pixel_error |

As expected the rgb long render spends more time sampling the brighter area where the bottle is full and the srgb long render spends more time sampling the darker area where the bottle is empty - the visible noise reflects this.

wilimitis:
An important thing to note for the rms calculation is that the variance is computed from the linear rgb image render against the linearized rgb reference image. I think we would need to perhaps add an option to specify whether the image render should be converted to srgb before running rms calculations.

dictoon:
> An important thing to note for the rms calculation is that the variance is computed from the linear rgb image render against the linearized rgb reference exr image.

If the reference image is in OpenEXR format, it's already linear so there is no linearization going on. Right?

> I think we would need to perhaps add an option to specify whether the image render should be converted to srgb before running rms calculations.

I'm not sure to follow, isn't it exactly what the code is doing right now, converting both the reference and the render to sRGB before computing the variance?

dictoon:
Thanks for the new comparisons! Results look a lot more logical.

The expected outcome of using the Adaptive Tile renderer is not less noise overall, it's _a more uniform level of noise across the image_. It seems to be what we observe in the Cornell Box scene.

The Cookies scene is more interesting: as you noticed, computing variance in sRGB leads to more samples in dark areas (background) to the detriment of brighter areas (milk bottle).

We are probably  still not using the right metric for visual noise. A more accurate measure of perceptual noise may lead to better results. There's a wealth of research and resources on the topic, however I have the impression that it's mostly geared toward video compression. Still, I'm sure there are lots of information available on the topic. For instance, maybe we should put more emphasis on luminance noise and less on chromatic noise.

I propose that we keep this topic in the back of our mind and put this PR on hold for a little while. Maybe we or someone will have an idea or we'll come across some new information that will allow us to move forward.

Thanks for your top quality work!

wilimitis:
> If the reference image is in OpenEXR format, it's already linear so there is no linearization going on. Right?

Right.

> isn't it exactly what the code is doing right now, converting both the reference and the render to sRGB before computing the variance?

Ah, in this case I'm referring to the rms as in: https://github.com/appleseedhq/appleseed/blob/master/src/appleseed/foundation/image/analysis.cpp#L131.

dictoon:
> Ah, in this case I'm referring to the rms as in: /src/appleseed/foundation/image/analysis.cpp@master#L131.

Yeah, we assume that the reference image is linear (and `Frame::image()` is always linear). I don't think we need to get too fancy here (linearizing the ref image if it's e.g. a PNG file) since this feature is targeted at developers who are supposed to know what they do :)

Issue #2023, opened 2018-05-05
Make Render stamp based on percentage


juancarlosgzrz:
When rendering a big image, the render stamp stays small, because the size of it is fixed. Based on image percentage should fix the problem ;D

BashPrince:
I'll have a go at this.

PR #2468, opened 2019-03-11
Implement render stamp scaling

BashPrince:
This if for Optionally scale render stamp based on render resolution [#2023
](https://github.com/appleseedhq/appleseed/issues/2023)
WIP: I created this so the current state can be reviewed and there are some issues.
First, I added a scale factor input by adding it to the metadata. I figured specifying the min and max allowed values there would limit them in the GUI as well, but the GUI seems to ignore these and allows setting any values.
Second, I noticed that bigger scale factors (e.g. 4.0) would introduce dark edges around the icon. I think this is because during the resize() using OpenImageIO, the scaling filter would weigh in some of the black pixels around the colored pixel regions. I am not sure how to prevent this, maybe simply using a higher resolution than the 64 x 61 of the appleseed_seeds_64[] float array would help.
This would also make more sense considering that the 64 x 61 version is only 4 times bigger than the original 16 x 15 which would only allow a scale factor of 4 before the image has to be upscaled (at the same relative icon size as originally). I could generate a appleseed_seeds_128 or maybe even 256 array in a .cpp file just like the 16, 32 and 64 versions. 256 would come in at at about 2MB.
Below is my commit message.

Add new scale factor control numeric to the stamp postprocessing
metadata.

Use the appleseed_seeds_64 float array as source data for the icon.
Scale the original 64 x 61 pixel image data to fit into the scaled area
using OpenImageIO's ImageBuf and resize function. Blit the scaled data
onto the frame. Also scale the text before drawing.

The mitchell filter is used for the resize function with the filter size
scaling according to the user scale factor as this gave best visual
results.

The default height of the destination region of interest is set to 14 to
best match the previous icon size.


src/appleseed/renderer/modeling/postprocessingstage/renderstamppostprocessingstage.cpp
```diff
@@ -207,6 +229,22 @@ DictionaryArray RenderStampPostProcessingStageFactory::get_input_metadata() cons
...
 
+    metadata.push_back(
+        Dictionary()
+            .insert("name", "scale_factor")
+            .insert("label", "Scale Factor")
+            .insert("type", "numeric")
+            .insert("min",
+                    Dictionary()
+                        .insert("value", "0.1")
+                        .insert("type", "soft"))
```
dictoon: By design, soft limits can be exceeded. Only hard limits (`type == "hard"`) are strictly enforced.

It would be nice if we could specify both soft and hard limits at the same time but our current system doesn't allow that yet.

src/appleseed/renderer/modeling/postprocessingstage/renderstamppostprocessingstage.cpp
```diff
@@ -142,31 +152,43 @@ namespace
...
+                FontColor,
+                MarginH + roi.width() + MarginH,
+                origin_y,
+                text.c_str());
+            
         }
 
       private:
         string m_format_string;
+        float m_scale_factor;
```
dictoon: Align members vertically.

src/appleseed/renderer/modeling/postprocessingstage/renderstamppostprocessingstage.cpp
```diff
@@ -45,15 +45,20 @@
...
 #include "foundation/utility/api/specializedapiarrays.h"
 #include "foundation/utility/containers/dictionary.h"
 #include "foundation/utility/string.h"
 
+// OpenImageIO headers.
+#include "OpenImageIO/imagebufalgo.h"
+
 // Standard headers.
 #include <string>
+#include <memory>
```
dictoon: Order alphabetically.

src/appleseed/renderer/modeling/postprocessingstage/renderstamppostprocessingstage.cpp
```diff
@@ -142,31 +152,43 @@ namespace
...
-                FontColor,
-                MarginH + appleseed_seeds_16_width + MarginH,
-                origin_y,
-                text.c_str());
+            // Calculate final icon size and scale the icon
+            ImageSpec icon_spec = m_icon.spec();
+            float aspect = static_cast<float>(icon_spec.width) / icon_spec.height;
+            ROI roi(0, m_scale_factor * aspect * DefaultIconHeight, 0, m_scale_factor * DefaultIconHeight, 0, 1, 0, m_icon.nchannels());
+            ImageBuf scaled_logo;
+            ImageBufAlgo::resize(scaled_logo, m_icon, "mitchell", m_scale_factor, roi);
```
dictoon: Dark edges may be caused by the Mitchell-Netravali filter since it features negative lobes. Do dark edges disappear with other filters?

dictoon:
There are also some warnings on Windows that cause AppVeyor builds to fail, see for instance https://ci.appveyor.com/project/appleseedhq/appleseed/builds/22968936/job/8udhc965e8g095al.

dictoon:
Also, feel free to add higher resolution versions of the logo; I don't think the size is much of a concern.

BashPrince:
Thanks, I'll get to soon.

BashPrince:
You were right about the dark borders, they were apparently used by the filter. I now use sharp gaussian instead.

BashPrince:
There still seem to be issues with Visual Studio builds related to OpenImageIO, I will have to look at that in more detail.

src/appleseed/foundation/resources/logo/appleseed-seeds-256.h
```diff
@@ -0,0 +1,57 @@
...
+// Standard headers.
+#include <cstddef>
+
+namespace foundation
+{
+
+//
+// appleseed "seeds" logo:
+//
+//   256x61 pixels
```
dictoon: Comment needs to be updated.

src/appleseed/foundation/resources/logo/appleseed-seeds-256.h
```diff
@@ -0,0 +1,57 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+#ifndef APPLESEED_FOUNDATION_RESOURCES_LOGO_APPLESEED_SEEDS_256_H
```
dictoon: Please use

    #pragma once

in new header files.

src/appleseed/renderer/modeling/postprocessingstage/renderstamppostprocessingstage.cpp
```diff
@@ -103,7 +115,7 @@ namespace
         {
             // Render stamp settings.
             const auto Font = TextRenderer::Font::UbuntuL;
-            const float FontHeight = 14.0f;
+            const float FontHeight = 14.0f * m_scale_factor;
```
dictoon: `FontHeight` is no longer a true constant (which would be marked as `constexpr` if all our compilers would support it), thus it must be spelled `font_height`.

src/appleseed/renderer/modeling/postprocessingstage/renderstamppostprocessingstage.cpp
```diff
@@ -142,31 +154,51 @@ namespace
...
-                frame.image(),
-                Font,
-                FontHeight,
-                FontColor,
-                MarginH + appleseed_seeds_16_width + MarginH,
-                origin_y,
-                text.c_str());
+            // Calculate final icon size and scale the icon
+            ImageSpec icon_spec = m_icon.spec();
+            float aspect = static_cast<float>(icon_spec.width) / static_cast<float>(icon_spec.height);
```
dictoon: Make const.

dictoon:
> There still seem to be issues with Visual Studio builds related to OpenImageIO, I will have to look at that in more detail.

@luisbarrancos opened a PR about that: #2474. Could it be related? #2474 will be merged in master as soon as CI builds passed.

BashPrince:
I would be surprised if it's related, the build error messages seem to be about type conversions and one undeclared identifier. I think I will try building on Visual Studio myself next to see if I get the same errors there.
For now, pngs of scale at 1x, 2x, 12x
scale1
scale2
scale12

BashPrince:
Is it an issue to push often to a pull request? Would it overload the CI servers?

BashPrince:
I assume this Travis build fail has nothing to do with my code?
`The command "if [ $TRAVIS_OS_NAME == osx ]; then chmod +x ./scripts/travis/build-macos.sh ; ./scripts/travis/build-macos.sh ; fi" exited with 1.`

BashPrince:
I finally got Appveyor Visual Studio to pass (at least VS17 for the moment). The build errors were caused by OpenImage TypeDesc's being defined as constexpr. I would not have found that, had you not mentioned it incidentally. I had to switch from using those to using static const members which are marked as deprecated. We are also using the static const TypeDescs in other code though, e.g. renderservices.cpp, so I think that should be fine for now.

I am still getting conversion warnings like: 
`c:\projects\appleseed\appleseed-deps-stage-win64-vs141-v18\stage\vc141\oiio-debug\include\openimageio\fmath.h(1131): warning C4244: 'initializing': conversion from 'int' to 'float', possible loss of data (compiling source file C:\projects\appleseed\src\appleseed\renderer\modeling\postprocessingstage\renderstamppostprocessingstage.cpp)`.
These are caused by stuff like:
`int q = fast_rint (x * float(M_1_PI));`
`float qf = q;`
in OpenImage headers. I think I can't really do anything about those without changing the header files.

dictoon:
> Is it an issue to push often to a pull request? Would it overload the CI servers?

It shouldn't be an issue as long as it's not every minute :)

src/appleseed/renderer/modeling/postprocessingstage/renderstamppostprocessingstage.cpp
```diff
@@ -168,7 +170,8 @@ namespace
                 m_icon.nchannels());
                 
             ImageBuf scaled_logo;
-            ImageBufAlgo::resize(scaled_logo, m_icon, "sharp-gaussian", 1.2f, roi);
+            const float scale_factor = OIIO::lerp(1.0f, 2.8f, (m_scale_factor - MinScaleFactor) / (MaxScaleFactor - MinScaleFactor));
```
dictoon: Does that work for scale factors smaller than 1?

BashPrince: Yes it linearly interpolates between filter width 1.0 for the smallest possible user supplied scale factor (0.1) and filter width 2.8 for the largest possible user scale factor (20.0). Here are results for a scale factor of 0.5 set in the GUI: 
scale_05
The const float scale_factor should actually be named filter_width for clarity, can I still change that?

dictoon: Sure!

src/appleseed/renderer/modeling/postprocessingstage/renderstamppostprocessingstage.cpp
```diff
@@ -170,8 +170,8 @@ namespace
                 m_icon.nchannels());
                 
             ImageBuf scaled_logo;
-            const float scale_factor = OIIO::lerp(1.0f, 2.8f, (m_scale_factor - MinScaleFactor) / (MaxScaleFactor - MinScaleFactor));
-            ImageBufAlgo::resize(scaled_logo, m_icon, "sharp-gaussian", scale_factor, roi);
+            const float filter_width = OIIO::lerp(1.0f, 2.8f, (m_scale_factor - MinScaleFactor) / (MaxScaleFactor - MinScaleFactor));
```
dictoon: It's better, I agree.

FYI, we have a nice little function, `foundation::fit()` (header `foundation/math/scalar.h`) to which you can pass the input range of a variable and the desired range for the output:

    const float filter_width = fit(m_scale_factor, MinScaleFactor, MaxScaleFactor, 1.0f, 2.8f);

You might need to specify the template types, I'm not sure:

    const float filter_width = fit<float, float>(m_scale_factor, MinScaleFactor, MaxScaleFactor, 1.0f, 2.8f);

You can keep the code you wrote, it's completely correct, but I thought I'd make you aware of existing utilities.

Issue #1508, opened 2017-07-06
Add progressive rendering time limit


gospodnetic:
Add an option in studio to use time as a rendering constraint instead of samples per pixel.

dictoon:
It would be nice to have this feature in appleseed.cli as well.

Idloj:
Where is the sample per pixel used as a limit? (my experience is that interactive rendering just runs indefinitely, but maybe I just haven't let it run enough. I think I gave it ~45 minutes)

dictoon:
It's used in the `ProgressiveFrameRenderer` class. The max number of samples is unlimited by default. This setting is not (yet) exposed in the Rendering Settings dialog.

Idloj:
And this issue is about exposing both samples per pixel and time limit in the Rendering Settings?

dictoon:
Yes absolutely!

dictoon:
We should also consider whether we can implement time-limited rendering in final mode.

PR #2471, opened 2019-03-12
Refactor max samples into average max spp for interactive mode

wilimitis:
Intended to be a first step towards exposing samples, samples per pixel, and time-based rendering constraints https://github.com/appleseedhq/appleseed/issues/1508.

Screen Shot 2019-03-12 at 4 34 09 AM

A default of 10,000 was chosen because this is the point at which the Cornell box begins to materialize. That said, we may want to bump it higher still.

Screen Shot 2019-03-12 at 4 34 51 AM


src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
```diff
@@ -652,6 +662,59 @@ namespace
...
+        {            
+            if (get_widget<bool>("general.unlimited_samples"))
+                config.get_parameters().remove_path("progressive_frame_renderer.max_samples");
+            else set_config(config, "progressive_frame_renderer.max_samples", get_widget<int>("general.max_samples"));
+        }
+
+      private:
+        void create_image_plane_sampling_general_settings(QVBoxLayout* parent)
+        {
+            QGroupBox* groupbox = new QGroupBox("General");
```
wilimitis: `Progressive` might be a more precise name for this group, however this seems implied by the configuration title, `interactive`. Maybe a group here isn't even worthwhile.

src/appleseed/renderer/kernel/rendering/progressive/progressiveframerenderer.cpp
```diff
@@ -763,6 +763,8 @@ Dictionary ProgressiveFrameRendererFactory::get_params_metadata()
         "max_samples",
         Dictionary()
             .insert("type", "int")
+            .insert("default", "100000")
+            .insert("unlimited", "true")
             .insert("label", "Max Samples")
             .insert("help", "Maximum number of samples per pixel"));
```
wilimitis: Hm, I think this needs to be changed to read `Maximum number of samples`, since this doesn't seem to be constraining maximum samples per pixel.

https://github.com/appleseedhq/appleseed/blob/master/src/appleseed/renderer/kernel/rendering/progressive/samplegeneratorjob.cpp#L127
The `ProgressiveFrameRenderer`'s termination condition hinges on a sample counter initialized with the total max sample count.

src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
dictoon: Indeed maybe we can get rid of the group.

Are these settings only exposed in interactive configurations?

src/appleseed/renderer/kernel/rendering/progressive/progressiveframerenderer.cpp
dictoon: That's correct, it currently controls the total number of samples. It's fine to expose those internal settings in appleseed.studio. In DCC apps we would probably expose "Max Samples per Pixel" instead since this is more useful to a user.

Retrospectively, maybe the actual control should have been "Max Average Samples per Pixel" both in the code, in appleseed.studio and in DCC apps because that makes the settings resolution-independent. It's not too late to change it :)

wilimitis: Oh nice, and "Max Average Samples per Pixel" would cause termination if total samples / total pixels exceeds it. Since I'm already in here, I can rework the computation then squash the commits.

(I'll double check that "Max samples" isn't incorrectly exposed as "Max samples per pixel" elsewhere too, though I didn't think it was on my first pass here)

src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
wilimitis: This change causes this specific setting to be exposed in interactive mode only.

Looking at the ui, there are other settings related to max samples exposed per the different sampler types in final mode, but none currently for max total number of samples.

dictoon: > This change causes this specific setting to be exposed in interactive mode only.

Ok, that's fine then.

> Looking at the ui, there are other settings related to max samples exposed per the different sampler types in final mode, but none currently for max total number of samples.

I'm not sure to understand what you mean here?

src/appleseed/renderer/kernel/rendering/progressive/progressiveframerenderer.cpp
dictoon: > Oh nice, and "Max Average Samples per Pixel" would cause termination if total samples / total pixels exceeds it.

Exactly.

(I'll double check that "Max samples" isn't incorrectly exposed as "Max samples per pixel" elsewhere too, though I didn't think it was on my first pass here)

👍

src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
wilimitis: I think I'm just repeating in saying that max samples is only exposed in interactive mode.

wilimitis: (sorry for the confusion, my understanding of the project is still growing, thus the comment in question seems naïve)

wilimitis:
I've pushed a commit to refactor this param into max average samples per pixel, based on the fact that
```
max average samples per pixel * total pixels = max total number of samples
```
as discussed briefly above.

Screen Shot 2019-03-19 at 3 39 50 AM
Screen Shot 2019-03-19 at 3 44 07 AM
Screen Shot 2019-03-19 at 3 49 32 AM

src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
dictoon: No problem!

src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
```diff
@@ -444,17 +444,27 @@ namespace
...
     //
-    // Image Plane Sampling panel.
+    // Image Plane Sampling panels.
     //
 
     class ImagePlaneSamplingPanel
       : public RenderSettingsPanel
+    {
+      public:
+        ImagePlaneSamplingPanel(QWidget* parent = nullptr)
```
dictoon: Should be marked as `explicit`.

src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
```diff
@@ -652,6 +662,65 @@ namespace
...
+            create_image_plane_sampling_general_settings(layout);
+
+            load_general_sampler(config);
+        }
+
+        void save_config(Configuration& config) const override
+        {            
+            if (get_widget<bool>("general.unlimited_samples"))
+                config.get_parameters().remove_path("progressive_frame_renderer.max_average_samples_per_pixel");
+            else set_config(
```
dictoon: Wrapping this in curly braces would look better.

src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
```diff
@@ -652,6 +662,65 @@ namespace
...
+            QGroupBox* groupbox = new QGroupBox("General");
+            parent->addWidget(groupbox);
+
+            QVBoxLayout* layout = new QVBoxLayout();
+            groupbox->setLayout(layout);
+
+            QFormLayout* sublayout = create_form_layout();
+            layout->addLayout(sublayout);
+
+            QSpinBox* max_average_samples_per_pixel = create_integer_input("general.max_average_samples_per_pixel", 1, 1000000, 10000);
```
dictoon: The last argument is the step, 10000 seems excessive now :) Not sure what would be a good value... 10?

src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
```diff
@@ -652,6 +662,65 @@ namespace
...
+            QCheckBox* unlimited_samples = create_checkbox("general.unlimited_samples", "Unlimited");
+            sublayout->addRow("Max Average Samples Per Pixel:", create_horizontal_group(max_average_samples_per_pixel, unlimited_samples));
+            connect(unlimited_samples, SIGNAL(toggled(bool)), max_average_samples_per_pixel, SLOT(setDisabled(bool)));
+        }
+
+        void load_general_sampler(const Configuration& config)
+        {
+            const int DefaultMaxSamples = 100000;
```
dictoon: Needs updating too... Maybe 64 like the default number of SPP in final mode?

src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
```diff
@@ -1522,12 +1591,13 @@ void RenderingSettingsWindow::create_panels(const Configuration& config)
...
 
-    if (!interactive)
-        m_panels.push_back(new ImagePlaneSamplingPanel(config));
-
-    if (interactive)
+    if (interactive) {
+        m_panels.push_back(new InteractiveImagePlaneSamplingPanel(config));
         m_panels.push_back(new InteractiveConfigurationLightingPanel(config));
-    else m_panels.push_back(new FinalConfigurationLightingPanel(config));
+    } else {
```
dictoon: Move braces to their own lines.

src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
```diff
@@ -1522,12 +1591,13 @@ void RenderingSettingsWindow::create_panels(const Configuration& config)
 
     m_panels.push_back(new GeneralSettingsPanel(config));
 
-    if (!interactive)
-        m_panels.push_back(new ImagePlaneSamplingPanel(config));
-
-    if (interactive)
+    if (interactive) {
```
dictoon: Here too.

src/appleseed/renderer/kernel/rendering/progressive/progressiveframerenderer.cpp
```diff
@@ -385,18 +387,19 @@ namespace
...
             const Spectrum::Mode        m_spectrum_mode;
             const SamplingContext::Mode m_sampling_mode;
-            const size_t                m_thread_count;         // number of rendering threads
-            const uint64                m_max_sample_count;     // maximum total number of samples to compute
-            const double                m_max_fps;              // maximum display frequency in frames/second
-            const bool                  m_perf_stats;           // collect and print performance statistics?
-            const bool                  m_luminance_stats;      // collect and print luminance statistics?
-            const string                m_ref_image_path;       // path to the reference image
+            const size_t                m_thread_count;                     // number of rendering threads
+            const uint64                m_max_average_samples_per_pixel;    // maximum average number of samples to compute per pixel
```
dictoon: I would be fine with using the "spp" abbreviation too :) `m_max_avg_spp` is clear enough IMHO.

src/appleseed/renderer/kernel/rendering/progressive/progressiveframerenderer.cpp
```diff
@@ -760,11 +764,13 @@ Dictionary ProgressiveFrameRendererFactory::get_params_metadata()
             .insert("help", "Maximum progressive rendering update rate in frames per second"));
 
     metadata.dictionaries().insert(
-        "max_samples",
+        "max_average_samples_per_pixel",
         Dictionary()
             .insert("type", "int")
-            .insert("label", "Max Samples")
-            .insert("help", "Maximum number of samples per pixel"));
+            .insert("default", "100000")
```
dictoon: Needs updating. 64?

src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
wilimitis: 👍

FWIW Clang-Tidy warns about implicit conversions for other constructors in this file as well, for instance
`RenderSettingsPanel(const QString& title, QWidget* parent = nullptr)`
since it is callable with a single argument.

src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
wilimitis: Yeah, I don't see the steps getting used often, though I am still ramping up here, so 10 seems good as any to me.

src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
wilimitis: Ah, yes.

wilimitis:
PTAL

Also, one thing I noticed is that this doesn't account for whether a crop window is set in studio. One way to account for this it seems is to expose a method on the frame to return total sample-able pixels and applying it here: https://github.com/appleseedhq/appleseed/pull/2471/files#diff-22f25c5c4f9bab4b8f5e8ec516306321R109.

src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
dictoon: They should be marked as explicit as well. Good catch!

dictoon:
> Also, one thing I noticed is that this doesn't account for whether a crop window is set in studio. One way to account for this it seems is to expose a method on the frame to return total sample-able pixels

Oh yeah, another good catch!

You can use `Frame::get_crop_window()::volume()` to get the number of pixels in the crop window. Please double-check that this does return exactly the number of pixels to render, I quickly checked and it seems fine, but you know :)

src/appleseed/renderer/kernel/rendering/progressive/progressiveframerenderer.cpp
```diff
@@ -417,13 +419,12 @@ namespace
...
                 SampleAccumulationBuffer&   buffer,
                 ITileCallback*              tile_callback,
-                const uint64                max_sample_count,
                 const double                max_fps,
                 IAbortSwitch&               abort_switch)
               : m_frame(frame)
               , m_buffer(buffer)
               , m_tile_callback(tile_callback)
-              , m_min_sample_count(min<uint64>(max_sample_count, 32 * 32 * 2))
+              , m_min_sample_count(min<uint64>(frame.get_crop_window().volume(), 32 * 32 * 2))
```
dictoon: This doesn't look correct as it's no longer taking into account the max number of samples/pixel.

dictoon:
PR will need to be rebased.

src/appleseed/renderer/kernel/rendering/progressive/progressiveframerenderer.cpp
wilimitis: Since this number is the minimum number of samples required to first display the render progress, the previous number made sense whereas now if we incorporate the max number of samples per pixel we won't call to display for sizes < 32 * 32 * 2 unless we reach the final rendering state, which doesn't seem correct.

Using `volume` here is stating that for sizes < 32 * 32 * 2 we should just display the render once we have a window's worth of pixel samples, which would be a roughly 45 x 45 window, which renders nearly immediately.

dictoon: I think that the multiplication by two in `32 * 32 * 2` meant 2 spp for a 32x32 window.

With your changes, we start displaying when we either have 2 spp at resolution 32x32, or 1 spp at 45x45 and up, which seems fine.

I guess the logic could be simplified / cleaned up further but it's not urgent.

dictoon:
Thanks!

You should read up about rebasing. The appropriate procedure here is to _fetch_ (and not _pull_) the master branch, then _rebase_ your branch on top.

What you just did is that you merged master into your branch. The result is similar and in that case it doesn't matter because I'll squash and rebase your PR on top of master anyway, but it's useful to know how these things work :)

wilimitis:
I don't mind rebasing, and I actually do it daily; I'm curious what is incorrect in this particular case by not rebasing.

dictoon:
Well, that's my point, you didn't rebase, you merged master into your branch.

dictoon:
Ah maybe I misunderstood what you mean, are you asking what's wrong with merging? It forces me to squash your PR into a single commit in order to get rid of the merge commit.

wilimitis:
Ahh, I see, I think I mistakenly thought this was going to be squashed regardless, in which case rebase + squash and merge + squash would both add only 1 commit to master.

Sorry for the confusion >.< ...

dictoon:
> Ahh, I see, I think I mistakenly thought this was going to be squashed regardless, in which case rebase + squash and merge + squash would both add only 1 commit to master.

> Sorry for the confusion >.< ...

No prob! In that case it's no big deal because I would squash the PR into a single commit anyway since all commits are related to the same feature. But other PRs might have a bunch of unrelated commits, or several commits that implement a feature in small pieces without ever going back to fix bugs, address review comments, etc. For _those_ PRs, I would _rebase_ the PR on top of master when merging instead of squashing it (which would loose the individual commits). And in _that_ case, merge commits are problematic and need to be avoided or removed by the PR's author.

Hope that clarifies somewhat the process!

src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
wilimitis: Updated.

dictoon:
The PR looks good. I realized yesterday that there's one last thing missing: we need to automatically update existing project files to the new progressive frame renderer parameter:

1. Bump the project file format revision in `modeling/project/projectformatrevision.h`.
2. Add a migration step in `modeling/project/projectfileupdater.cpp`.
3. Test by loading `* - lt.appleseed` scenes in sandbox/tests/test scenes/light tracing/`.

dictoon:
It seems that you changes from your commit "Mark panel ctors explicit" got somehow lost? Only one constructor is still marked as explicit, although you _did_ at some point mark them all as explicit.

wilimitis:
I did but then pulled it because it wasn't relevant to the changes in this PR, and since this is getting squashed because I failed to rebase, it seemed like opening a new PR for it would be correct. I don't mind re-adding it here though if it's no big deal.

dictoon:
Regarding the light tracing problem, the fix has just been merged so you if you pull master again you'll get it (and a bunch of other fixes).

src/appleseed/renderer/modeling/project/projectfileupdater.cpp
```diff
@@ -2017,6 +2018,55 @@ namespace
...
+        }
+
+      private:
+        void replace_max_samples_interactive_renderer_setting()
+        {
+            for (Configuration& config : m_project.configurations())
+            {
+                Dictionary& root = config.get_parameters();
+
+                if (root.dictionaries().exist("progressive_frame_renderer")) {
```
dictoon: Move braces to their own lines (here and on the lines below).

src/appleseed/renderer/modeling/project/projectfileupdater.cpp
```diff
@@ -2017,6 +2018,55 @@ namespace
...
+                    Dictionary& pfr = root.dictionaries().get("progressive_frame_renderer");
+
+                    if (pfr.strings().exist("max_samples")) {
+                        RENDERER_LOG_WARNING(
```
dictoon: I don't think we need this warning since we can migrate the project in a (mostly, admittedly) compatible manner.

The previous upgrade emitted a warning because there was no easy way to migrate the Adaptive Pixel Renderer settings to the new Adaptive Tile Renderer.

src/appleseed/renderer/modeling/project/projectfileupdater.cpp
```diff
@@ -2017,6 +2018,55 @@ namespace
...
+                            "the max samples setting of the progressive frame renderer was removed; "
+                            "migrating this project to use the max average spp setting instead.");
+
+                        uint64 max_samples = pfr.strings().get<uint64>("max_samples");
```
dictoon: Make const.

src/appleseed/renderer/modeling/project/projectfileupdater.cpp
```diff
@@ -2017,6 +2018,55 @@ namespace
...
+                        pfr.strings().remove("max_samples");
+
+                        Frame* frame = m_project.get_frame();
+                        if (frame) {
+                            // If max samples was previously set then preserve the nearest max average spp count.
+                            pfr.strings().insert(
+                                "max_average_spp", ceil(max_samples / frame->get_crop_window().volume()));
```
dictoon: It's probably a good idea to cast the result to an integer before inserting it into the dictionary.

wilimitis:
Yep! The light tests seem to run as expected now :) The max average spp value is picked up from the existing project file for the final mode.

PR #2472, opened 2019-03-12
Fix albedo AOV for Oren Nayar BRDF when roughness is zero

est77:


est77:
I have some test scenes in progress. I'll add them in a following PR.

PR #2474, opened 2019-03-13
Fix cryptomatte with OIIO2

luisbarrancos:
since ImageInput, ImageOutput are now unique pointers.

### DIFF
--- a/src/appleseed/renderer/modeling/aov/cryptomatteaov.cpp
+++ b/src/appleseed/renderer/modeling/aov/cryptomatteaov.cpp
@@ -54,17 +54,22 @@
 #include "foundation/utility/otherwise.h"
 #include "foundation/utility/string.h"
 
+// Boost headers.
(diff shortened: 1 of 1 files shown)


src/appleseed/renderer/modeling/aov/cryptomatteaov.cpp
```diff
@@ -244,10 +252,14 @@ namespace
         }
 
     private:
-        const ICanvas*      m_canvas;
-        OIIO::ImageSpec     m_spec;
-        OIIO::ImageOutput*  m_writer;
-        const char*         m_filename;
+        const ICanvas*                      m_canvas;
+        OIIO::ImageSpec                     m_spec;
+#if OIIO_VERSION > 20000
```
dictoon: >= no?

luisbarrancos: Fixed

PR #2477, opened 2019-03-14
Normalized the computed pixel variation in Pixel Variation AOV by the user's noise threshold 

### LINKED ISSUE #2404: Normalize Pixel Variation AOV wrt. the user's noise threshold
Divide computed variation by the noise threshold set by the user such that AOV = 1 means variation = noise threshold.

kushWithoutWax:
Divided max_variation by noise threshold to fix #2404 
- The noise threshold can have a min value of 0.0001 and so did not check for non-zero value.

### DIFF
--- a/src/appleseed/renderer/modeling/aov/pixelvariationaov.cpp
+++ b/src/appleseed/renderer/modeling/aov/pixelvariationaov.cpp
@@ -94,6 +94,7 @@ namespace
             // Normalize if a maximum was found.
             if (max_variation != 0.0f)
             {
+                max_variation = max_variation / m_params.get_required<float>("noise_threshold", 1.0f);
                 for (size_t y = crop_window.min.y; y <= crop_window.max.y; ++y)
                 {
                     for (size_t x = crop_window.min.x; x <= crop_window.max.x; ++x)


dictoon:
I'm not sure the change is correct. How did you test it? Can you provide comparison images (before and after the change)?

kushWithoutWax:
The change is not correct. I noticed  an error while testing 'required parameter "noise_threshold" not found; continuing using value "1"' (which I had not noticed). 
Could you give me a heads up on how I could access the noise threshold value set by the user, within the ```PixelVariationAOV class ?

I noticed that PixelVariationAOV class has a ```ParamArray reference ```m_params which it inherits from the class ```renderer::Entity .
In the ```AdaptiveTileRenderer class, the ```noise_threshold has been accessed from a ParamArray reference using the ```get_required  function. Apparently, I cannot do the same in the ```PixelVariationAOV class.  

The reason I cannot figure out how to access the noise_threshold is because I also cannot figure out where the actual ParamArray object is passed( to a frame object maybe).

dictoon:
`PixelVariationAOV` itself doesn't have any params so `m_params` is empty. The noise threshold is a parameter of the Adaptive Tile Renderer. You'll need to pass the parameter to the AOV somehow.

Maybe the Adaptive Tile Renderer should detect the Pixel Variation AOV and inject the parameter into it. I'm not 100% sure what's the best approach here and I'm open to suggestions.

PS: use _a single backtick around identifiers_ to mark them as code, your last comment is rather screwed up :)

oktomus:
Injecting parameters to the AOv is amready do for the PixelSampleCount AOV in adaptive renderders. It does not set m_params but it sets a normalization range for future post processing. It works well and im ok with that.
But im not sure how the AOV will look.

kushWithoutWax:
Injecting the noise threshold parameter into the `PixelVariationAOV` can be done in the same way.
There is one problem which is that the `PixelVariationAOV` class is not declared in the `pixelvariationaov.h` file and it is defined within an unnamed namespace which makes it impossible to create its pointer in the `AdaptiveTileRenderer` class or to access any of its methods, I think.

So, the simple solution could be to declare the `PixelVariationAOV` class in the `pixelvariationaov.h` file within the namespace `renderer` and to have a public function `set_noise_threshold ` that could set a noise threshold variable in the class. The `set_noise_threshold ` function could be called in the `AdaptiveTileRenderer` class.  
Is this correct?
Or is there a reason why the `PixelVariationAOV` class cannot be declared in the `renderer` namespace?

dictoon:
> the simple solution could be to declare the PixelVariationAOV class in the pixelvariationaov.h file within the namespace renderer and to have a public function set_noise_threshold that could set a noise threshold variable in the class. The set_noise_threshold function could be called in the AdaptiveTileRenderer class.

That's a good solution.

I would also suggest a change in `AdaptiveTileRenderer`'s constructor: instead of finding AOVs by name, we should find them by model (using `get_model()`) and then cast those of known models and inject the right parameters. Does that make sense?

kushWithoutWax:
Sorry for the late reply.

>I would also suggest a change in AdaptiveTileRenderer's constructor: instead of finding AOVs by name, we should find them by model (using get_model()) and then cast those of known models and inject the right parameters.

I don't exactly understand what you mean by this. 

As far as I've seen, I have to, in any case, search for the corresponding AOVs within the `AOVContainer` object of the `frame` object and obtain an index so that I could get the aov by index.

Do you mean that I should obtain the corresponding AOVs by the `get_by_name` function on the `AOVContainer` object? (that also calls the `get_index` function, so I don't see the difference it makes).
 
I saw that `get_model` would return me the name of the aov. 
So I could iterate through the `AOVContainer` object, get each aov , call `get_model` on them to check if it is the right AOV or not and then inject parameters. Is that what you mean?

oktomus:
It's not related to @FrancoisBeaune's suggestion but here is an example of how you can send informations to the AOV
https://github.com/appleseedhq/appleseed/blob/master/src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp#L260

kushWithoutWax:
@oktomus I am planning to do something like that. But then I would need to have a pointer to the PixelVariationAOV object for that, for which one would need to find it's index using the AOV's name(which is already done ). 
@dictoon said something here about not finding the AOV's by name but by their model (using `get_model()` ). This is what I do not understand.

PR #2479, opened 2019-03-15
Extract ColorMap class from ColorMapPostProcessingStage and use it for diagnostic AOVs

### LINKED ISSUE #2249: Extract ColorMap class from ColorMapPostProcessingStage and use it for diagnostic AOVs

achalpandeyy:
Fixes #2249 

I wasn't sure if we want to use the `ColorMap` class for `ColorMapPostProcessingStage` as well.
Currently, `ColorMap` is only used in the scalar diagnostic AOVs.

### DIFF
--- a/src/appleseed/CMakeLists.txt
+++ b/src/appleseed/CMakeLists.txt
@@ -136,6 +136,10 @@ set (foundation_image_sources
     foundation/image/analysis.h
     foundation/image/canvasproperties.h
     foundation/image/color.h
+    foundation/image/colormap.cpp
+    foundation/image/colormap.h
+    foundation/image/colormapdata.cpp
+    foundation/image/colormapdata.h
     foundation/image/colorspace.cpp
     foundation/image/colorspace.h
     foundation/image/drawing.cpp
@@ -1884,10 +1888,6 @@ source_group ("renderer\\modeling\\object" FILES
 )
 
 set (renderer_modeling_postprocessingstage_sources
-    renderer/modeling/postprocessingstage/colormap.cpp
-    renderer/modeling/postprocessingstage/colormap.h
-    renderer/modeling/postprocessingstage/colormapdata.cpp
-    renderer/modeling/postprocessingstage/colormapdata.h
     renderer/modeling/postprocessingstage/colormappostprocessingstage.cpp
     renderer/modeling/postprocessingstage/colormappostprocessingstage.h
     renderer/modeling/postprocessingstage/ipostprocessingstagefactory.cpp
--- a/src/appleseed/CMakeLists.txt
+++ b/src/appleseed/CMakeLists.txt
@@ -1884,6 +1884,8 @@ source_group ("renderer\\modeling\\object" FILES
 )
 
 set (renderer_modeling_postprocessingstage_sources
+    renderer/modeling/postprocessingstage/colormap.cpp
+    renderer/modeling/postprocessingstage/colormap.h
     renderer/modeling/postprocessingstage/colormapdata.cpp
     renderer/modeling/postprocessingstage/colormapdata.h
     renderer/modeling/postprocessingstage/colormappostprocessingstage.cpp
--- a/src/appleseed/foundation/image/colormap.cpp
+++ b/src/appleseed/foundation/image/colormap.cpp
@@ -29,14 +29,20 @@
 // Interface header.
 #include "colormap.h"
 
+// appleseed.foundation headers.
+#include "foundation/image/colorspace.h"
+#include "foundation/image/genericimagefilereader.h"
+
 // Standard headers.
 #include <cassert>
 #include <cstddef>
+#include <limits>
+#include <string>
 
 namespace foundation
 {
 
-void ColorMap::set_palette(const float* values, const size_t entry_count)
+void ColorMap::set_palette_from_array(const float* values, const size_t entry_count)
 {
     m_palette.resize(entry_count);
 
@@ -50,21 +56,31 @@ void ColorMap::set_palette(const float* values, const size_t entry_count)
     }
 }
 
-void ColorMap::remap_colors(
-    const AABB2u&       crop_window,
-    Image*              image,
-    const float         max_val,
-    const float         min_val)
+void ColorMap::set_palette_from_image_file(const std::string& filepath)
 {
-    const float min_value = static_cast<float>(min_val);
-    const float max_value = 
-        max_val == 0
-            ? get_max_value(image, crop_window)
-            : static_cast<float>(max_val);
+    GenericImageFileReader reader;
+    std::unique_ptr<Image> image(reader.read(filepath.c_str()));
 
+    const size_t image_width = image->properties().m_canvas_width;
+    m_palette.resize(image_width);
+
+    for (size_t i = 0; i < image_width; ++i)
+        image->get_pixel(i, 0, m_palette[i]);
+}
+
+void ColorMap::remap_red_channel(
+    Image*              image,
+    const AABB2u&       crop_window,
+    const float         min_value,
+    const float         max_value) const
+{
     if (max_value == 0.0f)
     {
-        fill_aov(image, crop_window, evaluate_palette(0.0f));
+        for (size_t y = crop_window.min.y; y <= crop_window.max.y; ++y)
+        {
+            for (size_t x = crop_window.min.x; x <= crop_window.max.x; ++x)
+                image->set_pixel(x, y, evaluate_palette(0.0f));
+        }
         return;
     }
 
@@ -84,9 +100,40 @@ void ColorMap::remap_colors(
     }
 }
 
+void ColorMap::remap_relative_luminance(
+    Image*          image,
+    const float     min_luminance, 
+    const float     max_luminance) const
+{
+    if (min_luminance == max_luminance)
+    {
+        for_each_pixel(image, [this](Color4f& color)
+        {
+            color.rgb() = evaluate_palette(0.0f);
+        });
+    }
+
+    else
+    {
+        for_each_pixel(image, [this, min_luminance, max_luminance](Color4f& color)
+        {
+            const float col_luminance = luminance(color.rgb());
+
+            const float x =
+                saturate(
+                    inverse_lerp(
+                        min_luminance,
+                        max_luminance,
+                        col_luminance));
+
+                color.rgb() = evaluate_palette(x);
+        });
+    }
+}
+
 Color3f ColorMap::evaluate_palette(float x) const
 {
-    // When AOV contains a single channel we do not set the color palette.
+    // When image contains a single channel we do not set the color palette.
     if (m_palette.empty())
         return Color3f(x);
 
@@ -98,33 +145,42 @@ Color3f ColorMap::evaluate_palette(float x) const
     return lerp(m_palette[ix], m_palette[ix + 1], w);
 }
 
-float ColorMap::get_max_value(const Image* image, const AABB2u& crop_window)
+void ColorMap::find_min_max_red_channel(
+    const Image*    image,
+    const AABB2u&   crop_window,
+    float&          min_val,
+    float&          max_val)
 {
-    float max_value = 0.0f;
+    max_val = 0.0f;
+    min_val = 0.0f;
 
     for (size_t y = crop_window.min.y; y <= crop_window.max.y; ++y)
     {
         for (size_t x = crop_window.min.x; x <= crop_window.max.x; ++x)
         {
-            Color3f value;
-            image->get_pixel(x, y, value);
-            max_value = max(value[0], max_value);
+            Color3f val;
+            image->get_pixel(x, y, val);
+            max_val = max(val[0], max_val);
         }
     }
-
-    return max_value;
 }
 
-void ColorMap::fill_aov(
-    Image*          image,
-    const AABB2u&   crop_window,
-    const Color3f&  color)
+void ColorMap::find_min_max_relative_luminance(
+    Image*      image,
+    float&      min_luminance,
+    float&      max_luminance)
 {
-    for (size_t y = crop_window.min.y; y <= crop_window.max.y; ++y)
+    Color4f min_color = Color4f(+std::numeric_limits<float>::max());
+    Color4f max_color = Color4f(-std::numeric_limits<float>::max());
+
+    for_each_pixel(image, [&min_color, &max_color](Color4f& color)
     {
-        for (size_t x = crop_window.min.x; x <= crop_window.max.x; ++x)
-            image->set_pixel(x, y, color);
-    }
+        min_color = component_wise_min(min_color, color);
+        max_color = component_wise_max(max_color, color);
+    });
+
+    min_luminance = luminance(min_color.rgb());
+    max_luminance = luminance(max_color.rgb());
 }
 
 }   // namespace foundation
--- a/src/appleseed/foundation/image/colormap.cpp
+++ b/src/appleseed/foundation/image/colormap.cpp
@@ -46,6 +46,9 @@ void ColorMap::find_min_max_red_channel(
     float&          min_val,
     float&          max_val)
 {
+    min_val = std::numeric_limits<float>::max();
+    max_val = -std::numeric_limits<float>::max();
+    
     for_each_pixel(image, crop_window, [&min_val, &max_val](const Color4f& val)
     {
         min_val = min(val[0], min_val);
@@ -59,8 +62,8 @@ void ColorMap::find_min_max_relative_luminance(
     float&              min_luminance,
     float&              max_luminance)
 {
-    min_luminance = luminance(Color4f(+std::numeric_limits<float>::max()).rgb());
-    max_luminance = luminance(Color4f(-std::numeric_limits<float>::max()).rgb());
+    min_luminance = std::numeric_limits<float>::max();
+    max_luminance = -std::numeric_limits<float>::max();
 
     for_each_pixel(image, crop_window, [&min_luminance, &max_luminance](const Color4f& color)
     {
@@ -150,10 +153,6 @@ void ColorMap::remap_relative_luminance(
 
 Color3f ColorMap::evaluate_palette(float x) const
 {
-    // When image contains a single channel we do not set the color palette.
-    if (m_palette.empty())
-        return Color3f(x);
-
     x *= m_palette.size() - 1;
 
     const size_t ix = min(truncate<size_t>(x), m_palette.size() - 2);
--- a/src/appleseed/foundation/image/colormap.cpp
+++ b/src/appleseed/foundation/image/colormap.cpp
@@ -27,34 +27,35 @@
 //
 
 // Interface header.
-#include "renderer/modeling/postprocessingstage/colormap.h"
+#include "colormap.h"
 
 // Standard headers.
 #include <cassert>
 #include <cstddef>
 
-using namespace foundation;
-
-namespace renderer
+namespace foundation
 {
 
-ColorMap::ColorMap()
+void ColorMap::set_palette(const float* values, const size_t entry_count)
 {
-}
+    m_palette.resize(entry_count);
 
-void ColorMap::set_palette(std::vector<Color3f> palette)
-{
-    m_palette = palette;
+    for (size_t i = 0; i < entry_count; ++i)
+    {
+        m_palette[i] =
+            Color3f(
+                values[i * 3 + 0],
+                values[i * 3 + 1],
+                values[i * 3 + 2]);
+    }
 }
 
 void ColorMap::remap_colors(
-    const Frame&        frame,
+    const AABB2u&       crop_window,
     Image*              image,
     const float         max_val,
     const float         min_val)
 {
-    const AABB2u& crop_window = frame.get_crop_window();
-
     const float min_value = static_cast<float>(min_val);
     const float max_value = 
         max_val == 0
@@ -63,9 +64,7 @@ void ColorMap::remap_colors(
 
     if (max_value == 0.0f)
     {
-        if (!m_palette.empty())
-            fill_aov(image, crop_window, evaluate_palette(0.0f));
-        
+        fill_aov(image, crop_window, evaluate_palette(0.0f));
         return;
     }
 
@@ -75,31 +74,21 @@ void ColorMap::remap_colors(
     {
         for (size_t x = crop_window.min.x; x <= crop_window.max.x; ++x)
         {
-            if (!m_palette.empty())
-            {
-                Color3f color;
-                image->get_pixel(x, y, color);
+            Color3f value;
+            image->get_pixel(x, y, value);
 
-                const float c = saturate(fit(color[0], min_value, max_value, 0.0f, 1.0f));
+            const float c = saturate(fit(value[0], min_value, max_value, 0.0f, 1.0f));
 
-                image->set_pixel(x, y, evaluate_palette(c));
-            }
-            else
-            {
-                float val;
-                image->get_pixel(x, y, &val);
-
-                val = saturate(fit(val, min_value, max_value, 0.0f, 1.0f));
-
-                image->set_pixel(x, y, &val);
-            }
+            image->set_pixel(x, y, evaluate_palette(c));
         }
     }
 }
 
-Color3f ColorMap::evaluate_palette(float x)
+Color3f ColorMap::evaluate_palette(float x) const
 {
-    assert(m_palette.size() > 1);
+    // When AOV contains a single channel we do not set the color palette.
+    if (m_palette.empty())
+        return Color3f(x);
 
     x *= m_palette.size() - 1;
 
@@ -113,24 +102,15 @@ float ColorMap::get_max_value(const Image* image, const AABB2u& crop_window)
 {
     float max_value = 0.0f;
 
-        for (size_t y = crop_window.min.y; y <= crop_window.max.y; ++y)
+    for (size_t y = crop_window.min.y; y <= crop_window.max.y; ++y)
+    {
+        for (size_t x = crop_window.min.x; x <= crop_window.max.x; ++x)
         {
-            for (size_t x = crop_window.min.x; x <= crop_window.max.x; ++x)
-            {
-                if (!m_palette.empty())
-                {
-                    Color3f color;
-                    image->get_pixel(x, y, color);
-                    max_value = max(color[0], max_value);
-                }
-                else
-                {
-                    float value;
-                    image->get_pixel(x, y, &value);
-                    max_value = max(value, max_value);
-                }
-            }
+            Color3f value;
+            image->get_pixel(x, y, value);
+            max_value = max(value[0], max_value);
         }
+    }
 
     return max_value;
 }
@@ -147,4 +127,4 @@ void ColorMap::fill_aov(
     }
 }
 
-}   // namespace renderer
+}   // namespace foundation
--- a/src/appleseed/foundation/image/colormap.cpp
+++ b/src/appleseed/foundation/image/colormap.cpp
@@ -53,16 +53,6 @@ void ColorMap::find_min_max_red_channel(
     {
         max_val = max(val[0], max_val);
     });
-
-    //for (size_t y = crop_window.min.y; y <= crop_window.max.y; ++y)
-    //{
-    //    for (size_t x = crop_window.min.x; x <= crop_window.max.x; ++x)
-    //    {
-    //        Color3f val;
-    //        image.get_pixel(x, y, val);
-    //        max_val = max(val[0], max_val);
-    //    }
-    //}
 }
 
 void ColorMap::find_min_max_relative_luminance(
@@ -115,11 +105,10 @@ void ColorMap::remap_red_channel(
 {
     if (max_value == min_value)
     {
-        for (size_t y = crop_window.min.y; y <= crop_window.max.y; ++y)
+        for_each_pixel(image, crop_window, [this](Color4f& color)
         {
-            for (size_t x = crop_window.min.x; x <= crop_window.max.x; ++x)
-                image.set_pixel(x, y, evaluate_palette(0.0f));
-        }
+            color.rgb() = evaluate_palette(0.0f);
+        });
     }
     else
     {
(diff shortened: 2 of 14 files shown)


src/appleseed/renderer/modeling/postprocessingstage/colormap.h
```diff
@@ -0,0 +1,73 @@
...
+
+class ColorMap
+{
+  public:
+    ColorMap();
+
+    void set_palette(std::vector<Color3f> palette);
+
+    void remap_colors(
+      const Frame&    frame,
```
dictoon: From the frame you only need the crop window, I suggest passing that instead. That will also allow moving that code to `foundation/`.

src/appleseed/renderer/modeling/postprocessingstage/colormap.h
```diff
@@ -0,0 +1,73 @@
...
+
+namespace renderer
+{
+
+class ColorMap
+{
+  public:
+    ColorMap();
+
+    void set_palette(std::vector<Color3f> palette);
```
dictoon: Pass by const ref instead of by value (which causes an unneeded copy).

src/appleseed/renderer/modeling/postprocessingstage/colormap.h
```diff
@@ -0,0 +1,73 @@
...
+
+using namespace foundation;
+
+namespace renderer
+{
+
+class ColorMap
+{
+  public:
+    ColorMap();
```
dictoon: As far as I can tell you don't need this constructor.

src/appleseed/renderer/modeling/postprocessingstage/colormap.h
```diff
@@ -0,0 +1,73 @@
...
+    void remap_colors(
+      const Frame&    frame,
+      Image*          image,
+      const float     max_val,
+      const float     min_val);
+
+  private:
+    Color3f evaluate_palette(float x);
+
+    float get_max_value(const Image* image, const AABB2u& crop_window);
```
dictoon: Can be static?

src/appleseed/renderer/modeling/postprocessingstage/colormap.h
```diff
@@ -0,0 +1,73 @@
...
+    void set_palette(std::vector<Color3f> palette);
+
+    void remap_colors(
+      const Frame&    frame,
+      Image*          image,
+      const float     max_val,
+      const float     min_val);
+
+  private:
+    Color3f evaluate_palette(float x);
```
dictoon: Must be const. Also mark `x` as const.

src/appleseed/renderer/modeling/postprocessingstage/colormap.h
```diff
@@ -0,0 +1,73 @@
...
+class ColorMap
+{
+  public:
+    ColorMap();
+
+    void set_palette(std::vector<Color3f> palette);
+
+    void remap_colors(
+      const Frame&    frame,
+      Image*          image,
```
dictoon: Must be prefixed by `foundation::` (but that won't be needed anymore when those files will be moved to `foundation/`).

src/appleseed/CMakeLists.txt
```diff
@@ -1880,6 +1880,8 @@ source_group ("renderer\\modeling\\object" FILES
 )
 
 set (renderer_modeling_postprocessingstage_sources
+    renderer/modeling/postprocessingstage/colormap.cpp
```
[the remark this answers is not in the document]
achalpandeyy: Done.

src/appleseed/renderer/modeling/postprocessingstage/colormap.cpp
```diff
@@ -0,0 +1,150 @@
...
+    const float w = x - ix;
+
+    return lerp(m_palette[ix], m_palette[ix + 1], w);
+}
+
+float ColorMap::get_max_value(const Image* image, const AABB2u& crop_window)
+{
+    float max_value = 0.0f;
+
+        for (size_t y = crop_window.min.y; y <= crop_window.max.y; ++y)
```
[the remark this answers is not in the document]
achalpandeyy: Done.

src/appleseed/renderer/modeling/postprocessingstage/colormap.h
achalpandeyy: Done.

src/appleseed/renderer/modeling/postprocessingstage/colormap.cpp
```diff
@@ -0,0 +1,150 @@
...
+        {
+            for (size_t x = crop_window.min.x; x <= crop_window.max.x; ++x)
+            {
+                if (!m_palette.empty())
```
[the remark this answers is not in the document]
achalpandeyy: You are right, this wasn't necessary. My bad.

src/appleseed/renderer/modeling/postprocessingstage/colormap.h
achalpandeyy: Removed.

src/appleseed/renderer/modeling/postprocessingstage/colormap.h
achalpandeyy: Yes. Done.

src/appleseed/renderer/modeling/postprocessingstage/colormap.h
achalpandeyy: Done.

Is marking `x` as const necessary? I modify it in the function's implementation.
`x *= m_palette.size() - 1;`

src/appleseed/renderer/modeling/postprocessingstage/colormap.h
achalpandeyy: Yes.

src/appleseed/renderer/modeling/postprocessingstage/colormap.h
achalpandeyy: Done.

src/appleseed/renderer/modeling/postprocessingstage/colormap.cpp
```diff
@@ -0,0 +1,150 @@
...
+                {
+                    Color3f color;
+                    image->get_pixel(x, y, color);
+                    max_value = max(color[0], max_value);
```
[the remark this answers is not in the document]
achalpandeyy: Here `color[0]` is not the red channel but spp in PixelSampleCountAOV, pixel time in PixelTimeAOV and variance in PixelVariationAOV, as you mentioned on Discord.
To illustrate this I changed `color` to `values` in the recent commit.

src/appleseed/renderer/modeling/postprocessingstage/colormap.h
dictoon: Ah, in that case it's fine!

src/appleseed/renderer/modeling/aov/pixelsamplecountaov.cpp
```diff
@@ -72,90 +74,12 @@ const char* PixelSampleCountAOV::get_model() const
...
-            m_image->get_pixel(x, y, color);
-
-            const float c = saturate(fit(color[0], min_spp, max_spp, 0.0f, 1.0f));
-
-            m_image->set_pixel(x, y, lerp(Blue, Red, c));
-        }
-    }
+    ColorMap color_map;
+    color_map.set_palette(InfernoColorMap, countof(InfernoColorMap) / 3);
+    color_map.remap_colors(crop_window, m_image, m_min_spp, m_max_spp);
```
dictoon: The argument ordering you chose for `remap_colors()` is so evil that even you fell in the trap here :)

src/appleseed/foundation/image/colormap.cpp
```diff
@@ -0,0 +1,130 @@
...
+{
+    float max_value = 0.0f;
+
+    for (size_t y = crop_window.min.y; y <= crop_window.max.y; ++y)
+    {
+        for (size_t x = crop_window.min.x; x <= crop_window.max.x; ++x)
+        {
+            Color3f value;
+            image->get_pixel(x, y, value);
+            max_value = max(value[0], max_value);
```
dictoon: We cannot just consider the red channel of the image in a class that is supposed to be general-purpose. Let's discuss our options on Discord.

achalpandeyy:
> Nice refactoring and attention to details. Beside a couple minor remarks, there's a slightly more problematic design issue, let's discuss our options on Discord.

Thanks for the review!
I'll make the changes ASAP.

src/appleseed/foundation/image/colormap.h
```diff
@@ -0,0 +1,66 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+#pragma once
+
+// appleseed.foundation headers.
+#include "foundation/image/color.h"
+#include "foundation/image/image.h"
```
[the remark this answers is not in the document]
achalpandeyy: Done.

src/appleseed/foundation/image/colormap.h
```diff
@@ -0,0 +1,66 @@
...
+      Image*            image,
+      const float       max_val,
+      const float       min_val);
+
+  private:
+    Color3f evaluate_palette(float x) const;
+
+    static float get_max_value(const Image* image, const AABB2u& crop_window);
+
+    static void fill_aov(
```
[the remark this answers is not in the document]
achalpandeyy: Done.

src/appleseed/foundation/image/colormap.h
```diff
@@ -0,0 +1,66 @@
...
+namespace foundation
+{
+
+class ColorMap
+{
+  public:
+    void set_palette(const float* values, const size_t entry_count);
+
+    void remap_colors(
+      const AABB2u&     crop_window,
```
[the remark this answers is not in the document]
achalpandeyy: Done 😅

src/appleseed/foundation/image/colormap.h
```diff
@@ -0,0 +1,66 @@
...
+      Image*            image,
+      const float       max_val,
+      const float       min_val);
+
+  private:
+    Color3f evaluate_palette(float x) const;
+
+    static float get_max_value(const Image* image, const AABB2u& crop_window);
```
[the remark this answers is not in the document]
achalpandeyy: Done.

src/appleseed/renderer/modeling/aov/pixelsamplecountaov.cpp
achalpandeyy: Damn! 😆

src/appleseed/foundation/image/colormap.cpp
achalpandeyy: Done.

src/appleseed/foundation/image/colormap.cpp
```diff
@@ -0,0 +1,130 @@
...
+    }
+}
+
+void ColorMap::remap_colors(
+    const AABB2u&       crop_window,
+    Image*              image,
+    const float         max_val,
+    const float         min_val)
+{
+    const float min_value = static_cast<float>(min_val);
```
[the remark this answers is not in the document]
achalpandeyy: Done.

src/appleseed/foundation/image/colormap.cpp
```diff
@@ -0,0 +1,187 @@
...
+    const float     max_luminance) const
+{
+    if (min_luminance == max_luminance)
+    {
+        for_each_pixel(image, [this](Color4f& color)
+        {
+            color.rgb() = evaluate_palette(0.0f);
+        });
+    }
+
```
dictoon: Remove blank line.

src/appleseed/foundation/image/colormap.h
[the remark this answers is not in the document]
dictoon: Hm, I still see the include. Or am I not looking at the latest version?

src/appleseed/renderer/modeling/camera/perspectivecamera.cpp
```diff
@@ -36,6 +36,7 @@
 #include "renderer/utility/paramarray.h"
 
 // appleseed.foundation headers.
+#include "foundation/image/image.h"
```
dictoon: Good catch. It's also missing `foundation/image/canvasproperties.h`.

src/appleseed/renderer/modeling/postprocessingstage/postprocessingstage.h
```diff
@@ -33,18 +33,11 @@
 #include "renderer/modeling/frame/frame.h"
```
dictoon: `Frame` can now be forward-declared.

src/appleseed/foundation/image/colormap.h
```diff
@@ -0,0 +1,110 @@
...
+#include <vector>
+
+namespace foundation
+{
+
+class ColorMap
+{
+  public:
+  	static void find_min_max_red_channel(
+		const Image*    image,
```
dictoon: All these methods should take an `Image&` instead of an `Image*` because you don't want `image` to be `nullptr`.

src/appleseed/foundation/image/colormap.h
```diff
@@ -0,0 +1,110 @@
...
+		float&          max_val);
+
+	static void find_min_max_relative_luminance(
+		Image*    	image,
+		float&  	min_luminance,
+		float&    	max_luminance);
+
+    void set_palette_from_array(const float* values, const size_t entry_count);
+
+    void set_palette_from_image_file(const std::string& filepath);
```
dictoon: Include `<string>`.

src/appleseed/foundation/image/colormap.h
```diff
@@ -0,0 +1,110 @@
...
+	{
+    	for (size_t tx = 0; tx < image_props.m_tile_count_x; ++tx)
+    	{
+        	foundation::Tile& tile = image->tile(tx, ty);
+
+        	for (size_t y = 0, th = tile.get_height(); y < th; ++y)
+        	{
+                for (size_t x = 0, tw = tile.get_width(); x < tw; ++x)
+                {
+                	foundation::Color4f color;
```
dictoon: Fix indentation.

src/appleseed/foundation/image/colormap.h
```diff
@@ -0,0 +1,110 @@
...
+		const AABB2u&   crop_window,
+		float&          min_val,
+		float&          max_val);
+
+	static void find_min_max_relative_luminance(
+		Image*    	image,
+		float&  	min_luminance,
+		float&    	max_luminance);
+
+    void set_palette_from_array(const float* values, const size_t entry_count);
```
dictoon: Include `<cstddef>`.

src/appleseed/foundation/image/colormap.cpp
```diff
@@ -0,0 +1,187 @@
...
+    {
+        m_palette[i] =
+            Color3f(
+                values[i * 3 + 0],
+                values[i * 3 + 1],
+                values[i * 3 + 2]);
+    }
+}
+
+void ColorMap::set_palette_from_image_file(const std::string& filepath)
```
dictoon: This method doesn't belong in this class whose sole purpose is to remap colors given a color palette, not to do I/O of any kind. Move it back to `ColorMapPostProcessingStage` (and remove header files that are no longer necessary).

src/appleseed/foundation/image/colormap.cpp
```diff
@@ -0,0 +1,187 @@
...
+        image->get_pixel(i, 0, m_palette[i]);
+}
+
+void ColorMap::remap_red_channel(
+    Image*              image,
+    const AABB2u&       crop_window,
+    const float         min_value,
+    const float         max_value) const
+{
+    if (max_value == 0.0f)
```
dictoon: This isn't quite correct, it should be

    if (min_value == max_value)

src/appleseed/foundation/image/colormap.cpp
```diff
@@ -0,0 +1,187 @@
...
+    {
+        for (size_t y = crop_window.min.y; y <= crop_window.max.y; ++y)
+        {
+            for (size_t x = crop_window.min.x; x <= crop_window.max.x; ++x)
+                image->set_pixel(x, y, evaluate_palette(0.0f));
+        }
+        return;
```
dictoon: Replace by `else`.

src/appleseed/foundation/image/colormap.cpp
```diff
@@ -0,0 +1,187 @@
...
+    assert(max_value > min_value);
+
+    for (size_t y = crop_window.min.y; y <= crop_window.max.y; ++y)
+    {
+        for (size_t x = crop_window.min.x; x <= crop_window.max.x; ++x)
+        {
+            Color3f value;
+            image->get_pixel(x, y, value);
+
+            const float c = saturate(fit(value[0], min_value, max_value, 0.0f, 1.0f));
```
dictoon: Include `foundation/math/scalar.h`.

src/appleseed/foundation/image/colormap.h
```diff
@@ -0,0 +1,110 @@
...
+class ColorMap
+{
+  public:
+  	static void find_min_max_red_channel(
+		const Image*    image,
+		const AABB2u&   crop_window,
+		float&          min_val,
+		float&          max_val);
+
+	static void find_min_max_relative_luminance(
```
dictoon: This method should also take a crop window. We want the API to be as orthogonal as possible, especially in appleseed.foundation, even if we don't happen to use the crop window when we remap luminance (why?).

src/appleseed/foundation/image/colormap.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+void ColorMap::remap_red_channel(
+    Image*              image,
+    const AABB2u&       crop_window,
+    const float         min_value,
+    const float         max_value) const
+{
+    if (max_value == 0.0f)
+    {
+        for (size_t y = crop_window.min.y; y <= crop_window.max.y; ++y)
```
dictoon: Use `for_each_pixel()`, to which you should pass the crop window.

src/appleseed/foundation/image/colormap.cpp
```diff
@@ -0,0 +1,187 @@
...
+#include <cassert>
+#include <cstddef>
+#include <limits>
+#include <memory>
+#include <string>
+
+namespace foundation
+{
+
+void ColorMap::set_palette_from_array(const float* values, const size_t entry_count)
```
dictoon: Follow the same order for method definitions as for declarations.

src/appleseed/renderer/modeling/aov/pixeltimeaov.cpp
```diff
@@ -181,38 +183,11 @@ namespace
         void post_process_image(const Frame& frame) override
         {
             const AABB2u& crop_window = frame.get_crop_window();
+            ColorMap color_map;
```
dictoon: Did you test this code? As far as I can tell, it won't work anymore since you're not loading a palette.

Please make sure you test all the code you modified.

src/appleseed/renderer/modeling/aov/pixelvariationaov.cpp
```diff
@@ -73,50 +75,13 @@ namespace
...
-                for (size_t y = crop_window.min.y; y <= crop_window.max.y; ++y)
-                {
-                    for (size_t x = crop_window.min.x; x <= crop_window.max.x; ++x)
-                    {
-                        m_image->set_pixel(x, y, Blue);
-                    }
-                }
-            }
+            float min_var, max_var;
+            color_map.find_min_max_red_channel(&frame.image(), crop_window, min_var, max_var);
```
dictoon: You're using `&frame.image()` here and `m_image` on the next line.

achalpandeyy:
Thanks for the review! I'll fix everything ASAP.

src/appleseed/foundation/image/colormap.h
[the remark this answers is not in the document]
achalpandeyy: You are looking at the recent version. The reason I included it back is that it was required for `for_each_pixel`.
I saw your comment about using `crop_window` for `remap_relative_luminance`. After making that change I'll deal with it accordingly. :)

src/appleseed/renderer/modeling/camera/perspectivecamera.cpp
achalpandeyy: Credit to the compiler.

src/appleseed/foundation/image/colormap.h
achalpandeyy: Right.

src/appleseed/renderer/modeling/aov/pixelvariationaov.cpp
achalpandeyy: Is there a problem? As far as I can tell both points to the same image object.
Regardless, I changed them both to `*m_image` (function takes in ref now) for consistency's sake.

src/appleseed/foundation/image/colormap.h
achalpandeyy: `remap_relative_luminance` should use the crop window. The whole frame was used in `ColorMapPostProcessingStage` so I didn't think about it much.

When I used the whole frame instead of just the crop window and selected a specific render region, everything outside that region which is supposed to be black turned purple-ish (in case of Inferno color map) which I believe is not the intended behavior. However, using the crop window fixes this which obviously makes sense and I think should be the correct behavior all along.

src/appleseed/renderer/modeling/aov/pixelvariationaov.cpp
dictoon: No problem, just a lack of consistency :)

src/appleseed/renderer/modeling/aov/pixeltimeaov.cpp
achalpandeyy: Yes, I tested this code for all three AOVs and Color map post-processing stage.
Moreover, this part of the code hasn't been changed _logically_ from the previous round of review. We didn't set the palette then and we didn't set it now. In fact, I even included a comment in `evaluate_palette` method regarding this.
`// When image contains a single channel we do not set the color palette.
    if (m_palette.empty())
        return Color3f(x);`

If you were expecting a segmentation fault because of this line `m_palette.empty()`, I tested it in a dummy C++ program, it worked so I put it there. I think this is because of `std::vector`'s default constructor gets called.

Moreover, the way things are implemented now in `colormappostprocessingstage.cpp`, we have a private member variable (`m_palette`) which is a `vector<Color3f>` and isn't explicitly initialized anywhere but we do call `resize()` on it when we set the palette (`set_palette_from_array`, `set_palette_from_image_file`) which I think is not much different than this line `m_palette.empty()`.

Sorry, but is there something I'm not getting?

I'm too anxious to push untested code even if the only thing a I change is a comment. 😅

src/appleseed/foundation/image/colormap.h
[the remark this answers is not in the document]
achalpandeyy: `Image` is being used for `for_each_pixel` method to get/set pixel color.

src/appleseed/renderer/modeling/aov/pixelsamplecountaov.cpp
```diff
@@ -72,90 +74,15 @@ const char* PixelSampleCountAOV::get_model() const
...
 void PixelSampleCountAOV::post_process_image(const Frame& frame)
 {
-    static const Color3f Blue(0.0f, 0.0f, 1.0f);
-    static const Color3f Red(1.0f, 0.0f, 0.0f);
-
     const AABB2u& crop_window = frame.get_crop_window();
+    ColorMap color_map;
+    color_map.set_palette_from_array(InfernoColorMap, countof(InfernoColorMap) / 3);
 
-    //
```
dictoon: You've removed an important bit of logic here... There are two ways to set the max SPP, you need to keep them...

src/appleseed/foundation/image/colormap.h
```diff
@@ -0,0 +1,109 @@
...
+void ColorMap::for_each_pixel(
+	Image& 			image,
+	const AABB2u& 	crop_window,
+	const Func& 	func)
+{
+	for (size_t y = crop_window.min.y; y <= crop_window.max.y; ++y)
+    {
+        for (size_t x = crop_window.min.x; x <= crop_window.max.x; ++x)
+        {
+			foundation::Color3f color;
```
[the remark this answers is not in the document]
achalpandeyy: Right.

src/appleseed/renderer/modeling/aov/pixelsamplecountaov.cpp
achalpandeyy: Yep you are right, we basically cast `m_max_spp` to float and set it to `max_spp`, if `m_max_spp` is 0 then we call `find_min_max` / `get_max_spp` methods.
Also, I set `min_spp` to 0 whereas it should be float casted `m_min_spp`.
Saw artifacts in the AOV because of this. Thanks!

src/appleseed/foundation/image/colormap.cpp
```diff
@@ -0,0 +1,170 @@
...
+    const AABB2u&       crop_window,
+    float&              min_luminance,
+    float&              max_luminance)
+{
+    Color3f min_color = Color3f(+std::numeric_limits<float>::max());
+    Color3f max_color = Color3f(-std::numeric_limits<float>::max());
+
+    for_each_pixel(image, crop_window, [&min_color, &max_color](Color3f& color)
+    {
+        min_color = component_wise_min(min_color, color);
```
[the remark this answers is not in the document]
achalpandeyy: Right, makes more sense.

achalpandeyy:
Addressed all the issues, please take a look.

src/appleseed/foundation/image/colormap.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+namespace foundation
+{
+
+void ColorMap::find_min_max_red_channel(
+    Image&          image,
+    const AABB2u&   crop_window,
+    float&          min_val,
+    float&          max_val)
+{
```
dictoon: You're not initializing `min_val` and `max_val` here, which is a problem.

src/appleseed/foundation/image/colormap.cpp
```diff
@@ -0,0 +1,165 @@
...
+    });
+}
+
+void ColorMap::find_min_max_relative_luminance(
+    Image&              image,
+    const AABB2u&       crop_window,
+    float&              min_luminance,
+    float&              max_luminance)
+{
+    min_luminance = luminance(Color4f(+std::numeric_limits<float>::max()).rgb());
```
dictoon: It would make more sense to directly initialize `min_luminance` and `max_luminance` to +/- `std::numeric_limits<float>::max()` rather than go through the `luminance()` function.

src/appleseed/renderer/modeling/aov/pixeltimeaov.cpp
```diff
@@ -181,38 +184,12 @@ namespace
...
 
-            // Find the maximum value.
+            float min_time = numeric_limits<float>::max();
```
dictoon: Once `find_min_max_red_channel()` properly initializes the min/max values, you will not need to initialize them here anymore.

src/appleseed/renderer/modeling/aov/pixelvariationaov.cpp
```diff
@@ -73,50 +76,14 @@ namespace
...
-            {
-                for (size_t y = crop_window.min.y; y <= crop_window.max.y; ++y)
-                {
-                    for (size_t x = crop_window.min.x; x <= crop_window.max.x; ++x)
-                    {
-                        m_image->set_pixel(x, y, Blue);
-                    }
-                }
-            }
+            float min_var = numeric_limits<float>::max();
```
dictoon: Same remark.

src/appleseed/renderer/modeling/aov/pixelsamplecountaov.cpp
```diff
@@ -72,90 +74,20 @@ const char* PixelSampleCountAOV::get_model() const
...
+    color_map.set_palette_from_array(InfernoColorMap, countof(InfernoColorMap) / 3);
+
+    float min_spp = static_cast<float>(m_min_spp);
+    float max_spp = 0.0f;
+    if (m_max_spp == 0)
+        color_map.find_min_max_red_channel(*m_image, crop_window, min_spp, max_spp);
+    else
+        max_spp = static_cast<float>(m_max_spp);
 
-    //
```
dictoon: Can you please keep this comment (and update it if needed)? It's quite important as it explains the logic behind the determination of max SPP.

src/appleseed/renderer/modeling/aov/pixelsamplecountaov.cpp
```diff
@@ -72,90 +74,20 @@ const char* PixelSampleCountAOV::get_model() const
...
 void PixelSampleCountAOV::post_process_image(const Frame& frame)
 {
-    static const Color3f Blue(0.0f, 0.0f, 1.0f);
-    static const Color3f Red(1.0f, 0.0f, 0.0f);
-
     const AABB2u& crop_window = frame.get_crop_window();
+    ColorMap color_map;
+    color_map.set_palette_from_array(InfernoColorMap, countof(InfernoColorMap) / 3);
+
+    float min_spp = static_cast<float>(m_min_spp);
```
dictoon: Same remark here.

src/appleseed/renderer/modeling/aov/pixeltimeaov.cpp
```diff
@@ -181,38 +184,12 @@ namespace
...
-                {
-                    float c;
-                    m_image->get_pixel(i, j, &c);
-
-                    c *= rcp_max_time;
-
-                    m_image->set_pixel(i, j, &c);
-                }
-            }
+            color_map.find_min_max_red_channel(*m_image, crop_window, min_time, max_time);
```
dictoon: I'm sorry but I've raised this point at twice already and you seem to ignore it:

**You are not loading a palette into the color map here.** I just don't understand how it can work. Am I missing something? (It can definitely happen.)

Can you please post renders for all the AOVS you modified, in particular Pixel Time?

src/appleseed/renderer/modeling/aov/pixelsamplecountaov.cpp
achalpandeyy: Sure :)

src/appleseed/renderer/modeling/aov/pixeltimeaov.cpp
achalpandeyy: There was a slight confusion, addressed on Discord.

achalpandeyy:
Addressed all the issues. Please take a look.

src/appleseed/foundation/image/colormap.cpp
```diff
@@ -150,10 +153,6 @@ void ColorMap::remap_relative_luminance(
 
 Color3f ColorMap::evaluate_palette(float x) const
 {
-    // When image contains a single channel we do not set the color palette.
-    if (m_palette.empty())
-        return Color3f(x);
-
     x *= m_palette.size() - 1;
```
dictoon: We should assert that `m_palette.size() >= 2`.

achalpandeyy:
All done, now it should be good enough.
Thank you for your awesome reviews! Glad I could help.

PR #2483, opened 2019-03-16
Initialize slab_normal to fix compilation warning

LZaw:
Leaving slab_normal uninitialized causes a warning (respectively error when compiling in release) when calling `sample_direction_given_cosine` (line 338) and `dot` (line 344), even though it gets poisoned.


LZaw:
Figured this doesn't happen with GCC-4.8 which is the one you officially support.

dictoon:
It doesn't happen with gcc 4.8 but it does happen with newer versions so we need to somehow fix the warning.

Performance-wise I'm not too concerned about initializing `slab_normal` to 0, 0, 0. What I'm more concerned about is whether there's an actual bug in this code or not.

PR #2484, opened 2019-03-16
Render stamp scaling rework

BashPrince:
I reworked my solution in #2468.
The icon did indeed not look nice if downsampled to the original icon size. I now unpremultiply and re-premultiply the icon before and after the resizing. I also changed the logo tint and filter size so the end result would match the appearance as closely as possible. It's still not a pixel perfect match, but a lot better I think.

I now also scale the margin of the stamp so the correct original proportions between stamp contents and stamp border are kept at any scale factor.

beta
[original version]

new2
[improved solution (this branch)]

master
[initial solution (master)]

[1024x1024 with stamp scale 2.0x]


src/appleseed/renderer/modeling/postprocessingstage/renderstamppostprocessingstage.cpp
```diff
@@ -143,44 +144,48 @@ namespace
...
 
             // Calculate final icon size and scale the icon
             const ImageSpec icon_spec = m_icon.spec();
+            const float stamp_height = text_height + 2 * margin_v;
+            const float icon_height = 0.68f * stamp_height;
             const float aspect = static_cast<float>(icon_spec.width) / static_cast<float>(icon_spec.height);
             const ROI roi(
                 0,
-                static_cast<int>(m_scale_factor * aspect * DefaultIconHeight),
+                static_cast<int>(round(icon_height * aspect)),
```
dictoon: The `static_cast<int>()` is redundant, this should be enough:

    round<int>(icon_weight * aspect)

src/appleseed/renderer/modeling/postprocessingstage/renderstamppostprocessingstage.cpp
```diff
@@ -120,9 +121,9 @@ namespace
             const float font_height = 14.0f * m_scale_factor;
             const Color4f FontColor(srgb_to_linear_rgb(Color3f(0.9f, 0.9f, 0.9f)), 1.0f);   // linear RGB
             const Color4f BackgroundColor(0.0f, 0.0f, 0.0f, 0.95f);                         // linear RGB
-            const Color4f LogoTint(1.0f, 1.0f, 1.0f, 0.8f);                                 // linear RGB
-            const float MarginH = 6.0f;
-            const float MarginV = 4.0f;
+            const Color4f LogoTint(0.85f, 0.85f, 0.85f, 0.8f);                              // linear RGB
+            const float margin_h = 6.0f * m_scale_factor;
```
dictoon: Good catch!

src/appleseed/renderer/modeling/postprocessingstage/renderstamppostprocessingstage.cpp
```diff
@@ -120,9 +121,9 @@ namespace
             const float font_height = 14.0f * m_scale_factor;
             const Color4f FontColor(srgb_to_linear_rgb(Color3f(0.9f, 0.9f, 0.9f)), 1.0f);   // linear RGB
             const Color4f BackgroundColor(0.0f, 0.0f, 0.0f, 0.95f);                         // linear RGB
-            const Color4f LogoTint(1.0f, 1.0f, 1.0f, 0.8f);                                 // linear RGB
-            const float MarginH = 6.0f;
-            const float MarginV = 4.0f;
+            const Color4f LogoTint(0.85f, 0.85f, 0.85f, 0.8f);                              // linear RGB
```
dictoon: I'm not sure what this is needed. The icons in `resources/logo/` are the canonical ones. 

In the final render with render stamp, the orange should be `#FF5A11` (255, 90, 17) and the gray should be `#C8C8C8` (200, 200, 200).

You should be able to get exactly that without any tinting.

dictoon: Note that these colors are expressed in sRGB.

dictoon:
You probably want to rebase this PR on top of master.

BashPrince:
I introduced the changed tint because the aliasing from the resizing made the end result look a bit brighter. Without changing the tint, for larger scales the icon colors in the center regions are now exactly where they are supposed to be. For the 1x scale the pixel values are marginally off (like 1 or 2 8bit RGB values), this is just caused by the heavy downsampling though I think. The aliasing still looks slightly different for scale 1x than the original. I've experimented quite a bit with the OpenImage filters and different widths now, this is as close as I can get it with this approach.

orig
[Original version]


[New at Scale 1.0x]


[Scale 10.0x]

BashPrince:
I'll try a rebase too, but I still have to read up on what to do exactly :'D

dictoon:
Results are definitely satisfactory!

dictoon:
GitHub was able to merge this PR without rebasing, that said you should read about it and experiment with it as it's a powerful tool.

Thanks for the quality work!

PR #2486, opened 2019-03-17
Improve rendering error analysis with fixes and AOV

wilimitis:
Please see individual commit comments for related implementation notes.

**Pixel Error AOV**
See https://www.ci.i.u-tokyo.ac.jp/~hachisuka/mbdpt.pdf for an elaborate reference example.

Generally useful for understanding the behavior of sampling or light transport optimizations. For instance, we would expect there to be lower error near hard to reach specular spots for some bidirectional methods (sppm, bdpt, ...). An obvious next step would be to enable comparing two renders side-by-side normalized against the highest error of a single render (or allow an upper bound param), but this might be better suited as a python script.

Directly relevant to the discussion in #2465 regarding the behavior of computing tile noise in sRGB space, where we might expect to see less variance in darker areas and higher variance in lighter areas for sRGB.

Usage
```
...
<frame name="beauty">
  ...
  <parameter name="reference_image" value="ref.exr" />
  <aovs>
    <aov model="pixel_error_aov" />
  </aovs>
</frame>
```

Left to right: SPPM reference, UDPT render
Screen Shot 2019-03-15 at 8 36 17 AM

Top to bottom: Linear RGB adaptive tiling, sRGB adaptive tiling
Untitled Diagram

### DIFF
--- a/src/appleseed/foundation/image/analysis.cpp
+++ b/src/appleseed/foundation/image/analysis.cpp
@@ -125,27 +125,15 @@ bool are_images_compatible(const Image& image1, const Image& image2)
 
     return
         props1.m_canvas_width == props2.m_canvas_width &&
-        props1.m_canvas_height == props2.m_canvas_height &&
-        props1.m_tile_width == props2.m_tile_width &&
-        props1.m_tile_height == props2.m_tile_height &&
-        props1.m_channel_count == props2.m_channel_count;
+        props1.m_canvas_height == props2.m_canvas_height;
 }
 
-namespace
+double compute_error_squared(const Color3f& c1, const Color3f& c2)
 {
-    double sum_pixel_components(
-        const Tile&     tile,
-        const size_t    i)
-    {
-        double sum = 0.0;
-
-        const size_t channel_count = tile.get_channel_count();
-
-        for (size_t c = 0; c < channel_count; ++c)
-            sum += tile.get_component<double>(i, c);
-
-        return sum;
-    }
+    return
+        square(c1.r - c2.r) +
+        square(c1.g - c2.g) +
+        square(c1.b - c2.b);
 }
 
 double compute_rms_deviation(const Image& image1, const Image& image2)
@@ -157,28 +145,23 @@ double compute_rms_deviation(const Image& image1, const Image& image2)
 
     double mse = 0.0;   // mean square error
 
-    for (size_t ty = 0; ty < props.m_tile_count_y; ++ty)
+    for (size_t y = 0; y < props.m_canvas_height; ++y)
     {
-        for (size_t tx = 0; tx < props.m_tile_count_x; ++tx)
+        for (size_t x = 0; x < props.m_canvas_width; ++x)
         {
-            const Tile& tile1 = image1.tile(tx, ty);
-            const size_t tile_width = tile1.get_width();
-            const size_t tile_height = tile1.get_height();
+            float image_color[4];
+            image1.get_pixel(x, y, image_color);
 
-            const Tile& tile2 = image2.tile(tx, ty);
-            assert(tile2.get_width() == tile_width);
-            assert(tile2.get_height() == tile_height);
+            float ref_color[4];
+            image2.get_pixel(x, y, ref_color);
 
-            for (size_t i = 0; i < tile_width * tile_height; ++i)
-            {
-                const double sum1 = sum_pixel_components(tile1, i);
-                const double sum2 = sum_pixel_components(tile2, i);
-                mse += square(sum1 - sum2);
-            }
+            mse += compute_error_squared(
+                Color3f::from_array(image_color),
+                Color3f::from_array(ref_color));
         }
     }
 
-    mse /= props.m_pixel_count * square(props.m_channel_count);
+    mse /= props.m_pixel_count * 3.0f;
 
     return sqrt(mse);
 }
--- a/src/appleseed/foundation/image/analysis.cpp
+++ b/src/appleseed/foundation/image/analysis.cpp
@@ -141,15 +141,15 @@ double compute_rms_deviation(const Image& image1, const Image& image2)
     {
         for (size_t x = 0; x < props.m_canvas_width; ++x)
         {
-            float image_color[4];
-            image1.get_pixel(x, y, image_color);
+            float color1[4];
+            image1.get_pixel(x, y, color1);
 
-            float ref_color[4];
-            image2.get_pixel(x, y, ref_color);
+            float color2[4];
+            image2.get_pixel(x, y, color2);
 
             mse += compute_error_squared(
-                Color3f::from_array(image_color),
-                Color3f::from_array(ref_color));
+                Color3f::from_array(color1),
+                Color3f::from_array(color2));
         }
     }
 
--- a/src/appleseed/foundation/image/analysis.cpp
+++ b/src/appleseed/foundation/image/analysis.cpp
@@ -128,14 +128,6 @@ bool are_images_compatible(const Image& image1, const Image& image2)
         props1.m_canvas_height == props2.m_canvas_height;
 }
 
-double compute_error_squared(const Color3f& c1, const Color3f& c2)
-{
-    return
-        square(c1.r - c2.r) +
-        square(c1.g - c2.g) +
-        square(c1.b - c2.b);
-}
-
 double compute_rms_deviation(const Image& image1, const Image& image2)
 {
     if (!are_images_compatible(image1, image2))
--- a/src/appleseed/foundation/image/analysis.h
+++ b/src/appleseed/foundation/image/analysis.h
@@ -30,6 +30,7 @@
 #pragma once
 
 // appleseed.foundation headers.
+#include "foundation/image/color.h"
 #include "foundation/core/exceptions/exception.h"
 
 // appleseed.main headers.
@@ -56,6 +57,9 @@ APPLESEED_DLLSYMBOL double compute_average_luminance(const Image& image);
 
 // Return whether two images are compatible and thus can be compared.
 APPLESEED_DLLSYMBOL bool are_images_compatible(const Image& image1, const Image& image2);
+
+// Compute the squared error between two linear rgb colors.
+double compute_error_squared(const Color3f& c1, const Color3f& c2);
 
 // Exception thrown by comparison functions when two images are incompatible.
 struct ExceptionIncompatibleImages
--- a/src/appleseed/foundation/image/analysis.h
+++ b/src/appleseed/foundation/image/analysis.h
@@ -58,9 +58,6 @@ APPLESEED_DLLSYMBOL double compute_average_luminance(const Image& image);
 // Return whether two images are compatible and thus can be compared.
 APPLESEED_DLLSYMBOL bool are_images_compatible(const Image& image1, const Image& image2);
 
-// Compute the squared error between two linear rgb colors.
-double compute_error_squared(const Color3f& c1, const Color3f& c2);
-
 // Exception thrown by comparison functions when two images are incompatible.
 struct ExceptionIncompatibleImages
   : public Exception
--- a/src/appleseed/foundation/image/analysis.h
+++ b/src/appleseed/foundation/image/analysis.h
@@ -30,8 +30,8 @@
 #pragma once
 
 // appleseed.foundation headers.
-#include "foundation/image/color.h"
 #include "foundation/core/exceptions/exception.h"
+#include "foundation/image/color.h"
 
 // appleseed.main headers.
 #include "main/dllsymbol.h"
--- a/src/appleseed/foundation/image/color.h
+++ b/src/appleseed/foundation/image/color.h
@@ -347,6 +347,7 @@ Color<T, 3> integer_to_color3(const Int i);
 template <typename T>
 T compute_error_squared(const Color<T, 3>& c1, const Color<T, 3>& c2);
 
+
 //
 // N-dimensional color implementation.
 //
--- a/src/appleseed/foundation/image/color.h
+++ b/src/appleseed/foundation/image/color.h
@@ -345,7 +345,7 @@ Color<T, 3> integer_to_color3(const Int i);
 
 // Compute the squared error between two linear rgb colors.
 template <typename T>
-T compute_error_squared(const Color<T, 3>& c1, const Color<T, 3>& c2);
+inline T compute_error_squared(const Color<T, 3>& c1, const Color<T, 3>& c2);
 
 
 //
--- a/src/appleseed/foundation/image/color.h
+++ b/src/appleseed/foundation/image/color.h
@@ -345,7 +345,7 @@ Color<T, 3> integer_to_color3(const Int i);
 
 // Compute the squared error between two linear rgb colors.
 template <typename T>
-inline T compute_error_squared(const Color<T, 3>& c1, const Color<T, 3>& c2);
+T compute_error_squared(const Color<T, 3>& c1, const Color<T, 3>& c2);
 
 
 //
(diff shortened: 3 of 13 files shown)


wilimitis:
Looks like I did mention them but didn't add the file to the last commit, thank you! 😄

dictoon:
> Left to right: SPPM reference, UDPT render

Looking at the left image, I can readily tell it wasn't rendered with SPPM :)

src/appleseed/foundation/image/analysis.cpp
```diff
@@ -157,28 +145,23 @@ double compute_rms_deviation(const Image& image1, const Image& image2)
...
+    for (size_t y = 0; y < props.m_canvas_height; ++y)
     {
-        for (size_t tx = 0; tx < props.m_tile_count_x; ++tx)
+        for (size_t x = 0; x < props.m_canvas_width; ++x)
         {
-            const Tile& tile1 = image1.tile(tx, ty);
-            const size_t tile_width = tile1.get_width();
-            const size_t tile_height = tile1.get_height();
+            float image_color[4];
+            image1.get_pixel(x, y, image_color);
```
dictoon: You can do the same with Color4f:

    Color4f color1;
    image1.get_pixel(x, y, color1);

Also, if you call the arguments of this function `image1` and `image2`, there's no such thing as a "ref image", just call the colors `color1` and `color2`.

src/appleseed/foundation/image/analysis.cpp
```diff
@@ -157,28 +145,23 @@ double compute_rms_deviation(const Image& image1, const Image& image2)
...
+            image2.get_pixel(x, y, ref_color);
 
-            for (size_t i = 0; i < tile_width * tile_height; ++i)
-            {
-                const double sum1 = sum_pixel_components(tile1, i);
-                const double sum2 = sum_pixel_components(tile2, i);
-                mse += square(sum1 - sum2);
-            }
+            mse += compute_error_squared(
+                Color3f::from_array(image_color),
```
dictoon: And then here you can use `color1.rgb()` etc.

src/appleseed/foundation/image/analysis.cpp
```diff
@@ -157,28 +145,23 @@ double compute_rms_deviation(const Image& image1, const Image& image2)
 
     double mse = 0.0;   // mean square error
```
dictoon: I think making `mse` a double is a very sensible choice. That doesn't mean that `compute_error_squared()` should return a double too.

src/appleseed/renderer/modeling/aov/pixelerroraov.cpp
```diff
@@ -0,0 +1,170 @@
...
+// Interface header.
+#include "pixelerroraov.h"
+
+// appleseed.renderer headers.
+#include "renderer/kernel/aov/aovaccumulator.h"
+#include "renderer/modeling/aov/aov.h"
+#include "renderer/modeling/frame/frame.h"
+
+// appleseed.foundation headers.
+#include <foundation/image/analysis.h>
```
dictoon: Use double quotes.

wilimitis:
Hah, let's try that again: left is with caustics on, right is without ;)

src/appleseed/renderer/kernel/rendering/progressive/progressiveframerenderer.cpp
```diff
@@ -276,7 +252,7 @@ namespace
                     *m_buffer.get(),
                     m_params.m_perf_stats,
                     m_params.m_luminance_stats,
-                    m_ref_image.get(),
+                    &m_project.get_frame()->ref_image(),
```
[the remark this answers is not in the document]
wilimitis: As I understand it, `m_ref_image`, which is an `Integer*`, is assigned a null pointer extracted from the `frame`'s `unique_ptr`. Hence the existing `if (foo)` checks inside the function behave as before.

src/appleseed/foundation/image/analysis.h
```diff
@@ -57,6 +58,9 @@ APPLESEED_DLLSYMBOL double compute_average_luminance(const Image& image);
 // Return whether two images are compatible and thus can be compared.
 APPLESEED_DLLSYMBOL bool are_images_compatible(const Image& image1, const Image& image2);
 
+// Compute the squared error between two linear rgb colors.
+double compute_error_squared(const Color3f& c1, const Color3f& c2);
```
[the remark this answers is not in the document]
wilimitis: Much nicer, updated.

src/appleseed/foundation/image/analysis.cpp
wilimitis: I opted to use the component route so that we could relax the constraint that channel counts must match to perform this analysis since the error computation is rgb specific.

For instance, this seems to be the method used by https://github.com/appleseedhq/appleseed/blob/master/src/appleseed/foundation/math/sampling/imageimportancesampler.h#L308 to support RGB/RGBA.

(Maybe we don't care to support RGB in this case, and enforcing RGBA is okay..)

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -301,6 +320,28 @@ Image& Frame::image() const
     return *impl->m_image.get();
 }
 
+Image& Frame::ref_image() const
+{
+    return *impl->m_ref_image.get();
+}
+
+bool Frame::validate_ref_image() const
```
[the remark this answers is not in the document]
wilimitis: Yeah, I wasn't fond of it for those reasons as well - your feedback and a second pass have helped simplify the way I'm thinking about it, I've pushed an update.

I think I was getting hung up on the studio workflow, and expected users to see meaningful output if, for instance, the user changes the resolution in studio and then wonders why the AOV returns blank. However, there isn't actually UI to even support selecting a reference images yet, if there was then it might be worth considering being more verbose.

src/appleseed/foundation/image/analysis.cpp
dictoon: I see, that's a good point.

src/appleseed/foundation/image/analysis.h
```diff
@@ -30,6 +30,7 @@
 #pragma once
 
 // appleseed.foundation headers.
+#include "foundation/image/color.h"
```
dictoon: Order alphabetically.

src/appleseed/foundation/image/color.h
```diff
@@ -343,6 +343,9 @@ typedef Color<double,   4> Color4d;
 template <typename T, typename Int>
 Color<T, 3> integer_to_color3(const Int i);
 
+// Compute the squared error between two linear rgb colors.
+template <typename T>
+T compute_error_squared(const Color<T, 3>& c1, const Color<T, 3>& c2);
```
dictoon: Leave two blank lines before block comments.

src/appleseed/foundation/image/color.h
```diff
@@ -1183,4 +1186,13 @@ Color<T, 3> integer_to_color3(const Int i)
         static_cast<T>(z) * (1.0f / 4294967295.0f));
 }
 
+template <typename T>
+T compute_error_squared(const Color<T, 3>& c1, const Color<T, 3>& c2)
```
dictoon: Mark inline.

src/appleseed/foundation/meta/tests/test_analysis.cpp
```diff
@@ -119,4 +119,44 @@ TEST_SUITE(Foundation_Image_Analysis)
...
+
+        image1.clear(Color4f(0.0f));
+        image2.clear(Color4f(1.0f));
+
+        const double rmsd = compute_rms_deviation(image1, image2);
+
+        EXPECT_FEQ(1.0, rmsd);
+    }
+
+    TEST_CASE(ComputeRMSDeviation_GivenRedImageAndBlueImage_ReturnsAlmostRootTwoThirds)
```
dictoon: 😂

src/appleseed/renderer/modeling/aov/pixelerroraov.cpp
```diff
@@ -0,0 +1,169 @@
...
+                    float image_color[4];
+                    frame.image().get_pixel(x, y, image_color);
+
+                    float ref_color[4];
+                    frame.ref_image().get_pixel(x, y, ref_color);
+
+                    float error = sqrt(compute_error_squared(
+                        Color3f::from_array(image_color),
+                        Color3f::from_array(ref_color)));
+                    max_error = max(max_error, error);
```
dictoon: Include `<algorithm>`.

src/appleseed/renderer/modeling/aov/pixelerroraov.cpp
```diff
@@ -0,0 +1,169 @@
...
+            {
+                for (size_t x = crop_window.min.x; x <= crop_window.max.x; ++x)
+                {
+                    float image_color[4];
+                    frame.image().get_pixel(x, y, image_color);
+
+                    float ref_color[4];
+                    frame.ref_image().get_pixel(x, y, ref_color);
+
+                    float error = sqrt(compute_error_squared(
```
dictoon: Make const.

src/appleseed/renderer/modeling/aov/pixelerroraov.cpp
```diff
@@ -0,0 +1,169 @@
...
+                        Color3f::from_array(image_color),
+                        Color3f::from_array(ref_color)));
+                    max_error = max(max_error, error);
+
+                    const Color3f color(error, 0, 0);
```
dictoon: Spell out floating point literals: 0.0f

src/appleseed/renderer/modeling/aov/pixelerroraov.h
```diff
@@ -0,0 +1,74 @@
...
+// appleseed.foundation headers.
+#include "foundation/utility/autoreleaseptr.h"
+
+// appleseed.main headers.
+#include "main/dllsymbol.h"
+
+// Forward declarations.
+namespace foundation    { class Dictionary; }
+namespace foundation    { class DictionaryArray; }
+namespace renderer      { class AOV; }
```
dictoon: You have both a fwd decl and an include for `AOV`, maybe you can get rid of the include? I'm not entirely sure if you can, with `auto_release_ptr<>`...

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -172,6 +177,23 @@ Frame::Frame(
...
+
+        GenericImageFileReader reader;
+        impl->m_ref_image.reset(reader.read(
+            search_paths.qualify(impl->m_ref_image_path).c_str()));
+
+        if (!has_valid_ref_image())
+        {
+            RENDERER_LOG_ERROR(
+                "the reference image is not compatible with the output frame "
+                "(different dimensions, tile size or number of channels).");
```
dictoon: The "tile size" part is no longer true, AFAICT.

src/appleseed/renderer/modeling/frame/frame.cpp
```diff
@@ -301,6 +325,17 @@ Image& Frame::image() const
     return *impl->m_image.get();
 }
 
+Image& Frame::ref_image() const
+{
+    return *impl->m_ref_image.get();
```
dictoon: I would return an `Image*` given that the ref image might not be there (and then document that behavior in the header file).

src/appleseed/renderer/modeling/frame/frame.h
```diff
@@ -102,6 +103,12 @@ class APPLESEED_DLLSYMBOL Frame
     // Access the main underlying image.
     foundation::Image& image() const;
 
+    // Access the reference image.
```
dictoon: As per my previous comment, add "Returns nullptr if there is no reference image".

src/appleseed/renderer/kernel/rendering/progressive/progressiveframerenderer.cpp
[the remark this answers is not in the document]
dictoon: Except that `ref_image()` in this case dereferenced a null pointer (which is undefined behavior in C++) and returned an invalid reference.

src/appleseed/renderer/modeling/frame/frame.cpp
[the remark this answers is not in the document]
dictoon: It's much better now!

src/appleseed/foundation/meta/tests/test_analysis.cpp
wilimitis: Oddly enough `FEQ(sqrt(2.0f / 2.0f), ...)` passed on most platforms but failed on lower `gcc` versions.

src/appleseed/renderer/modeling/aov/pixelerroraov.cpp
wilimitis: 👍

Is the idea that even though we have access to it indirectly (`color.h`) we should include here in case things change upstream?

src/appleseed/renderer/modeling/aov/pixelerroraov.cpp
wilimitis: Ugh, and not doing this is a pet peeve of mine in glsl shader code too... 😆

src/appleseed/renderer/modeling/frame/frame.cpp
wilimitis: Yep, channel size neither.

src/appleseed/renderer/modeling/frame/frame.cpp
wilimitis: Right, and then we don't need to dereference what would potentially be a `nullptr` as you've mentioned above.

Thank you for this advice!

src/appleseed/renderer/kernel/rendering/progressive/progressiveframerenderer.cpp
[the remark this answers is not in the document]
wilimitis: Ouch, good point.. Updated, per your comments below.

src/appleseed/foundation/image/color.h
wilimitis: I'm admittedly not clear on why we want this `inline`d, though this is what I worked through
- This function is used in multiple cpp files and it compiles okay (not necessarily a good reason not to `inline` though)
- Perhaps we really want to hint to the compiler that code should be generated inline for this to reduce function call overhead since it is called within expensive loops

wilimitis:
PTAL, thank you for working through this with me btw.

src/appleseed/foundation/meta/tests/test_analysis.cpp
dictoon: Actually, the correct way to do this is

    EXPECT_FEQ(sqrt(2.0f / 3.0f), rmsd);

or

    EXPECT_FEQ_EPS(sqrt(2.0f / 3.0f), rmsd, 1.0e-6f);

(using a reasonable epsilon value)

src/appleseed/foundation/image/color.h
dictoon: > perhaps we want to hint to the compiler that code should be generated inline for this to reduce function call overhead since it is called within expensive loops.

That's exactly the reason, although admittedly I'm not sure that any compiler actually takes the hint seriously anymore.

src/appleseed/renderer/modeling/aov/pixelerroraov.cpp
dictoon: Yes, every file must include the headers it relies on, even if those are already indirectly included.

The only exception is .h/.cpp pairs: if a header file already includes something then the corresponding .cpp file shouldn't include it again.

This rule has been popularized by Google under the moniker "Include What You Use" (IWYU in short): https://include-what-you-use.org/

src/appleseed/foundation/meta/tests/test_analysis.cpp
wilimitis: Ah, right, so I had actually tried the first way originally (which failed).

Neat, the second way is much better, updated.

Issue #42, opened 2011-10-17
Move Frame::write() and Frame::archive() to separate classes


dictoon:
For instance `renderer::FrameWriter` and `renderer::FrameArchiver`.

jsxs0:
Hey, is this issue still available to work?

jsxs0:
Thanks, working over it!

PR #2487, opened 2019-03-18
Appleseed x Move Frame::write() and Frame::archive() #1

jsxs0:
Draft PR, to move Frame::write() and Frame::archive() to separate classes #42
Before I could get working on the issue of ***Move Frame::write() and Frame::archive() to separate classes #42***, I had couple of doubts.
If these can resolved or answered by any of the moderators here, I'd be more than happy to complete this draft PR, and get to the code.
Thanks in advance!

1. Can I use a method of calling itself once before rendering each frame. Basically like, it'd return true on success, false otherwise.

2. And, maybe with the help of a function like
`int Framearchive(const char* name)
`and then, moving this function into the main class `FrameWriter` class.
If it works this way, I can go write a code pertaining to it


dictoon:
Your PR only contains commits by @est77.

jsxs0:
I guess I've created a PR in the wrong branch of Appleseed x. I'll rectify this error, and get back to you ASAP

Issue #2480, opened 2019-03-15
Add OpenImageIO's idiff tool to appleseed distribution


dictoon:
Prerequisite of #2481.

slongle:
If nobody work on it, I'd like to do it.

dictoon:
Go ahead.

slongle:
I'm not sure about my thought about it now.
I just want to add the idiff.exe into appleseed\scripts\tools with the new folder named idiff.
The structure is
appleseed
| -- scripts
|       | -- tools
|       |        | -- cloc
|       |        |        `--cloc.exe
|       |        | -- idiff
|       |        |        `--idiff.exe
.....
I think then it can be used by the **.py
Is this what you want?

est77:
We already bundle a tool called oiiotool from OpenImageIO. Just copy what we do for oiiotool.

slongle:
Thanks est, I get it.

slongle:
For the linux, there isn't an idiff in prebuilt-linux-deps.
So, can I update that?

est77:
I added it to the deps.

PR #2492, opened 2019-03-19
Add OpenImageIO's idiff tool to appleseed distribution

slongle:
#2480 
Do the same things like oiiotool.


dictoon:
Looks good to me, but I'd like @est77 to have a look too.

dictoon:
Fails on Travis because as @slongle [noted](https://github.com/appleseedhq/appleseed/issues/2480#issuecomment-474258696), idiff is not available in prebuilt Linux dependencies.

est77:
Added idiff to linux deps.

PR #2494, opened 2019-03-19
Moving Frame::write() and Frame::archive()

jsxs0:
```
bpy::object archive_frame(const FrameArchiver* frameArchiver, const char* directory)
    {
        char* output = 0;
        if (frameArchiver.archive(directory, &output))
        {
            const bpy::str path(output);
            foundation::free_string(output);
            return path;
        }
        return bpy::object();
    }

```
OR
Maybe, I can create the file name with 
```
int addDir(const char* name) {
    std::string fileName = name;
    fileName += "/.";

    if (g_debugLevel > 0) {
      std::cout << "Creating dir: " << fileName << std::endl;
    }

    std::string nametemp= BASE_PATH;
    nametemp += name;
    int res = emulate_vfs_mkdir(nametemp.c_str(), O_CREAT);
    if (res < 0) {
      std::cerr << "Cannot create dir" << std::endl;
    }
    return 0;
}
``` 
and subsequent operations can follow

### DIFF
--- a/Moving Frame::write() and Frame::archive() into separate classes
+++ b/Moving Frame::write() and Frame::archive() into separate classes
@@ -1,0 +1,2 @@
+Using bpy::object archive_frame(const FrameArchiver* frameArchiver, const char* directory) function or int addDir(const char* name)
+and subsquent functions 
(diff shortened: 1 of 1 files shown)


rbrune:
Please implement your suggested code changes and additions into the actual files, compile it, test it - and then let us then review the code.

dictoon:
Alright, this is not going to go anywhere.

Not knowing how to code in C++ (or how to code at all) is perfectly fine. _Pretending_ to know, however, is not, at least in my book.

Thanks for your interest and good luck.

Issue #2420, opened 2019-02-11
Add "Object ID" override shading


Narann:
For now, there is only "Object Instances". But to check which objects are instanciated we need an "Object ID" override shading:

![image](https://user-images.githubusercontent.com/659708/52599357-b3336380-2e58-11e9-8d85-67ee350f4523.png)

This override would return the ID color of the object (not the object instance) used by the shape.

dictoon:
Makes a lot of sense. To be consistent with the other names, I suggest to:

- Call it simply "Objects"
- Add "Assemblies"

So we would end up with:

- Assemblies
- Assembly Instances
- Objects
- Object Instances

in that order.

Himanshu-Garg:
is anyone working on this ?

dictoon:
Nope, good ahead!

jsxs0:
Is this issue available, to work upon?

Himanshu-Garg:
Yes, not working on it rightnow

jsxs0:
> 
> 
> Yes, not working on it rightnow

Okay, thanks. Will work on it

bassemhermina:
So, I've been fiddling around with this issue as my entrance for Appleseed, and somehow I figured out how to solve it, so is anybody still working on it, or can I submit my fix?

jsxs0:
Yes, even I was finding my way to work about it but eventually found another issue to work on. You can go ahead, goodluck!

bassemhermina:
Thank you! Then I’ll have it.

dictoon:
Done.

PR #2498, opened 2019-03-21
Added 'object' and 'assembly' override shaders

bassemhermina:
Solving issue #2420 
"Objects" override shader can be easily tested by instantiating any of the objects in the cornelBox project as it's instantiated so far and large-scaled so colors can be easily compared. (is that an issue?)
but for "Assemblies" shader, instantiate the assembly with name "assembly_inst1" then run this to translate it a little bit for it to be visible.
(running the python code while interactive render is activated crashes appleseed.studio instantly on linux ubuntu 16.04, is that an issue too?)

from appleseed.studio import *
import appleseed as asr
scene = current_project().get_scene()

assembly = scene.assembly_instances().get_by_name("assembly_inst1")
matrix = asr.Matrix4d([1.0, 0.0, 0.0, 0.3,
                       0.0, 1.0, 0.0, 0.3,
                       0.0, 0.0, 1.0, 0.0,
                       0.0, 0.0, 0.0, 1.0])
assembly.transform_sequence().set_transform(0.0, asr.Transformd(matrix))

### DIFF
--- a/src/appleseed/renderer/modeling/surfaceshader/diagnosticsurfaceshader.cpp
+++ b/src/appleseed/renderer/modeling/surfaceshader/diagnosticsurfaceshader.cpp
@@ -583,7 +583,7 @@ void DiagnosticSurfaceShader::evaluate(
       case Assemblies:
         set_shading_result(
             shading_result,
-            integer_to_color3<float>(shading_point.get_assembly_instance().find_assembly()->get_uid()));
+            integer_to_color3<float>(shading_point.get_assembly_instance().get_assembly().get_uid()));
         break;
 
       case AssemblyInstances:
@@ -595,7 +595,7 @@ void DiagnosticSurfaceShader::evaluate(
       case Objects:
         set_shading_result(
             shading_result,
-            integer_to_color3<float>(shading_point.get_object_instance().find_object()->get_uid()));
+            integer_to_color3<float>(shading_point.get_object_instance().get_object().get_uid()));
         break;
 
       case ObjectInstances:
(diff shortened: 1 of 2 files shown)


src/appleseed/renderer/modeling/surfaceshader/diagnosticsurfaceshader.cpp
```diff
@@ -576,11 +580,23 @@ void DiagnosticSurfaceShader::evaluate(
         }
         break;
 
+      case Assemblies:
+        set_shading_result(
+            shading_result,
+            integer_to_color3<float>(shading_point.get_assembly_instance().find_assembly()->get_uid()));
```
dictoon: Use `ShadingPoint::get_assembly()` which is very fast. `AssemblyInstance::find_assembly()` is very slow, it's used to find the assembly outside of rendering.

src/appleseed/renderer/modeling/surfaceshader/diagnosticsurfaceshader.cpp
```diff
@@ -576,11 +580,23 @@ void DiagnosticSurfaceShader::evaluate(
...
       case AssemblyInstances:
         set_shading_result(
             shading_result,
             integer_to_color3<float>(shading_point.get_assembly_instance().get_uid()));
         break;
+      
+      case Objects:
+        set_shading_result(
+            shading_result,
+            integer_to_color3<float>(shading_point.get_object_instance().find_object()->get_uid()));
```
dictoon: Similarly, use `ShadingPoint::get_object()`.

src/appleseed/renderer/modeling/surfaceshader/diagnosticsurfaceshader.cpp
```diff
@@ -583,7 +583,7 @@ void DiagnosticSurfaceShader::evaluate(
       case Assemblies:
         set_shading_result(
             shading_result,
-            integer_to_color3<float>(shading_point.get_assembly_instance().find_assembly()->get_uid()));
+            integer_to_color3<float>(shading_point.get_assembly_instance().get_assembly().get_uid()));
```
dictoon: Not what I suggested, please read again my comments :)

bassemhermina: ahh ahh ah sorry for that! fixing it now!

bassemhermina: done! sorry still getting used to the codebase :)

PR #2467, opened 2019-03-10
Implement filter importance sampling

est77:
This PR implements filter importance sampling, as a replacement for the sample splatting we were using before.
This removes the need for tile margins and removes many filter evaluations. It also makes noise independent for each pixel, since samples are not shared anymore and will make certain denoisers work better with appleseed rendered images.

I noticed that we are using a mix of signed and unsigned integers to represent pixel coordinates in different places. I think we can switch all to unsigned now that there is no padding on tiles.
I didn't do it here because the PR was getting pretty big already.

Issue #2354, opened 2018-12-11
Allow driving image space sampling with a B&W texture


dictoon:
In the black and white texture, black pixels would mean 0 sample for the corresponding pixel in the final render, 255 would mean full sampling rate for the corresponding pixel, and intermediate value would allow controlling the sampling rate.

This is probably best implemented in a separate pixel renderer. `MapPixelRenderer`? Think of a better name.

BashPrince:
I'll start work on this.

dictoon:
Done.

PR #2499, opened 2019-03-23
Add texture-controlled pixel renderer

BashPrince:
For #2354.

This is the current state of my implementation.

settings

gradient-render
gradient-tex

spiral-render
spiral-tex

This will still have to be merged with #2467 since there are changes introduced to the UniformPixelRenderer which this new renderer is largely based on and should also adopt (e.g. remove the decorrelate pixels option). Can I already merge that PR into my branch while it is still pending?

Commit message:
Add new pixel renderer class TextureControlledPixelRenderer created by
new class TextureControlledPixelRendererFactory.
TextureControlledPixelRenderer determines the number of iterations
of the sample loop by querying a texture for that pixel position.
Textures can be of any resolution, and are resized to exactly match the
frame dimensions for a 1:1 frame/texture correspondence. Resizing will
also change the aspect ratio of the texture if it doesn't match the
frame, possibly squishing the texture in one dimension.

Since the texture driving the sampling should be black and white anyway,
only the red channel of the texture is used to retrieve values.

Add new setting controls to the GUI to set min/max samples, the force
anti-aliasing parameter and a file path for the
TextureControlledPixelRenderer. The decorrelate option is ignored since
it is set to be removed in a pending PR (#2467).


BashPrince:
Somehow the black regions in the renders where zero samples were placed appear white in my browser. They are correctly displayed in black when the files are downloaded and viewed offline though.

BashPrince:
I figured it out now, the regions I'm talking about are actually fully transparent, because at the moment, when the number of samples is computed as zero, I simply skip the sampling loop all together, which leaves the framebuffer values for color and alpha of main and AOVs initialized to zero. The white display color on github is just the background color of the site coming through I guess.
I could just construct an opaque black shading result for all framebuffer contents when the sampling texture gives back zero so those regions are simply displayed black. Alternatively we could set the minimum allowed samples to one whichever makes more sense.

src/appleseed/renderer/kernel/rendering/renderercomponents.cpp
```diff
@@ -347,6 +355,47 @@ bool RendererComponents::create_pixel_renderer_factory()
...
+
+        if (tex_path.empty())
+        {
+            RENDERER_LOG_ERROR("no texture path was specified for the texture controlled pixel renderer.");
+            return false;
+        }
+
+        // Create a new dedicated ImageCache instead of using the global ImageCache.
+        // This is to prevent errors during ImageBuf access if the file behind tex_path was overwritten on disk.
+        ImageCache* image_cache = ImageCache::create();
```
dictoon: You should use facilities in `foundation/image/` for image I/O instead of using OIIO directly.

dictoon:
I think you can also remove the Force Antialiasing option and consider it's always ON. This option is only relevant when the number of SPP is 1

BashPrince:
I just went with the original recommendation in #2354 for the new pixel renderer, but I agree that it might be better to just fuse the two into one. I could keep the GUI options and based on wether the user selects uniform or texture sampler I could pass an optional texture to the UniformPixelRenderer, similarly to how I do now with the TextureControlledPixelRenderer. Then in the render pixel loop based on wether the image is null or not, I take the uniform sample value or get the value from the texture.

I don't think it would clutter things up too much since the only addition would be how the the sample number is calculated, and a few extra member variables. And it prevents future contributors from having to remember to implement every change they make in the uniform renderer also in the texture renderer. 

I'll get back to this once I submitted my GSoC proposal.

oktomus:
Pixel decorolation was removed by Esteban in #2467. Maybe that can simplify your PR once Esteban's PR gets merged.

src/appleseed/renderer/kernel/rendering/renderercomponents.cpp
BashPrince: Hey, so the reason why I want to use OIIO directly here, is because I want to resize the supplied texture to exactly match the frame dimensions and I found that the easiest way to do that would be with OIIO ImageBuf and ImageBufAlgo::resize(). I could use the GenericImageFileReader in `foundation/image/` but this would return type Image and I guess the only way to get an ImageBuf from that is to then loop over each pixel of the Image to get each pixel value and set it in the ImageBuf and then proceed with the resize. That seems like a bit of a detour that I would rather avoid.
Maybe we already have some custom resizing functionality implemented somewhere that I am not seeing?

BashPrince:
Thanks oktomus, yeah I saw Esteban's PR.

src/appleseed/CMakeLists.txt
```diff
@@ -1212,6 +1212,8 @@ set (renderer_kernel_rendering_final_sources
     renderer/kernel/rendering/final/pixelsampler.h
     renderer/kernel/rendering/final/uniformpixelrenderer.cpp
     renderer/kernel/rendering/final/uniformpixelrenderer.h
+    renderer/kernel/rendering/final/texturecontrolledpixelrenderer.cpp
```

src/appleseed/renderer/kernel/rendering/renderercomponents.cpp
dictoon: Ah, I see.

> Maybe we already have some custom resizing functionality implemented somewhere that I am not seeing?

I don't think we do.

Another remark: Image loading should be done by a method [1] in `TextureControlledPixelRendererFactory`; `RendererComponents` is only about creating and connecting components together, it shouldn't contain other logic.

[1] Since you want to be able to react to I/O errors, it's best to do this in a separate method vs doing it in the constructor...

src/appleseed/renderer/kernel/rendering/renderercomponents.cpp
```diff
@@ -368,6 +376,47 @@ bool RendererComponents::create_pixel_renderer_factory()
 
         return true;
     }
+    else if (name == "texture")
+    {
+        if (m_sample_renderer_factory.get() == nullptr)
+        {
+            RENDERER_LOG_ERROR("cannot use the texture controlled pixel renderer without a sample renderer.");
```
dictoon: *texture-controlled*

src/appleseed/renderer/kernel/rendering/renderercomponents.cpp
```diff
@@ -368,6 +376,47 @@ bool RendererComponents::create_pixel_renderer_factory()
...
+            return false;
+        }
+
+        ParamArray tex_sampler_params = get_child_and_inherit_globals(m_params, "texture_controlled_pixel_renderer");
+        const string tex_path = tex_sampler_params.get_optional<string>("file_path", "");
+
+        if (tex_path.empty())
+        {
+            RENDERER_LOG_ERROR("no texture path was specified for the texture controlled pixel renderer.");
```
dictoon: *texture-controlled*

src/appleseed/renderer/kernel/rendering/renderercomponents.cpp
```diff
@@ -44,6 +44,7 @@
 #include "renderer/kernel/rendering/ephemeralshadingresultframebufferfactory.h"
 #include "renderer/kernel/rendering/final/adaptivetilerenderer.h"
 #include "renderer/kernel/rendering/final/uniformpixelrenderer.h"
+#include "renderer/kernel/rendering/final/texturecontrolledpixelrenderer.h"
```
dictoon: Order alphabetically.

src/appleseed/renderer/kernel/rendering/final/uniformpixelrenderer.cpp
```diff
@@ -85,7 +85,6 @@ namespace
           : m_params(params)
           , m_sample_renderer(factory->create(thread_index))
           , m_sample_count(m_params.m_samples)
-          , m_sqrt_sample_count(round<int>(sqrt(static_cast<double>(m_params.m_samples))))
```
dictoon: 👍

src/appleseed/renderer/kernel/rendering/final/texturecontrolledpixelrenderer.h
```diff
@@ -0,0 +1,89 @@
+
+//
+// This source file is part of appleseed.
+// Visit https://appleseedhq.net/ for additional information and resources.
+//
+// This software is released under the MIT license.
+//
+// Copyright (c) 2010-2013 Francois Beaune, Jupiter Jazz Limited
```
dictoon: Replace those two lines by a single line:

    Copyright (c) 2019 Stephen Agyemang, The appleseedhq Organization

src/appleseed/renderer/kernel/rendering/final/texturecontrolledpixelrenderer.cpp
```diff
@@ -0,0 +1,339 @@
+
+//
+// This source file is part of appleseed.
+// Visit https://appleseedhq.net/ for additional information and resources.
+//
+// This software is released under the MIT license.
+//
+// Copyright (c) 2010-2013 Francois Beaune, Jupiter Jazz Limited
```
dictoon: Same here (see comment in header file).

src/appleseed/renderer/kernel/rendering/final/texturecontrolledpixelrenderer.h
```diff
@@ -0,0 +1,89 @@
...
+// Forward declarations.
+namespace foundation    { class Dictionary; }
+namespace renderer      { class Frame; }
+namespace renderer      { class ISampleRendererFactory; }
+
+namespace renderer
+{
+
+//
+// Texture controlled pixel renderer.
```
dictoon: *Texture-controlled*

src/appleseed/renderer/kernel/rendering/final/texturecontrolledpixelrenderer.h
```diff
@@ -0,0 +1,89 @@
...
+    // Delete this instance.
+    void release() override;
+
+    // Return a new texture controlled pixel renderer instance.
+    IPixelRenderer* create(
+        const size_t                thread_index) override;
+
+  private:
+    const Frame&                          m_frame;
+    std::shared_ptr<OIIO::ImageBuf>       m_texture;
```
dictoon: Do we actually need a `shared_ptr<>` here? Just wondering.

src/appleseed/renderer/kernel/rendering/final/texturecontrolledpixelrenderer.h
```diff
@@ -0,0 +1,89 @@
...
+    TextureControlledPixelRendererFactory(
+        const Frame&                                frame,
+        const std::unique_ptr<const OIIO::ImageBuf> texture,
+        ISampleRendererFactory*                     factory,
+        const ParamArray&                           params);
+    
+    // Delete this instance.
+    void release() override;
+
+    // Return a new texture controlled pixel renderer instance.
```
dictoon: *texture-controlled*

src/appleseed/renderer/kernel/rendering/final/texturecontrolledpixelrenderer.cpp
```diff
@@ -0,0 +1,339 @@
...
+
+        void release() override
+        {
+            delete this;
+        }
+
+        void print_settings() const override
+        {
+            RENDERER_LOG_INFO(
+                "texture controlled pixel renderer settings:\n"
```
dictoon: *texture-controlled*

src/appleseed/renderer/kernel/rendering/final/texturecontrolledpixelrenderer.cpp
```diff
@@ -0,0 +1,339 @@
...
+            for (size_t i = 0; i < sample_count; ++i)
+            {
+                // Generate a uniform sample in [0,1)^2.
+                const Vector2d s =
+                    m_max_sample_count > 1 || m_params.m_force_aa
+                        ? sampling_context.next2<Vector2d>()
+                        : Vector2d(0.5);
+
+                // Sample the pixel filter.
+                const auto &filter_table = frame.get_filter_sampling_table();
```
dictoon: Move `&` to the type's side (in the code you copied from too, possibly).

src/appleseed/renderer/kernel/rendering/final/texturecontrolledpixelrenderer.h
BashPrince: I make it shared, because the factory passes the same texture to each created renderer.

BashPrince:
Finally got around to making the requested changes. I now actually think it makes more sense to not merge this with the uniform pixel renderer, that way you can clearly separate the params and don't have to introduce `if(texture_loaded)` or similar to decide how to calculate the sample count.
I moved image loading into a class method, but kept the custom loading functionality, is that ok?

src/appleseed/renderer/kernel/rendering/final/texturecontrolledpixelrenderer.cpp
```diff
@@ -303,26 +302,41 @@ Dictionary TextureControlledPixelRendererFactory::get_params_metadata()
...
   , m_factory(factory)
   , m_params(params)
 {
+}
+
+bool TextureControlledPixelRendererFactory::load_texture(const std::string &texture_path)
+{
+    // Create a new dedicated ImageCache instead of using the global ImageCache.
+    // This is to prevent errors during ImageBuf access if the file behind tex_path was overwritten on disk.
+    ImageCache *image_cache = ImageCache::create();
```
dictoon: Move `*` to the type's side.

src/appleseed/renderer/kernel/rendering/final/texturecontrolledpixelrenderer.cpp
```diff
@@ -303,26 +302,41 @@ Dictionary TextureControlledPixelRendererFactory::get_params_metadata()
...
+    std::unique_ptr<ImageBuf> texture(new ImageBuf(texture_path, 0, 0, image_cache));
+    // Force the read to immediately do the entire disk I/O for the file.
+    // We can then safely destroy the cache since the entire image data has been read into the ImageBuf.
+    const bool read_successful = texture->read(0, 0, true);
+    ImageCache::destroy(image_cache);
+
+    if(!read_successful)
```
dictoon: Insert space after `if`.

src/appleseed/renderer/kernel/rendering/final/texturecontrolledpixelrenderer.cpp
```diff
@@ -303,26 +302,41 @@ Dictionary TextureControlledPixelRendererFactory::get_params_metadata()
...
-    const std::unique_ptr<const ImageBuf> texture,
     ISampleRendererFactory*               factory,
     const ParamArray&                     params)
   : m_frame(frame)
   , m_factory(factory)
   , m_params(params)
 {
+}
+
+bool TextureControlledPixelRendererFactory::load_texture(const std::string &texture_path)
```
dictoon: Move `&` to the type's side.

src/appleseed/renderer/kernel/rendering/final/texturecontrolledpixelrenderer.h
```diff
@@ -68,14 +67,16 @@ class TextureControlledPixelRendererFactory
     // Constructor.
     TextureControlledPixelRendererFactory(
         const Frame&                                frame,
-        const std::unique_ptr<const OIIO::ImageBuf> texture,
         ISampleRendererFactory*                     factory,
         const ParamArray&                           params);
     
+    // Load an image texture from a file
```
dictoon: End full line comments with a period.

dictoon:
> I moved image loading into a class method, but kept the custom loading functionality instead of using our image loading functionality for the reasons listed above, is that ok?

Yes.

src/appleseed/renderer/kernel/rendering/final/texturecontrolledpixelrenderer.h
dictoon: Ok.

BashPrince:
Sorry, been away last week, thanks for the fixes.

dictoon:
Thanks for your work!

PR #2503, opened 2019-03-24
Extend GCC compiler optimization for Linux.

### LINKED ISSUE #1098: Build with -funroll-loops on Linux / macOS
Apparently loops are not unrolled with gcc, even at -O3. Watch this https://youtu.be/w5Z4JlMJ1VQ?t=1906 for more details.

LZaw:
Partly fixes #1098. 
Only covers Linux GCC (4.8.5) optimizations. Due to the lack of Mac devices, I can't verify the compilerflags for Mac.
In addition to the desired loop-unroll, some more loop optimizations have been added, which are trying to optimize loops. Generally, optimization flags have been chosen to increase runtime-speed, while maintaining precision (that's why e.g. -fast-math is not included) and sacrificing compile-speed and filesize.
Flags chosen from [GNU Optimization Options](https://gcc.gnu.org/onlinedocs/gcc-4.8.5/gcc/Optimize-Options.html).
GCC 4.8.5 has been chosen, since it's the latest 4.8 version available.

Code compiles and is still able to render simple scenes (i.e. the cornell box) with these flags.

### DIFF
--- a/cmake/config/linux-gcc-clang.txt
+++ b/cmake/config/linux-gcc-clang.txt
@@ -167,7 +167,11 @@ endif ()
 # Release configuration.
 set (c_compiler_flags_release
     -O3                                                 # optimization level
-    -funroll-loops                                      # not included in O3
+    -funroll-loops                                      # Unrolls loops
+    -fgcse-las                                          # Eliminates stores followed by a load to the same positions.
+    -fgcse-sm                                           # Moves load/save instructions out of loops, if possible.
+    -fgcse-after-reload                                 # Cleans different fgcse passes up.
+    -ftree-loop-ivcanon                                 # Supports unrolling loops by determining iteration numbers of complex loops.
 )
 
 
--- a/cmake/config/linux-gcc-clang.txt
+++ b/cmake/config/linux-gcc-clang.txt
@@ -167,6 +167,7 @@ endif ()
 # Release configuration.
 set (c_compiler_flags_release
     -O3                                                 # optimization level
+    -funroll-loops                                      # not included in O3
 )
 
 
--- a/cmake/config/linux-gcc-clang.txt
+++ b/cmake/config/linux-gcc-clang.txt
@@ -170,7 +170,6 @@ set (c_compiler_flags_release
     -funroll-loops                                      # Unrolls loops
     -fgcse-las                                          # Eliminates stores followed by a load to the same positions.
     -fgcse-sm                                           # Moves load/save instructions out of loops, if possible.
-    -fgcse-after-reload                                 # Cleans different fgcse passes up.
     -ftree-loop-ivcanon                                 # Supports unrolling loops by determining iteration numbers of complex loops.
 )
 
--- a/cmake/config/linux-gcc-clang.txt
+++ b/cmake/config/linux-gcc-clang.txt
@@ -167,10 +167,8 @@ endif ()
 # Release configuration.
 set (c_compiler_flags_release
     -O3                                                 # optimization level
-    -funroll-loops                                      # Unrolls loops
     -fgcse-las                                          # Eliminates stores followed by a load to the same positions.
     -fgcse-sm                                           # Moves load/save instructions out of loops, if possible.
-    -ftree-loop-ivcanon                                 # Supports unrolling loops by determining iteration numbers of complex loops.
 )
 
 


est77:
What difference do they make the new flags?

The most important factor is speed, but also check the size of libappleseed.so before and after.

cmake/config/linux-gcc-clang.txt
```diff
@@ -167,6 +167,11 @@ endif ()
 # Release configuration.
 set (c_compiler_flags_release
     -O3                                                 # optimization level
+    -funroll-loops                                      # Unrolls loops
+    -fgcse-las                                          # Eliminates stores followed by a load to the same positions.
+    -fgcse-sm                                           # Moves load/save instructions out of loops, if possible.
```
dictoon: Do I understand correctly that `-fgcse-lm` is already enabled by `-fgcse` which is automatically enabled thanks to `-O3`?

cmake/config/linux-gcc-clang.txt
```diff
@@ -167,6 +167,11 @@ endif ()
...
+    -fgcse-after-reload                                 # Cleans different fgcse passes up.
```
dictoon: Already enabled by `-O3`?

LZaw: Oh, that one slipped through. In the description of -fgcse-after-reload it doesn't say anything about being enabled automatically, but on the top at O3. Gonna remove that one.

cmake/config/linux-gcc-clang.txt
LZaw: Yes, -fgcse-lm is enabled when -fgcse is enabled, which gets activated automatically by O2, O3 and Os. Therefore, I didn't include it explicitely.

LZaw:
Sorry for not mentioning the precise use of each flag. I selected them with speed in mind:
- funroll-loops:
> Unroll loops whose number of iterations can be determined at compile time or upon entry to the loop. -funroll-loops implies -frerun-cse-after-loop. This option makes code larger, and may or may not make it run faster. 

Loops with a known (const expr) number of iterations will be unrolled into single statements. The compiler would convert
`for(int i = 0; i < 10; i++)`
`{`
`    a[i]++;`
`}`
into
`a[0]++;`
`a[1]++;`
`a[2]++;`
`...`
- ftree-loop-ivcanon
> Create a canonical counter for number of iterations in loops for which determining number of iterations requires complicated analysis. Later optimizations then may determine the number easily. Useful especially in connection with unrolling. 

Supports the unrolling, by enabling it possibly to unroll more loops, especially the ones not having a const expr as iteration number, but which still can be determined at compiletime.
- fgcse-las
> When -fgcse-las is enabled, the global common subexpression elimination pass eliminates redundant loads that come after stores to the same memory location (both partial and full redundancies). 

When you have the following code
`a[i] = x;`
`x = a[i];`
you can skip the second line. This is a very simple example, but it illustrates the idea.
- fgcse-sm
> When -fgcse-sm is enabled, a store motion pass is run after global common subexpression elimination. This pass attempts to move stores out of loops. When used in conjunction with -fgcse-lm, loops containing a load/store sequence can be changed to a load before the loop and a store after the loop. 

Since loading may take a while and you may benefit from doing it before entering a loop, such that the loop itself doesn't need to wait for each load, it can be beneficial to perform loads and stores before and after the loop. If everything is loaded before the loop happens and stored afterwards, the loop can benefit a lot more from pipelining.

Sorry for the bad code formatting, I'm not sure how to insert linebreaks into a code-block.

dictoon:
Use triple backticks for formatting:

'''
your code
'''

(Replace single quotes by backticks).

LZaw:
Allright, it took some time, but I got some numbers.
About the testsetup:
- Each time I changed a compileflag I compiled the whole system anew (make clean && make)
- I disabled the Warnings-to-Errors flag
- I used the appleseed.cli to render
- While rendering I didn't touch the computer at all
- While rendering only a terminal was opened, no other user application was running (besides the desktop environment).
- The system runs Ubuntu 18.04 LTS on a Core i7-4770k@3.5GHz (x8) and 16GB of RAM.

I checked different flag configurations. All configurations had the O3 included:
- No additional flags
- funroll-loops, fgcse-las, fgcse-sm, ftree-loop-ivcanon (All additional flags)
- funroll-loops (no ftree-loop-ivcanon)
- fgcse-las, fgcse-sm

**Unit tests:**
All configurations passed all unit tests with summaries like the following:
```
2019-03-25T21:24:56.813932Z <001>   124 MB info    |   suites      : 179 executed, 0 failed (0.0%)
2019-03-25T21:24:56.813932Z <001>   124 MB info    |   cases       : 1,529 executed, 0 failed (0.0%)
2019-03-25T21:24:56.813952Z <001>   124 MB info    |   assertions  : 389,572 executed, 0 failed (0.0%)
```
Only difference between the outputs was the time and the size info.

**Results**
The following table will refer to the size of libappleseed.so as Size. The rendering times will have three columns. One for the best time, one for average time and one for worst time.

| Configuration | Size | Worst Timing | Average Timing | Best Timing |
| --- | --- | --- | --- | --- |
| No additional flags | 22.580.624 Bytes | 13:17.267 | 13:16.214 | 13:15.161 |
| All additional flags | 24.960.400 Bytes | 13:37.888 | 13:32.430 | 13:26.933 |
| Only funroll-loops | 24.956.304 Bytes |  13:37.320 | 13:37.577 | 13:37.789 |
| Two fgcse flags | 22.584.720 Bytes | 13:14.022 | 13:12.284 | 13:10.977 |

From my point of view, the size is acceptable for any configuration.
Interestingly, unrolling loops worsens the whole process (which might be the reason, it's not included by default into O3). Nevertheless, shifting loads and stores out of loops seems to bring a slight benefit.

Therefore, I removed the funroll-loops and ftree-loop-ivcanon flags, but kept the two fgcse ones.

dictoon:
Thanks for the excellent and thorough work!

Just one question:

> I disabled the Warnings-to-Errors flag.

Why did you have to do this? Were you getting new warnings with the new optimization flags?

Regarding your PR: it contains a lot of unrelated files, we cannot merge it as is. Here's one solution:
1. _Fetch_ (not pull) `master`.
2. Create a new branch at the tip of `master`.
3. Cherry-pick interesting changes from this branch and commit them into the new branch.
4. Push your new branch to `LZaw:Issue1098` on GitHub to update this PR (no need to close it and open a new one).
5. Delete your old branch.

And here's another:
1. _Fetch_ (not pull) `master`.
2. Rebase this branch _interactively_ on top of `master`. Delete the merge commit (`@LZaw
Merge branch 'Issue1098'...`).
3. Force-push to `LZaw:Issue1098`.

LZaw:
That rebase didn't work out as inteded. Really sorry about the mess. I will try to clean up, may take a small while. -.-

About the errors: I got them from the first day, regardless of the flags. I disabled them to get the Release build don, respectively, to get the Release flags set. One of these errors has been discussed in [#2483](https://github.com/appleseedhq/appleseed/pull/2483) as well as [#2325](https://github.com/appleseedhq/appleseed/issues/2325). The warnings during compiling didn't change when I added or removed any flags.

LZaw:
Allright, it should be fine now. Sorry again for the mess.

Issue #2393, opened 2019-01-26
Optimize light paths visualization (OpenGL)


dictoon:
We're currently using old-school immediate mode OpenGL to visualize light paths (and the scene). Move toward a more modern (VBO-based?) approach.

The relevant code is in `src/appleseed.studio/mainwindow/rendering/lightpathswidget.cpp`.

termhn:
Going to be taking a look at this in the next few days hopefully

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Issue #2184, opened 2018-08-30
Migrate appleseed.studio from Qt 4 to Qt 5


Blackhart:
According to the [VFX reference](https://www.vfxplatform.com/), we have to migrate appleseed.studio from Qt 4 to Qt 5.6.latest.

dictoon:
Done.

PR #2496, opened 2019-03-21
Use modern OpenGL to draw light paths and scene representation

termhn:
Fixes #2393 

Requires #2509 in order to work

Uses OpenGL 3.3 core to draw the light paths visualization. Drastically improves performance on more complex scenes and scenes with many light paths.

* Uses instanced drawing to draw multiple instances of the same object in the same draw call
* Currently **does not update positions or other data of the scene objects based on changes in appleseed.studio after the first data load has completed.** (This is for the scene object representations only; light paths are updated dynamically based on user input). If this is a significant issue to have this merged then I can fix it first.
* Uses Qt 5 builtins to load GL and GL function pointers.
* Adds a `triangle_count_hint` to `begin_object` on `ObjectRasterizer` which allows the rasterizer to preallocate a buffer of the appropriate size.
* Adds `get_vertex_count` to `LightPathRecorder` to get the total number of vertices stored in all light paths, also for use in preallocating space for a buffer

Issue #2239, opened 2018-10-16
OpenGL render in Light Paths tab is glitchy on macOS 10.14 (Mojave)


dictoon:
On a MacBook Pro, it either shows the content beneath appleseed.studio's window, or it's black:

![image](https://user-images.githubusercontent.com/321290/47048882-fd9f1f00-d19b-11e8-9849-8653098e20ab.png)

On a macOS VM, it flickers between black and the correct render.

**Important: it used to work fine with appleseed 1.9.0-beta.**

pjessesco:
I couldn't install macOS 10.10 on my mac, I need some time to solve it :(

dictoon:
No worry.

I don't think you want to install macOS 10.10 on your Mac, why would you want to do that?

pjessesco:
I found 2.0.0 beta package that you uploaded on dropbox is build with clang 10.0(macOS 10.14). So I just want to check if master branch's code works well when it built with clang 7.0 (that is, on macOS 10.10) like 1.9 release package one.

pjessesco:
Update : I build with clang 7.0.2, OpenGL renderer seems work well.

![screen shot 2018-11-02 at 12 27 20 am](https://user-images.githubusercontent.com/11532321/47861994-71e1ef80-de37-11e8-91cc-d720993064e7.png)

dictoon:
Oh that's interesting! Are you sure it's related to clang's version? Isn't it more about macOS's version?

pjessesco:
I'm not sure because I have no idea if I can install different version of clang.

pjessesco:
I installed clang 7 on macOS 10.14, it still doesn't work. 
<img width="835" alt="2018-11-07 12 20 00" src="https://user-images.githubusercontent.com/11532321/48323844-7f864900-e671-11e8-8d8b-71dc1682aed4.png">

White area is not blank, it actually filled with white color. I couldn't install clang 10 on macOS 10.10. I guess it is about macOS version as you said..

dictoon:
Thanks for checking! Yes I think it's logical that it's related to the macOS version and not the compiler. It would be interesting to dig deeper. Also, we have deprecation warnings on OpenGL calls when building on macOS 10.14 (with the clang version provided with the developer tools), we need to shut those off for now.

pjessesco:
> we need to shut those off for now.

Yes. I can fix it, and this issue is only for macOS 10.14.

pjessesco:
[...]

PR #2509, opened 2019-03-27
Upgrade to Qt 5

termhn:
#2184 

This is now working! This is a required change for #2393 to work on macOS. Required changes to build:

## Windows

In order to build with `WITH_DISNEY_MATERIAL=OFF`, the only change needed is to first download and build (or download a precompiled version of) Qt 5, I tested with the latest Qt 5.12.2. I recommend building from source or using this precompiled version https://github.com/martinrotter/qt5-minimalistic-builds (this is what I tested with) because the official builds are massive.

With that done, the cmake command needs to be modified slightly, removing `-DQT_QMAKE_EXECUTABLE` and replacing it with `-DCMAKE_PREFIX_PATH`, which should point to the root qt5 folder. For example, for me that is `-DCMAKE_PREFIX_PATH=G:\qt5`.

Finally, in the Visual Studio solution, the debugging environment variables need to be changed slightly (I'm not sure if this is true with the official builds but it is with the precompiled one I linked), we need to now include `qt5\lib` and `qt5\bin` in `PATH`. (This is under Environment when you right-click a target and to go properties then select All Configurations).

Building with `WITH_DISNEY_MATERIAL=ON` requires a patched SeExpr, and I've made a version of `appleseed-deps` which supports this, https://github.com/termhn/appleseed-deps on the `qt5` branch. The `BuildAll.bat` should work. You can also just comment out everything except the SeExpr part of that build script and copy it over to your current appleseed-deps if you have one.

## macOS

If you built using the recommended way, you'll need to do a couple of homebrew things.

```
brew tap-unpin cartr/qt4
brew uninstall pyqt@4 qt@4 # There was another thing that needed to be uninstalled that said it was dependent of qt@4 but I forget what it was called, also uninstall that, you also might be able to just brew unlink instead of uninstall
brew install qt5
```

Then when building appleseed without Disney material support, the only change needed is to add the cmake variable `-DCMAKE_PREFIX_PATH`, which should point to the root qt5 folder. For the homebrew installation method, this is `/usr/local/opt/qt`, i.e. `-DCMAKE_PREFIX_PATH=/usr/local/opt/qt`

In order to build with `WITH_DISNEY_MATERIAL=ON`, you'll also need a patched SeExpr. You should be able to do the normal thing with this but just use my fork and check out the `appleseed-qt5` branch https://github.com/termhn/SeExpr The only difference is you'll need again need to set `-DCMAKE_PREFIX_PATH` to the root of qt5, which for the homebrew installation method is `/usr/local/opt/qt`

```
git clone https://github.com/termhn/SeExpr
git checkout appleseed-qt5
mkdir build
cd build
cmake -DCMAKE_PREFIX_PATH=/usr/local/opt/qt ..
sudo make install
```

## Linux

I haven't tested or attempted, but you should be able to basically follow the macOS instructions but you'll need to figure out how to install or build Qt 5 yourself.


termhn:
Alright, with just a couple extra changes this works quite well with `-DUSE_DISNEY_MATERIAL=OFF` after initial testing, running on macOS! 
Screenshot 2019-03-27 12 48 02

termhn:
Oh, also, you need to set `CMAKE_PREFIX_PATH` to the root of wherever your qt installation is. If you install qt5 with homebrew on mac (you need to uninstall qt4 as well) then that would be `-DCMAKE_PREFIX_PATH=/usr/local/opt/qt`.

oktomus:
As long as it's a work in progress, you should add the WIP label and once it's ready, ask for a review.

termhn:
@oktomus Didn't know I had permissions to label it myself :P

oktomus:
About what you said on `CMAKE_PREFIX_PATH`, can't we add FindQt5 directly in appleseed ? What do you think @est77 ?

termhn:
For `CMAKE_PREFIX_PATH`, there are several(many) cmake scripts that it needs to be able to find, not just one `FindQt5.cmake`, so I think it makes most sense to leave it with Qt. We could attempt to set `CMAKE_PREFIX_PATH` to some common install locations by default and then just make people set it if they have it in a nonstandard location, though.

termhn:
Oh, and also, on Windows I would recommend building your own qt5 or using one of these builds: https://github.com/martinrotter/qt5-minimalistic-builds because the standard installers are *massive* (3.7GB download, what the heck??)

oktomus:
When building on windows for instance, we specify `QT_QMAKE_EXE_PATH` or something, can't we extract the QT `CMAKE_PREFIX_PATH` from that ?

termhn:
Yes you could, but I don't necessarily see a reason to do that, the `CMAKE_PREFIX_PATH` replaces the old `QT_QMAKE_EXE` variable (don't need that anymore).

Also, evidently my Windows building knowledge is just poor because I just built my modified version of SeExpr and then built appleseed.studio with `WITH_DISNEY_MATERIAL=ON` on macOS and it worked without a hitch (I didn't actually test its functionality yet, but it runs!)

termhn:
Alright, I figured out windows as well and it's working with SeExpr too :) 
image

I will wrap up some loose ends and update original PR then ask for review.

termhn:
Okay I think this is ready for testing/review. Updated top level comment with instructions. CC @dictoon @est77 @oktomus

pjessesco:
It seems this PR also solves #2239.

termhn:
@pjessesco Sweet! It should be trivial to also test #2496 if you've gotten this PR working :D

pjessesco:
Ah, light paths works well first time but it doesn't works after that.

oktomus:
Travis linux builds are failing. I'm not sure but you may need to update the travis build script.
See https://github.com/appleseedhq/appleseed/blob/master/scripts/travis/build-linux.sh#L82

Here is where it failed https://travis-ci.org/appleseedhq/appleseed/jobs/512359848#L796

oktomus:
Oh you haven't tested for Qt5 on linux. Maybe I can help with that.

src/appleseed.studio/CMakeLists.txt
```diff
@@ -652,13 +652,6 @@ if (WITH_DISNEY_MATERIAL)
     link_against_seexpreditor (appleseed.studio)
 endif ()
```
dictoon: Keep two blank lines before block comments.

src/appleseed.studio/CMakeLists.txt
```diff
@@ -42,11 +42,8 @@ endif ()
 # OpenGL.
 find_package (OpenGL REQUIRED)
 
-# Qt 4.
-find_package (Qt4 REQUIRED)
-set (QT_USE_QTOPENGL TRUE)
-include (${QT_USE_FILE})
-
+# Qt 5.
+find_package (Qt5 COMPONENTS Core Concurrent OpenGL Widgets REQUIRED)
```
dictoon: Keep two blank lines before block comments.

src/appleseed.studio/main/main.cpp
```diff
@@ -271,16 +271,16 @@ namespace
...
-    void message_handler(QtMsgType type, const char* msg)
+    void message_handler(QtMsgType type, const QMessageLogContext& context, const QString& msg)
     {
 #ifdef __APPLE__
         // Under certain circumstances (under an macOS virtual machine?), a bogus warning
         // message is repeatedly printed to the console. Disable this warning message.
         // See https://github.com/appleseedhq/appleseed/issues/254 for details.
         if (type == QtWarningMsg &&
-            strcmp(msg, "QCocoaView handleTabletEvent: This tablet device is unknown (received no proximity event for it). Discarding event.") == 0)
+            msg == "QCocoaView handleTabletEvent: This tablet device is unknown (received no proximity event for it). Discarding event.")
```
dictoon: I wonder if this kludge is still necessary. I'll check once I had it running on a macOS VM.

src/appleseed.studio/main/main.cpp
```diff
@@ -290,26 +290,26 @@ namespace
...
-            g_previous_message_handler(type, msg);
+            g_previous_message_handler(type, context, msg);
             return;
         }
 
         switch (type)
         {
           case QtDebugMsg:
-            fprintf(stderr, "Debug: %s\n", msg);
+            fprintf(stderr, "Debug: %s\n", msg.toLocal8Bit().constData());
```
dictoon: I'm not sure we want to use `toLocal8Bit()` / `fromLocal8Bit()`. Ideally appleseed would internally only handle UTF-8 strings.

termhn: Could convert these to `toUtf8()` / `fromUtf8()` instead... I'm not exactly sure the implications of one or the other to be completley honest, the Local8Bit versions seemed safter but I could be off base on that assumption :)

dictoon:
This is ready to merge, but we need to wait until https://github.com/appleseedhq/appleseed-deps/pull/4 in appleseed-deps is in.

dictoon:
Once this is merged, we will need to update the build guides.

oktomus:
Dont we need to make it support linux before merging it?

termhn:
We should probably test it, yes... it should Just Work TM by pointing it to qt 5 instead of qt4, but... worth a test.

oktomus:
In any case, the travis linux build script needs to be updated.

termhn:
OK, I in theory updated the linux and macos travis build scripts; however, for linux, https://github.com/appleseedhq/prebuilt-linux-deps needs a new release with the patched version of SeExpr

dictoon:
The prerequisite PR (https://github.com/appleseedhq/appleseed-deps/pull/4) has been merged.

dictoon:
It looks like the Linux dependency package needs to be updated:
```
In file included from /home/travis/build/appleseedhq/appleseed/src/appleseed.studio/mainwindow/project/expressioneditorwindow.cpp:52:0:
/home/travis/build/appleseedhq/appleseed/prebuilt-linux-deps/include/SeExprEditor/SeExprEditor.h:26:30: fatal error: QtGui/QTextBrowser: No such file or directory
 #include <QtGui/QTextBrowser>
```

termhn:
Yep. Let me know if/how I can help with doing so; it should just be a simple rebuild of SeExpr based on my fork.

dictoon:
I'm testing this PR.

I'm getting this warning when running CMake:

```
CMake Warning (dev) at CMakeLists.txt:236 (find_package):
  Policy CMP0074 is not set: find_package uses <PackageName>_ROOT variables.
  Run "cmake --help-policy CMP0074" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.

  Environment variable Boost_ROOT is set to:

    N:\appleseed\boost_1_69_0

  For compatibility, CMake is ignoring the variable.
```

We need to silence it.

Later, CMake fails to to find Qt 5:

```
CMake Error at src/appleseed.studio/CMakeLists.txt:46 (find_package):
  By not providing "FindQt5.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Qt5", but
  CMake did not find one.

  Could not find a package configuration file provided by "Qt5" with any of
  the following names:

    Qt5Config.cmake
    qt5-config.cmake

  Add the installation prefix of "Qt5" to CMAKE_PREFIX_PATH or set "Qt5_DIR"
  to a directory containing one of the above files.  If "Qt5" provides a
  separate development package or SDK, be sure it has been installed.
```

dictoon:
I fixed the second error about finding Qt 5 by providing the right path to Qt 5.

dictoon:
It builds and runs correctly.

A few small issues:

- The About dialog is missing a title bar and cannot be moved:

image

- List header rows are too thick:

image

- There's a bright horizontal line at the bottom of the window when both the Log and Python Console tabs are shown:

image

termhn:
👍🏻 I will investigate probably tomorrow as today is quite busy for me.

dictoon:
Created two tickets for the issues I mentioned earlier: https://github.com/appleseedhq/appleseed/issues/2535 https://github.com/appleseedhq/appleseed/issues/2536

Issue #2507, opened 2019-03-26
Fix compiler warnings


LZaw:
**Setup**
- Ubuntu 18.04 LTS.
- gcc-4.8.
- Additional to the flags set as in the [wiki](https://github.com/appleseedhq/appleseed/wiki/Building-appleseed-on-Linux), I set 
```
-DCMAKE_C_COMPILER=gcc-4.8
-DCMAKE_CXX_COMPILER=g++-4.8
-DCMAKE_BUILD_TYPE=Release
```

**GCC**
When asking gcc-4.8 about its version, it tells me:
```
$ gcc-4.8 -v
Using built-in specs.
COLLECT_GCC=gcc-4.8
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-linux-gnu/4.8/lto-wrapper
Target: x86_64-linux-gnu
Configured with: ../src/configure -v --with-pkgversion='Ubuntu 4.8.5-4ubuntu8' --with-bugurl=file:///usr/share/doc/gcc-4.8/README.Bugs --enable-languages=c,c++,go,d,fortran,objc,obj-c++ --prefix=/usr --program-suffix=-4.8 --enable-shared --enable-linker-build-id --libexecdir=/usr/lib --without-included-gettext --enable-threads=posix --with-gxx-include-dir=/usr/include/c++/4.8 --libdir=/usr/lib --enable-nls --with-sysroot=/ --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --enable-gnu-unique-object --disable-libmudflap --enable-plugin --with-system-zlib --enable-objc-gc --enable-multiarch --disable-werror --with-arch-32=i686 --with-abi=m64 --with-multilib-list=m32,m64,mx32 --with-tune=generic --enable-checking=release --build=x86_64-linux-gnu --host=x86_64-linux-gnu --target=x86_64-linux-gnu
Thread model: posix
gcc version 4.8.5 (Ubuntu 4.8.5-4ubuntu8) 
```
Almost the same for `g++-4.8 -v`, except the `COLLECT_GCC=gcc-4.8` becomes `COLLECT_GCC=g++-4.8`.

**Compiling**
Having the system setup in this way, and calling for make, yields errors (to create this log, I disabled the Warnings-to-Errors flag, otherwise I would be stuck at the first warning/error):
```
...
[...]

PR #2511, opened 2019-03-27
Suppress slab_normal warning.

LZaw:
Fixes second half of #2507.
Implemented the `always_poison` for vectors and used them on the slab_normal to suppress the compiler warning.
`always_poison` as well as `PoisonImpl<T>::do_always_posion` are only implemented for
- `float`
- `double`
- `Vector`

All the other classes are still open to implement this, as soon as they need it.

This Pullrequest may still contain changes made by #2508.


oktomus:
The appveyor job completely failed. We should restart it. @FrancoisBeaune ? :)

src/appleseed/foundation/utility/poison.h
```diff
@@ -112,6 +121,10 @@ class PoisonImpl<float>
     {
         x = FP<float>::snan();
     }
+    static void do_always_poison(float& x)
```
oktomus: Add a blank line above this.

src/appleseed/foundation/utility/poison.h
```diff
@@ -122,6 +135,9 @@ class PoisonImpl<double>
     {
         x = FP<double>::snan();
     }
+    static void do_always_poison(double& x){
```
oktomus: Also add a blank line above this.

src/appleseed/foundation/utility/poison.h
```diff
@@ -122,6 +135,9 @@ class PoisonImpl<double>
...
+        x = FP<double>::snan();
```
oktomus: What's the difference with do_poison ?

LZaw: There is none, technically one could also wrap a poison call here.

src/appleseed/foundation/utility/poison.h
LZaw: Done

src/appleseed/foundation/utility/poison.h
LZaw: Done

src/appleseed/foundation/utility/poison.h
```diff
@@ -82,6 +85,12 @@ class PoisonImpl
     }
 };
 
+template <typename T>
+inline void always_poison(T& x)
+{
+    PoisonImpl<T>::do_always_poison(x);
```
dictoon: You should simply use `PoisonImpl<T>::do_poison(x)` here. That way we don't have to add duplicates of all the poisoning methods.

LZaw: I didn't do that on purpose. You never know how encapsuled the templates are. Let's say I'm using a `Vector<Vector<float, 3>>` for whatever reason (it's not the best example). In that case, the outer Vector would be called to do the `always_poison` and the inner Vector would still be stuck in the regular `poison`, resulting in no poison.
I'm relatively unhappy with this doubled infrastructure as well, but I don't want to restrict users to use built-in types at the first template-level.

LZaw:
> I would also rename the `poison()` _free function_ to `debug_poison()` or something like that.

Will do, that's gonna take a small while, until I found all the poison calls, which have to be renamed then. :D

src/appleseed/foundation/utility/poison.h
dictoon: I'm not sure to understand your argument.

`poison<T>()` is just a wrapper for `PoisonImpl<T>::do_poison(x)`, except that it doesn't do anything in debug. All we want is a new variant of it that _always_ call `PoisonImpl<T>::do_poison(x)`, regardless of the configuration.

The poisoning itself is done by `PoisonImpl<T>` and it must do the same whether it's called via `debug_poison()` or `always_poison()`.

LZaw: Let's unwrap the chain:
In `src/appleseed/renderer/modeling/bssrdf/randomawlkbssrdf.cpp:242` the function `always_poison` gets called. This results in the following chain
```
foundation/utility/poison.h:89 (always_poison)
foundation/math/vector.h:507 (PoisonImpl<Vector<float, 3>>::do_always_poison)
foundation/utility/poision.h:125 (PoisonImpl<float>::do_always_poison)
```
In this scenario it doesn't make a difference if we call `do_always_poison` or `do_poison`, because for floats there is no debug macro (floats will always be poisoned, regardless of the build type).
Let's assume I got a `Vector<Vector<float, 3>>` and call `poison` on it. Then the call chain would look like this:
```
foundation/utility/poison.h:89 (always_poison)
foundation/math/vector.h:507 (PoisonImpl<Vector<Vector<float, 3>, 3>>::do_always_poison)
foundation/utility/poison.h:89 (always_poison)
foundation/math/vector.h:507 (PoisonImpl<Vector<float, 3>>::do_always_poison)
foundation/utility/poison.h:89 (always_poison)
foundation/utility/poision.h:125 (PoisonImpl<float>::do_always_poison)
```
This results in all floats of the inner vector to be poisoned.

But if we exchange the `do_always_poison` call for a regular `do_poison` call (as suguested), the chain looks like this:
```
foundation/utility/poison.h:89 (always_poison)
foundation/math/vector.h:500 (PoisonImpl<Vector<T, N>>::do_poison)
foundation/utility/poison.h:95 (poison)
```
The calls stop there, because the function `poison.h:95 (poison)` has the debug macro in it. We won't run into this problem, as long as the template parameter of Vectors will always be doubles or floats. But as soon as we have more complex structures, leading to a call-chain like the above mentioned one, `always_poison` will not result in a poisoned variable, which I want to prevent. From my point of view an `always` function should do its job **always**. In the above mentioned scenario it wouldn't do so.

I highly agree that this kind of double-structure is nasty and I'm thinking about how to get rid of it, without running into trouble.

The codelines of this comment refer to the as-is state in this branch. I haven't changed the `poison` functions to `debug_poison` yet. Also, I left out the `src/appleseed/` in the mutli-line structure, to increase readability.

dictoon: Oh, good point, I understand now.

Maybe a nicer solution would be to ask that all `PoisonImpl` specializations use `always_poison()` in their implementation?

LZaw:
Allright,
The old `poison` has been renamed to `debug_poison`. The `PoisonImpl<>::do_poison` now calls `always_posion`.
This way, the only difference is in the initial call to `always_poison` or `debug_poison`, with `debug_poison` having the `DEBUG` macro (and thus only doing something in debug mode).

Furthermore, I have mirrored the poison tests, so there are poison tests using `always_poison` as well. These tests will always be called, regardless of the build mode, while the `debug_poison` tests will only be available in Debug mode (just like before).
The tests have been renamed, to indicate which poison they are testing (debug or always)

All tests passed in Release build, including the "new" `always_poison` ones:
```
2019-04-17T21:44:50.965100Z <001>   113 MB info    | Foundation_Utility_Always_Poison:
2019-04-17T21:44:50.965115Z <001>   113 MB info    |   [passed] Always_Poison_Int8
2019-04-17T21:44:50.965130Z <001>   113 MB info    |   [passed] Always_Poison_UInt8
2019-04-17T21:44:50.965147Z <001>   113 MB info    |   [passed] Always_Poison_Int32
2019-04-17T21:44:50.965161Z <001>   113 MB info    |   [passed] Always_Poison_UInt32
2019-04-17T21:44:50.965173Z <001>   113 MB info    |   [passed] Always_Poison_VoidPointer
2019-04-17T21:44:50.965188Z <001>   113 MB info    |   [passed] Always_Poison_IntPointer
2019-04-17T21:44:50.965203Z <001>   113 MB info    |   [passed] Always_Poison_Float
2019-04-17T21:44:50.965220Z <001>   113 MB info    |   [passed] Always_Poison_Double
2019-04-17T21:44:50.965234Z <001>   113 MB info    |   [passed] Always_Poison_Bool
2019-04-17T21:44:50.965248Z <001>   113 MB info    |   [passed] Always_Poison_Enum

...

2019-04-17T21:45:00.041542Z <001>   123 MB info    | unit testing summary:
2019-04-17T21:45:00.041581Z <001>   123 MB info    |   suites      : 180 executed, 0 failed (0.0%)
2019-04-17T21:45:00.041581Z <001>   123 MB info    |   cases       : 1,539 executed, 0 failed (0.0%)
2019-04-17T21:45:00.041630Z <001>   123 MB info    |   assertions  : 389,581 executed, 0 failed (0.0%)
```

src/appleseed/foundation/utility/poison.h
LZaw: Done (Details below).

src/appleseed/foundation/utility/poison.h
```diff
@@ -65,7 +65,10 @@ namespace foundation
 //
 
 template <typename T>
-void poison(T& x);
+void debug_poison(T& x);
```
dictoon: The comment above needs updating both regarding which free function to use when, and how to implement `do_poison()`.

src/appleseed/foundation/utility/poison.h
```diff
@@ -122,6 +131,7 @@ class PoisonImpl<double>
     {
         x = FP<double>::snan();
     }
+
```
dictoon: Revert this change.

src/appleseed/foundation/utility/poison.h
LZaw: Ah, sorry. On it.

PR #2515, opened 2019-03-30
Update mis.h

hritikthube:
"Assertion failed: q2 >= T(0.0), file N:\appleseed\appleseed\src\appleseed\foundation/math/mis.h, line 145 "
resolved

### DIFF
--- a/src/appleseed/foundation/math/mis.h
+++ b/src/appleseed/foundation/math/mis.h
@@ -34,6 +34,7 @@
 #include "foundation/platform/compiler.h"
 
 // Standard headers.
+# define NDEBUG
 #include <algorithm>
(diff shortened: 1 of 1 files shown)


dictoon:
This makes no sense at all; in fact it's borderline insulting.

1. `NDEBUG` is defined in the build configuration for optimized builds, you're not supposed to define it yourself.
2. Defining `NDEBUG` in a header file means that not only assertions are disabled in that header file, they're also disabled in all header files that include this one. Ouch!
3. The parentheses around `q1 + q2` are useless, precedence was correct.

The problem is not in `mis_power2()` which is fully correct; there's a logic error in calling code.

PR #2516, opened 2019-03-30
Fix elusive corner case in knn query

dictoon:


oktomus:
Your code is so clean. How many times do you review it before sending a PR? :p

dictoon:
Thanks for the compliment @oktomus! I review the code once or twice before submitting a PR.

PR #2518, opened 2019-03-31
Refactor cpp and python basic sample

### LINKED ISSUE #1989: Rename basic sample
Name suggestions: render, integration

mandeep:
fixes #1989 

I liked @dictoon's suggestion of render.py so I basically renamed everything from basic to render.

Since there isn't really much to this issue, if there's anything that can be added please let me know.


mandeep:
One other thought is, do we want to keep the parent directory name `samples` or should we rename it to `examples`?

dictoon:
Hi @mandeep, thanks for the contribution, it looks like you got all of it, nice!

Regarding naming, lately I was thinking of a somewhat less ambiguous name, maybe "HelloWorld". This has become a bit of a convention for "starter examples". What do you think?

Out of curiosity, how did you regenerate the PDF file?

I also like your idea of renaming `samples` to `examples`, but I'm curious about your motivation behind this change?

mandeep:
@dictoon I had a similar thought once I started working on the cpp example. Maybe instead of render, we should call it box or cube since that is what is in the scene.

Regarding the pdf, there were a few steps I had to take since pandoc did not correctly generate the code blocks. First I converted README.md to HTML using a Python library called `grip`. Then I converted from HTML to pdf using `wkhtmltopdf`. 

The motivation behind changing `samples` to `examples` is that I think "examples" is the more general term. When I think of the word "samples", I think of something having to do with image sampling rather than examples or demos that the user can try out.

dictoon:
I agree about `samples` vs. `examples`, let's do that change.

A quick way to convert Markdown to PDF is to use Visual Studio Code with the `Markdown PDF` extension. In addition, VS Code is a decent Markdown editor.

Regarding naming: my ambition is to make the demo scene less ugly in the future thanks to using a nicer model and better materials and lighting, but still keeping things simple enough. I would settle for HelloWorld.

mandeep:
Thanks for letting me know about VS Code! I'll check it out.

I've renamed render to helloworld, let me know if you want me to squash these commits.

dictoon:
Thanks for your contribution!

mandeep:
Thanks!

PR #2522, opened 2019-03-31
Set fullscreen menu item as checkable

### LINKED ISSUE #2519: Fullscreen checkbox not shown when in full screen
In appleseed.studio when entering Fullscreen via View->Fullscreen, there is no indication (besides the layout changing) that the user is in full screen mode. In Fullscreen when the user clicks on View->Fullscreen, there isn't a checkbox next to Fullscreen indicating that the user is in this mode.

mandeep:
fixes #2519 

I have experience with PyQt, but this is my first with Qt in C++. I believe all that was needed was to set the fullscreen QAction as checkable. If not, please let me know!

### DIFF
--- a/src/appleseed.studio/mainwindow/mainwindow.cpp
+++ b/src/appleseed.studio/mainwindow/mainwindow.cpp
@@ -144,7 +144,6 @@ MainWindow::MainWindow(QWidget* parent)
     build_connections();
 
     slot_load_application_settings();
-    slot_check_fullscreen();
 
     update_project_explorer();
     update_workspace();
@@ -397,6 +396,13 @@ void MainWindow::build_menus()
     connect(m_ui->log->toggleViewAction(), SIGNAL(triggered()), SLOT(slot_check_fullscreen()));
     connect(m_ui->python_console->toggleViewAction(), SIGNAL(triggered()), SLOT(slot_check_fullscreen()));
     connect(fullscreen_action, SIGNAL(triggered()), SLOT(slot_fullscreen()));
+
+    if (any_of(m_minimize_buttons.cbegin(),
+               m_minimize_buttons.cend(),
+               [](MinimizeButton* button) {return button->is_on();})) {
+
+        fullscreen_action->setChecked(false);
+    }
 
     //
     // Rendering menu.
--- a/src/appleseed.studio/mainwindow/mainwindow.cpp
+++ b/src/appleseed.studio/mainwindow/mainwindow.cpp
@@ -144,6 +144,7 @@ MainWindow::MainWindow(QWidget* parent)
     build_connections();
 
     slot_load_application_settings();
+    slot_check_fullscreen();
 
     update_project_explorer();
     update_workspace();
@@ -390,6 +391,11 @@ void MainWindow::build_menus()
     QAction* fullscreen_action = m_ui->menu_view->addAction("Fullscreen");
     fullscreen_action->setCheckable(true);
     fullscreen_action->setShortcut(Qt::Key_F11);
+    fullscreen_action->setObjectName("Fullscreen");
+    connect(m_ui->project_explorer->toggleViewAction(), SIGNAL(triggered()), SLOT(slot_check_fullscreen()));
+    connect(m_ui->attribute_editor->toggleViewAction(), SIGNAL(triggered()), SLOT(slot_check_fullscreen()));
+    connect(m_ui->log->toggleViewAction(), SIGNAL(triggered()), SLOT(slot_check_fullscreen()));
+    connect(m_ui->python_console->toggleViewAction(), SIGNAL(triggered()), SLOT(slot_check_fullscreen()));
     connect(fullscreen_action, SIGNAL(triggered()), SLOT(slot_fullscreen()));
 
     //
@@ -2137,6 +2143,24 @@ void MainWindow::slot_fullscreen()
 
     for (MinimizeButton* button : m_minimize_buttons)
         button->set_fullscreen(m_fullscreen);
+}
+
+void MainWindow::slot_check_fullscreen()
+{
+    QAction* fullscreen = m_ui->menu_view->findChild<QAction*>("Fullscreen");
+
+    if (all_of(m_minimize_buttons.cbegin(),
+               m_minimize_buttons.cend(),
+               [](MinimizeButton* button) {return button->is_on();})) {
+
+        fullscreen->setChecked(true);
+    }
+
+    else {
+        if (fullscreen->isChecked()) {
+            fullscreen->setChecked(false);
+        }
+    }
 }
 
 void MainWindow::slot_show_application_settings_window()
--- a/src/appleseed.studio/mainwindow/mainwindow.cpp
+++ b/src/appleseed.studio/mainwindow/mainwindow.cpp
@@ -388,6 +388,7 @@ void MainWindow::build_menus()
     m_ui->menu_view->addSeparator();
 
     QAction* fullscreen_action = m_ui->menu_view->addAction("Fullscreen");
+    fullscreen_action->setCheckable(true);
     fullscreen_action->setShortcut(Qt::Key_F11);
     connect(fullscreen_action, SIGNAL(triggered()), SLOT(slot_fullscreen()));
 
--- a/src/appleseed.studio/mainwindow/mainwindow.cpp
+++ b/src/appleseed.studio/mainwindow/mainwindow.cpp
@@ -391,7 +391,6 @@ void MainWindow::build_menus()
     m_action_fullscreen = m_ui->menu_view->addAction("Fullscreen");
     m_action_fullscreen->setCheckable(true);
     m_action_fullscreen->setShortcut(Qt::Key_F11);
-    m_action_fullscreen->setObjectName("Fullscreen");
     connect(m_ui->project_explorer->toggleViewAction(), SIGNAL(triggered()), SLOT(slot_check_fullscreen()));
     connect(m_ui->attribute_editor->toggleViewAction(), SIGNAL(triggered()), SLOT(slot_check_fullscreen()));
     connect(m_ui->log->toggleViewAction(), SIGNAL(triggered()), SLOT(slot_check_fullscreen()));
--- a/src/appleseed.studio/mainwindow/mainwindow.cpp
+++ b/src/appleseed.studio/mainwindow/mainwindow.cpp
@@ -387,15 +387,15 @@ void MainWindow::build_menus()
     m_ui->menu_view->addAction(m_ui->python_console->toggleViewAction());
     m_ui->menu_view->addSeparator();
 
-    QAction* action_fullscreen = m_ui->menu_view->addAction("Fullscreen");
-    action_fullscreen->setCheckable(true);
-    action_fullscreen->setShortcut(Qt::Key_F11);
-    action_fullscreen->setObjectName("Fullscreen");
+    m_action_fullscreen = m_ui->menu_view->addAction("Fullscreen");
+    m_action_fullscreen->setCheckable(true);
+    m_action_fullscreen->setShortcut(Qt::Key_F11);
+    m_action_fullscreen->setObjectName("Fullscreen");
     connect(m_ui->project_explorer->toggleViewAction(), SIGNAL(triggered()), SLOT(slot_check_fullscreen()));
     connect(m_ui->attribute_editor->toggleViewAction(), SIGNAL(triggered()), SLOT(slot_check_fullscreen()));
     connect(m_ui->log->toggleViewAction(), SIGNAL(triggered()), SLOT(slot_check_fullscreen()));
     connect(m_ui->python_console->toggleViewAction(), SIGNAL(triggered()), SLOT(slot_check_fullscreen()));
-    connect(action_fullscreen, SIGNAL(triggered()), SLOT(slot_fullscreen()));
+    connect(m_action_fullscreen, SIGNAL(triggered()), SLOT(slot_fullscreen()));
 
     //
     // Rendering menu.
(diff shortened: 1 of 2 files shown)


oktomus:
Does it check and uncheck correctly? Even with you leabe fullscreen by other ways than the buton?

mandeep:
Right now fullscreen can be entered by clicking on the Fullscreen menu item or by hitting the F11 shortcut key. Looking at the code, it looks like deselecting all of the menu items also enters fullscreen mode. Is this the case? If so, I'll push another commit for that.

oktomus:
To be honest, I have no idea :') The best is just to test it.

mandeep:
Sorry if I wasn't clear. The checkbox works in the scenarios I mentioned. However, if all buttons are unchecked, the checkbox for the Fullscreen remains unchecked. I'm not sure if the checkbox should be checked in this case.

oktomus:
Great. I don't know what other buttons you are talking about though :/ I don't have the UI in mind.

dictoon:
> Right now fullscreen can be entered by clicking on the Fullscreen menu item or by hitting the F11 shortcut key. Looking at the code, it looks like deselecting all of the menu items also enters fullscreen mode. Is this the case? If so, I'll push another commit for that.

It's indeed the case.

mandeep:
@dictoon Thanks, I'll push a commit fixing this.

mandeep:
I've added a slot that iterates through the minimize buttons and checks the Fullscreen box if all of the buttons are unchecked. In order to do this, a new slot was added and an objectName was given to the Fullscreen action which allows us to find the Fullscreen action through findChildren(). 

I'm not sure if this is the best way to do this since we are iterating through all of the buttons each time the slot is triggered. I'm very open to other ideas on how to solve this.

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -388,9 +388,22 @@ void MainWindow::build_menus()
     m_ui->menu_view->addSeparator();
 
     QAction* fullscreen_action = m_ui->menu_view->addAction("Fullscreen");
+    fullscreen_action->setCheckable(true);
     fullscreen_action->setShortcut(Qt::Key_F11);
+    fullscreen_action->setObjectName("Fullscreen");
```
dictoon: The name should be `action_fullscreen` in order to be consistent with other names.

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -2138,6 +2151,24 @@ void MainWindow::slot_fullscreen()
         button->set_fullscreen(m_fullscreen);
 }
 
+void MainWindow::slot_check_fullscreen()
+{
+    QAction* fullscreen = m_ui->menu_view->findChild<QAction*>("Fullscreen");
+
+    if (all_of(m_minimize_buttons.cbegin(),
+               m_minimize_buttons.cend(),
+               [](MinimizeButton* button) {return button->is_on();})) {
```
dictoon: In our coding conventions, all braces go on their own lines.

dictoon: Is the header file declaring `all_of()` included?

dictoon: The logic seems convoluted, what about something like

    const bool is_fullscreen = all_of(...);
    fullscreen->setChecked(is_fullscreen);

Also, are you sure all buttons must be `on`? Shouldn't it be off?

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -388,9 +388,22 @@ void MainWindow::build_menus()
...
+    fullscreen_action->setCheckable(true);
     fullscreen_action->setShortcut(Qt::Key_F11);
+    fullscreen_action->setObjectName("Fullscreen");
+    connect(m_ui->project_explorer->toggleViewAction(), SIGNAL(triggered()), SLOT(slot_check_fullscreen()));
+    connect(m_ui->attribute_editor->toggleViewAction(), SIGNAL(triggered()), SLOT(slot_check_fullscreen()));
+    connect(m_ui->log->toggleViewAction(), SIGNAL(triggered()), SLOT(slot_check_fullscreen()));
+    connect(m_ui->python_console->toggleViewAction(), SIGNAL(triggered()), SLOT(slot_check_fullscreen()));
     connect(fullscreen_action, SIGNAL(triggered()), SLOT(slot_fullscreen()));
 
+    if (any_of(m_minimize_buttons.cbegin(),
```
dictoon: Simply call `slot_check_fullscreen()` to avoid logic duplication.

src/appleseed.studio/mainwindow/mainwindow.cpp
mandeep: Sounds good. I've pushed a commit that addresses this.

src/appleseed.studio/mainwindow/mainwindow.cpp
mandeep: all_of is coming from using namespace std. 

The buttons work opposite than what's expected. On returns 0 and off returns 1. This can be changed but I think there's more potential for new bugs. Do you want me to dig into this?

src/appleseed.studio/mainwindow/mainwindow.cpp
mandeep: This actually doesn't work anymore. I'm not sure what exactly changed between Qt4 and Qt5, but now the checkboxes load after everything else. I tried calling `slot_check_fullscreen` elsewhere in the code, but it is always setting the screen to fullscreen because the other checkboxes haven't loaded. Is there somewhere you recommend me look?

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -2138,6 +2144,17 @@ void MainWindow::slot_fullscreen()
         button->set_fullscreen(m_fullscreen);
 }
 
+void MainWindow::slot_check_fullscreen()
+{
+    QAction* fullscreen = m_ui->menu_view->findChild<QAction*>("Fullscreen");
```
dictoon: It would be cleaner IMHO to store the action in the class and use it, instead of looking for it by name every time.

src/appleseed.studio/mainwindow/mainwindow.cpp
dictoon: > all_of is coming from using namespace std.

Yes but that's not the question I was asking :) `std::all_of()` is declared in `<algorithm>`. It so happens that it's already included so we're good here.

src/appleseed.studio/mainwindow/mainwindow.cpp
dictoon: Do I understand correctly that as it stands now (with the two new commits you made recently), it doesn't work?

mandeep: Everything works except during startup. I believe something changed in the Qt4 to Qt5 transition. I had programmed it so that the Fullscreen checkbox would be checked in the following scenario:

1. User unchecks all checkboxes
2. User closes appleseed.studio
3. User reopens appleseed.studio and finds that all checkboxes are unchecked except for the Fullscreen checkbox

However, now the Fullscreen checkbox is not checked in this case.

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -387,9 +387,15 @@ void MainWindow::build_menus()
     m_ui->menu_view->addAction(m_ui->python_console->toggleViewAction());
     m_ui->menu_view->addSeparator();
 
-    QAction* fullscreen_action = m_ui->menu_view->addAction("Fullscreen");
-    fullscreen_action->setShortcut(Qt::Key_F11);
-    connect(fullscreen_action, SIGNAL(triggered()), SLOT(slot_fullscreen()));
+    QAction* action_fullscreen = m_ui->menu_view->addAction("Fullscreen");
```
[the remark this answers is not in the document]
mandeep: The argument that is passed to addAction is also the text of the item as it appears in the view menu. If we change it to `action_fullscreen` the text in the view menu will also read `action_fullscreen`.

src/appleseed.studio/mainwindow/mainwindow.cpp
mandeep: Good idea!

src/appleseed.studio/mainwindow/mainwindow.cpp
mandeep: Got it, thanks!

src/appleseed.studio/mainwindow/mainwindow.cpp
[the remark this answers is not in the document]
dictoon: Oh, you're right, sorry.

src/appleseed.studio/mainwindow/mainwindow.cpp
dictoon: Isn't the problem simply that you check the status of the checkboxes _before_ `slot_load_application_settings()` is called? It's called in `MainWindow`'s constructor.

mandeep: No, initially in my tests I called `slot_check_fullscreen()` after `slot_load_application_settings()` in the constructor, however this came with its own problems.

The checkboxes aren't checked when the MainWindow is constructed, thus the Fullscreen checkbox always starts as checked. To debug this I tried calling `slot_check_fullscreen()` after startup and everything worked as expected. This leads me to believe this problem has something to do with how things are initialized at startup.

mandeep: @dictoon I have fixed the issue in the latest commit. The problem was that I was calling the MinimizeButton `is_on()` method, but the method was not in sync with the QDockWidget `isHidden()` method. 

Now `findChildren() is being used to find all of the QDockWidgets. I'm not sure this should be an instance variable like m_action_fullscreen, but if you want me to make it one I will do so.

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -387,9 +388,15 @@ void MainWindow::build_menus()
     m_ui->menu_view->addAction(m_ui->python_console->toggleViewAction());
     m_ui->menu_view->addSeparator();
 
-    QAction* fullscreen_action = m_ui->menu_view->addAction("Fullscreen");
-    fullscreen_action->setShortcut(Qt::Key_F11);
-    connect(fullscreen_action, SIGNAL(triggered()), SLOT(slot_fullscreen()));
+    m_action_fullscreen = m_ui->menu_view->addAction("Fullscreen");
+    m_action_fullscreen->setCheckable(true);
+    m_action_fullscreen->setShortcut(Qt::Key_F11);
+    m_action_fullscreen->setObjectName("Fullscreen");
```
dictoon: Remember my previous comment about the name of the action? It turns out I had simply commented on the wrong line :) The argument to `setObjectName()` should be `action_fullscreen`. But I wonder if we really need a name for that action; I think not, in which case please remove the call altogether.

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -2138,6 +2145,17 @@ void MainWindow::slot_fullscreen()
         button->set_fullscreen(m_fullscreen);
 }
 
+void MainWindow::slot_check_fullscreen()
+{
+    QList<QDockWidget*> dock_widgets = findChildren<QDockWidget*>();
```

dictoon:
As for `findChildren()`, I think it's fine like you did it...

src/appleseed.studio/mainwindow/mainwindow.cpp
mandeep: Nope, we don't need it so I removed it. :smile:

src/appleseed.studio/mainwindow/mainwindow.cpp
mandeep: Sounds good!

mandeep:
@dictoon Thanks for taking the time to review this small feature.

dictoon:
Thanks for the quality work and your patience through the review process, much appreciated :)

dictoon:
PR approved, will be merged when CI completes.

I was wondering if this code could be made more generic (and more robust to changes) thanks the `findChildren()` trick?

```cpp
connect(m_ui->project_explorer->toggleViewAction(), SIGNAL(triggered()), SLOT(slot_check_fullscreen()));
connect(m_ui->attribute_editor->toggleViewAction(), SIGNAL(triggered()), SLOT(slot_check_fullscreen()));
connect(m_ui->log->toggleViewAction(), SIGNAL(triggered()), SLOT(slot_check_fullscreen()));
connect(m_ui->python_console->toggleViewAction(), SIGNAL(triggered()), SLOT(slot_check_fullscreen()));
connect(m_action_fullscreen, SIGNAL(triggered()), SLOT(slot_fullscreen()));
```

mandeep:
@dictoon From my understanding we need to call connect on each object.

dictoon:
Merging despite macOS build failing on Travis which is unrelated to this PR (it's about the OIIO Homebrew package that got updated from 1.8.17 to 2.0.7).

dictoon:
> @dictoon From my understanding we need to call connect on each object.

Why can't we simply loop on the dockable widgets and call `connect()` on them?

mandeep:
When we loop through the dockable widgets, the last widget in the QList will be the one connected to the slot.

dictoon:
Hm. What about Qt 5's new `connect()` syntax?

mandeep:
I gave the new syntax a try and faced the same problem. If you want I can submit a PR that refactors the old syntax to the new syntax anyway.

dictoon:
We'll most likely want to make a pass over the code to switch all `connect()` calls to the new syntax but it's not a priority.

I'm still not quite sure why calling `connect()` in a loop doesn't work. I've searched on the web but couldn't find anything related to that.

Could you post the code you tried?

mandeep:
This is what I tried:
```cpp
    const QList<QDockWidget*> dock_widgets = findChildren<QDockWidget*>();
    for (const QDockWidget* dock : dock_widgets)
    {
        connect(dock->toggleViewAction(), &QAction::triggered, this, &MainWindow::slot_check_fullscreen);
    }
```

I'll continue digging and see if I can come up with a solution to this.

dictoon:
The code looks good. There must be a good reason why Qt doesn't support this but right now it eludes me.

You could post this code on Stack Overflow.

mandeep:
I'm starting to think that the code for this is correct, but there may be a problem with `slot_fullscreen()`. I'll test some more and submit a PR.

PR #2523, opened 2019-04-01
Change the pause/resume text based on rendering state

mandeep:
When the render is rendering, the Pause/Resume menu item
can lead to confusion as it is not clear which state the renderer
is in based on a checkbox. Does the checkbox mean that the renderer
is running, or does it mean that it is paused? This change makes it
clear which state the renderer is in. If the renderer is running
the menu item will say 'Pause Rendering', likewise if the renderer
is paused the menu item will say 'Resume Rendering'. When the rendering
is canceled or finished, the menu item will revert back to
'Pause/Resume Rendering'.

### DIFF
--- a/src/appleseed.studio/mainwindow/mainwindow.cpp
+++ b/src/appleseed.studio/mainwindow/mainwindow.cpp
@@ -1685,11 +1685,13 @@ void MainWindow::slot_apply_application_settings()
 
 void MainWindow::slot_start_interactive_rendering()
 {
+    m_ui->action_rendering_pause_resume_rendering->setText("Pause Rendering");
     start_rendering(InteractiveRendering);
 }
 
 void MainWindow::slot_start_final_rendering()
 {
+    m_ui->action_rendering_pause_resume_rendering->setText("Pause Rendering");
     start_rendering(FinalRendering);
 }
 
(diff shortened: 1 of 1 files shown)


pjessesco:
Do we need `Pause/Resume Rendering`? Maybe we can set `Pause Rendering` as default.

mandeep:
I think `Pause Rendering` as default could be confusing as it could lead people to believe that there is a render currently in process.

oktomus:
I agree with @mandeep  for `Pause Rendering`.

mandeep:
I guess it's possible to use `Pause Rendering`, but in order to do so you we would probably want to grey out the option if there is not a render in progress. I'm happy either way.

dictoon:
I think I would also prefer not to alter the menu item's text on the fly and just settle for

    [ ] Pause Rendering

That menu item is anyway grayed out when a render is not in progress.

mandeep:
I'm not sure I like keeping it as `Pause Rendering` because then the user has to click on something that says Pause when they want to resume their render. If changing on the fly is off the table, I think we should keep it as is.

dictoon:
> I'm not sure I like keeping it as Pause Rendering because then the user has to click on something that says Pause when they want to resume their render.

I understand your point, but if it's a checkbox, it should be pretty clear that clicking on the Pause checkbox _unchecks it and thus resumes the render_, no?

mandeep:
I'm not sure this is that obvious, but let's hope that it is. :smiley:

dictoon:
Ok so let's go with simply `[ ] &Pause Rendering` (the PR will need to be changed accordingly, it should no longer be needed to alter the item's title on the fly). It will already be an improvement over what we have right now. Thanks!

PR #2525, opened 2019-04-02
Remove curly brace initialization to support VS2012

usakhelo:


src/appleseed.studio/mainwindow/rendering/materialdrophandler.cpp
```diff
@@ -78,6 +78,12 @@ namespace
     {
         std::string                     m_name;
         renderer::ObjectInstance::Side  m_side;
+
+        explicit MaterialSlot(const std::string name, const renderer::ObjectInstance::Side side)
```
dictoon: I'm surprised that this compiles without a default constructor, but if it does, then fine :)

usakhelo: Hi @dictoon , 
Yes it does. Do you want me to rebase this PR?

usakhelo: I checked recent commits and it looks like we are going to drop 2012. Do you want me to delete this PR instead?

dictoon: Yes, it's not longer needed. Thanks for your work though!

dictoon:
VS 2012 support dropped.

PR #2529, opened 2019-04-04
Expose OIIO make_texture function to Python bindings.

jdent02:


jdent02:
I'm sure there's a better way to add the libraries other than listing every one but I'm still a relative newb on CMake.

src/appleseed.python/bindutility.cpp
```diff
@@ -6,7 +6,7 @@
 // This software is released under the MIT license.
 //
 // Copyright (c) 2012-2013 Esteban Tovagliari, Jupiter Jazz Limited
-// Copyright (c) 2014-2018 Esteban Tovagliari, The appleseedhq Organization
+// Copyright (c) 2014-2019 The appleseedhq Organization
```
oktomus: Leave esteban's name, just change the year

src/appleseed.python/CMakeLists.txt
```diff
@@ -213,6 +213,11 @@ target_link_libraries (appleseed.python
     appleseed
     ${Boost_PYTHON_LIBRARY}             # all Boost libraries but Boost.Python3
     ${PYTHON_LIBRARIES}
+    ${APPLESEED_DEPS_STAGE_DIR}/oiio-release/lib/OpenImageIO.lib
```
dictoon: That's Windows-specific and won't compile on other platforms (as reported by Travis).

Use `link_against_oiio (appleseed.python)` instead. Check how it's done in the case of appleseed's core library.

src/appleseed.python/CMakeLists.txt
```diff
@@ -222,6 +227,11 @@ if (WITH_PYTHON3_BINDINGS)
         appleseed
         ${Boost_PYTHON3_LIBRARY}        # all Boost libraries but Boost.Python
         ${PYTHON3_LIBRARY}
+        ${APPLESEED_DEPS_STAGE_DIR}/oiio-release/lib/OpenImageIO.lib
```
dictoon: `link_against_ocio (appleseed.python)`

src/appleseed.python/bindutility.cpp
```diff
@@ -34,9 +34,59 @@
...
 using namespace renderer;
+using namespace std;
+
+void oiio_make_texture(
+    const string&   in_filename,
+    const string&   out_filename,
+    const string&   in_colorspace,
+    const string&   out_depth)
+{
+    unordered_map<string, TypeDesc> out_depth_map{
```
dictoon: Brace initialization is not supported by VS 2012.

src/appleseed.python/bindutility.cpp
```diff
@@ -34,9 +34,59 @@
...
+        {"uint8", TypeDesc::UINT8},
+        {"uint16", TypeDesc::UINT16},
+        {"sint16", TypeDesc::INT16},
+        {"half", TypeDesc::HALF},
+        {"float", TypeDesc::FLOAT}};
+
+    ImageSpec spec;
+
+    if (out_depth != "default")
+    {
```
dictoon: We usually leave braces out for single line `if` bodies.

src/appleseed.python/bindutility.cpp
```diff
@@ -34,9 +34,59 @@
...
+    spec.attribute("maketx:incolorspace", in_colorspace);
+    spec.attribute("maketx:outcolorspace", "linear");
+    spec.attribute("maketx:fixnan", "box3");
+
+    ImageBufAlgo::MakeTextureMode mode = ImageBufAlgo::MakeTxTexture;
+
+    stringstream s;
+
+    if (!make_texture(mode, in_filename, out_filename, spec, &s))
+    {
```
dictoon: Same here.

src/appleseed.python/bindutility.cpp
```diff
@@ -34,9 +34,59 @@
...
+#include <sstream>
+
 namespace bpy = boost::python;
 using namespace foundation;
+using namespace OIIO;
 using namespace renderer;
+using namespace std;
+
+void oiio_make_texture(
+    const string&   in_filename,
```
dictoon: Include `<string>`.

src/appleseed.python/bindutility.cpp
```diff
@@ -34,9 +34,59 @@
...
+    spec.attribute("maketx:unpremult", 1);
+    spec.attribute("maketx:incolorspace", in_colorspace);
+    spec.attribute("maketx:outcolorspace", "linear");
+    spec.attribute("maketx:fixnan", "box3");
+
+    ImageBufAlgo::MakeTextureMode mode = ImageBufAlgo::MakeTxTexture;
+
+    stringstream s;
+
+    if (!make_texture(mode, in_filename, out_filename, spec, &s))
```
dictoon: I'm surprised that this works without the `ImageBufAlgo` namespace qualifier... How come?

src/appleseed.python/CMakeLists.txt
jdent02: I already did, just hadn't pushed it yet.

src/appleseed.python/CMakeLists.txt
jdent02: same here

src/appleseed.python/bindutility.cpp
jdent02: Not sure, ReSharper marked it as redundant and it compiled just fine without it.  Looking at the imagebufalgo header shows that it is in that namespace.

So I dunno.......

dictoon: What's even more surprising is that in other places such as `renderer/modeling/postprocessingstage/renderstamppostprocessingstage.cpp` where we use functions from `ImageBufAlgo`, removing the namespace qualifiers clearly no longer compiles...

jdent02: I can add it in just for consistency

jdent02:
Think I covered everything

src/appleseed.python/bindutility.cpp
```diff
@@ -34,9 +34,56 @@
...
 using namespace renderer;
+using namespace std;
+
+void oiio_make_texture(
+    const string&   in_filename,
+    const string&   out_filename,
+    const string&   in_colorspace,
+    const string&   out_depth)
+{
+    unordered_map<string, TypeDesc> out_depth_map;
```
dictoon: You need to include `<unordered_map>`, missed it the first time, sorry.

src/appleseed.python/bindutility.cpp
```diff
@@ -34,9 +34,56 @@
...
+    spec.attribute("maketx:updatemode", 1);
+    spec.attribute("maketx:constant_color_detect", 1);
+    spec.attribute("maketx:monochrome detect", 1);
+    spec.attribute("maketx:opaque detect", 1);
+    spec.attribute("maketx:unpremult", 1);
+    spec.attribute("maketx:incolorspace", in_colorspace);
+    spec.attribute("maketx:outcolorspace", "linear");
+    spec.attribute("maketx:fixnan", "box3");
+
+    ImageBufAlgo::MakeTextureMode mode = ImageBufAlgo::MakeTxTexture;
```
dictoon: Make const.

src/appleseed.python/CMakeLists.txt
```diff
@@ -208,6 +208,8 @@ endif ()
 #--------------------------------------------------------------------------------------------------
 
 link_against_platform (appleseed.python)
+link_against_ocio(appleseed.python)
```
dictoon: Insert one space before `(`.

PR #2530, opened 2019-04-05
Normalized Pixel Variation AOV wrt the user noise threshold

### LINKED ISSUE #2404: Normalize Pixel Variation AOV wrt. the user's noise threshold

kushWithoutWax:
Fixes #2404 

Issue: After applying the post-processing stage, the image before and after this change look the same.

### DIFF
--- a/src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
+++ b/src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
@@ -542,7 +542,7 @@ namespace
                         {
                             Color3f variation;
                             m_variation_aov_tile->get_pixel(pt.x, pt.y, variation);
-                            variation[0] += pb.m_block_error;
+                            variation[0] += pb.m_block_error / m_params.m_noise_threshold;
                             m_variation_aov_tile->set_pixel(pt.x, pt.y, variation);
                         }
                     }


src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
```diff
@@ -552,7 +552,7 @@ namespace
                         {
                             Color3f variation;
                             m_variation_aov_tile->get_pixel(pt.x, pt.y, variation);
-                            variation[0] += pb.m_block_error;
+                            variation[0] += pb.m_block_error / m_params.m_noise_threshold;
```
oktomus: What if the noise threshold is zero ?

kushWithoutWax: The noise threshold has a minimum value of 0.0001 . It cannot be set to 0 either in the .appleseed file or through appleseed studio.
https://github.com/appleseedhq/appleseed/blob/127bbc39b2bd9a4cbcaff2b94552757aec6f8058/src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp#L1014 ensures it I think.

oktomus: Ok nice

oktomus: Ok nice

oktomus: Could you share some images of what this change does ?

kushWithoutWax: Screenshot from 2019-04-05 07-16-09
The scene 
NormBuild
Using thresholds of 0.004, 0.3, 0.6
NormalWithoutPost
Without the post processing stage

The problem with this is that after the post processing stage, the aov image with my changes on and those without are exactly the same. I am trying to figure out what is causing the problem

oktomus: ok. I think i already tried this when I first implented it and it wasnt that great. Though I was normalizing while rendering whereas here you normalize it only for the final aov.

PR #2532, opened 2019-04-06
Normalized the Pixel Variation AOV wrt the user's noise threshold

kushWithoutWax:
Fixes #2477 

Discussion on this issue on #2530 and #2477.

### DIFF
--- a/src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
+++ b/src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
@@ -496,7 +496,8 @@ namespace
 
             size_t tile_converged_pixel_count = 0;
             float average_noise_level = 0.0f;
-
+            const float normalizing_factor = 1.0f / m_params.m_noise_threshold;
+            
             for (size_t i = 0, n = rendering_blocks.size(); i < n; ++i)
             {
                 const PixelBlock& pb = rendering_blocks[i];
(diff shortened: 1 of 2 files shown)


src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
```diff
@@ -542,7 +542,7 @@ namespace
                         {
                             Color3f variation;
                             m_variation_aov_tile->get_pixel(pt.x, pt.y, variation);
-                            variation[0] += pb.m_block_error;
+                            variation[0] += pb.m_block_error / m_params.m_noise_threshold;
```
dictoon: That's actually a pretty nice solution as it doesn't require injecting the noise threshold into the AOV.

You can precompute `1.0f / m_params.m_noise_threshold`.

src/appleseed/renderer/modeling/aov/pixelvariationaov.cpp
```diff
@@ -79,9 +79,9 @@ namespace
             ColorMap color_map;
             color_map.set_palette_from_array(InfernoColorMap, countof(InfernoColorMap) / 3);
 
-            float min_var, max_var;
-            color_map.find_min_max_red_channel(*m_image, crop_window, min_var, max_var);
-            color_map.remap_red_channel(*m_image, crop_window, min_var, max_var);
+            //Clamps the pixel variation in the red channel between 0 and 1.
```
dictoon: Insert spaces after `//`.

kushWithoutWax: Done.

src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
kushWithoutWax: Thanks @dictoon . I made the change.
I realized that we did not need to inject the noise threshold into the AOV since the variation was computed in the Adaptive Tile renderer.

PR #2537, opened 2019-04-09
Add 3ds max parameter ID meta-data to each OSL shader parameter

usakhelo:
Hi guys,
I was holding on this PR for too long. Finally decided to publish it. It's nothing huge, just adds each OSL parameter an ID for 3ds max plugin. By using these IDs 3ds max plugin will use the same param_id-s for each parameter and adding new parameters won't break scene compatibility.

Right now numbers reflect current 3ds max param_id-s. If plugin uses these IDs right now parameters will get the same param_id-s as before, maintaining compatibility.

I have an issue with parameters in helper .h files, such as `MAYA_COLORMANAGEMENT_PARAMETERS`. They appear in different places in plugins and get different id-s. I think shaders with that parameters are not widely used and probably I could assign some high numbers to these parameters to not make conflicts to any of the plugins.

Thanks,
Sergo.

### DIFF
--- a/src/appleseed.shaders/src/appleseed/as_attributes.osl
+++ b/src/appleseed.shaders/src/appleseed/as_attributes.osl
@@ -46,179 +46,208 @@ shader as_attributes
     [[
         string as_maya_attribute_name = "objectInstanceId",
         string as_maya_attribute_short_name = "oid",
-        string label = "Object Instance ID"
+        string label = "Object Instance ID",
+        int as_max_param_id = 0
     ]],
     output int out_object_instance_index = 0
     [[
         string as_maya_attribute_name = "objectInstanceIndex",
         string as_maya_attribute_short_name = "odx",
-        string label = "Object Instance Index"
+        string label = "Object Instance Index",
+        int as_max_param_id = 0
     ]],
     output string out_object_instance_name = ""
     [[
         string as_maya_attribute_name = "objectInstanceName",
         string as_maya_attribute_short_name = "oin",
-        string label = "Object Instance Name"
+        string label = "Object Instance Name",
+        int as_max_param_id = 0
     ]],
     output string out_object_name = ""
     [[
         string as_maya_attribute_name = "objectName",
         string as_maya_attribute_short_name = "onn",
-        string label = "Object Name"
+        string label = "Object Name",
+        int as_max_param_id = 0
     ]],
     output int out_assembly_instance_id = 0
     [[
         string as_maya_attribute_name = "assemblyInstanceId",
         string as_maya_attribute_short_name = "aid",
-        string label = "Assembly Instance ID"
+        string label = "Assembly Instance ID",
+        int as_max_param_id = 0
     ]],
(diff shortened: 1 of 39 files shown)


src/appleseed.shaders/src/appleseed/as_attributes.osl
```diff
@@ -46,179 +46,208 @@ shader as_attributes
...
         string as_maya_attribute_short_name = "oid",
-        string label = "Object Instance ID"
+        string label = "Object Instance ID",
+        int as_max_param_id = 0
     ]],
     output int out_object_instance_index = 0
     [[
         string as_maya_attribute_name = "objectInstanceIndex",
         string as_maya_attribute_short_name = "odx",
-        string label = "Object Instance Index"
```
dictoon: Same ID for all parameters?

src/appleseed.shaders/src/appleseed/as_blackbody.osl
```diff
@@ -224,25 +238,29 @@ shader as_blackbody
...
         string as_maya_attribute_short_name = "om",
-        string widget = "null"
+        string widget = "null",
+        int as_max_param_id = 21
     ]],
     output color out_blackbody_color = color(0)
     [[
         string as_maya_attribute_name = "outBlackbodyColor",
         string as_maya_attribute_short_name = "bbc",
-        string widget = "null"
```
dictoon: Multiple parameters with same ID?

usakhelo: Output parameters get the same id because right now the are listed using single parameter in 3ds max.

src/appleseed.shaders/src/appleseed/as_attributes.osl
luisbarrancos: @usakhelo why the label removal?

src/appleseed.shaders/src/appleseed/as_blackbody.osl
luisbarrancos: @usakhelo how can i find the IDs that max gives to the parameters?

usakhelo: Hi @luisbarrancos ,
I added debug print statements to osl plugin to get ids to each parameter. Is that the question?
Also in normal plugins IDs are assigned in `ParamBlockDesc2` constructor.

dictoon: Thanks for the explanation @usakhelo.

Are you reviewing this PR @luisbarrancos?

luisbarrancos: @dictoon i was taking a look yes, since we have https://github.com/appleseedhq/appleseed/pull/2426 draft on hold due to new parameters being added. I'll be updating the PR soon and once done perhaps @usakhelo can try the draft PR with Max. So far i was unable to get it working on Wine and unable to get GPU pass-through for a windows guest on KVM.

PR #2542, opened 2019-04-10
Many fixes, improvements and refactorings

dictoon:


src/appleseed/foundation/image/conversion.cpp
```diff
@@ -0,0 +1,150 @@
...
+namespace foundation
+{
+
+bool is_linear_image_file_format(const bf::path& path)
+{
+    const string extension = lower_case(path.extension().string());
+
+    return
+        extension == ".exr"  ||
+        extension == ".tiff" ||
```
est77: We cannot assume that TIFFs contain floating point pixels. Most do not.
Checking in Frame, we are saving linear 16 bits TIFFs, which is wrong. 
Images with integer pixels should be sRGB (I opened an issue to track that).

dictoon: Ok.

dictoon:
Thanks for the review.

src/appleseed/foundation/image/conversion.cpp
luisbarrancos: I'm going to be pedantic here, but you **do** have enough precision with uint16 TIFFs to have a smooth tonal progression without banding ([Barten threshold](https://www.mysterybox.us/blog/2016/10/19/hdr-video-part-3-hdr-video-terms-explained)), though using a perceptual encoding would make better use of the format.

PR #2543, opened 2019-04-11
Fix gcc 4 errors when including QtConcurrent

est77:


src/appleseed.studio/mainwindow/project/projectmanager.cpp
```diff
@@ -89,8 +91,14 @@ void ProjectManager::load_project_async(const string& filepath)
 
     m_async_io_filepath = QString::fromStdString(filepath);
 
+#if defined(__GNUC__) && (__GNUC__ > 4)
```
dictoon: This code has been there for ages, did it not compile with gcc 4.8?

src/appleseed.studio/mainwindow/project/projectmanager.cpp
```diff
@@ -89,8 +91,14 @@ void ProjectManager::load_project_async(const string& filepath)
...
     m_async_io_future_watcher.setFuture(
         QtConcurrent::run(this, &ProjectManager::do_load_project, filepath));
+#else
+    const bool successful = do_load_project(filepath);
```
dictoon: Is this actually working? An alternative would be to spawn a thread, it's basically what QtConcurrent does except that it picks the thread from a thread pool.

dictoon: Is this actually working? An alternative would be to spawn a thread, it's basically what QtConcurrent does except that it picks the thread from a thread pool.

src/appleseed.studio/mainwindow/project/projectmanager.cpp
est77: It does not compile when using Qt 5.12.

est77: It does not compile when using Qt 5.12. We cannot include QtConcurrent when using gcc 4.8.

src/appleseed.studio/mainwindow/project/projectmanager.cpp
est77: In my tests it worked. It is doing the same thing as before but on the main thread, no?

src/appleseed.studio/mainwindow/project/projectmanager.cpp
est77: Does not compile when using Qt 5.12. Since this is the only problem, I tried a quick workaround.

dictoon: Ok.

src/appleseed.studio/mainwindow/project/projectmanager.cpp
dictoon: Ok.

> It is doing the same thing as before but on the main thread, no?

Yes. I guess the only difference is that log messages emitted during loading won't be visible until the project is actually fully loaded (or has failed to load).

dictoon:
Why did you close this PR?

Issue #768, opened 2015-06-17
Review Bézier curve intersection code


dictoon:
Compare our implementation with the one in PBRT:
https://github.com/mmp/pbrt-v3/blob/master/src/shapes/curve.cpp

Reproduce the issue reported (and fixed) by Koji, and create a test scene that demonstrates the fix:
https://github.com/appleseedhq/appleseed/pull/767#issuecomment-112521098

lorentzo:
H!
I would like to look more into this issue and hopefully find a solution.

dictoon:
Done.

PR #2552, opened 2019-04-14
bezier curve intersection test scene

lorentzo:
Hello!

I have created test scene for bezier curve intersection issue #768 
I have used blender with blenderseed plugin to create scene as described in #767 (comment). Finally I have
exported scene in applessed format which is file in this pull request.

Please, let mi know if this is OK!


dictoon:
Hm, as far as I can tell your scene using a `*.binarymesh` file which is a triangle mesh and not a Bezier curve.

lorentzo:
Oh, yes! Sorry, I have used curve to mesh conversion before exporting. 
I will set this right.

PR #2559, opened 2019-04-17
Implement BDPT density caching

wilimitis:
As mentioned in PBRT, iteratively considering hypothetical subpaths results in an _O(n^4)_ runtime. This PR attempts to implement density caching as discussed in PBRT by only considering the density ratios between adjacent paths.

A primary aim of this PR is to prove that no actual rendering changes were introduced. I believe this has been achieved as a 32 sample Cornell box test results in `0.0004` rmse between the proposed change and the existing implementation (this number, I believe, results from `float`<->`double` conversions, and decreases as sample size increases; I can further elaborate if needed). There may be more test results worth exploring. To debug subtle errors I rendered and examined isolated layers for combinations of `s = [0, ..., 3]` and `t = [2, ..., 5]`.

Old
c,samples=32

New
cli


src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.cpp
```diff
@@ -93,19 +105,42 @@ namespace
...
             if (dist2 == 0.0)
                 return 0.0;
             const double rcp_dist2 = 1.0 / dist2;
-            pdf *= max(dot(vertex.m_geometric_normal, w * sqrt(rcp_dist2)), 0.0);
-            return pdf * rcp_dist2;
+            pdf *= std::abs(dot(next.m_geometric_normal, w * sqrt(rcp_dist2)));
+            return static_cast<float>(pdf * rcp_dist2);
+        }
+
+        float pdf_light(const BDPTVertex* vertex)
```
dictoon: Make method const.

Replace pointer by reference since it cannot be null.

src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.cpp
```diff
@@ -93,19 +105,42 @@ namespace
...
         {
-            const Vector3d w = m_position - vertex.m_position;
+            const Vector3d w = next.m_position - m_position;
             const double dist2 = square_norm(w);
             if (dist2 == 0.0)
                 return 0.0;
             const double rcp_dist2 = 1.0 / dist2;
-            pdf *= max(dot(vertex.m_geometric_normal, w * sqrt(rcp_dist2)), 0.0);
-            return pdf * rcp_dist2;
+            pdf *= std::abs(dot(next.m_geometric_normal, w * sqrt(rcp_dist2)));
```
dictoon: `std::` prefixes are redundant in this file since there's a `using` clause at the top.

src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.cpp
```diff
@@ -93,19 +105,42 @@ namespace
...
 
-        double convert_density(double pdf, const BDPTVertex& vertex) const
+        // Returns the density converted from probability per steridian to probability per unit area (Veach 8.2.2.2).
+        float convert_density(double pdf, const BDPTVertex& next) const
         {
-            const Vector3d w = m_position - vertex.m_position;
+            const Vector3d w = next.m_position - m_position;
             const double dist2 = square_norm(w);
             if (dist2 == 0.0)
                 return 0.0;
```
dictoon: `0.0f` would be slightly clearer.

src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.cpp
```diff
@@ -214,126 +249,60 @@ namespace
...
             BDPTVertex*                 light_vertices,
             BDPTVertex*                 camera_vertices,
             const size_t                s,
-            const size_t                t,
-            const size_t                p,
-            const size_t                q)
+            const size_t                t)
         {
-            assert(p + q == s + t);
+            float sum_ri = 0.0f;  // The sum of the ratios of path densities for adjacent sampling strategies.
```
dictoon: End-of-line comments should not start with a capital nor end with a period (they're not really sentences, more annotations).

src/appleseed/renderer/kernel/lighting/bdpt/bdptlightingengine.cpp
```diff
@@ -93,19 +105,42 @@ namespace
...
-            const Vector3d w = m_position - vertex.m_position;
+            const Vector3d w = next.m_position - m_position;
             const double dist2 = square_norm(w);
             if (dist2 == 0.0)
                 return 0.0;
             const double rcp_dist2 = 1.0 / dist2;
-            pdf *= max(dot(vertex.m_geometric_normal, w * sqrt(rcp_dist2)), 0.0);
-            return pdf * rcp_dist2;
+            pdf *= std::abs(dot(next.m_geometric_normal, w * sqrt(rcp_dist2)));
+            return static_cast<float>(pdf * rcp_dist2);
```
dictoon: `pdf` has a `sqrt(rcp_dist2)` term in numerator, maybe we can simplify and simply divide by `sqrt(rcp_dist2)` here (and remove the term from the numerator)?

PR #2570, opened 2019-04-26
Increase precision of floating point values in text files

### LINKED ISSUE #2514: Increase precision of floating point values in text files
According to `std::numeric_limits<float>::max_digits10`, 9 decimals are required to guarantee a float -> string -> float round-trip and 17 for a double. See https://stackoverflow.com/a/22458961/393756 for details.

Increase precision everywhere we store floating point values to disk (at least project files and settings files).

LZaw:
This PR fixes #2514 

Increasing the precision of floats and doubles with respect to the `std::numeric_limits<>::max_digits10` values. Wherever possible and useful, I tried to use non-fixed values (i.e. ask the `numeric_limits`). That way, the precision obeys the precision given by the machine.

Where it wasn't possible or suitable (e.g. in macros), i used the values given by [cppreference.com](https://en.cppreference.com/w/cpp/types/numeric_limits/max_digits10).

In the attachment you can find two project files, each showing the Cornell Box (I left out the mesh files). One has been created using the changes made in this PR (HighPrecision.appleseed), the other one using the current master branch (LowPrecision.appleseed).
A `diff HighPrecision.appleseed LowPrecision.appleseed` shows very well how the precision has increased, e.g.
```
177c177
<                     1.000000000
---
>                     1.000000
```

I couldn't find any problem/issue with the precision in the settings file, so I can't provide any changes there.
[Precision.zip](https://github.com/appleseedhq/appleseed/files/3120997/Precision.zip)

### DIFF
--- a/src/appleseed/foundation/meta/tests/test_qmc.cpp
+++ b/src/appleseed/foundation/meta/tests/test_qmc.cpp
@@ -50,6 +50,7 @@
 #include <cmath>
 #include <cstddef>
 #include <cstdio>
+#include <limits>
 #include <string>
 #include <vector>
 
@@ -549,8 +550,9 @@ TEST_SUITE(Foundation_Math_QMC)
             for (size_t d = 0; d < Dimension; ++d)
             {
                 const double s = radical_inverse<double>(Primes[d], i);
-
-                fprintf(file, "%.17f", s);
+                
+                std::string fmt = "%." + std::to_string(std::numeric_limits<double>::max_digits10) + "f";
+                fprintf(file, fmt.c_str(), s);
 
                 if (d < Dimension - 1)
                     fprintf(file, ", ");
--- a/src/appleseed/foundation/meta/tests/test_qmc.cpp
+++ b/src/appleseed/foundation/meta/tests/test_qmc.cpp
@@ -549,7 +549,7 @@ TEST_SUITE(Foundation_Math_QMC)
             for (size_t d = 0; d < Dimension; ++d)
             {
                 const double s = radical_inverse<double>(Primes[d], i);
-                
+
                 fprintf(file, "%.17f", s);
 
                 if (d < Dimension - 1)
--- a/src/appleseed/foundation/meta/tests/test_qmc.cpp
+++ b/src/appleseed/foundation/meta/tests/test_qmc.cpp
@@ -550,8 +550,7 @@ TEST_SUITE(Foundation_Math_QMC)
             {
                 const double s = radical_inverse<double>(Primes[d], i);
                 
-                std::string fmt = "%.17f";
-                fprintf(file, fmt.c_str(), s);
+                fprintf(file, "%.17f", s);
 
                 if (d < Dimension - 1)
                     fprintf(file, ", ");
--- a/src/appleseed/foundation/meta/tests/test_qmc.cpp
+++ b/src/appleseed/foundation/meta/tests/test_qmc.cpp
@@ -550,7 +550,7 @@ TEST_SUITE(Foundation_Math_QMC)
             {
                 const double s = radical_inverse<double>(Primes[d], i);
 
-                fprintf(file, "%.16f", s);
+                fprintf(file, "%.17f", s);
 
                 if (d < Dimension - 1)
                     fprintf(file, ", ");
--- a/src/appleseed/foundation/meta/tests/test_qmc.cpp
+++ b/src/appleseed/foundation/meta/tests/test_qmc.cpp
@@ -50,7 +50,6 @@
 #include <cmath>
 #include <cstddef>
 #include <cstdio>
-#include <limits>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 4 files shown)


src/appleseed/foundation/meta/tests/test_qmc.cpp
```diff
@@ -549,8 +550,9 @@ TEST_SUITE(Foundation_Math_QMC)
             for (size_t d = 0; d < Dimension; ++d)
             {
                 const double s = radical_inverse<double>(Primes[d], i);
-
-                fprintf(file, "%.16f", s);
+                
+                std::string fmt = "%." + std::to_string(std::numeric_limits<double>::max_digits10) + "f";
```
dictoon: Even if we're in a unit test, the multiple memory allocations (to create the multiple strings, probably at least 3) for each SampleCount * Dimension is a bit too much on the heavy side.

I think we can live with `"%.16f"` here since we're generating values, not round-tripping from/to a file.

src/appleseed/foundation/resources/logo/generatelogofiles.cpp
```diff
@@ -134,8 +135,9 @@ TEST_SUITE(Foundation_Resources_Logos)
                 image->get_pixel(x, y, color);
 
                 color.rgb() = srgb_to_linear_rgb(color.rgb());
-
-                fprintf(file, "%.4ff, %.4ff, %.4ff, %.4ff", color[0], color[1], color[2], color[3]);
+                
+                std::string fmt = "%." + std::to_string(std::numeric_limits<float>::max_digits10) + "ff";
```
dictoon: Same remark here.

src/appleseed/renderer/modeling/project/projectfilewriter.cpp
```diff
@@ -113,8 +114,8 @@ namespace renderer
 namespace
 {
     // Floating-point formatting settings.
-    const char* MatrixFormat     = "%.15f";
-    const char* ColorValueFormat = "%.6f";
+    std::string MatrixFormat     = "%." + std::to_string(std::numeric_limits<double>::max_digits10) + "f";
```
dictoon: I appreciate the effort here but I'd be happy to keep the simpler, lighter-weight solution we had before, just with updated precision. The values of `std::numeric_limits<double>::max_digits10()` and `std::numeric_limits<float>::max_digits10()` are fixed and equal on all platforms (right?).

src/appleseed/foundation/meta/tests/test_qmc.cpp
dictoon: Or `"%.17f"` while we're at it :)

src/appleseed/renderer/modeling/project/projectfilewriter.cpp
LZaw: As long as the IEEE 754 standard is used (which is recommended by the C++ standard, but not required!) it is. If you want to keep it simple one could argue, there are just too few platforms (which might also not be able to run appleseed at all) which differ here.
I thought, that way we are safe, if things change somewhen in the future (not predictable yet), or we got that one person trying to render on his washing mashine. :D

LZaw: Moreover, `std::numeric_limits<T>::max_digits10` isn't a function, but a `static constexpr` and thus should benefit a lot from optimization.
I will run some tests tomorrow to check how strong the performance is affected by using these calls in comparison to a fixed expression.

src/appleseed/foundation/meta/tests/test_qmc.cpp
LZaw: Done.

src/appleseed/foundation/resources/logo/generatelogofiles.cpp
LZaw: Done.

src/appleseed/renderer/modeling/project/projectfilewriter.cpp
```diff
@@ -191,7 +192,7 @@ namespace
             const Vec&          v,
             const size_t        size,
             const size_t        columns,
-            const char*         fmt)
+            const std::string&         fmt)
```
dictoon: Fix vertical alignment.

src/appleseed/renderer/modeling/project/projectfilewriter.cpp
dictoon: Performance won't be affected. It's just that there's little value in using that more complex expression when `%.17f` is shorter, clearer and sufficient :) But the thoroughness is definitely appreciated!

LZaw: Allright, changed to the more readable version.

src/appleseed/renderer/modeling/project/projectfilewriter.cpp
LZaw: Solved.

PR #2571, opened 2019-04-27
Add option to disable support for spectral colors

est77:
For simple scenes, it is already 10% faster.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -130,6 +130,7 @@ option (WITH_PYTHON3_BINDINGS               "Build Python bindings for Python 3"
 option (WITH_DISNEY_MATERIAL                "Build Disney material"                                     OFF)
 option (WITH_EMBREE                         "Include support for Embree intersection backend"           OFF)
 option (WITH_GPU                            "Build GPU support"                                         OFF)
+option (WITH_SPECTRAL_SUPPORT               "Include support for spectral colors"                       ON)
 option (WITH_DOXYGEN                        "Generate API reference with Doxygen"                       ON)
 option (INSTALL_HEADERS                     "Install header files"                                      ON)
 option (INSTALL_TESTS                       "Install unit tests and benchmarks"                         ON)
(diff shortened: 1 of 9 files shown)


src/appleseed/renderer/utility/settingsparsing.cpp
```diff
@@ -56,20 +56,35 @@ Spectrum::Mode get_spectrum_mode(const ParamArray& params)
...
+#ifdef APPLESEED_WITH_SPECTRAL_COLOR
     return
         spectrum_mode == "rgb"
             ? Spectrum::RGB
             : Spectrum::Spectral;
+#else
+    if (spectrum_mode == "spectral")
+    {
+        RENDERER_LOG_WARNING(
+            "color pipeline set the spectral but spectral color support "
```
dictoon: set **to** \\"spectral\\"

CMakeLists.txt
```diff
@@ -129,6 +129,7 @@ option (WITH_PYTHON2_BINDINGS               "Build Python bindings for Python 2"
 option (WITH_PYTHON3_BINDINGS               "Build Python bindings for Python 3"                        OFF)
 option (WITH_DISNEY_MATERIAL                "Build Disney material"                                     OFF)
 option (WITH_EMBREE                         "Include support for Embree intersection backend"           OFF)
+option (WITH_SPECTRAL_COLOR                 "Include support for spectral colors"                       ON)
```
dictoon: I find `APPLESEED_WITH_SPECTRAL_COLOR` somewhat confusing since the pipeline still supports RGB colors when defined. `APPLESEED_WITH_SPECTRAL_SUPPORT` would be clearer and would help avoid the confusion that is certain to happen.

src/appleseed/renderer/modeling/surfaceshader/nprsurfaceshaderhelper.cpp
```diff
@@ -64,9 +64,11 @@ void NPRSurfaceShaderHelper::evaluate(
     const Material* material = shading_point.get_material();
     const ShaderGroup* sg = material->get_render_data().m_shader_group;
 
+#ifdef APPLESEED_WITH_SPECTRAL_COLOR
```
dictoon: A slightly better option is to remove the `#ifdef` and check if `get_mode()` returns something else than `Spectrum::RGB`.

est77: There is not Spectrum::Spectral when spectral colors are not enabled

dictoon: I know, that's why I'm suggesting to do

    if (shading_components.m_beauty.get_mode() != Spectrum::RGB)

PR #2573, opened 2019-04-28
Allow input file of mitsuba2appleseed.py to be anywhere

mororo250:
Now, when converting mitsuba scene to applessed we need to have the scene and mitsuba2appleseed.py in the same directory. This allows the mitsuba scene to be in any other directory.

### DIFF
--- a/sandbox/share/mitsuba2appleseed.py
+++ b/sandbox/share/mitsuba2appleseed.py
@@ -949,8 +949,8 @@ def main():
 
     project = convert(tree)
 
-    asr.ProjectFileWriter().write(project, args.output_file,
-    asr.ProjectFileWriterOptions.OmitHandlingAssetFiles)
+    asr.ProjectFileWriter().write(project, args.output_file, 
+                                  asr.ProjectFileWriterOptions.OmitHandlingAssetFiles)
 
 if __name__ == '__main__':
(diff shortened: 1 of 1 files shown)


sandbox/share/mitsuba2appleseed.py
```diff
@@ -941,9 +942,14 @@ def main():
...
 
+    for child in tree.getroot():
+        filepath = child.find("string[@name='filename']")
+        if filepath is not None:
+            filepath.attrib["value"] = os.path.join(os.path.dirname(args.input_file), filepath.attrib["value"])
+
     project = convert(tree)
 
-    asr.ProjectFileWriter().write(project, args.output_file,
+    asr.ProjectFileWriter().write(project, args.output_file,
```
dictoon: Remove trailing space.

sandbox/share/mitsuba2appleseed.py
```diff
@@ -941,9 +942,14 @@ def main():
     except IOError:
         fatal("Failed to load {0}".format(args.input_file))
 
+    for child in tree.getroot():
```
dictoon: Add a short comment explaining what this code block does.

mororo250:
Thank you for the review.

PR #2576, opened 2019-05-01
Add uniform sampling on analytic rectangles

oktomus:
02 - rectangle


src/appleseed/renderer/kernel/lighting/lighttypes.cpp
```diff
@@ -161,6 +194,20 @@ void EmittingShape::sample_uniform(
         // Set the world space geometric normal.
         light_sample.m_geometric_normal = m_geom.m_triangle.m_geometric_normal;
     }
+    else if (shape_type == RectangleShape)
+    {
+        // Set the barycentric coordinates.
+        light_sample.m_bary = s;
```
est77: Maybe not for this PR, but it would make sense to rename m_bary now that they are not always barycentric coordinates.
Maybe m_param?

src/appleseed/renderer/kernel/intersection/intersector.cpp
```diff
@@ -384,6 +384,65 @@ void Intersector::make_triangle_shading_point(
     shading_point.m_members = 0;
 }
 
+void Intersector::make_procedural_surface_shading_point(
+    ShadingPoint&                       shading_point,
+    const ShadingRay&                   shading_ray,
+    const foundation::Vector2f&         uv,
```
dictoon: Rename to `bary` to match the name in the header file.

src/appleseed/renderer/kernel/intersection/intersector.h
```diff
@@ -116,6 +116,22 @@ class Intersector
         const size_t                        primitive_index,
         const TriangleSupportPlaneType&     triangle_support_plane) const;
 
+    // Manufacture a procedural surface hit "by hand".
```
dictoon: I can't comment there but please update the comment for the `make_triangle_shading_point()` method above (it should at least mention the word "triangle").

src/appleseed/renderer/kernel/intersection/intersector.cpp
dictoon: `foundation::` prefixes are unnecessary in this file.

src/appleseed/renderer/kernel/lighting/lightsample.h
```diff
@@ -97,7 +97,7 @@ class LightSample
 
     // Data for a light-emitting shape sample.
     const EmittingShape*        m_shape;
-    foundation::Vector2f        m_bary;                         // barycentric coordinates of the sample
+    foundation::Vector2f        m_param;                        // parametric coordinates of the sample
```
dictoon: As suggested on Discord, `m_parametric` may be less confusing since we frequently use `m_params` for entity parameters.

src/appleseed/renderer/kernel/lighting/lighttypes.cpp
```diff
@@ -161,6 +194,20 @@ void EmittingShape::sample_uniform(
...
+    {
+        // Set the parametric coordinates.
+        light_sample.m_param = s;
+
+        light_sample.m_point =
+            m_geom.m_rectangle.m_origin +
+            static_cast<double>(s[0]) * m_geom.m_rectangle.m_x +
+            static_cast<double>(s[1]) * m_geom.m_rectangle.m_y;
+
+        // Set the world space shading and geometric normal.
```
dictoon: *normals*

src/appleseed/renderer/modeling/object/rectangleobject.cpp
```diff
@@ -143,6 +143,21 @@ double RectangleObject::get_uncached_height() const
     return m_params.get_optional<double>("height", 1.0);
 }
 
+void RectangleObject::get_origin_and_axes(
```
dictoon: The implementation of this method doesn't appear to match the description from the header file:

    The rectangle is centered at the origin and lies in the X-Z plane.

If that was the case, `origin` would always be (0, 0, 0), no?

src/appleseed/renderer/kernel/lighting/lighttypes.cpp
```diff
@@ -108,6 +108,39 @@ EmittingShape EmittingShape::create_triangle_shape(
...
+    shape.m_geom.m_rectangle.m_width = norm(x);
+    shape.m_geom.m_rectangle.m_height = norm(y);
+    shape.m_geom.m_rectangle.m_geometric_normal = n;
+    shape.m_geom.m_rectangle.m_plane_dist = -dot(p, n);
+
+    shape.m_area = static_cast<float>(
+        shape.m_geom.m_rectangle.m_width * shape.m_geom.m_rectangle.m_height);
+
+    if (shape.m_area != 0.0f)
+        shape.m_rcp_area = 1.0f / shape.m_area;
```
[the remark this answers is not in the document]
oktomus: Isn't it better to propagate NaNs so that we can assert and detect bugs ?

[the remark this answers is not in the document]
dictoon: Sure, but where are you setting `shape.m_rcp_area` to NaN?

PR #2577, opened 2019-05-01
Move oiio_make_texture to libappleseed

est77:
Avoids having OpenImageIO symbols in appleseed.python.


dictoon:
We need to appleseed against OCIO, and removing linking the Python bindings to OCIO.

PR #2578, opened 2019-05-02
Pad array allocations to 16 bytes.

est77:
Required by Embree, useful for SIMD code.

### DIFF
--- a/src/appleseed/foundation/array/arrayallocator.h
+++ b/src/appleseed/foundation/array/arrayallocator.h
@@ -1,0 +1,163 @@
+
+//
+// This source file is part of appleseed.
+// Visit https://appleseedhq.net/ for additional information and resources.
+//
+// This software is released under the MIT license.
+//
+// Copyright (c) 2019 Esteban Tovagliari, The appleseedhq Organization
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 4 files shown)


src/appleseed/foundation/array/arrayallocator.h
```diff
@@ -0,0 +1,168 @@
...
+    template <typename U>
+    struct rebind
+    {
+        typedef ArrayAllocator<U> other;
+    };
+
+    ArrayAllocator() = default;
+
+    template <typename U>
+    ArrayAllocator(const ArrayAllocator<U>&)
```
dictoon: Why no `= default` here?

src/appleseed/foundation/array/arrayallocator.h
```diff
@@ -0,0 +1,168 @@
...
+        new(p) value_type(x);
+    }
+
+    void destroy(pointer p)
+    {
+        p->~value_type();
+    }
+
+  private:
+    size_type round_up_size(const size_type size)
```
dictoon: Can `foundation::next_multiple()` do the job?

Otherwise, make static.

src/appleseed/foundation/array/arrayallocator.h
```diff
@@ -0,0 +1,168 @@
...
+    template <typename U>
+    struct rebind
+    {
+        typedef ArrayAllocator<U> other;
+    };
+
+    ArrayAllocator() = default;
+
+    template <typename U>
+    ArrayAllocator(const ArrayAllocator<U>&)
```
dictoon: Why no `= default` here?

src/appleseed/foundation/array/arrayallocator.h
est77: Does not work for templated copy constructors.

src/appleseed/foundation/array/arrayallocator.h
est77: It doesn't work

src/appleseed/foundation/array/arrayallocator.h
dictoon: Ok.

PR #2579, opened 2019-05-02
Fixes

dictoon:


usakhelo:
Hi @dictoon ,
There's one more `const GGXMDF mdf;` left in `plasticbrdf.cpp`, line 156.

dictoon:
> There's one more const GGXMDF mdf; left in plasticbrdf.cpp, line 156.

Good catch, thanks! I'll do another pass to make sure I'm not missing any other.

PR #2581, opened 2019-05-04
Implement new EC paper, part I

est77:
- Remove precomputed EC tables.
- Implement new EC paper for glossy and metal BRDFs.
- Glass BSDF simplifications.
- Remove wrong EC code from Glass BSDF.


src/appleseed/renderer/modeling/bsdf/glassbsdf.cpp
```diff
@@ -1130,63 +824,72 @@ namespace
...
+                        const float G = GGXMDF::G(wi, wo, m, alpha, alpha, 0.0f);
+                        R += (1.0f - F) * G * rcp_G1;
+                    }
+                }
             }
 
             return min(R / static_cast<float>(SampleCount), 1.0f);
         }
+
+        static float safe_rcp(const float x)
```
dictoon: This could be moved to `foundation/math/scalar.h` together with the other `safe_*()` methods. The epsilon value should be passed in argument and it should have `default_eps<T>()` as a default value.

est77: I'll include it in the next PR

PR #2583, opened 2019-05-05
Add uniform sampling on analytic spheres and disks

oktomus:
04 - sphere
05 - disk
03 - rectangle and spheres


src/appleseed/renderer/kernel/intersection/intersector.cpp
```diff
@@ -387,9 +387,9 @@ void Intersector::make_triangle_shading_point(
 void Intersector::make_procedural_surface_shading_point(
     ShadingPoint&                       shading_point,
     const ShadingRay&                   shading_ray,
-    const foundation::Vector2f&         uv,
+    const Vector2f&                     bary,
```
est77: uv was correct here.

src/appleseed/renderer/kernel/lighting/lightsamplerbase.cpp
```diff
@@ -343,14 +343,14 @@ void LightSamplerBase::collect_emitting_shapes(
             const RectangleObject& rectangle = static_cast<const RectangleObject&>(object);
 
             // Retrieve object instance space geometry of the rectangle.
-            Vector3d o, x, y, n;
-            rectangle.get_origin_and_axes(o, x, y, n);
+            Vector3d c, x, y, n;
+            rectangle.get_corner_and_axes(c, x, y, n);
```
est77: Personally, I prefer origin. I think it is more accurate

src/appleseed/renderer/kernel/lighting/lighttypes.cpp
```diff
@@ -274,6 +315,30 @@ void EmittingShape::make_shading_point(
             m_geom.m_rectangle.m_x,
             cross(m_geom.m_rectangle.m_x, m_geom.m_rectangle.m_geometric_normal));
     }
+    else if (shape_type == SphereShape)
+    {
+        const double theta = static_cast<double>(bary[0]);
```
est77: In this method, we can rename bary to param_coords

PR #2587, opened 2019-05-07
Fix sampling bug, improve code

dictoon:


src/appleseed/foundation/math/cdf.h
```diff
@@ -53,14 +50,16 @@ namespace foundation
...
-  : public NonCopyable
 {
   public:
     typedef std::pair<Item, Weight> ItemWeightPair;
 
     // Constructor.
     CDF();
 
+    // Return the number of items in the CDF.
+    size_t size() const;
```
oktomus: `std::size_t` and include `<cstddef>` i suppose ?

src/appleseed/foundation/math/cdf.h
```diff
@@ -145,6 +144,12 @@ inline CDF<Item, Weight>::CDF()
 {
 }
 
+template <typename Item, typename Weight>
+inline size_t CDF<Item, Weight>::size() const
```
oktomus: `std::size_t`

src/appleseed/foundation/math/cdf.h
```diff
@@ -185,7 +190,7 @@ inline void CDF<Item, Weight>::insert(const Item& item, const Weight weight)
 }
 
 template <typename Item, typename Weight>
-inline const std::pair<Item, Weight>& CDF<Item, Weight>::operator[](const size_t i) const
+inline const typename CDF<Item, Weight>::ItemWeightPair& CDF<Item, Weight>::operator[](const size_t i) const
```
oktomus: `std::size_t`

dictoon:
> There is only one thing, are we suppose to prefix size_t with std:: in header files ? If so, you have some changes to do in cdf.h.

We will. Until now, we don't prefix `size_t` anywhere but it's not very logical. At some point we'll do a big search & replace and make sure it is prefixed everywhere. Until then, and out of consistency, I prefer not to add the prefixes.

Thanks for the review!

PR #2593, opened 2019-05-12
Include build options header first in API headers

### LINKED ISSUE #2592: Include renderer/api/buildoptions.h in all API headers
It needs to be the first `#include` clause in _all_ appleseed.renderer API headers.

est77:
Fixes #2592


est77:
Closing. We don't have a nice way to avoid the redefinitions warnings.

PR #2602, opened 2019-05-13
Expose analytical objects in API

usakhelo:
Hi,
I had to add these changes in order to access analytical objects from max plugin.
Thanks for review.


usakhelo:
Do we ignore failed macos build? Should I merge this?

dictoon:
> Do we ignore failed macos build? Should I merge this?

We don't normally ignore any build failure.

PR #2605, opened 2019-05-14
Fix Cryptomatte accumulator assertion errors.

jdent02:


src/appleseed/renderer/modeling/aov/cryptomatteaov.cpp
```diff
@@ -528,7 +528,9 @@ namespace
                         }
 
                         const size_t num_channels = (m_num_layers * 2) + 3;
-                        for (size_t i = ranked_vector.size(); i < num_channels; ++i)
+
+                        // Fill remaining channels with black.
+                        for (size_t i = ranked_vector.size() * 2 + 3; i < num_channels; ++i)
```
oktomus: Why  `* 2 + 3` ? Can't it be a parameter or something ?

jdent02: This seemed like the most direct way to do it. I suppose it could be a parameter

oktomus: Just adding a comment that explains why it's `* 2 + 3` is enough i think

jdent02: I'll see what I can come up with.  When you take this bit in context with the code above it (that doesn't show up) it seemed fairly obvious what it was doing.

dictoon: I must say I'm not entirely clear what the `* 2` and `+ 3` mean here :) I can see that the same calculations are involved in computing `num_channels`. I guess `* 2` is because we have rank + coverage channels for each item in `ranked_vector` and `+ 3` is for the RGB preview channels?

Unrelated to this commit, but just reading this code it appears that there's an obvious optimization: remove this code (lines 512-514)

```
// Remove background contribution.
if (ranked_vector.size() > 1 && m3hash_preview == 0)
    ranked_vector.erase(ranked_vector.begin());
```

and instead skip the first item when iterating over `ranked_vector`.

jdent02: Okay you guys win, I'll add some more comments explaining what it's doing.

BTW, Franz you are right about what the numbers are calling.  Prior to this fix the accumulator was only registering one channel per ranked vector and none of the RGB previews so the pixel_values vector ended up being four floats too large (19 vs 15).  When rendering through .studio it wasn't causing any obvious issues but for whatever reason it caused the render to randomly quit under Python.

dictoon:
I'm confused by this PR, its title doesn't match the sole commit that is in there.

jdent02:
Yeah looks like I goofed. I’ll get that last commit removed

jdent02:
Closing due to mangling my Github branches

PR #2606, opened 2019-05-15
Make use of shape proabilities consistent

oktomus:
`shape_prob` argument of `sample_uniform` contains exactly the same value as `m_shape_prob`. This PR makes sure we don't have two variables for the same thing.

In the future, we would like to completly remove the shape prob (`m_shape_prob`)  from the emitting shape.

### DIFF
--- a/src/appleseed/renderer/kernel/lighting/lighttypes.h
+++ b/src/appleseed/renderer/kernel/lighting/lighttypes.h
@@ -43,6 +43,8 @@
 // Forward declarations.
 namespace renderer  { class AssemblyInstance; }
 namespace renderer  { class BackwardLightSampler; }
+namespace renderer  { class ForwardLightSampler; }
+namespace renderer  { class LightSamplerBase; }
 namespace renderer  { class Intersector; }
 namespace renderer  { class Light; }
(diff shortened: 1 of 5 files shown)


src/appleseed/renderer/kernel/lighting/lighttypes.h
```diff
@@ -43,6 +43,8 @@
 // Forward declarations.
 namespace renderer  { class AssemblyInstance; }
 namespace renderer  { class BackwardLightSampler; }
+namespace renderer  { class ForwardLightSampler; }
+namespace renderer  { class LightSamplerBase; }
```
dictoon: Order alphabetically.

oktomus: Isn't it ?

dictoon: Nope :)

oktomus: image
Wait, what ? XD

oktomus: Ah! I wasn't seeing what was under your comment...

PR #2607, opened 2019-05-15
Remove seexpr.h include from API

Mango-3:
- Removes the include of seexpr.h from the API

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@
 # This software is released under the MIT license.
 #
 # Copyright (c) 2010-2013 Francois Beaune, Jupiter Jazz Limited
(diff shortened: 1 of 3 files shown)


Mango-3:
This fixes a problem causing build errors of the appleseed-Maya plugin when the new build options header `renderer/api/buildoptions.h` is included together with `renderer/api/utility.h`

Mango-3:
Ok, I will have a look later. Probably splitting out the SeExpression class into its own header file so that the rest of seexpr.h will not get imported.

dictoon:
> Probably splitting out the SeExpression class into its own header file so that the rest of seexpr.h will not get imported.

To be clear, the `SeExpression` class is from SeExpr, not appleseed.

PR #2608, opened 2019-05-16
Remove assertion errors in cryptomatteaov

jdent02:
Optimize ranked vector translation

### DIFF
--- a/src/appleseed/renderer/modeling/aov/cryptomatteaov.cpp
+++ b/src/appleseed/renderer/modeling/aov/cryptomatteaov.cpp
@@ -511,13 +511,8 @@ namespace
 
                         // Remove background contribution.
                         size_t ranked_vector_start = 0;
-                        size_t ranked_vector_size = ranked_vector.size();
-
                         if (ranked_vector.size() > 1 && m3hash_preview == 0)
-                        {
                             ranked_vector_start = 1;
-                            ranked_vector_size -= 1;
-                        }
 
                         // Ranked channels.
                         for (size_t i = ranked_vector_start, e = min(ranked_vector.size(), m_num_layers); i < e; ++i)
@@ -538,9 +533,9 @@ namespace
                         // 3 channels for the preview image and subtracting that from the
                         // total number of AOV channels.
                         const size_t num_channels = (m_num_layers * 2) + 3;
-                        const size_t filled_pixels = ranked_vector_size * 2 + 3;
+                        const size_t filled_channels = (ranked_vector.size() - ranked_vector_start) * 2 + 3;
 
-                        for (size_t i = filled_pixels; i < num_channels; ++i)
+                        for (size_t i = filled_channels; i < num_channels; ++i)
                             pixel_values.push_back(0.0f);
 
                         m_aov_image.set_pixel(rx, ry, pixel_values.data(), pixel_values.size());
--- a/src/appleseed/renderer/modeling/aov/cryptomatteaov.cpp
+++ b/src/appleseed/renderer/modeling/aov/cryptomatteaov.cpp
@@ -510,11 +510,17 @@ namespace
                         pixel_values.push_back(b);
 
                         // Remove background contribution.
+                        size_t ranked_vector_start = 0;
+                        size_t ranked_vector_size = ranked_vector.size();
+
                         if (ranked_vector.size() > 1 && m3hash_preview == 0)
-                            ranked_vector.erase(ranked_vector.begin());
+                        {
+                            ranked_vector_start = 1;
+                            ranked_vector_size -= 1;
+                        }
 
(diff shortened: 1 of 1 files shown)


src/appleseed/renderer/modeling/aov/cryptomatteaov.cpp
```diff
@@ -510,11 +510,17 @@ namespace
                         pixel_values.push_back(b);
 
                         // Remove background contribution.
+                        size_t ranked_vector_start = 0;
+                        size_t ranked_vector_size = ranked_vector.size();
+
                         if (ranked_vector.size() > 1 && m3hash_preview == 0)
```
dictoon: You could use `ranked_vector_size` here.

src/appleseed/renderer/modeling/aov/cryptomatteaov.cpp
```diff
@@ -510,11 +510,17 @@ namespace
...
-                            ranked_vector.erase(ranked_vector.begin());
+                        {
+                            ranked_vector_start = 1;
+                            ranked_vector_size -= 1;
+                        }
 
                         // Ranked channels.
-                        for (size_t i = 0, e = min(ranked_vector.size(), m_num_layers); i < e; ++i)
+                        for (size_t i = ranked_vector_start, e = min(ranked_vector.size(), m_num_layers); i < e; ++i)
```
dictoon: This is still using `ranked_vector.size()` here, is it intended?

src/appleseed/renderer/modeling/aov/cryptomatteaov.cpp
```diff
@@ -510,11 +510,17 @@ namespace
                         pixel_values.push_back(b);
 
                         // Remove background contribution.
+                        size_t ranked_vector_start = 0;
+                        size_t ranked_vector_size = ranked_vector.size();
+
```
dictoon: I'd suggest to remove this blank line so that the comment above applies to the whole block.

src/appleseed/renderer/modeling/aov/cryptomatteaov.cpp
```diff
@@ -527,8 +533,14 @@ namespace
                             pixel_values.push_back(coverage);
                         }
 
+                        // Set the remaining channels of the pixel to black.
+                        // This is determined by taking 2 channels per ranked vector plus
+                        // 3 channels for the preview image and subtracting that from the
+                        // total number of AOV channels.
                         const size_t num_channels = (m_num_layers * 2) + 3;
-                        for (size_t i = ranked_vector.size(); i < num_channels; ++i)
+                        const size_t filled_pixels = ranked_vector_size * 2 + 3;
```
dictoon: I guess what would be clearer and simpler is to get rid of `ranked_vector_size` and use ` ranked_vector.size() - ranked_vector_start` here.

src/appleseed/renderer/modeling/aov/cryptomatteaov.cpp
```diff
@@ -527,8 +533,14 @@ namespace
                             pixel_values.push_back(coverage);
                         }
 
+                        // Set the remaining channels of the pixel to black.
```
dictoon: 👍

src/appleseed/renderer/modeling/aov/cryptomatteaov.cpp
dictoon: I think `filled_pixels` isn't quite right, this isn't about pixels but about channels, right?

jdent02: Good point.  Got my terms crossed

src/appleseed/renderer/modeling/aov/cryptomatteaov.cpp
jdent02: Yes

src/appleseed/renderer/modeling/aov/cryptomatteaov.cpp
jdent02: Good point

src/appleseed/renderer/modeling/aov/cryptomatteaov.cpp
jdent02: More thorough: we're skipping over the first ranked vector but we still need to iterate over rest of them.  If I used 'ranked_vecotr_size' in this context it would stop at the last one and not iterate over it.

src/appleseed/renderer/modeling/aov/cryptomatteaov.cpp
jdent02: I see what you're saying with regards to the 'ranked_vector_size' comment.  I'll modify

PR #2613, opened 2019-05-18
appleseed Python updates

jdent02:


src/appleseed.python/bindaov.cpp
```diff
@@ -91,6 +94,21 @@ namespace
 
         return channels;
     }
+
+    Image& get_crypto_image(AOV* aov)
```
dictoon: It should be possible to make this a const pointer.

src/appleseed.python/bindaov.cpp
```diff
@@ -91,6 +94,21 @@ namespace
...
+        const string model = aov->get_model();
+
+        if (model == "cryptomatte_object_aov" || model == "cryptomatte_material_aov")
+            return static_cast<CryptomatteAOV*>(aov)->get_crypto_image();
+
+        PyErr_SetString(PyExc_RuntimeError, "AOV is not a Cryptomatte AOV");
+        bpy::throw_error_already_set();
+
+        // If AOV is not a Cryptomatte type, return a blank pointer for the image.
+        Image* dummy = nullptr;
```
dictoon: This is going to crash since you're dereferencing a null pointer.

If `get_crypto_image()` must be able to fail, it could return a pointer instead of a reference.

src/appleseed.python/bindaov.cpp
```diff
@@ -103,7 +121,8 @@ void bind_aov()
         .def("get_channel_count", &AOV::get_channel_count)
         .def("get_channel_names", &get_channel_names)
         .def("has_color_data", &AOV::has_color_data)
-        .def("get_image", &AOV::get_image, bpy::return_value_policy<bpy::reference_existing_object>());
+        .def("get_image", &AOV::get_image, bpy::return_value_policy<bpy::reference_existing_object>())
+        .def("get_crypto_image", &get_crypto_image, bpy::return_value_policy<bpy::reference_existing_object>());
```
dictoon: Can you rename "crypto" to "cryptomatte" everywhere please? It disturbs me 😂

src/appleseed/renderer/modeling/aov/cryptomatteaov.h
```diff
@@ -26,8 +26,7 @@
 // THE SOFTWARE.
 //
 
-#ifndef APPLESEED_RENDERER_MODELING_AOV_CRYPTOMATTEAOV_H
-#define APPLESEED_RENDERER_MODELING_AOV_CRYPTOMATTEAOV_H
+#pragma once
```
dictoon: Good catch.

src/appleseed.python/bindaov.cpp
jdent02: The static cast doesn’t work if it’s a const pointer

src/appleseed.python/bindaov.cpp
jdent02: I’ll have to see how to do that with Boost Python. Right now it fails to compile info try to return a pointer

src/appleseed.python/bindaov.cpp
jdent02: Oh okay

src/appleseed.python/bindaov.cpp
jdent02: Never mind, guess it works fine.

src/appleseed.python/bindaov.cpp
jdent02: Never mind, I wasn't doing the cast correctly

jdent02:
Fixes in

PR #2615, opened 2019-05-18
Microfacet-related code cleanups

est77:
- Do not compute the glass albedo table at startup.
- Remove the gamma parameter from all MDF classes that don't use it.

### DIFF
--- a/src/appleseed/foundation/math/scalar.h
+++ b/src/appleseed/foundation/math/scalar.h
@@ -178,6 +178,10 @@ T cube(const T x);
 template <typename T>
 T rcp(const T x);
 
+// Return 1 / x or eps if the absolute value of x is less than eps.
+template <typename T>
+T safe_rcp(const T x, const T eps);
+
 // Return the square root of x or 0 if x is negative.
 template <typename T>
 T safe_sqrt(const T x);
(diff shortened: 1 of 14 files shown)


src/appleseed/foundation/math/scalar.h
```diff
@@ -178,6 +178,10 @@ T cube(const T x);
 template <typename T>
 T rcp(const T x);
 
+// Return 1 / x or eps if the absolute value of x is less than eps.
+template <typename T>
+T safe_rcp(const T x, const T eps);
```
dictoon:
```
T safe_rcp(const T x, const T eps = default_eps<T>());
```

src/appleseed/foundation/meta/tests/test_microfacet.cpp
```diff
@@ -114,10 +169,10 @@ TEST_SUITE(Foundation_Math_Microfacet)
 
     template <typename MDF>
     void weak_white_furnace_test(
+        const MDF&                  mdf,
```
dictoon: What's the point of passing `mdf` here since everything is static anyway?

src/appleseed/foundation/math/scalar.h
est77: I cannot do that because default_eps is not defined at this point.
I am also not sure if default_eps is the correct default value.

src/appleseed/foundation/meta/tests/test_microfacet.cpp
est77: The StdMDFAdaptor class defined at the begining of the file.

dictoon: Ok, makes sense, thanks.

PR #2621, opened 2019-05-19
Fix ray parallelogram intersection tests

### LINKED ISSUE #2612: New rectangle_object size/position errors
Got a good one for you.

Recently I switched the area lamps in Blender over to the new analytic shapes.  Spheres and discs work as expected, but there are some issues with the rectangles.

Here is an array of disc lights with a 0.5 radius:
disc_0 5

Here's the same with 0.25:
disc_0 25

As you can see, the size changes correctly and the objects remain in place.

Here's rectangles at 1x1:
1x1

Here's the same thing at 1x2:
1x2

As you can see, the shapes themselves have shifted from the proper locations and shrunk in size.  The matrices of each object were not changed, only the width parameter on the object entity.

Here's 1x3:
1x3

It get's even worse there.

Here's the scene files:
[as_test_file.zip](https://github.com/appleseedhq/appleseed/files/3192948/as_test_file.zip)

oktomus:
fix #2612


usakhelo:
Hi @oktomus ,
This PR fixes rectangle size issue in max plugin:
image

PR #2629, opened 2019-05-28
Light paths recording annotations and tweaks

termhn:
* Annotates light path creation code
* Uses reinhard tonemapping (`x/(1+x)`) for displaying the radiance as color when drawing. This gives much more helpful color results than previously when displaying light paths, since it gives a sense of intensity as well as just hue:

old: 
image 
image

new: 
image 
image

* Takes into account secondary emitter radiance in the middle of a light path. Previously if a path intersected with an emitter that was also a reflector that was not the origin of the path, the emitted radiance of said emitter would not be added to the path's total radiance.

### DIFF
--- a/src/appleseed.studio/mainwindow/rendering/lightpathswidget.cpp
+++ b/src/appleseed.studio/mainwindow/rendering/lightpathswidget.cpp
@@ -385,7 +385,7 @@ void LightPathsWidget::render_light_path(const size_t light_path_index) const
         light_path_recorder.get_light_path_vertex(i, v1);
 
         auto radiance = Color3f::from_array(v1.m_radiance);
-        radiance /= sum_value(radiance);
+        radiance /= radiance + Color3f(1.0);
         radiance = linear_rgb_to_srgb(radiance);
 
         glColor(radiance);
--- a/src/appleseed/renderer/kernel/lighting/lightpathstream.cpp
+++ b/src/appleseed/renderer/kernel/lighting/lightpathstream.cpp
@@ -267,9 +267,19 @@ void LightPathStream::create_path_from_hit_emitter(const size_t emitter_event_in
             reflector_vertex.m_radiance = current_radiance;
             m_vertices.push_back(reflector_vertex);
 
+            // If the hit is an emitter, add the radiance to current
+            if (event.m_type == EventType::HitEmitter)
+            {
+                const auto& event = m_events[event_index];
+                current_radiance += m_hit_emitter_data[event.m_data_index].m_emitted_radiance;
+            }
+
             // Update current radiance.
             const auto& throughput = event_data.m_path_throughput;
+            // Multiply by previous throughput to attenuate by the next hit
             current_radiance *= prev_throughput;
+            // Divide by throughput before previous so that we isolate only throughput from the light source to current vertex,
+            // since throughput is cumulative in reverse
             current_radiance /= throughput;
             prev_throughput = throughput;
         }
@@ -333,9 +343,19 @@ void LightPathStream::create_path_from_sampled_emitter(const size_t emitter_even
             reflector_vertex.m_radiance = current_radiance;
             m_vertices.push_back(reflector_vertex);
 
+            // If the hit is an emitter, add the radiance to current
+            if (event.m_type == EventType::HitEmitter)
+            {
+                const auto& event = m_events[event_index];
+                current_radiance += m_hit_emitter_data[event.m_data_index].m_emitted_radiance;
+            }
+
             // Update current radiance.
             const auto& throughput = event_data.m_path_throughput;
+            // Multiply by previous throughput to attenuate by the next hit
             current_radiance *= prev_throughput;
+            // Divide by throughput before previous so that we isolate only throughput from the light source to current vertex,
+            // since throughput is cumulative in reverse
             current_radiance /= throughput;
             prev_throughput = throughput;
         }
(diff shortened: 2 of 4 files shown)


LZaw:
Looks good,
One (you?) could possibly refactor the three functions
```
void LightPathStream::create_path_from_hit_emitter(const size_t emitter_event_index)
void LightPathStream::create_path_from_sampled_emitter(const size_t emitter_event_index)
void LightPathStream::create_path_from_sampled_environment(const size_t env_event_index)
```
since they got a lot of overlapping code.

src/appleseed/renderer/kernel/lighting/pathvertex.h
```diff
@@ -70,7 +70,7 @@ class PathVertex
     // Path properties.
     size_t                      m_path_length;
     int                         m_scattering_modes;
-    Spectrum                    m_throughput;
+    Spectrum                    m_throughput;   // Cumulative throughput (percent of incoming light which is reflected as outgoing light) of this path, up to but excluding this vertex
```
LZaw: Maybe indent the comment such that it's right of all variable names (i.e. `m_scattering_modes`).
Also the codeguidlines state
`Comments on the same line as code start with a lower case letter and don't end with a period.`

src/appleseed.studio/mainwindow/rendering/lightpathswidget.cpp
```diff
@@ -385,7 +385,7 @@ void LightPathsWidget::render_light_path(const size_t light_path_index) const
         light_path_recorder.get_light_path_vertex(i, v1);
 
         auto radiance = Color3f::from_array(v1.m_radiance);
-        radiance /= sum_value(radiance);
+        radiance /= radiance + Color3f(1.0);
```
LZaw: Possibly clamp this value, such that it doesn't get too low (and thus won't be seeable anymore).

src/appleseed/renderer/kernel/lighting/lightpathstream.cpp
```diff
@@ -267,9 +267,19 @@ void LightPathStream::create_path_from_hit_emitter(const size_t emitter_event_in
             reflector_vertex.m_radiance = current_radiance;
             m_vertices.push_back(reflector_vertex);
 
+            // If the hit is an emitter, add the radiance to current
+            if (event.m_type == EventType::HitEmitter)
+            {
+                const auto& event = m_events[event_index];
```
LZaw: Why do you need this declaration? It's already defined before and doesn't change.

src/appleseed/renderer/kernel/lighting/lightpathstream.cpp
```diff
@@ -267,9 +267,19 @@ void LightPathStream::create_path_from_hit_emitter(const size_t emitter_event_in
             reflector_vertex.m_radiance = current_radiance;
             m_vertices.push_back(reflector_vertex);
 
+            // If the hit is an emitter, add the radiance to current
+            if (event.m_type == EventType::HitEmitter)
```
LZaw: Wouldn't it be better to have this whole addition part after calculating the reflectance? I.e. add the emitting light last.

src/appleseed.studio/mainwindow/rendering/lightpathswidget.cpp
dictoon: Good observation, however I fear that any non-zero value may misrepresent the fact that the path effectively carries zero radiance. I feel like we're better off with black rays which anyway will be visible on non-black background.

src/appleseed/renderer/kernel/lighting/lightpathstream.cpp
```diff
@@ -267,9 +267,19 @@ void LightPathStream::create_path_from_hit_emitter(const size_t emitter_event_in
             reflector_vertex.m_radiance = current_radiance;
             m_vertices.push_back(reflector_vertex);
 
+            // If the hit is an emitter, add the radiance to current
```
dictoon: Nice catch, that was a serious bug...

src/appleseed/renderer/kernel/lighting/lightpathstream.cpp
dictoon: Yep it's redundant, below too.

src/appleseed/renderer/kernel/lighting/lightpathstream.cpp
dictoon: Not sure what you mean here @LZaw?

LZaw: Sorry, the line of code doesn't fit the comment very well. I should have positioned it further down, my bad.
In the if statement, I marked (L271-L275) the emitter addition happens. But a bit further down (afterwards), the reflectance of the current surface is being considered (L277-L284). This means, in the current version, the emitting light is being added and then reduced by the reflection value of its own surface.
I suggest to first reduce the incoming light by the reflectance of the surface and add the emitted light then. (Simply speaking, an emitter doesn't get affected by its own reflectance value.)

termhn: Yea myself and @LZaw were discussing this in DMs and I'm pretty sure having it at the end of the function (as it is in the commit I just pushed) is correct.

termhn:
Ready for review again :)

src/appleseed.studio/mainwindow/rendering/lightpathswidget.cpp
oktomus: Should be `Color3f(1.0f)`

src/appleseed/renderer/kernel/lighting/lightpathstream.cpp
```diff
@@ -269,9 +269,18 @@ void LightPathStream::create_path_from_hit_emitter(const size_t emitter_event_in
 
             // Update current radiance.
             const auto& throughput = event_data.m_path_throughput;
+            // Multiply by previous throughput to attenuate by the next hit
```
oktomus: End your comments with a dot.

PR #2632, opened 2019-05-30
Add spectral and gpu build info in studio about dialog

Mango-3:
- Adds build options information for spectral and GPU support enabled/disabled to the about window in appleseed.studio 
- Renamed build option info `Disney material` to `Disney SeExpr` to avoid confusing it with the asDisney shader.

### DIFF
--- a/src/appleseed.studio/help/about/aboutwindow.cpp
+++ b/src/appleseed.studio/help/about/aboutwindow.cpp
@@ -127,11 +127,27 @@ void AboutWindow::set_library_features()
         false;
 #endif
 
+    const bool WithSpectralSupport =
+#ifdef APPLESEED_WITH_SPECTRAL_SUPPORT
+        true;
+#else
+        false;
+#endif
+
+    const bool WithGPUSupport =
+#ifdef APPLESEED_WITH_GPU
+        true;
+#else
+        false;
+#endif
+
     QString details;
     details += "This build of appleseed has the following features:\n\n";
     details += QString("  Instruction sets: %1\n").arg(Appleseed::get_lib_cpu_features());
-    details += QString("  Disney material: %1\n").arg(to_enabled_disabled(WithDisneyMaterial));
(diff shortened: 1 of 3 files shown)


src/appleseed.studio/help/about/aboutwindow.cpp
```diff
@@ -127,11 +127,27 @@ void AboutWindow::set_library_features()
...
+        true;
+#else
+        false;
+#endif
+
     QString details;
     details += "This build of appleseed has the following features:\n\n";
     details += QString("  Instruction sets: %1\n").arg(Appleseed::get_lib_cpu_features());
-    details += QString("  Disney material: %1\n").arg(to_enabled_disabled(WithDisneyMaterial));
+    details += QString("  Disney SeExpr: %1\n").arg(to_enabled_disabled(WithDisneyMaterial));
```
dictoon: "Disney SeExpr" isn't very clear either. Maybe "Disney built-in material"?

Mango-3: It should not lead to confusion with our Disney shaders. So I would prefer the name to not contain `material`. 
As I understood it, enabling this is only needed when one wants to use SeExpr. It does not affect the building of the asDisneyMaterial OSL shader or legacy appleseedDisney one.

dictoon: Then maybe "Disney material with SeExpr support"? My problem right now is that "Disney" alone doesn't mean much, and that "Disney SeExpr" is not very clear.

Mango-3: This is indeed better. I changed it.

dictoon:
Nice!

Just realized the same changes need to be applied to `appleseed.shared/application/commandlinehandlerbase.cpp` for appleseed.cli.

src/appleseed.studio/help/about/aboutwindow.cpp
Mango-3: Added the changes also for the cli version info.

dictoon:
Thanks!

PR #2633, opened 2019-05-31
Small tweaks

dictoon:


src/appleseed.studio/mainwindow/rendering/lightpathswidget.cpp
```diff
@@ -455,7 +455,7 @@ void LightPathsWidget::load_light_paths_data()
                 light_path_recorder.get_light_path_vertex(vertex_idx, curr);
 
                 auto piece_radiance = Color3f::from_array(curr.m_radiance);
-                piece_radiance /= sum_value(piece_radiance);
+                piece_radiance /= piece_radiance + Color3f(1.0);    // Reinhard tone mapping
```
termhn: `Color3f(1.0f)`?

dictoon: Yes, good catch.

PR #2639, opened 2019-06-05
Use noexcept for auto_release_ptr

est77:
Fix some incorrect uses of throw()

### DIFF
--- a/src/appleseed/foundation/utility/autoreleaseptr.h
+++ b/src/appleseed/foundation/utility/autoreleaseptr.h
@@ -29,6 +29,9 @@
 
 #pragma once
 
+// appleseed.foundation headers.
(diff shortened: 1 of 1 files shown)


dictoon:
Can we use `noexcept` directly now that we're using VS 2015 and later? What about gcc?

est77:
> Can we use `noexcept` directly now that we're using VS 2015 and later? What about gcc?

I think it is supported on gcc 4.8. I'll check that. 

Probably we should open a new PR, checking which of the compiler feature macros we want to drop and replacing them by the standard keyword everywhere.

dictoon:
> Probably we should open a new PR, checking which of the compiler feature macros we want to drop and replacing them by the standard keyword everywhere.

Yes. Feel free to merge this one in the meantime.

PR #2642, opened 2019-06-11
Misc changes

est77:
- Address basic CUDA classes PR comments.
- Bring back energy compensation tables.
- Small changes to CopyOnWrite class.
- Arrays code cleanup and API improvements.


src/appleseed/renderer/modeling/bsdf/energycompensation.cpp
```diff
@@ -129,6 +129,15 @@ AlbedoTable2D::AlbedoTable2D()
     m_avg_table = m_albedo_table + TableSize * TableSize;
 }
 
+AlbedoTable2D::AlbedoTable2D(const float* table)
```
dictoon: Why are we passing a `const float*` if we need to `const_cast<>` it away immediately after?

src/appleseed/foundation/array/array.cpp
```diff
@@ -44,7 +44,109 @@ namespace foundation
 // Array class implementation.
 //
 
-Array::Concept::~Concept()
+template <typename T>
+Array::Concept* Array::Model<T>::copy() const
```
dictoon: Mark these methods as inline? Not sure it makes much of a difference...

src/appleseed/renderer/modeling/bsdf/energycompensation.cpp
est77: The array is a const float* but to make this class work with both external tables and computed tables I need the const cast

src/appleseed/foundation/array/array.cpp
est77: Probably not. All these methods are virtual.

PR #2645, opened 2019-06-14
Increase precision of noise level log output

Mango-3:
Increases the precision of the displayed average noise level value in the debug render log to 4 digits.
One significant digit (current form) is not sufficient, more converged render will show 0.0 noise level values while noise is still visible (see screenshot).

There is probably an easier method to do this with stats.insert() but it didn't accept a std::streamsize precision specifier when I tried.  

Capture

### DIFF
--- a/src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
+++ b/src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
@@ -565,7 +565,7 @@ namespace
                     pretty_uint(tile_converged_pixel_count),
                     converged_pixels_string));
             stats.insert("samples/pixel", m_spp);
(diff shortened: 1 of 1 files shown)


src/appleseed/renderer/kernel/rendering/final/adaptivetilerenderer.cpp
```diff
@@ -565,12 +565,14 @@ namespace
                     pretty_uint(tile_converged_pixel_count),
                     converged_pixels_string));
             stats.insert("samples/pixel", m_spp);
```
dictoon: A better solution that comes to mind is to insert a string that you generated using `foundation::pretty_scalar()`.

Mango-3: Thanks, that worked.

PR #2646, opened 2019-06-16
Arrays and transforms work

est77:


src/appleseed/foundation/array/keyframedarray.cpp
```diff
@@ -0,0 +1,216 @@
...
+}
+
+const Array* KeyFramedArray::begin() const
+{
+    return m_keys;
+}
+
+const Array* KeyFramedArray::end() const
+{
+    return m_keys + m_key_count;
```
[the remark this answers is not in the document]
est77: m_keys is an array of Arrays. m_keys + N will give you the Nth Array in the array.

src/appleseed/renderer/utility/transformsequencestack.h
```diff
@@ -0,0 +1,121 @@
...
+    m_stack.pop();
+    return seq;
+}
+
+inline void TransformSequenceStack::clear()
+{
+    while (!m_stack.empty())
+        pop();
+
+    m_stack.push(TransformSequence());
```
dictoon: Why push a transform here? It's kind of weird that after a call to `clear()`, `size()` will return 1.

src/appleseed/renderer/utility/transformsequencestack.h
```diff
@@ -0,0 +1,121 @@
...
+    push(seq);
+}
+
+inline TransformSequence TransformSequenceStack::pop()
+{
+    assert(!m_stack.empty());
+
+    TransformSequence seq(std::move(m_stack.top()));
+    m_stack.pop();
+    return seq;
```
dictoon: I suppose that we are relying on RVO/NRVO here to avoid a copy?

src/appleseed/foundation/meta/tests/test_keyframedarray.cpp
```diff
@@ -0,0 +1,106 @@
...
+        EXPECT_FALSE(y.get_key(0).empty());
+        EXPECT_EQ(UInt32Type, y.type());
+        EXPECT_EQ(101, y.get_key(0).size());
+    }
+
+    TEST_CASE(CopyConstruct)
+    {
+        KeyFramedArray k(UInt32Type);
+
+        const uint32 items[] = {1, 5, 7, 11, 17, 21, 23};
```
dictoon: Include `foundation/platform/types.h` for `uint32` (at some point we'll switch to `std::uint32_t` & co. everywhere).

PR #2656, opened 2019-07-01
Path guiding

BashPrince:
This PR will be continuously updated and should not be merged at this point.

Here is the current state of the implementation of "Practical Path Guiding" [Müller et.al. 2017]. The reference implementation for Mitsuba can be found here https://github.com/Tom94/practical-path-guiding.

To test the current state the Guided Path Tracing Lighting Engine has to be selected in the UI and a sufficiently high number of passes has to be set (this can be arbitrarily high, the guided path tracer will stop when the sample budget has been reached).


BashPrince:
The current state contains all changes for my GSoC path guiding project and is ready for review. Settings can be controlled from the Path Guiding panel.

Settings

- _Samples per pass_: Controls the number of samples of one path guiding pass. The actual number of samples per pixel is set as usual in the Uniform Sampler settings. For expected results the user has to make sure that the number of passes in the sampler settings is sufficiently high for the sample number.
Example: 1024 samples per pixel at 4 samples per pass = 256 passes in the sampler.
It's all a bit convoluted for now I'd still like to improve that.
- _Spatial Filter_: Sets the spatial S-Tree filter type. Recommended is stochastic.
- _Directional Filter_: Sets the directional D-Tree filter type. Recommended is box.
- _BSDF Sampling Fraction Mode_: Set the ratio mode between ordinary sampling and SD-Tree sampling to a fixed fraction or utilize the automatic learning of the ideal fraction.
- _Fixed BSDF Sampling Fraction_: Set the fraction if sampling fraction mode is _Fixed_
- _Learnig Rate_: Set the learning rate if sampling fraction mode is _Learn_
- _Iteration Progression_: Either combine the images across iteration based on their estimated inverse variance (_Combine Iterations_) or initiate a final iteration when the projected variance of the final image increases (_Automatic Cut-off_)
- _Guided Bounce Mode_: To fit within our path tracing routine a scattering mode should be chosen at a path extension. Since the path guided bounces replace ordinary BSDF sampling, a mode has to be selected. This setting gives some control over how that should happen.

"Learned Distribution" will calculate how spread out over the sphere the incoming radiance at any D-Tree's spatial location is. If >40% of the energy in the last iteration are contained within <= 10% of the spherical directions the D-Tree returns Glossy for a direction sample and Diffuse otherwise. Ideally fireflies caused by spikey D-tree distributions can be prevented (if caustics are disabled).
The remaining modes will strictly (_Strictly Diffuse_ / _Strictly Specular_) assign a user chosen mode on path guided bounces (or terminate if the remaining path scattering modes do not allow bounces of that type) or preferably assign the chosen mode and assign the other mode otherwise (_Prefer Diffuse_ / _Prefer Specular_).

One issue with assigning the scattering mode based on the local distribution is that this assignment is likely to change across iterations so you would terminate paths differently across different iterations (unless all bounce types are unlimited) and introduce brightness changes that will be combined in the final image weighting. If a strict mode is chosen it might be not the ideal mode with regard to terminating caustics and also the regular path tracer would be starved for bounces of that mode. I think none of these options are ideal. This will require some more future thought.

At the current moment rendering checkpoints will not work with path guiding since this will require a method to integrate storing and loading the SD-tree to / from disk. Saving/loading is not really be an issue but doing so will require integration with the rest of the checkpoint mechanic.

There are still some issues with the path guiding algorithm that I will have to investigate further. One such issue are prominent fireflies that show up with path guiding.

BashPrince:
Last commit was accidental and should be removed.

Aaron19960821:
Hi @BashPrince Are you still working on merging this commit into the master branch? I saw **Path Guiding** is a topic in the idea list of GsOC 2020 and want to make it my topic. I am curious what's going on your project.

BashPrince:
Hey,
Yes I'm still working on it, currently I'm trying to extend the algorithm for my bachelor thesis (working on it in a separate branch). I'll then still have to integrate all of the results. The core algorithm already works but still needs quite a bit of code overhaul.

PR #2657, opened 2019-07-01
Fixes + test suite updates

dictoon:


src/appleseed/foundation/meta/tests/test_objmeshfilereader.cpp
```diff
@@ -121,4 +139,41 @@ TEST_SUITE(Foundation_Mesh_OBJMeshFileReader)
         EXPECT_EQ(4, mesh.m_tex_coords.size());
         EXPECT_EQ(1, mesh.m_faces.size());
     }
+
+#if 0
```
oktomus: Why do we keep this ?

src/appleseed/renderer/modeling/project/projectfileupdater.cpp
```diff
@@ -2109,12 +2110,19 @@ namespace
...
-                        if (frame)
+                        if (max_samples < numeric_limits<uint64>::max())
                         {
-                            // If max samples was previously set then preserve the nearest max average spp count.
-                            pfr.strings().insert(
-                                "max_average_spp", static_cast<uint64>(ceil(max_samples / frame->get_crop_window().volume())));
+                            Frame* frame = m_project.get_frame();
+                            if (frame)
+                            {
+                                // If max samples was previously set then preserve the nearest max average spp count.
```
oktomus: Maybe warn the user when this happen ?

src/appleseed/foundation/meta/tests/test_objmeshfilereader.cpp
dictoon: This code is useful when we need to convert an image to code data. I want to keep it around and not have to rewrite it every time or fish it from an outdated branch. I don't think it really hurts to keep it here.

dictoon:
> I don't think the name of the commit Expose --disable-abort-dialogs even in Release builds is not relevant to its changes ?

Yes it is. Before the option was only exposed in Debug builds, now it's also exposed in Release builds.

dictoon:
> Ok so i checked your code but not the scene files (too much diff 🤣). I added comments but no requests.

Ahah yeah, no worries.

Thanks for the review!

src/appleseed/renderer/modeling/project/projectfileupdater.cpp
dictoon: I missed that comment. Maybe it's a good idea yes.

PR #2659, opened 2019-07-09
Fix regression: crash when using light tree and motion blur

oktomus:


src/appleseed/renderer/kernel/lighting/lightsamplerbase.cpp
```diff
@@ -633,19 +633,39 @@ void LightSamplerBase::store_object_area_in_shadergroups(
     }
 }
 
+void LightSamplerBase::sample_emitting_shape(
+    const ShadingRay::Time&             time,
+    const Vector3f&                     s,
```
dictoon: I think it would be clearer if this method took a `Vector2f` for `s`.

src/appleseed/renderer/kernel/lighting/lightsamplerbase.cpp
```diff
@@ -633,19 +633,39 @@ void LightSamplerBase::store_object_area_in_shadergroups(
...
+}
+
 void LightSamplerBase::sample_emitting_shapes(
     const ShadingRay::Time&             time,
     const Vector3f&                     s,
     LightSample&                        light_sample) const
 {
     assert(m_emitting_shapes_cdf.valid());
 
+    // Fetch the emitthing shape.
```
dictoon: Typo: emitting

PR #2660, opened 2019-07-10
Implement Order-Independent Transparency

termhn:
WIP. Going to split the latest commit up a bit.


oktomus:
image
There is only one thing stopping your code from building on linux :)

PR #2661, opened 2019-07-10
Unified Viewport

termhn:
* Fixes some ways to get into invalid states by going into and out of light path display toggle/base layer modes and restarting rendering
* No longer does a light path picking event when moving the camera in OpenGL view
* No longer does a light path picking event at all when in OpenGL view until I implement better filtering modes/scene space picking
* Dims the final render when you have at least one light path selected

To be implemented in this PR:
* Selected individual light paths are not drawn solo but rather the rest of the picked light paths have lower transparency.

=======

https://www.grayolson.me/blog/posts/gsoc-2019/


oktomus:
Closed in favor of #2748

PR #2663, opened 2019-07-11
Allow Travis OSX build to fail

est77:
The build is broken too frequently. Tracking random configuration and dependency version changes is just too time consuming for a platform we don't really support.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,6 +67,10 @@ matrix:
       compiler: gcc
       env: GCC_VERSION=7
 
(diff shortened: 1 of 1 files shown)


dictoon:
I don't think this is a good idea. It's reassuring to know that appleseed builds on macOS at all times, it makes my life a lot easier when comes release time.

The failure we're getting right now (https://travis-ci.org/appleseedhq/appleseed/jobs/558781873) is legitimate. I'm fixing it.

PR #2668, opened 2019-07-22
Fix crash in appleseed.studio; update test suite

dictoon:


sandbox/tests/test scenes/analytic lights/01 - rectangle - mesh.appleseed
```diff
@@ -82,14 +86,14 @@
                 <assign_material slot="default" side="front" material="material1" />
                 <assign_material slot="default" side="back" material="material1" />
             </object_instance>
-            <object_instance name="triangle" object="triangle_obj.mesh">
+            <object_instance name="rect" object="rect_obj">
```
oktomus: `rectangle` would have been better

PR #2670, opened 2019-07-26
Fix diffuse btdf aov components

### LINKED ISSUE #2372: Review albedo initialization in BSDFs
It seems that some BSDFs don't return the correct albedo, for instance the Diffuse BTDF (`diffusebtdf.cpp`) returns `values->m_transmittance` while it should probably return `sample.m_value.m_diffuse`:

    // Compute the BRDF value.
    sample.m_value.m_diffuse = values->m_transmittance;
    sample.m_value.m_diffuse *= values->m_transmittance_multiplier * RcpPi<float>();
    sample.m_value.m_beauty = sample.m_value.m_diffuse;

    // ...

    sample.m_aov_components.m_albedo = values->m_transmittance;

Same problem with the Diffuse BRDF, and possibly others.

mandeep:
fixes #2372 

Not much discussion happened on the issue tracker for this issue so I'm just going to submit this and we can discuss further if needed.

### DIFF
--- a/src/appleseed/renderer/modeling/bsdf/diffusebtdf.cpp
+++ b/src/appleseed/renderer/modeling/bsdf/diffusebtdf.cpp
@@ -137,9 +137,10 @@ namespace
 
                 // Compute the BRDF value.
                 sample.m_value.m_diffuse = values->m_transmittance;
-                sample.m_value.m_diffuse *= values->m_transmittance_multiplier * RcpPi<float>();
-                sample.m_value.m_beauty = sample.m_value.m_diffuse;
+                sample.m_value.m_diffuse *= values->m_transmittance_multiplier;
                 sample.m_aov_components.m_albedo = sample.m_value.m_diffuse;
+                sample.m_value.m_diffuse *= RcpPi<float>();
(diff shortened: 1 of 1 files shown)


src/appleseed/renderer/modeling/bsdf/diffusebtdf.cpp
```diff
@@ -139,7 +139,7 @@ namespace
                 sample.m_value.m_diffuse = values->m_transmittance;
                 sample.m_value.m_diffuse *= values->m_transmittance_multiplier * RcpPi<float>();
                 sample.m_value.m_beauty = sample.m_value.m_diffuse;
-                sample.m_aov_components.m_albedo = values->m_transmittance;
+                sample.m_aov_components.m_albedo = sample.m_value.m_diffuse;
```
est77: What we really want is values->m_transmittance * values->m_transmittance_multiplier, 
without the RcpPi<float>() factor.

What you can do is compute the albedo first and then:

diffuse = albedo;
diffuse *= RcpPi<float>();

mandeep: I submitted a new change, but we want albedo = diffuse right?

PR #2675, opened 2019-08-02
Fix procedural shape normals transformation

oktomus:
Scene `09 - lit sphere - mesh` renders a scene with a translate sphere on Z and a rotate assembly on X by 90 degrees. The mesh render looks like this.
image

The same scene is rendered but with a procedural sphere. Because normals are not correctly transformed, the render is wrong.
image

This PR fixes this by transforming procedural shapes normals in assembly instance space. Now the render looks like this. There is still self-intersections that will need to be fixed.
image


src/appleseed/renderer/kernel/intersection/assemblytree.cpp
```diff
@@ -791,9 +792,17 @@ bool AssemblyLeafVisitor::visit(
                     m_shading_point.m_object_instance_index = object_instance_index_pair.second;
                     m_shading_point.m_primitive_index = 0;
                     m_shading_point.m_primitive_pa = result.m_material_slot;
-                    m_shading_point.m_geometric_normal = object_instance_transform.normal_to_parent(result.m_geometric_normal);
-                    m_shading_point.m_original_shading_normal = object_instance_transform.normal_to_parent(result.m_shading_normal);
+                    m_shading_point.m_geometric_normal =
```
dictoon: That looks good.

Could you add a simple test scene that demonstrates the bug and shows that normals are now properly transformed?

src/appleseed/renderer/kernel/shading/shadingpoint.h
```diff
@@ -611,9 +611,9 @@ inline const foundation::Vector3d& ShadingPoint::get_offset_point(const foundati
     assert(m_members & HasRefinedPoints);
 
     return
-        foundation::dot(m_asm_geo_normal, direction) > 0.0
-            ? m_front_point
-            : m_back_point;
+        foundation::dot(m_asm_inst_geo_normal, direction) > 0.0
```
dictoon: What space is `direction` in? It should be in assembly instance space (and it should be documented as such in the function _declaration_). However, if it's in world space, then we have a bug here.

oktomus:
I will create another PR.

src/appleseed/renderer/kernel/shading/shadingpoint.h
oktomus: It's in assembly instance space

oktomus: The only call is the following (with direction being defined on the first line).
` 
// Transform the ray direction to assembly instance space.
        output_ray.m_dir = assembly_instance_transform.vector_to_local(input_ray.m_dir);

        // Compute the ray origin in assembly instance space.
        if (parent_sp &&
            parent_sp->get_assembly_instance().get_uid() == assembly_instance.get_uid() &&
            parent_sp->get_object_instance().get_ray_bias_method() == ObjectInstance::RayBiasMethodNone)
        {
            // The caller provided the previous intersection, and we are about
            // to intersect the assembly instance that contains the previous
            // intersection. Use the properly offset intersection point as the
            // origin of the child ray.
            output_ray.m_org = parent_sp->get_offset_point(output_ray.m_dir);
`

sandbox/tests/test scenes/procedural objects/09 - lit sphere - mesh.appleseed
```diff
@@ -0,0 +1,124 @@
...
+            <object_instance name="sphere_inst" object="sphere">
+                <transform>
+                    <translation value="0.0 0.0 -1.0" />
+                </transform>
+                <assign_material slot="default" side="front" material="material1" />
+            </object_instance>
+        </assembly>
+        <assembly_instance name="assembly_inst" assembly="assembly">
+                <transform>
+                    <rotation axis="1.0" angle="90.0" />
```
dictoon: I'm surprised that this works at all, `axis` is supposed to contain three values (X, Y, Z).

dictoon: Also, could you load and save again the file in appleseed.studio (or using `projecttool`) please?

sandbox/tests/test scenes/procedural objects/10 - lit sphere - analytic.appleseed
```diff
@@ -0,0 +1,121 @@
...
+            <object_instance name="sphere_inst" object="sphere">
+                <transform>
+                    <translation value="0.0 0.0 -1.0" />
+                </transform>
+                <assign_material slot="default" side="front" material="material1" />
+            </object_instance>
+        </assembly>
+        <assembly_instance name="assembly_inst" assembly="assembly">
+                <transform>
+                    <rotation axis="1.0" angle="90.0" />
```
dictoon: Same problem.

src/appleseed/renderer/kernel/shading/shadingpoint.h
```diff
@@ -158,8 +158,8 @@ class ShadingPoint
     // Return the intersection point in world space, with per-object-instance ray bias applied.
     foundation::Vector3d get_biased_point(const foundation::Vector3d& direction) const;
```
dictoon: What about `direction` here, in what space is it?

sandbox/tests/test scenes/procedural objects/09 - lit sphere - mesh.appleseed
oktomus: I did and now it's a marix

src/appleseed/renderer/kernel/shading/shadingpoint.h
oktomus: It's in world space

oktomus: image
image
image

dictoon: Got it, thanks!

PR #2677, opened 2019-08-15
Fix rare crash of appleseed on exit

dictoon:


src/appleseed/renderer/kernel/intersection/intersector.cpp
```diff
@@ -1,389 +1,389 @@
...
-
+
+//
+// This source file is part of appleseed.
+// Visit https://appleseedhq.net/ for additional information and resources.
+//
+// This software is released under the MIT license.
+//
+// Copyright (c) 2010-2013 Francois Beaune, Jupiter Jazz Limited
+// Copyright (c) 2014-2018 Francois Beaune, The appleseedhq Organization
```
oktomus: 2019

dictoon:
> It's hard to see what changed in intersector.cpp. It seems you changed the order of some declarations

Line endings were busted, I fixed them. Nothing else has changed.

src/appleseed/renderer/kernel/intersection/intersector.cpp
dictoon: At some point I'll do a pass on all files to update copyright notices. Usually I do that at the beginning of each year but this time I've been lazy. Might as well wait until January 2020 at this point :)

dictoon:
Thanks for the review @oktomus!

PR #2689, opened 2019-09-09
Remove using namespace std; from appleseed.python

LZaw:
This PR part of a series.
Fixes Issue#2641 for appleseed.python.
I tried to keep the formatting up and also to detect all missing `std::`s, which are in Windows/macOS-only code.

### DIFF
--- a/src/appleseed.python/bindtexture.cpp
+++ b/src/appleseed.python/bindtexture.cpp
@@ -44,7 +44,6 @@
 namespace bpy = boost::python;
 using namespace foundation;
 using namespace renderer;
-using namespace std;
 
 // Work around a regression in Visual Studio 2015 Update 3.
 #if defined(_MSC_VER) && _MSC_VER == 1900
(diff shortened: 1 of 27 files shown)


src/appleseed.python/bindtexture.cpp
```diff
@@ -114,9 +113,9 @@ namespace
     }
 
     auto_release_ptr<TextureInstance> create_texture_instance(
-        const string&                   name,
+        const std::string&                   name,
```
dictoon: Fix indentation.

src/appleseed.python/bindvolume.cpp
```diff
@@ -58,8 +57,8 @@ namespace boost
 namespace
 {
     auto_release_ptr<Volume> create_volume(
-        const string&    model,
-        const string&    name,
+        const std::string&    model,
```
dictoon: Fix indentation.

PR #2690, opened 2019-09-09
Remove using namespace std; from appleseed.studio

LZaw:
This PR part of a series.
Fixes #2641 for appleseed.studio.

See also: #2689, #2691, #2692, #2693, #2694

PR #2691, opened 2019-09-09
Remove 'using namespace std' from appleseed/renderer/modeling

### LINKED ISSUE #2641: Remove using namespace std directives

LZaw:
This PR part of a series.
Fixes #2641 for appleseed/renderer/modeling.

See also: #2689, #2690, #2692, #2693, #2694


dictoon:
Technically, `cos()`, `acos()`, etc. would need to be qualified with `std::`. Right now it builds fine because these are imported into the global namespace by `<cmath>` but it is not a guarantee. We can and should add all missing qualifiers over time. Merging this now. Thanks!

Issue #2688, opened 2019-09-04
Remove microfacet distribution params from OSL shaders


dictoon:
For instance, in `src/appleseed.shaders/src/appleseed/as_metal.osl` we currently have this:

```
int in_distribution = 0
[[
    string as_maya_attribute_name = "distribution",
    string as_maya_attribute_short_name = "mdf",
    string widget = "mapper",
    string options = "Beckmann:0|GGX:1",
    string label = "Distribution",
    string page = "Specular",
    int as_maya_attribute_connectable = 0,
    int as_maya_attribute_keyable = 0,
    int as_maya_attribute_hidden = 1,
    int as_blender_input_socket = 0,
    int gafferNoduleLayoutVisible = 0,
    int divider = 1,
    int as_deprecated = 1
]],
```

But the value of this parameter is now ignored.

There are likely several other shaders that require updating.

**Important:** We need to make sure that removing these parameters don't break backward-compatibility. If that's not possible, then we should at least make it clear that they are deprecated.

luisbarrancos:
@dictoon i think it's only those 3 (glass, metal, plastic), they are marked as deprecated. Removing them will break existing Maya scenes (one can add attributes, but removing them breaks scenes). I have no idea if in Max the scenario is the same. Iterating over the shader graph to find offending shaders (v1) and disconnect the attrs, remap to new (v2) mapping and rewire everything again would be one solution, but for this i would wait until some functionality is in place in order to write v2 shaders (layered closures mostly, volume closures in OSL).

est77:
Parameters can be removed from Maya nodes. You have a warning, but scenes don't break.

luisbarrancos:
@est77 great, i'll remove this and add a ticket for a script that rewires shading networks. The parameter never allowed upstream connections anyway so it should be safe.

dictoon:
> Parameters can be removed from Maya nodes. You have a warning, but scenes don't break.
Unknown attributes are skipped.

That's great! I don't think these shaders can be used in Max, at least not officially, so we should be good there. I'm not sure about Blender though. @jdent02 Will removing deprecated parameters break existing Blender scenes that rely on those shaders?

jdent02:
I don't think so.  The parameter would be saved in any existing .blend but I think it just gets ignored if the shader no longer has it.

dictoon:
Thanks for the info Jon. We should test this.

Issue #2680, opened 2019-08-20
Set Tonemap Color to off by default for emission in asStandardSurface


Mango-3:
`Tonemap Color` limits the brightness range of the emission when an asBlackbody node is used to drive the emission color. This confuses new users because by default it is enabled and gives the impression that the brightness cannot be adjusted. 
See for instance user Ralf Dietrich's comment in our forum:
https://forum.appleseedhq.net/t/problem-trying-to-create-a-light-emitting-material-in-blender/507/5

dictoon:
Seems reasonable. @luisbarrancos?

luisbarrancos:
@dictoon sure, and apologies for the late reply.

PR #2699, opened 2019-09-13
Remove MDF choice, disable tonemapping EDF

luisbarrancos:
Addressing #2688 and #2680

### DIFF
--- a/src/appleseed.shaders/src/appleseed/as_glass.osl
+++ b/src/appleseed.shaders/src/appleseed/as_glass.osl
@@ -87,21 +87,6 @@ shader as_glass
         int gafferNoduleLayoutVisible = 0,
         int divider = 1
     ]],
-    string in_distribution = "ggx"
-    [[
-        string as_maya_attribute_name = "distribution",
-        string as_maya_attribute_short_name = "dd",
-        string label = "Distribution",
-        string widget = "popup",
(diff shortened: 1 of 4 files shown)


src/appleseed.shaders/src/appleseed/as_glass.osl
```diff
@@ -87,21 +87,6 @@ shader as_glass
         int gafferNoduleLayoutVisible = 0,
         int divider = 1
     ]],
-    string in_distribution = "ggx"
```
dictoon: To be honest, I hadn't noticed that these parameters were marked as deprecated. That makes them less of a problem than I initially thought.

src/appleseed.shaders/src/appleseed/as_metal.osl
```diff
@@ -328,26 +312,6 @@ shader as_metal
         return;
     }
 
-    string distribution;
```
dictoon: That's clearly code that we can remove without breaking anything, regardless of our decision of what to do with the deprecated parameters.

dictoon:
In #2688 I've asked what happens to existing Blender scenes. If removing deprecated params don't break them then we can merge this PR as is. Otherwise we'll need to restore the deprecated params.

PR #2701, opened 2019-09-19
Limit self-intersection warnings to 20 per thread

oktomus:
Inspired from adaptive tile renderer invalid sample warnings

### DIFF
--- a/src/appleseed/renderer/kernel/intersection/intersector.cpp
+++ b/src/appleseed/renderer/kernel/intersection/intersector.cpp
@@ -224,13 +224,27 @@ namespace
         const ShadingPoint&         shading_point,
         const ShadingPoint*         parent_shading_point)
     {
+        static const size_t MaxWarningsPerThread = 100;
+        static size_t warnings_count = 0;
+
         if (shading_point.hit_surface() &&
             parent_shading_point &&
             same_triangle(*parent_shading_point, shading_point))
         {
-            RENDERER_LOG_WARNING(
-                "self-intersection detected, distance %e.",
-                shading_point.get_distance());
+            if (warnings_count < MaxWarningsPerThread)
+            {
+                RENDERER_LOG_WARNING(
+                    "self-intersection detected, distance %e.",
+                    shading_point.get_distance());
+
+                warnings_count += 1;
+            }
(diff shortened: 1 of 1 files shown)


src/appleseed/renderer/kernel/intersection/intersector.cpp
```diff
@@ -224,13 +224,27 @@ namespace
         const ShadingPoint&         shading_point,
         const ShadingPoint*         parent_shading_point)
     {
+        static const size_t MaxWarningsPerThread = 100;
```
dictoon: Replace `static const` by `constexpr`.

src/appleseed/renderer/kernel/intersection/intersector.cpp
```diff
@@ -224,13 +224,27 @@ namespace
         const ShadingPoint&         shading_point,
         const ShadingPoint*         parent_shading_point)
     {
+        static const size_t MaxWarningsPerThread = 100;
+        static size_t warnings_count = 0;
```
dictoon: If you put "count" after something, that something must be singular: `warning_count`.

src/appleseed/renderer/kernel/intersection/intersector.cpp
```diff
@@ -224,13 +224,27 @@ namespace
...
-            RENDERER_LOG_WARNING(
-                "self-intersection detected, distance %e.",
-                shading_point.get_distance());
+            if (warnings_count < MaxWarningsPerThread)
+            {
+                RENDERER_LOG_WARNING(
+                    "self-intersection detected, distance %e.",
+                    shading_point.get_distance());
+
+                warnings_count += 1;
```
dictoon: `++warnings_count` :)

src/appleseed/renderer/kernel/intersection/intersector.cpp
```diff
@@ -224,13 +224,27 @@ namespace
...
+            }
+            else if (warnings_count == MaxWarningsPerThread)
+            {
+                RENDERER_LOG_WARNING("more self-intersections detected, omitting warning messages for brevity.");
+
+                warnings_count += 1;
```
dictoon: Same here.

PR #2703, opened 2019-09-21
Updates, fixes, refactorings

dictoon:


scripts/appleseed.package.py
```diff
@@ -874,8 +878,9 @@ def postprocess_stage(self):
         self.__make_executable(os.path.join("appleseed/bin", "oslc"))
         self.__make_executable(os.path.join("appleseed/bin", "oslinfo"))
         self.__add_dependencies_to_stage()
-        self.__set_runtime_paths_on_binaries()
-        self.__clear_runtime_paths_on_libraries()
+        self.__set_runtime_paths_on_binaries("appleseed/bin/", "$ORIGIN/../lib")
+        self.__set_runtime_paths_on_libraries("appleseed/bin/", "$ORIGIN/../../lib")  # Qt plugins
```
oktomus: Is the / at the end wanted?

scripts/appleseed.package.py
```diff
@@ -916,14 +936,29 @@ def get_dependencies_for_file(self, filepath):
...
+            for lib in libs:
+                trace("      {0}".format(lib))
+        else:
+            trace("  Dependencies for {0}: None".format(filepath))
+
         return libs
 
+    def __is_shared_lib(self, filename):
+        ext = os.path.splitext(filename)[1]
+        return ext == ".so" or ".so." in filename
```
oktomus: i propose the following : `filename.endswith(".so") || ".so" in filename`

dictoon:
> Thanks! I think your PR changed while i was reading it 🤔

I'm just rebasing it on top of master and fixing conflicts.

scripts/appleseed.package.py
dictoon: Yep.

scripts/appleseed.package.py
dictoon: That's not quite correct, the second dot in `".so."` is important (otherwise it would pick things like `file.something.ext`).

oktomus: My bad, then it is 


`filename.endswith(".so") || ".so." in filename`

dictoon: Ok.

PR #2707, opened 2019-09-25
Add missing std:: qualifiers

LZaw:
As discussed in #2691 we can't rely on the cmath functions working without an explicit 'std::'. I added 'std::'s for every cmath function, which wasn't already present in `foundation::`.


src/appleseed/renderer/modeling/bssrdf/randomwalkbssrdf.cpp
```diff
@@ -495,7 +495,7 @@ namespace
             assert(abs(cosine) <= 1.0f);
```
dictoon: `std::abs()`

src/appleseed/renderer/modeling/bssrdf/randomwalkbssrdf.cpp
```diff
@@ -495,7 +495,7 @@ namespace
...
             const Basis3f basis(normal);
             const Vector2f tangent = sample_circle_uniform(s);
-            const float sine = sqrt(max(1.0f - cosine * cosine, 0.0f));
+            const float sine = std::sqrt(max(1.0f - cosine * cosine, 0.0f));
```
dictoon: `std::max()`

Feel free to fix these ones (those not from `<cmath>`) in another PR.

src/appleseed/renderer/modeling/light/maxspotlight.cpp
```diff
@@ -109,10 +109,10 @@ namespace
...
             const double tilt_angle = deg_to_rad(m_params.get_optional<double>("tilt_angle", 0.0));
 
-            m_cos_inner_half_angle = cos(inner_half_angle);
-            m_cos_outer_half_angle = cos(outer_half_angle);
-            m_rcp_screen_half_size = 1.0 / tan(outer_half_angle);
-            m_up = Vector3d(sin(tilt_angle), cos(tilt_angle), 0.0);
+            m_cos_inner_half_angle = std::cos(inner_half_angle);
+            m_cos_outer_half_angle = std::cos(outer_half_angle);
+            m_rcp_screen_half_size = 1.0 / std::tan(outer_half_angle);
+            m_up = Vector3d(sin(tilt_angle), std::cos(tilt_angle), 0.0);
```
dictoon: `std::sin()`

src/appleseed/renderer/modeling/light/spotlight.cpp
```diff
@@ -113,10 +113,10 @@ namespace
...
             const double tilt_angle = deg_to_rad(m_params.get_optional<double>("tilt_angle", 0.0));
 
-            m_cos_inner_half_angle = cos(inner_half_angle);
-            m_cos_outer_half_angle = cos(outer_half_angle);
-            m_rcp_screen_half_size = 1.0 / tan(outer_half_angle);
-            m_up = Vector3d(sin(tilt_angle), cos(tilt_angle), 0.0);
+            m_cos_inner_half_angle = std::cos(inner_half_angle);
+            m_cos_outer_half_angle = std::cos(outer_half_angle);
+            m_rcp_screen_half_size = 1.0 / std::tan(outer_half_angle);
+            m_up = Vector3d(sin(tilt_angle), std::cos(tilt_angle), 0.0);
```
dictoon: Same here.

src/appleseed/renderer/modeling/light/maxspotlight.cpp
LZaw: Ah, I had a script doing the work, but only considered ` sin` but not `(sin`. :D
I will modify and rerun.

src/appleseed/renderer/modeling/bssrdf/randomwalkbssrdf.cpp
LZaw: I thought we had an `abs` in `foundation::`, but rereading the foundation code makes me think it's not meant to replace `std::abs`, will include it too.

src/appleseed/renderer/modeling/bssrdf/randomwalkbssrdf.cpp
LZaw: I think `max` can be included here, even if it - strictly speaking - would not be part of `cmath`. But since we got e.g. `fmax` in it, I think it's fine to include `max`/`min` here as well.

LZaw:
I now added all of the cmath functions, plus `min` and `max`. Also, I included the functions which had a `(` in front (instead of a space).

src/appleseed/renderer/kernel/intersection/intersector.cpp
```diff
@@ -1,563 +1,563 @@
...
-
-    vec.insert(
-        "triangle tree access cache statistics",
-        make_dual_stage_cache_stats(m_triangle_tree_cache));
-
-    return vec;
-}
-
-}   // namespace renderer
+
```
dictoon: Not sure what happened to this file, did you change the line endings convention maybe?

LZaw: Don't know either what happened there. There are two or three files having this kind of issue, but this is the only one that actually has a change in it (somewhere in these lines there is an `std::` being added). I checked it locally and every IDE told me, only one line has been changed, while `git diff` told me, the whole file has been changed. As I stated before, I had a script add the `std::`s, so I'm not a 100% sure what's going on here.
I will investigate.

LZaw: There was something at the line endings. I have no clue what it was, but it got changed during the file analysis/automatic adaption. The file is back to normal and the respective `std::`s have been added.

PR #2714, opened 2019-09-29
Fix directions normalization when using procedural objects

oktomus:


src/appleseed/renderer/kernel/lighting/lightsamplerbase.cpp
```diff
@@ -355,7 +355,8 @@ void LightSamplerBase::collect_emitting_shapes(
             o = global_transform.point_to_parent(o);
             x = global_transform.vector_to_parent(x);
             y = global_transform.vector_to_parent(y);
-            n = global_transform.normal_to_parent(n);
+            n = normalize(global_transform.normal_to_parent(n));
+            assert(is_normalized(n));
```
oktomus: I suppose the assert is unnecessary here.

dictoon: Yep it is :)

src/appleseed/renderer/kernel/lighting/lightsamplerbase.cpp
```diff
@@ -494,7 +495,9 @@ void LightSamplerBase::collect_emitting_shapes(
             // Transform disk to world space.
             x = global_transform.vector_to_parent(x);
             y = global_transform.vector_to_parent(y);
-            n = global_transform.normal_to_parent(n);
+            n = normalize(global_transform.normal_to_parent(n));
+            assert(is_normalized(n));
```
dictoon: Here too.

PR #2721, opened 2019-10-08
Fix missing xy|xz choice for aniso vector fields

luisbarrancos:


src/appleseed.shaders/src/appleseed/as_anisotropy_vector_field.osl
```diff
@@ -168,6 +168,11 @@ shader as_anisotropy_vector_field
     {
         vector vector_map = normalize((vector) in_color * 2.0 - 1.0);
 
+        if (in_field_mode == 1)
```
dictoon: Ahah, I was about to submit a fix for that :)

My version was slightly more verbose:

```
if (in_field_mode == 0)
{
    tangent = normalize(
        vector_map[0] * normalize(Tn) +
        vector_map[1] * normalize(Bn) +
        vector_map[2] * Nn);
}
else
{
    tangent = normalize(
        vector_map[0] * normalize(Tn) +
        vector_map[2] * normalize(Bn) +
        vector_map[1] * Nn);
}
```

luisbarrancos: @dictoon i left some comments in the PR regarding the anisotropy amount range.

Regarding the 3rd channel, that was actually intentional. If one is exporting or painting a vector anisotropy map and doesn't set the Z channel to 0, or what's much much more common: using maps where RG encoded XY instead of RB encoding XY and using the wrong choice in the shader - then the mistake becomes glaringly obvious. I'm ok with removing it or adding a note in the documentation for this case, let me know your preference.

dictoon: > i left some comments in the PR regarding the anisotropy amount range.

Gotcha, answering there then.

> Regarding the 3rd channel

If the only reason for taking the normal into account is to spot errors more easily, I feel this doesn't justify the loss of compatibility with maps that encode no contribution as 0 instead of 128), so I would vote for no longer taking the normal into account. I can't imagine a case where we would legitimately want that.

If you choose to remove it, then feel free to merge this PR and do it in another one.

luisbarrancos: Got it, i'll remove it.

PR #2723, opened 2019-10-15
appleseed.cli's --samples option now takes a single argument

### LINKED ISSUE #2716: Change appleseed.cli's --samples command line argument to take a single value
Instead of two, the second one of which being ignored anyway.

mandeep:
fixes #2716 

Saw this as I was going through the tracker and figured I'd submit a PR real quick.

### DIFF
--- a/src/appleseed.cli/commandlinehandler.cpp
+++ b/src/appleseed.cli/commandlinehandler.cpp
@@ -104,9 +104,9 @@ CommandLineHandler::CommandLineHandler()
         &m_samples
             .add_name("--samples")
             .add_name("-s")
-            .set_description("set the minimum and maximum numbers of samples per pixel")
-            .set_syntax("min max")
-            .set_exact_value_count(2));
+            .set_description("set the number of samples per pixel")
+            .set_syntax("n")
+            .set_exact_value_count(1));
 
     parser().add_option_handler(
         &m_passes
--- a/src/appleseed.cli/main.cpp
+++ b/src/appleseed.cli/main.cpp
@@ -264,7 +264,7 @@ namespace
         {
             params.insert_path(
                 "uniform_pixel_renderer.samples",
-                g_cl.m_samples.values());
+                g_cl.m_samples.value());
         }
 
         if (g_cl.m_passes.is_set())
--- a/src/appleseed.cli/main.cpp
+++ b/src/appleseed.cli/main.cpp
@@ -264,7 +264,7 @@ namespace
         {
             params.insert_path(
                 "uniform_pixel_renderer.samples",
-                g_cl.m_samples.values()[1]);
+                g_cl.m_samples.values());
         }
 
         if (g_cl.m_passes.is_set())


src/appleseed.cli/commandlinehandler.cpp
```diff
@@ -104,9 +104,9 @@ CommandLineHandler::CommandLineHandler()
         &m_samples
             .add_name("--samples")
             .add_name("-s")
-            .set_description("set the minimum and maximum numbers of samples per pixel")
-            .set_syntax("min max")
-            .set_exact_value_count(2));
+            .set_description("set the number of samples per pixel")
+            .set_syntax("n")
+            .set_exact_value_count(1));
```
oktomus: I'm not sure this will work. Check out `apply_rendering_settings_command_line_options` in `main.cpp`.

oktomus: This is ok but you need to update places where `m_samples` is used.

mandeep: I didn't see any conflicts with m_samples, but I'll take a look again. Is there anywhere that immediately seemed like a change was necessary?

oktomus: Not sure but the line `g_cl.m_samples.values()[1]` won't work anymore i suppose ? We now expect only one value and you access the second value with `[1]`.

dictoon: Yes, for sure the code that uses this option needs to be updated.

mandeep: True g_cl.m_samples.values()[1] would work, but it would return the 2nd digit in the integer which is not what we want. I'll go through and see what needs to change.

mandeep: It looks like `src/appleseed.cli/main.cpp` was the only user of `m_samples`. Other occurrences of `m_samples` seem to be a different type. Please correct me if I'm wrong.

src/appleseed.cli/main.cpp
```diff
@@ -264,7 +264,7 @@ namespace
         {
             params.insert_path(
                 "uniform_pixel_renderer.samples",
-                g_cl.m_samples.values()[1]);
+                g_cl.m_samples.values());
```
dictoon: That doesn't look right. You're inserting a `std::vector` as a value to the `"uniform_pixel_renderer.samples"` param. You probably need to use `value()` instead (to be checked).

How did you test your changes?

dictoon: The test suite is in `sandbox/tests/test suite/` but it wouldn't cover that. Manually running `appleseed.cli` and checking the render and/or the log should be enough. Can't hurt adding a couple `printf()`'s if needed.

I think this works because we actually _do_ support automatic conversion of `std::vector` to string. Still, it would be nicer to use `value()` (which returns a scalar) instead of `values()` (which in this case returns a `std::vector` of size 1).

mandeep:
No problem! Thanks for taking the time to review. :smiley:

PR #2742, opened 2019-11-28
Add max param ID metadata to OSL shaders: as_subsurface, as_triplanar

usakhelo:
This PR adds metada with existing 3ds max param ID values.

### DIFF
--- a/src/appleseed.shaders/src/appleseed/as_subsurface.osl
+++ b/src/appleseed.shaders/src/appleseed/as_subsurface.osl
@@ -57,14 +57,16 @@ shader as_subsurface
         int as_maya_attribute_hidden = 1,
         int as_blender_input_socket = 0,
         int gafferNoduleLayoutVisible = 0,
-        int divider = 1
+        int divider = 1,
+        int as_max_param_id = 0
     ]],
     color in_color = color(0.87, 0.31, 0.12)
     [[
         string as_maya_attribute_name = "color",
         string as_maya_attribute_short_name = "c",
         string label = "Reflectance",
-        string page = "Subsurface"
+        string page = "Subsurface",
+        int as_max_param_id = 1
     ]],
     float in_sss_amount = 1.0
     [[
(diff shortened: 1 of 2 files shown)


luisbarrancos:
@usakhelo so we're adding the parameter IDs in order to allow the addition/removal of shader parameters? I.e, https://github.com/appleseedhq/appleseed/pull/2426

dictoon:
> @usakhelo so we're adding the parameter IDs in order to allow the addition/removal of shader parameters? I.e, #2426

That's correct, we figured that was the easiest (to understand) and simplest (to implement) solution. Does that sound reasonable to you? The conversation took place on Discord, sorry that you weren't part of it.

src/appleseed.shaders/src/appleseed/as_subsurface.osl
```diff
@@ -272,7 +294,8 @@ shader as_subsurface
         string label = "Output Matte",
         int as_maya_attribute_hidden = 1,
         int as_blender_input_socket = 0,
-        int gafferNoduleLayoutVisible = 0
+        int gafferNoduleLayoutVisible = 0,
+        int as_max_param_id = 35
```
dictoon: Same param ID as previous param.

src/appleseed.shaders/src/appleseed/as_subsurface.osl
```diff
@@ -74,14 +76,16 @@ shader as_subsurface
         float max = 1.0,
         string label = "Subsurface Weight",
         string page = "Subsurface",
-        int divider = 1
+        int divider = 1,
+        int as_max_param_id = 3
```
dictoon: What's the rationale with the param ID numbering? Are you leaving space for new params in between existing ones?

In that case, I would suggest a step size of 10 instead of 2. I would also suggest not making an exception for the first two params (that currently have IDs 0 and 1, i.e. they don't follow the scheme used for the next params).

usakhelo: IDs are the same that 3ds max plugin assigns to these parameters right now. So it should not be any difference between using provided IDs or generated by max.
All the new parameters should have ID greater than the greatest number in the shader.

usakhelo: Also output parameters have the same ID in max since they are all under one UI parameter (combobox)

dictoon: Understood!

usakhelo:
> 
> 
> @usakhelo so we're adding the parameter IDs in order to allow the addition/removal of shader parameters? I.e, #2426
> Anything to ellaborate?

Hi Luis,
Yes as @dictoon said we decided to add 3ds max param ID as metadata. That will allow to change OSL shaders. I hope in any possible way - rearrange parameters, add and remove parameters and so on. I think first we will need to mark all the parameters of the shader that we decide to update with param ID metadata. And then it can be safely updated.

PR #2745, opened 2019-12-04
Fix documentation URL in OSL shaders

dictoon:


luisbarrancos:
@dictoon i'll merge and remove the labels as discussed earlier.

PR #2746, opened 2019-12-06
Add as_max_param_id metadata to as_globals.osl

usakhelo:
Related to #2426

### DIFF
--- a/src/appleseed.shaders/src/appleseed/as_globals.osl
+++ b/src/appleseed.shaders/src/appleseed/as_globals.osl
@@ -45,66 +45,76 @@ shader as_globals
     point Pref = P
     [[
         int lockgeom = 0,
-        string widget = "null"
+        string widget = "null",
+        int as_max_param_id = 0
     ]],
     normal Nref = N
     [[
         int lockgeom = 0,
-        string widget = "null"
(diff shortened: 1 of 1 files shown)


src/appleseed.shaders/src/appleseed/as_globals.osl
```diff
@@ -45,104 +45,119 @@ shader as_globals
...
+        string widget = "null",
+        int as_max_param_id = 1
     ]],
     output point out_world_P = P
     [[
         string as_maya_attribute_name = "worldP",
         string as_maya_attribute_short_name = "wp",
-        string label = "Surface Position"
+        string label = "Surface Position",
+        int as_max_param_id = 2
```
dictoon: So all `as_max_param_id` on output parameters have the same value?

usakhelo: Yes, and the reason is because all the outputs go under single multi-string combo-box parameter in max paramblock.
It is probably enough to set it once in one output but I didn't check that option.

dictoon: That makes sense, thanks for confirming.

PR #2748, opened 2019-12-20
Unified Viewport & Light path filtering

oktomus:
I rebased the code from @termhn and I will fix some issues before merging it. This PR brings the following:

* Fixes some ways to get into invalid states by going into and out of light path display toggle/base layer modes and restarting rendering
* No longer does a light path picking event when moving the camera in OpenGL view
* No longer does a light path picking event at all when in OpenGL view until I implement better filtering modes/scene space picking
* Dims the final render when you have at least one light path selected

To be implemented in this PR:
* Selected individual light paths are not drawn solo but rather the rest of the picked light paths have lower transparency.

=======

https://www.grayolson.me/blog/posts/gsoc-2019/


dictoon:
Notes taken during live demo and discussion with @oktomus (some of these bugs may not be related to this PR and already exist in master):

- [ ] Change "Display Light Paths Overlay" to icon.
- [ ] Make light paths thinner and keep thickness constant.
- [ ] Assert when clicking on "Next light path".
- [ ] What is the shading/lighting model of the OpenGL render? It used to be based on the facing ratio but now there appears to be some kind of lighting.
- [ ] A texture store is created every time a light path is selected?
- [ ] Cannot select light paths in OpenGL visualization mode.
- [ ] Move OpenGL visualization mode to separate tab (like the current "RGB" tab) and get rid of "Scene Display Mode" combo.
- [ ] What does Alt+LMB do? Can we disable it?
- [ ] Shift+LMB shouldn't select light paths.
- [ ] We need a way to clear light paths selection (Escape?).
- [ ] Zooming too much in the render (with the mouse wheel) leads to black screen
- [ ] Clicking to select an area selects and shows light paths on the first click (instead of when the area has been defined).
- [ ] Can the light paths antialiasing be nicer?
- [ ] Save Light Paths button's tooltip should show the total number of available light paths, not the number of selected light paths.
- [ ] Backface culling button should only be present in the OpenGL tab.
- [ ] If a render is aborted, are the old light paths properly discarded? It seems that we are still able to visualize the old light paths, in particular in tiles that weren't even rendered.
- [ ] Are we always 100% precise when clicking on a pixel?
- [ ] "Display Light Paths Overlay" gets disabled whenever a new render is triggered.
- [ ] Light paths-related messages are displayed in the log even when light paths recording is disabled.
- [ ] Clear Frame should also discard light paths.

Improvements for separate PRs:

- [ ] Allow the user to choose the number of light paths to show.
- [ ] Save light paths asynchronously to avoid blocking the UI.
- [ ] Use transparency to convey light paths intensity.

src/appleseed.studio/mainwindow/rendering/renderlayer.h
```diff
@@ -145,6 +145,7 @@ class RenderLayer
     GLuint                              m_shader_program;
     GLint                               m_mult_loc;
     bool                                m_gl_initialized;
+    bool                                m_refresh_gl_texture;
```
dictoon: It needs to be initialized to `false`! :)

PR #2751, opened 2019-12-25
Adapt C++ examples to recent changes in appleseed

Mango-3:
- Removed obsolete appleseed integer types
- Added override of pure virtual function of derived class to be able to instantiate it.

These changes allow building of the example files though I'm not sure it is the most elegant way to do it (adding the override). 
Also there is still an issue with parts of the objects appearing black when rendering (see screenshot of sphereobject)
sphereobject

### DIFF
--- a/sandbox/examples/cpp/distancefieldobject/distancefieldobject.cpp
+++ b/sandbox/examples/cpp/distancefieldobject/distancefieldobject.cpp
@@ -44,7 +44,6 @@
 #include "foundation/math/rng/xoroshiro128plus.h"
 #include "foundation/math/scalar.h"
 #include "foundation/math/vector.h"
-#include "foundation/platform/types.h"
 #include "foundation/utility/api/specializedapiarrays.h"
 #include "foundation/utility/containers/dictionary.h"
 #include "foundation/utility/job/iabortswitch.h"
@@ -59,6 +58,7 @@
 #include <algorithm>
 #include <cmath>
 #include <cstddef>
+#include <cstdint>
 
 namespace asf = foundation;
 namespace asr = renderer;
@@ -151,6 +151,13 @@ namespace
             return raymarch(ray, t, p);
         }
 
+        void refine_and_offset(
+            const foundation::Ray3d& obj_inst_ray,
+            foundation::Vector3d& obj_inst_front_point,
+            foundation::Vector3d& obj_inst_back_point,
+            foundation::Vector3d& obj_inst_geo_normal) const override
+        {}
+
       private:
 
         //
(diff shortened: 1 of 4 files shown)


oktomus:
Was it already black before ?

Mango-3:
I haven't run this example before so I cannot say with certainty. But I doubt the black part is correct. Looks like a bug.

dictoon:
Thanks!

It wasn't black before. This almost certainly is a self-intersection issue: `refine_and_offset()` must return valid front & back intersection points.

sandbox/examples/cpp/distancefieldobject/distancefieldobject.cpp
```diff
@@ -151,6 +151,13 @@ namespace
             return raymarch(ray, t, p);
         }
 
+        void refine_and_offset(
+            const foundation::Ray3d& obj_inst_ray,
```
dictoon: Use `asf::` qualifier.

sandbox/examples/cpp/distancefieldobject/distancefieldobject.cpp
```diff
@@ -181,12 +188,12 @@ namespace
         {
             return
                 asf::Xoroshiro128plus(
-                    asf::hash_uint64(asf::binary_cast<asf::uint64>(ray.m_org.x)) ^
-                    asf::hash_uint64(asf::binary_cast<asf::uint64>(ray.m_org.y)) ^
-                    asf::hash_uint64(asf::binary_cast<asf::uint64>(ray.m_org.z)),
-                    asf::hash_uint64(asf::binary_cast<asf::uint64>(ray.m_dir.x)) ^
-                    asf::hash_uint64(asf::binary_cast<asf::uint64>(ray.m_dir.y)) ^
-                    asf::hash_uint64(asf::binary_cast<asf::uint64>(ray.m_dir.z)));
+                    asf::hash_uint64(asf::binary_cast<uint64_t>(ray.m_org.x)) ^
```
dictoon: Prefix integral types with `std::`.

sandbox/examples/cpp/distancefieldobject/distancefieldobject.cpp
dictoon: Align arguments.

dictoon:
Now that appleseed has a built-in sphere object, the sphere object plugin is kind of redundant even though it's a nice, simple example of how to write an object plugin (but maybe we could implement another shape).

Anyway, you could have a look at how the built-in sphere object implements refining if you want to fix the sphere object plugin:

https://github.com/appleseedhq/appleseed/blob/master/src/appleseed/renderer/modeling/object/sphereobject.cpp#L136-L188

PR #2757, opened 2020-01-08
Fix crash when rendering ends after texture was modified

dictoon:


dictoon:
Approved by @oktomus.

PR #2768, opened 2020-02-02
Fix OpenGL lighting regression

### LINKED ISSUE #2755: Fix lighting regression in OpenGL mode
How the lighting should be (appleseed 2.0.0)
image

How the lighting is now (appleseed 2.1.0)
image

oktomus:
Fix #2755

The facing ratio lighting model was removed with the introduction of OpenGL 3.3 (commit 9fee17f8e31550de476b0a24073e4206b070765c).

Future potential improvements:
- Computing the normal matrix in the vertex shader is expensive, we can cache it and compute it on the CPU.

image

### DIFF
--- a/src/appleseed.studio/mainwindow/rendering/lightpathswidget.cpp
+++ b/src/appleseed.studio/mainwindow/rendering/lightpathswidget.cpp
@@ -156,7 +156,6 @@ void LightPathsWidget::set_transform(const Transformd& transform)
 {
     m_camera_matrix = transform.get_parent_to_local();
     m_gl_view_matrix = transpose(m_camera_matrix);
-    m_camera_position = Vector3f(m_camera_matrix.extract_translation());
 }
 
 void LightPathsWidget::set_light_paths(const LightPathArray& light_paths)
@@ -615,7 +614,6 @@ void LightPathsWidget::initializeGL()
 
     m_scene_view_mat_location = m_gl->glGetUniformLocation(m_scene_shader_program, "u_view");
     m_scene_proj_mat_location = m_gl->glGetUniformLocation(m_scene_shader_program, "u_proj");
-    m_scene_camera_pos_location = m_gl->glGetUniformLocation(m_scene_shader_program, "u_camera_pos");
 
     m_light_paths_view_mat_location = m_gl->glGetUniformLocation(m_light_paths_shader_program, "u_view");
     m_light_paths_proj_mat_location = m_gl->glGetUniformLocation(m_light_paths_shader_program, "u_proj");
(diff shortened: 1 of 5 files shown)


src/appleseed.studio/mainwindow/rendering/lightpathswidget.cpp
```diff
@@ -229,7 +228,7 @@ void LightPathsWidget::slot_synchronize_camera()
 {
     m_camera.transform_sequence().clear();
     m_camera.transform_sequence().set_transform(0.0f,
-        Transformd::from_local_to_parent(inverse(m_camera_matrix)));
+    Transformd::from_local_to_parent(inverse(m_camera_matrix)));
```
dictoon: Restore indentation.

src/appleseed.studio/resources/shaders/scene.frag
```diff
@@ -28,37 +28,26 @@
...
 void main()
 {
-    vec3 acc = vec3(0.0);
-
-    vec3 N = normalize(v_norm);
-    vec3 L = normalize(LIGHT_POS - v_world_pos);
-    vec3 V = normalize(u_camera_pos - v_world_pos);
-    vec3 H = normalize(L + V);
+    vec3 view_dir = normalize(-frag_pos);
+    vec3 reflect_dir = reflect(-view_dir, frag_norm);
```
dictoon: Remove trailing whitespaces.

src/appleseed.studio/resources/shaders/scene.frag
```diff
@@ -28,37 +28,26 @@
...
-    vec3 N = normalize(v_norm);
-    vec3 L = normalize(LIGHT_POS - v_world_pos);
-    vec3 V = normalize(u_camera_pos - v_world_pos);
-    vec3 H = normalize(L + V);
+    vec3 view_dir = normalize(-frag_pos);
+    vec3 reflect_dir = reflect(-view_dir, frag_norm);  
 
-    vec3 specular_color = mix(vec3(1.0), MATERIAL_BASE_COLOR, MATERIAL_METALNESS_FACTOR);
+    float diffuse_ratio = max(0.0, dot(view_dir, frag_norm));
+    float specular_ratio = pow(max(dot(view_dir, reflect_dir), 0.0), MATERIAL_SHININESS);
```
dictoon: Be consistent with the order of arguments to `max()`.

src/appleseed.studio/resources/shaders/scene.frag
dictoon: I'm kind of a noob when it comes to GPU shader writing, but since `frag_norm` is interpolated, doesn't it require a normalization here?

oktomus: Indeed, it does. I didn't normalize it because I was looking for a flat look. But it's not correct. The flat look comes from the duplication of the normals.

PR #2772, opened 2020-02-09
Implement SPPM importons

dictoon:


src/appleseed/foundation/math/knn/knn_anyquery.h
```diff
@@ -0,0 +1,397 @@
...
+        const VectorType* APPLESEED_RESTRICT point_end = point_ptr + node->get_point_count();
+
+        while (point_ptr < point_end)
+        {
+            FOUNDATION_KNN_QUERY_STATS(++tested_point_count);
+
+            const ValueType square_dist = square_distance(*point_ptr++, query_point);
+
+            if (square_dist <= query_max_square_distance)
+                goto found;
```
LZaw: `goto` statements are badly to read as well. Also, the CI is complaining about it as well.
Maybe split the `run` function into several inlined subfunctions and make this a function call?

src/appleseed/foundation/math/knn/knn_anyquery.h
```diff
@@ -0,0 +1,397 @@
...
+#define ORDER_NODE_ENTRIES(LhsIndex, RhsIndex)                                                  \
+    if (node_queue[node_queue_size - (LhsIndex)].m_dvec_square_norm <                           \
+        node_queue[node_queue_size - (RhsIndex)].m_dvec_square_norm)                            \
+    {                                                                                           \
+        const NodeEntry tmp = node_queue[node_queue_size - (LhsIndex)];                         \
+        node_queue[node_queue_size - (LhsIndex)] = node_queue[node_queue_size - (RhsIndex)];    \
+        node_queue[node_queue_size - (RhsIndex)] = tmp;                                         \
+    }
+
+    const size_t NodeQueueSize = 128;
```
LZaw: `constexpr` can be faster (compilewise) here than `const`.

src/appleseed/foundation/math/knn/knn_anyquery.h
```diff
@@ -0,0 +1,397 @@
...
+    NodeEntry node_queue[NodeQueueSize];
+    size_t node_queue_size = 0;
+
+    // Start at the root.
+    node = nodes;
```
LZaw: Same as above

src/appleseed/foundation/math/knn/knn_anyquery.h
```diff
@@ -0,0 +1,397 @@
...
+
+#undef ORDER_NODE_ENTRIES
+
+    FOUNDATION_KNN_QUERY_STATS(stats.m_fetched_nodes.insert(fetched_node_count));
+    FOUNDATION_KNN_QUERY_STATS(stats.m_visited_leaves.insert(visited_leaf_count));
+    FOUNDATION_KNN_QUERY_STATS(stats.m_tested_points.insert(tested_point_count));
+
+    return false;
+
+  found:
```
LZaw: Addition to the `goto` statement. This is badly readable, maybe rewrite it into an inlined function.

src/appleseed/foundation/math/knn/knn_anyquery.h
```diff
@@ -0,0 +1,397 @@
...
+
+    //
+    // Step 4:
+    //
+    //   Visit the nodes in the node stack, from the closest to the farthest,
+    //   updating the stack with new nodes to visit later. We terminate as soon
+    //   as the closest node is farther than our current maximum search distance.
+    //
+
+    while (node_queue_size-- > 0)
```
LZaw: One can do that, but generally I'd recommend splitting the comparison and the decrementation into two lines. Speedwise it shouldn't make any difference (in contrast to a precrement).

src/appleseed/foundation/math/knn/knn_anyquery.h
```diff
@@ -0,0 +1,397 @@
...
+    if (node_queue[node_queue_size - (LhsIndex)].m_dvec_square_norm <                           \
+        node_queue[node_queue_size - (RhsIndex)].m_dvec_square_norm)                            \
+    {                                                                                           \
+        const NodeEntry tmp = node_queue[node_queue_size - (LhsIndex)];                         \
+        node_queue[node_queue_size - (LhsIndex)] = node_queue[node_queue_size - (RhsIndex)];    \
+        node_queue[node_queue_size - (RhsIndex)] = tmp;                                         \
+    }
+
+    const size_t NodeQueueSize = 128;
+    NodeEntry node_queue[NodeQueueSize];
```
LZaw: Might be a bit nitpicky, but this is an array, not a queue. :D

src/appleseed/foundation/math/knn/knn_anyquery.h
```diff
@@ -0,0 +1,397 @@
...
+            break;
+
+        // Push the node that we don't visit now to the node stack.
+        const ValueType square_split_dist = square(split_dist);
+        if (square_split_dist < query_max_square_distance)
+        {
+            VectorType dvec(0.0);
+            dvec[split_dim] = split_dist;
+
+            // Push the node to the node stack.
```
LZaw: The node array, which is called `node_queue` gets referred here to as a node stack. I'd recommend unified naming.

src/appleseed/foundation/math/knn/knn_anyquery.h
```diff
@@ -0,0 +1,397 @@
...
+    // Step 4:
+    //
+    //   Visit the nodes in the node stack, from the closest to the farthest,
+    //   updating the stack with new nodes to visit later. We terminate as soon
+    //   as the closest node is farther than our current maximum search distance.
+    //
+
+    while (node_queue_size-- > 0)
+    {
+        const NodeEntry* top_entry = node_queue + node_queue_size;
```
LZaw: Again nitpicky. The `top_entry` is the lowest one in the tree you found and the last one added in the stack/queue/array. Maybe rename this variable, to indicate you're currently checking the smallest node (as close as possible to leafs) that's still a candidate.
This would make the code-snipplet better readable without reading through step 1-3.

src/appleseed/foundation/math/knn/knn_anyquery.h
```diff
@@ -0,0 +1,397 @@
...
+        const VectorType* APPLESEED_RESTRICT point_end = point_ptr + node->get_point_count();
+
+        while (point_ptr < point_end)
+        {
+            FOUNDATION_KNN_QUERY_STATS(++tested_point_count);
+
+            const ValueType square_dist = square_distance(*point_ptr++, query_point);
+
+            if (square_dist <= query_max_square_distance)
+                goto found;
```
LZaw: Again `goto`

src/appleseed/foundation/math/knn/knn_query.h
```diff
@@ -184,18 +184,22 @@ inline void Query<T, N>::run(
     const VectorType* APPLESEED_RESTRICT points = &m_tree.m_points.front();
     const NodeType* APPLESEED_RESTRICT nodes = &m_tree.m_nodes.front();
     const size_t max_answer_size = m_answer.m_max_size;
+    const NodeType* APPLESEED_RESTRICT node;
+
+    const size_t IdealLeafSize = 10;
```
LZaw: Same as the `const`/`constexpr` from above.

src/appleseed/foundation/math/knn/knn_query.h
```diff
@@ -184,18 +184,22 @@ inline void Query<T, N>::run(
...
-    //   allow us to compute an initial maximum search distance, which will speed up
-    //   the "real" search later on.
+    //   This descent is very fast since no node is pushed to the stack, and it allows
+    //   us to compute an initial maximum search distance which that speed up the real
+    //   search later on.
     //
 
-    const NodeType* APPLESEED_RESTRICT node = nodes;
+    // Start at the root.
+    node = nodes;
```
LZaw: Same as above.

src/appleseed/foundation/math/knn/knn_query.h
```diff
@@ -318,20 +323,23 @@ inline void Query<T, N>::run(
     NodeEntry node_queue[NodeQueueSize];
     size_t node_queue_size = 0;
 
+    // Start at the root.
     node = nodes;
```
LZaw: Same again.

src/appleseed/foundation/math/knn/knn_anyquery.h
```diff
@@ -0,0 +1,397 @@
+
+//
+// This source file is part of appleseed.
+// Visit https://appleseedhq.net/ for additional information and resources.
+//
+// This software is released under the MIT license.
+//
+// Copyright (c) 2010-2013 Francois Beaune, Jupiter Jazz Limited
```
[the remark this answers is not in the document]
dictoon: Good point!

src/appleseed/foundation/math/knn/knn_anyquery.h
```diff
@@ -0,0 +1,397 @@
...
+{
+    assert(!m_tree.empty());
+
+    FOUNDATION_KNN_QUERY_STATS(++stats.m_query_count);
+    FOUNDATION_KNN_QUERY_STATS(size_t fetched_node_count = 0);
+    FOUNDATION_KNN_QUERY_STATS(size_t visited_leaf_count = 0);
+    FOUNDATION_KNN_QUERY_STATS(size_t tested_point_count = 0);
+
+    const VectorType* APPLESEED_RESTRICT points = &m_tree.m_points.front();
+    const NodeType* APPLESEED_RESTRICT nodes = &m_tree.m_nodes.front();
```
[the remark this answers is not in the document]
dictoon: > What is the benefit of having a pointer to a reference?

There's no such thing as a pointer to a reference in C++. A reference has no existence of its own (unlike a pointer which is a real object that takes space). Here we're simply taking a pointer to the front node.

> Also, why not make nodes a const std::vector<NodeType>&?

That would work but it would mean using `&nodes[0]` whenever we need to construct a pointer to one of the child node. It's likely that it would lead to the same generated code. It's a style issue I guess?

> If you want to stick to the pointer to a reference, I'd recommend renaming nodes to root, since this is, what front() actually returns.

`root` would be a good name too, maybe even better than `nodes`. The reason why it's called `nodes` is because it's an array of nodes (and you can do things like `nodes[nodes_index]` to reach a particular node), at least that's how I thought of it when writing this code.

src/appleseed/foundation/math/knn/knn_anyquery.h
```diff
@@ -0,0 +1,397 @@
...
+    //
+    //   Find the deepest node of the tree (leaf node or interior node) that contains
+    //   the query point and enough other points to compute a maximum search distance.
+    //   This descent is very fast since no node is pushed to the stack, and it allows
+    //   us to compute an initial maximum search distance which that speed up the real
+    //   search later on.
+    //
+
+    // Start at the root.
+    node = nodes;
```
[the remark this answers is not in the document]
dictoon: I think that weighs in favor of renaming `nodes` to `root` as you suggested above.

src/appleseed/foundation/math/knn/knn_anyquery.h
dictoon: Yeah I've removed the `goto`. I use them extremely rarely, but they occasionally come handy in performance critical code when needing to exit nested loops. But since CI complained I went for a cleaner solution.

src/appleseed/foundation/math/knn/knn_anyquery.h
dictoon: In pratice it will be the same. Also `constexpr` doesn't guarantee that the value will be a pure compile-time one, just that it _can_ be used at compile-time.

In any case, you're right, the right semantic here is `constexpr` since we want to express a true constant and not a variable that is merely const. We need to progressively replace `const` by `constexpr` in the code base.

src/appleseed/foundation/math/knn/knn_anyquery.h
dictoon: It's actually a (partially ordered) queue that happens to be stored in an array.

src/appleseed/foundation/math/knn/knn_anyquery.h
dictoon: Well ok, you're right, it's actually a partially ordered stack :) Will stick to stack then.

src/appleseed/foundation/math/knn/knn_anyquery.h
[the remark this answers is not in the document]
LZaw: > It's a style issue I guess?
Yes, most of the other comments are style/readability issues. Readabilitywise I'd prefer the `&nodes[0]`, but then again, it might not be worth the effort of changing everything here.

> it's an array of nodes
Totally makes sense. I just wanted to raise another point of view on it. If you decide to stick to the current version, that's fine as well.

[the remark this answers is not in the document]
dictoon: I've renamed `nodes` to `root_node`.

src/appleseed/foundation/math/knn/knn_anyquery.h
dictoon: By "top" here I mean that it's the top item on the stack. Since the stack is only partially ordered, it's not guaranteed to be (and typically won't be) the closest one.

src/appleseed/foundation/math/knn/knn_anyquery.h
```diff
@@ -0,0 +1,397 @@
...
+            const ValueType     dvec_square_norm)
+          : m_dvec_square_norm(dvec_square_norm)
+          , m_node(node)
+          , m_dvec(dvec)
+        {
+        }
+
+        bool operator<(const NodeEntry& rhs) const
+        {
+            return m_dvec_square_norm > rhs.m_dvec_square_norm;
```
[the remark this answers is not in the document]
dictoon: That's actually dead code, I removed it. Thanks for noticing.

src/appleseed/foundation/math/knn/knn_anyquery.h
dictoon: This is a fairly common C idiom. I used it for compactness, not for performance (as you've noted).

Anyway, for readability's sake, I've split the statement.

LZaw:
Looking very good, thank you.
The approval is meant for the KNN part, since I didn't read up on the SPPM so far, so please don't merge yet.

LZaw:
The SPPM part looks good. From my side you can merge.

PR #2773, opened 2020-02-14
Fix file path to texture sampler's mask file to naive format

### LINKED ISSUE #2695: File path to texture-controlled pixel sampler's mask file should be converted to native format
On Windows, it should use backslashes (`\`) instead of forward slashes:

image

qingqhua:
Fix #2695

### DIFF
--- a/src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
+++ b/src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
@@ -834,6 +834,7 @@ namespace
 
             if (!filepath.isEmpty())
             {
+                filepath = QDir::toNativeSeparators(filepath);
                 m_path_line_edit->setText(filepath);
             }
         }


src/appleseed.studio/mainwindow/renderingsettingswindow.cpp
```diff
@@ -834,6 +834,7 @@ namespace
 
             if (!filepath.isEmpty())
             {
+                filepath = QDir::toNativeSeparators(filepath);
```
oktomus: I don't think the change should be done here. I think it's better to do this in `get_open_filename`

dictoon: That's a good point. This should be moved to `get_open_filename()` and removed from all call sites to `get_open_filename()`. This can be done in a separate PR.

dictoon: -> https://github.com/appleseedhq/appleseed/issues/2778

PR #2774, opened 2020-02-15
Fix texture sampler panel display when loading project

### LINKED ISSUE #2698: After loading project that uses texture-controlled sampler, wrong sampler is shown in settings
It shows `Uniform` instead of `Texture`.

qingqhua:
Fix #2698 , the bug is because of the code checks if the current sampler is `"adaptive"` by `"pixel_renderer" value`, which is already been checked by `"tile_renderer" value` in previous if statement.


oktomus:
Thanks for the fix @qingqhua  :)

PR #2775, opened 2020-02-16
Do not include the scenepicker in API headers.

est77:
The scene picker includes indirectly OSL headers and force all plugins to have a dependency on OSL.


dictoon:
Thanks!

PR #2779, opened 2020-02-22
Test file writing for multiple image file formats

ChinYing-Li:
issue #2170.


src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
```diff
@@ -0,0 +1,217 @@
+// appleseed.foundation headers.
```
oktomus: Can you add a copyright notice on the top of the file with your name and `2020`, like we do in other files.

ChinYing-Li: copyright notice added. Thank you!

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
```diff
@@ -0,0 +1,245 @@
+
+//
+// This source file is part of appleseed.
+// Visit https://appleseedhq.net/ for additional information and resources.
+//
+// This software is released under the MIT license.
+//
+// Copyright (c) 2017-2018 Esteban Tovagliari, The appleseedhq Organization
```
oktomus: Change to `// Copyright (c) 2020 ChinYing-Li, The appleseedhq Organization`

oktomus:
Hi @ChinYing-Li, Thanks for updating your PR.

Regarding naming, I was referring to the commits and not the PR title. You should name your commits correctly following those rules :

- Short and descriptive (max 50 chars)
- Capitalized
- In imperative present tense
- Not end with period

You can use `git rebase` to rename your commits and merge them together. More info [here](https://git-scm.com/book/en/v2/Git-Branching-Rebasing).

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
```diff
@@ -0,0 +1,245 @@
...
+            }
+        }
+    }
+    
+    TEST_CASE(WriteTIFF_CorrectlyWritesImagePixels)
+    {
+        const char* ImageFilePath = "unit tests/outputs/test_imagefilewriter.tiff";
+        
+        {
+            Image image(16, 16, 16, 16, 4, PixelFormatUInt8);
```
oktomus: Why you make a 16x16 image here but you use 2x2 everywhere else ?

ChinYing-Li: Hi, the reason is that an error was thrown when writing tiff files: "tile size have to be multiple of 16"

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
ChinYing-Li: sorry for overlooking this, I will change that.

ChinYing-Li:
Sorry that I did not modify the title of previous commits yesterday; my laptop had some hardware issues.

ChinYing-Li:
Now I think the email address is set for ChinYing-Li::master; please let me know if there's any other issue.
Sorry for causing troubles by not following the rules!

oktomus:
Hi @ChinYing-Li, Thanks for the changes !

No worries about the rules. It's a habit that you will take very easily :)

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
```diff
@@ -0,0 +1,245 @@
...
+    {
+        const char* ImageFilePath = "unit tests/outputs/test_imagefilewriter.png";
+        
+        {
+            Image image(2, 2, 2, 2, 4, PixelFormatFloat);
+            Color4b color(50, 100, 150, 42);
+            color.unpremultiply_in_place();
+            color.rgb() = fast_linear_rgb_to_srgb(color.rgb());
+            color = saturate(color);
+            color.premultiply_in_place();
```
oktomus: Can you explain why you do this only for `png` and `tiff` ?

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
```diff
@@ -0,0 +1,245 @@
...
+            GenericImageFileWriter writer(ImageFilePath);
+            writer.append_image(&image);
+            writer.write();
+        }
+        
+        {
+            GenericImageFileReader reader;
+            std::unique_ptr<Image> image(reader.read(ImageFilePath));
+
+            for (size_t y = 0; y < 16; ++y)
```
oktomus: There is a lot of duplicated code . I'm sure we can create functions for this. @FrancoisBeaune What do you think ? 

The process is always:
1. Create a color 
2. Write an image of given size with this color
3. Read the image and assert that all pixels are == to the color

ChinYing-Li: I totally agree there's too much duplicate code. Though I wonder if we should keep the test code explicit.

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
ChinYing-Li: error_PNG
Without these few lines, the actual written value will have small discrepancy from expected value. So far I could not figure out a plausible cause, so I thought it could be a colorspace issue. (however, colorspace conversion should already be handled by other source code).

oktomus: I think you should add a comment in the code to explain this.

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
oktomus: It can stay explicit if you add functions. Take a look at some existing tests, we already do this kind of stuff :)

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
dictoon: A short comment explaining this would be useful.

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
```diff
@@ -0,0 +1,245 @@
...
+using namespace foundation;
+
+TEST_SUITE(Foundation_Image_ImageFileWriter)
+{
+    TEST_CASE(WriteBMP_CorrectlyWritesImagePixels)
+    {
+        const char* ImageFilePath = "unit tests/outputs/test_imagefilewriter.bmp";
+        
+        {
+            Image image(2, 2, 2, 2, 3, PixelFormatUInt16);
```
dictoon: You're using 16-bit integer channels for your images but that's not supported by BMP. That may be the reason why you're getting differences between what you write and what you read back.

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
```diff
@@ -0,0 +1,245 @@
+
```
dictoon: You need to add that file to `CMakeLists.txt` otherwise it won't even be compiled.

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
```diff
@@ -0,0 +1,245 @@
...
+// appleseed.foundation headers.
+#include "foundation/image/color.h"
+#include "foundation/image/colorspace.h"
+#include "foundation/image/genericimagefilereader.h"
+#include "foundation/image/genericimagefilewriter.h"
+#include "foundation/image/image.h"
+#include "foundation/image/imageattributes.h"
+#include "foundation/image/pixel.h"
+#include "foundation/math/rng/distribution.h"
```
dictoon: Do you need any of the RNG stuff? If not, remove the headers. Make sure to include all required headers, but only required headers.

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
```diff
@@ -0,0 +1,245 @@
...
+    }
+    
+    TEST_CASE(WriteJPG_CorrectlyWritesImagePixels)
+    {
+        const char* ImageFilePath = "unit tests/outputs/test_imagefilewriter.jpg";
+        
+        {
+            Image image(2, 2, 2, 2, 3, PixelFormatFloat);
+            
+            image.clear(Color3f(0.6f, 0.7f, 0.8f));
```
dictoon: I would prefer that we stick to integer components, especially for JPEG which doesn't support floating point components. You can use `std::abs()` in conjunction with `EXPECT_TRUE()` to test that values that you read back are close enough to input values.

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
dictoon: That doesn't make sense. PNG is lossless, you should be able to read back exactly what you wrote. The problem is that you're now using a floating point image (`PixelFormatFloat`) with a format that doesn't support floating point values.

dictoon:
Thanks for the contribution!

I would suggest checking which features are supported by each of the image format we want to test. For instance, PNG supports 8-bit and 16-bit integer pixel components but not floating point components. BMP only supports 8-bit integer components. Etc. With that information you'll be able to use an image using the right pixel format, and that'll get rid of most of the discrepancies you were observing.

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
ChinYing-Li: Done.

ChinYing-Li:
Hi, thank you for the suggestions. Here are the steps I took when I was writing the test:
1. I first looked up supported pixel formats for each image formats, and wrote the tests according to this. But PNG and JPEG failed due to the discrepancy in pixel values. I have not yet find an explanation for this after digging into appleseed source code or oiio source code.
2. So I tried different combinations of pixel formats and color formats to see which works. 
- image with PixelFormatUInt8 or PixelFormatUInt16 will fail the JPEG test even if I use EXPECT_FEQ.
- image with PixelFormatFloat can pass the JPEG test if we use EXPECT_FEQ.
- image with PixelFormatUInt8 or PixelFormatUInt16 will fail the PNG test due to discrepancy.

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
ChinYing-Li: Using PixelFormatUInt16 works for BMP, and there's no discrepancy. But is this unexpected behaviour?

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
ChinYing-Li: PixelFormatUInt8 & PixelFormatUInt16 image do not work for PNG. The testcase "WritePNG_CorrectlyWritesPixels_pixelB_colorB" used PixelFormatUInt8. 
```
TEST_CASE(WritePNG_CorrectlyWritesImagePixels_pixelB_colorB)
    {
        const char* ImageFilePath = "unit tests/outputs/test_imagefilewriter_pixelB_colorB.png";
        
        {
            Image image(2, 2, 2, 2, 4, PixelFormatUInt8);
            
            image.clear(Color4b(50, 100, 150, 42));
            
            GenericImageFileWriter writer(ImageFilePath);
            writer.append_image(&image);
            writer.write();
        }
        
        {
            GenericImageFileReader reader;
            std::unique_ptr<Image> image(reader.read(ImageFilePath));

            for (size_t y = 0; y < 2; ++y)
            {
                for (size_t x = 0; x < 2; ++x)
                {
                    Color4b c;
                    image->get_pixel(x, y, c);
                    EXPECT_EQ(Color4b(50, 100, 150, 42), c);
                }
            }
        }
    }
```

ChinYing-Li: Here's the source files that I wrote for the experiment: https://github.com/ChinYing-Li/Foundation_Image_IO/tree/master/debug

I will change PixelFormatUInt16 to UInt8 as necessary and upload the results soon.

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
oktomus: BMP support 8 bits colors https://en.wikipedia.org/wiki/BMP_file_format

Take this for example : Write a gradient in 16 bit in a bmp, open it, the colors will be interpreted as 8bit and the gradient will not be the same

oktomus: The problem is the same with PNG and floating points

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
oktomus: @ChinYing-Li Your file is not compiled and not used at the moment

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
```diff
@@ -0,0 +1,219 @@
...
+#include <cmath>
+#include <cstddef>
+#include <cstring>
+#include <exception>
+#include <memory>
+
+using namespace foundation;
+
+/*
+// Default export formats:
```
oktomus: Comments lack of context, you can explain what is the global goal of the test file here, and what each column of your table contain

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
```diff
@@ -0,0 +1,219 @@
...
+// BMP       .bmp          4-channel    8-bit (std::uint8_t)                    sRGB
+// PNG       .png          4-channel    8-bit (std::uint8_t)                    sRGB
+// JPEG      .jpg/.jpe/    3-channel    8-bit (std::uint8_t)                    sRGB
+//           .jpeg/.jif/
+//           .jfif/.jfi
+*/
+
+TEST_SUITE(Foundation_Image_ImageFileWriter)
+{
+    bool write_image(const std::string ImageFilePath)
```
oktomus: `write_image` should not return a `bool`. 

args and variables names should be formatted as such : `image_file_path`

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
```diff
@@ -0,0 +1,219 @@
...
+    {
+        const std::string FileExtension = ImageFilePath.substr(ImageFilePath.find_last_of("."));
```
oktomus: Unused variable, remove this

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
```diff
@@ -0,0 +1,219 @@
...
+    
+        Image image(16, 16, 16, 16, 4, PixelFormatUInt8);
```
oktomus: I think the size, pixel format and default color should be a parameter of the function.

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
```diff
@@ -0,0 +1,219 @@
...
+    
+    bool read_and_compare_RGBA(const std::string ImageFilePath)
+    {
+        
+        const std::string FileExtension = ImageFilePath.substr(ImageFilePath.find_last_of("."));
+        
+        GenericImageFileReader reader;
+        std::unique_ptr<Image> image(reader.read(ImageFilePath.c_str()));
+        
+        for (size_t y = 0; y < 16; ++y)
```
oktomus: The code that checks every pixel should stay in the test itself and you should use EXPECT_ instead of returning a value. Otherwise, how can you tell which pixel is wrong ? 

Though, you can make a method that simply read the image and return a `std::unique_ptr<Image>`. But that's not really useful.

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
```diff
@@ -0,0 +1,219 @@
...
+                image->get_pixel(x, y, c);
+                if(Color4b(50, 100, 150, 42) != c) return false;
+            }
+        }
+        
+        return true;
+    }
+                        
+    
+    TEST_CASE(WriteBMP_CorrectlyWritesImagePixels)
```
oktomus: Here is an example of how you could write one test case:
```cpp
TEST_CASE(WriteBMP_CorrectlyWriteAndReadImagePixels)
{
    const string image_file_path = "unit tests/outputs/test_imagefilewriter.bmp";
    csont Color4b fill_color(50, 100, 150, 42);

    // Write image file.
    // BMP contains 8-bit RGBA colors.
    write_image(image_file_path, 16, 16, 16, 16, PixelFormatUInt8, fill_color);

    // Check the content of the image.
    const std::unique_ptr<Image> image = read_image(image_file_path);

    for (size_t y = 0; y < 16; ++y)
    {
        for (size_t x = 0; x < 16; ++x)
        {
                Color4b c;
                image->get_pixel(x, y, c);
                EXPECT_EQ(fill_color, c);
        }
    }    
}

File: src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp (hunk 10 above)
Comment (oktomus) [reply src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp]: Or remove this comment block and comment each test, as I did in the example I sent you.

File: src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp [hunk 16]
```diff
@@ -0,0 +1,232 @@
...
+                    {
+                        EXPECT_TRUE((c[i]-fill_color[i]) < 2 || (fill_color[i] - c[i]) < 2);
+                    }
+                }
+            }
+        }
+    }
+    
+    // Test case for PNG does not work. see PR #2779 for more details.
+    TEST_CASE(WritePNG_CorrectlyWritesImagePixels)
```
ChinYing-Li: 螢幕截圖 2020-02-29 下午3 43 13
The test for PNG fails. Have been looking into this issue for a while.
The written .png looks like this: 
test_imagefilewriter

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
```diff
@@ -20,200 +47,186 @@
...
         {
-            EXPECT_TRUE(read_and_compare_RGBA(ImageFilePath));
+            GenericImageFileReader reader;
+            std::unique_ptr<Image> image(reader.read(image_file_path.c_str()));
+            
+            for (size_t y = 0; y < s; ++y)
+            {
+                for (size_t x = 0; x < s; ++x)
+                {
+                        Color4b c;
```
oktomus: Indent is wrong here

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
oktomus: Ok, i'm not sure EXPECT_EQ is the correct thing to use.
Have you tried float comparison (feq, look in other tests) on each component (c.r, c.b, c.g, c.a) ?

ChinYing-Li:
Thank you for reviewing, and I apologize for keep making stupid mistakes (not strictly following the rules etc. ). I definitely should keep the good habit to write maintainable code. 
It turns out that, .png files can be "approximately" correctly written only if ```Color4f``` is used. (Whether it is ```PixelFormatUInt8``` or ```PixelFormatFloat``` does not matter)
Below is an example, written using ```Color4f``` and ```PixelFormatUInt8```  
test_imagefilewriter_Cf_PInt_OneByOne

ChinYing-Li:
So it's not a coincidence that all other test that involves png all used ```Color4f```.
I had tried comparing the actuall written ```Color4f ``` and the expected ```Color4f``` using ```feq``` at the very beginning of this pull issue. 
Since ```appleseed``` only supports writing png using ```PixelFormatUInt8```, ```Color4f ``` and the expected ```Color4f``` will never be exactly the same unless elements of ```Color4f ``` are multiples of ```0.00390625``` (```1/256```).

oktomus:
@ChinYing-Li No problem, it's normal to do this kind of mistake when doing your first pull requests. 

All the comments we make can be annoying, but after a few contributions, you will write clean code without having to think about it. Thanks to that, your code will become easier/faster to read and maintain :)

Regarding the png issue, I think @dictoon should have a look. But you haven't addressed the last comments yet (tell me if i'm wrong). Also, be sure to remove those `.DSStore` files.

Thanks again !

ChinYing-Li:
I have to take back the previous comment after trying writing different values. Here's the result of my experiment: 
_Case 1_
```
fill_color = Color4f(0.9f, 0.7f, 0.2f, 0.9f)
Actual written color = Color4f(0.901961, 0.698039, 0.196078, 0.901961)
Difference = (0.001961, -0.001961, -0.003922, 0.001961)
```
_Case 2_
```
fill_color = Color4f(0.0f, 0.05f, 0.15f, 0.95f)
Actual written color = Color4f(0.0, 0.0470588,0.145098, 0.94902)
Difference = (0.0, -0.0029412, -0.004902, -0.00098)
```
_Case 3_
```
fill_color = Color4f(0.555555f, 0.777777f, 0.222222f, 0.89999f)
Actual written color = Color4f(0.552941, 0.772549, 0.219608, 0.898039)
Difference = (0.002614, 0.005228, 0.002614, 0.001951)
```
_Case 4_
```
fill_color = Color4f(0.0f, 1.0f, 0.0f, 1.0f)
Actual written color = Color4f(0.0f, 1.0f, 0.0f, 1.0f)
Difference = (0.0, 0.0, .0.0, 0.0)
```

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
```diff
@@ -0,0 +1,224 @@
...
+                    for (size_t i = 0; i < 3; ++i)
+                    {
+                        EXPECT_TRUE((c[i]-fill_color[i]) < 2 || (fill_color[i] - c[i]) < 2);
+                    }
+                }
+            }
+        }
+    }
+    
+    // Used Color3b instead of Color4b; Color4b does not work. See PR #2779
```
oktomus: I suggest you remove this command and create an issue on Github for this

ChinYing-Li: Sure, I will do that!

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
```diff
@@ -0,0 +1,197 @@
...
+        std::unique_ptr<Image> image(reader.read(image_file_path.c_str()));
+            
+        for (size_t y = 0; y < s; ++y)
+        {
+            for (size_t x = 0; x < s; ++x)
+            {
+                Color3b c;
+                image->get_pixel(x, y, c);
+                
+                EXPECT_TRUE((c[i]-fill_color[i]) < 2 || (fill_color[i] - c[i]) < 2);
```
oktomus: This is not clear, maybe you could write your own method or explain this. Also, a clearer way of doing this is the following :
```cpp
bool equal = std::abs(c[i] - fill_color[i]) <= 1;
```
Also, the colors are not supposed to be exactly equals ?

ChinYing-Li: Thanks for the suggestions. I will modify this and try to squash the commits together (did not squash successfully yesterday)

oktomus: If you need help with git, don't hesitate to ask on our discord :)

ChinYing-Li: 螢幕截圖 2020-03-07 上午5 45 31
As you can see here, pixel values in jpg files could be off by one.

ChinYing-Li: I don't think we can directly apply```std::abs```, as they are unsigned char( so 1-2=-1%255 = 254).

oktomus: Indeed, you're right

oktomus: @dictoon What do you think about this test case ?

oktomus:
@ChinYing-Li your code is not compiling 
image

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
```diff
@@ -0,0 +1,201 @@
...
+#include <memory>
+
+using namespace foundation;
+
+TEST_SUITE(Foundation_Image_ImageFileWriter)
+{
+    template<typename Color>
+    void write_image(const Color fill_color, const PixelFormat pixel_format, const size_t s, const std::string image_file_path)
+    {
+        size_t channels = fill_color.Components;
```
dictoon: This variable isn't terribly useful, I would suggest to remove it.

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
```diff
@@ -0,0 +1,201 @@
...
+#include <cstddef>
+#include <cstring>
+#include <exception>
+#include <memory>
+
+using namespace foundation;
+
+TEST_SUITE(Foundation_Image_ImageFileWriter)
+{
+    template<typename Color>
```
dictoon: Insert space after `template`.

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
```diff
@@ -0,0 +1,201 @@
...
+        
+        GenericImageFileWriter writer(image_file_path.c_str());
+        writer.append_image(image.get());
+        writer.write();
+    }
+    
+    TEST_CASE(WriteBMP_CorrectlyWritesImagePixels)
+    {
+        const std::string image_file_path = "unit tests/outputs/test_imagefilewriter.bmp";
+        const size_t s = 16;
```
dictoon: Use `std::size_t` in new code (fix in all new code introduced in this PR).

Rename `s` to something clearer, maybe `ImageSize`?

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
dictoon: I would introduce a function (a lambda for instance) that returns the difference between two colors, and I would then use it together with `EXPECT_LT()` for better error messages if the assertion is violated.

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
```diff
@@ -0,0 +1,200 @@
...
+            for (std::size_t x = 0; x < image_size; ++x)
+            {
+                Color3b c;
+                image->get_pixel(x, y, c);
+                
+                // the actual written and expected pixel value may be off by 1 in JPG format
+                for (std::size_t i = 0; i < 3; ++i)
+                {
+                    int difference = std::min(c[i]-fill_color[i], fill_color[i] - c[i]);
+                    EXPECT_LT(difference, 2);
```
dictoon: This should be

    EXPECT_LT(2, difference);

Right now you're checking that `2 < difference`.

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
```diff
@@ -0,0 +1,200 @@
...
+#include <cstring>
+#include <exception>
+#include <memory>
+
+using namespace foundation;
+
+TEST_SUITE(Foundation_Image_ImageFileWriter)
+{
+    template <typename Color>
+    void write_image(const Color fill_color, const PixelFormat pixel_format, const std::size_t image_size, const std::string image_file_path)
```
dictoon: Pass `image_file_path` by const ref to avoid an unnecessary copy. Performance is not really a concern here but it's just a good habit to take.

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
```diff
@@ -0,0 +1,200 @@
...
+        image->clear(fill_color);
+        
+        GenericImageFileWriter writer(image_file_path.c_str());
+        writer.append_image(image.get());
+        writer.write();
+    }
+    
+    TEST_CASE(WriteBMP_CorrectlyWritesImagePixels)
+    {
+        const std::string image_file_path = "unit tests/outputs/test_imagefilewriter.bmp";
```
dictoon: We spell constants in CamelCase.

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
```diff
@@ -0,0 +1,200 @@
...
+        const std::size_t image_size = 16;
```
dictoon: CamelCase, also mark as `static constexpr` (same in all test cases).

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
```diff
@@ -0,0 +1,200 @@
...
+                EXPECT_EQ(fill_color, c);
+            }
+        }
+    }
+    
+    TEST_CASE(WriteEXR_CorrectlyWritesImagePixels)
+    {
+        const std::string image_file_path = "unit tests/outputs/test_imagefilewriter.exr";
+        const std::size_t image_size = 16;
+        const Color4b fill_color(50, 100, 150, 42);
```
dictoon: For the OpenEXR format it would make sense to read and write floating point values since that's what the format natively supports.

src/appleseed/foundation/meta/tests/test_imagefilewriter.cpp
```diff
@@ -0,0 +1,200 @@
...
+        {
+            for (std::size_t x = 0; x < image_size; ++x)
+            {
+                Color3b c;
+                image->get_pixel(x, y, c);
+                
+                // the actual written and expected pixel value may be off by 1 in JPG format
+                for (std::size_t i = 0; i < 3; ++i)
+                {
+                    int difference = std::min(c[i]-fill_color[i], fill_color[i] - c[i]);
```
dictoon: Insert missing spaces around `-` operator.

Make const.

dictoon:
Also, as I suggested in https://github.com/appleseedhq/appleseed/pull/2779#discussion_r384052148, a short comment explaining the limitation where TIFF files must have a tile size be multiple of 16 would be useful.

PR #2783, opened 2020-02-27
Fix warning

Aaron19960821:
In previous implementation, use if(...) if(...) will cause warning when we build this project.

```
/home/yucwang/develop/appleseed/src/appleseed/renderer/modeling/bsdf/hairbsdf.cpp: In member function 'virtual void renderer::{anonymous}::HairBSDFImpl::sample(renderer::SamplingContext&, const void*, bool, bool, const renderer::BSDF::LocalGeometry&, const Dual3f&, int, renderer::BSDFSample&) const':
/home/yucwang/develop/appleseed/src/appleseed/renderer/modeling/bsdf/hairbsdf.cpp:601:83: warning: 'sin_theta_i' may be used uninitialized in this function [-Wmaybe-uninitialized]
             const float cos_theta_i = std::sqrt(std::max(0.0f, 1.0f - sin_theta_i * sin_theta_i));
```

We replace it with if(...) else if(...) to mitigate this issue.

### DIFF
--- a/src/appleseed/renderer/modeling/bsdf/hairbsdf.cpp
+++ b/src/appleseed/renderer/modeling/bsdf/hairbsdf.cpp
@@ -573,7 +573,7 @@ namespace
 
             // Sample longitudinal function for given component of scattering.
 
-            float sin_theta_i = 0.0;
+            float sin_theta_i;
             if (p == 0)
             {
                 sin_theta_i = sample_longitudinal(
@@ -590,7 +590,7 @@ namespace
                     values->m_precomputed.m_v * 0.25f,
                     sample_M);
             }
-            else if (p == 2)
+            else
             {
                 sin_theta_i = sample_longitudinal(
                     std::sin(theta_o_TRT),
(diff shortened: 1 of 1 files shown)


src/appleseed/renderer/modeling/bsdf/hairbsdf.cpp
```diff
@@ -582,15 +582,15 @@ namespace
...
+            else if (p == 1)
             {
                 sin_theta_i = sample_longitudinal(
                     std::sin(theta_o_TT),
                     std::cos(theta_o_TT),
                     values->m_precomputed.m_v * 0.25f,
                     sample_M);
             }
-            if (p == 2)
+            else if (p == 2)
```
oktomus: You can just do `else` here. Like this you don't need to do `sin_thetat_i = 0.0f` and `sin_theta_i` will always be initialized.

Aaron19960821: Thanks for your commenting, have resolved.

Issue #1182, opened 2016-12-28
Draw bucket highlights in inverted colors


dictoon:
To make sure they are visible when the rendered image is white.

aytekaman:
Are we still interested in this? It looks like a self-contained, small issue I can deal with quickly:)

aytekaman:
Yes it looks interesting. I will give it a try.

dictoon:
Awesome! Don't hesitate to describe your approach or let us know if you're stuck.

PR #2785, opened 2020-02-28
Tile highlights are now colored

laurelkeys:
Addresses https://github.com/appleseedhq/appleseed/issues/1182.

Added _thread index_ and _number of threads_ as arguments to `on_tile_begin` and `highlight_tile`.

From these values, one color can be uniquely selected per thread, using one of [Inigo Quilez](http://iquilezles.org/www/articles/palettes/palettes.htm)'s color palettes.

 (See https://www.shadertoy.com/view/wlKXDm for a demo)

### DIFF
--- a/src/appleseed.bench/mainwindow/rendering/qttilecallback.h
+++ b/src/appleseed.bench/mainwindow/rendering/qttilecallback.h
@@ -65,8 +65,8 @@ class QtTileCallback
         const renderer::Frame*  frame,
         const size_t            tile_x,
         const size_t            tile_y,
-        const size_t            thread_index = -1,
-        const size_t            nb_threads = -1) override;
+        const size_t            thread_index = ~size_t(0),
+        const size_t            thread_count = ~size_t(0)) override;
 
     void on_tile_end(
         const renderer::Frame*  frame,
--- a/src/appleseed.bench/mainwindow/rendering/qttilecallback.h
+++ b/src/appleseed.bench/mainwindow/rendering/qttilecallback.h
@@ -64,7 +64,9 @@ class QtTileCallback
     void on_tile_begin(
         const renderer::Frame*  frame,
         const size_t            tile_x,
-        const size_t            tile_y) override;
+        const size_t            tile_y,
+        const size_t            thread_index = -1,
+        const size_t            nb_threads = -1) override;
 
     void on_tile_end(
         const renderer::Frame*  frame,
--- a/src/appleseed.bench/mainwindow/rendering/qttilecallback.h
+++ b/src/appleseed.bench/mainwindow/rendering/qttilecallback.h
@@ -65,8 +65,8 @@ class QtTileCallback
         const renderer::Frame*  frame,
         const size_t            tile_x,
         const size_t            tile_y,
-        const size_t            thread_index = ~size_t(0),
-        const size_t            thread_count = ~size_t(0)) override;
+        const size_t            thread_index,
+        const size_t            thread_count) override;
 
     void on_tile_end(
         const renderer::Frame*  frame,
--- a/src/appleseed.bench/mainwindow/rendering/renderwidget.cpp
+++ b/src/appleseed.bench/mainwindow/rendering/renderwidget.cpp
@@ -176,7 +176,9 @@ namespace
 void RenderWidget::highlight_tile(
     const Frame&    frame,
     const size_t    tile_x,
-    const size_t    tile_y)
+    const size_t    tile_y,
+    const size_t    thread_index,
+    const size_t    nb_threads)
 {
     QMutexLocker locker(&m_mutex);
 
@@ -205,7 +207,16 @@ void RenderWidget::highlight_tile(
 
     // Draw a bracket around the tile.
     const size_t BracketExtent = 5;
-    const std::uint8_t BracketColor[3] = { 255, 255, 255 };
+    std::uint8_t BracketColor[3] = { 255, 255, 255 }; // white
+    if (thread_index >= 0) {
+        // Choose one color per thread (see https://www.shadertoy.com/view/wlKXDm).
+        float t = TwoPi<float>() * (thread_index / static_cast<float>(nb_threads));
+        float cos_t = std::cos(t);
+        float sin_t = std::sin(t);
+        BracketColor[0] = 128 + static_cast<std::uint8_t>(127.5f * cos_t);
+        BracketColor[1] = 128 + static_cast<std::uint8_t>(127.5f * (-0.5f * cos_t - 0.866f * sin_t));
+        BracketColor[2] = 128 + static_cast<std::uint8_t>(127.5f * (-0.5f * cos_t + 0.866f * sin_t));
+    }
     draw_bracket(
         dest,
         width,
--- a/src/appleseed.bench/mainwindow/rendering/renderwidget.cpp
+++ b/src/appleseed.bench/mainwindow/rendering/renderwidget.cpp
@@ -178,7 +178,7 @@ void RenderWidget::highlight_tile(
     const size_t    tile_x,
     const size_t    tile_y,
     const size_t    thread_index,
-    const size_t    nb_threads)
+    const size_t    thread_count)
 {
     QMutexLocker locker(&m_mutex);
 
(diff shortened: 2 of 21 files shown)


src/appleseed.bench/mainwindow/rendering/renderwidget.cpp
```diff
@@ -176,7 +176,9 @@ namespace
 void RenderWidget::highlight_tile(
     const Frame&    frame,
     const size_t    tile_x,
-    const size_t    tile_y)
+    const size_t    tile_y,
+    const size_t    thread_index,
+    const size_t    nb_threads)
```
oktomus: As far as I understand, only the `RenderWidget` needs `thread_index` and `nb_threads`. Have you tried to give those informations to the `RenderWidget` without affecting all tile callbacks (such as `stdout`) ?

src/appleseed.bench/mainwindow/rendering/renderwidget.cpp
```diff
@@ -205,7 +207,16 @@ void RenderWidget::highlight_tile(
 
     // Draw a bracket around the tile.
     const size_t BracketExtent = 5;
-    const std::uint8_t BracketColor[3] = { 255, 255, 255 };
+    std::uint8_t BracketColor[3] = { 255, 255, 255 }; // white
+    if (thread_index >= 0) {
+        // Choose one color per thread (see https://www.shadertoy.com/view/wlKXDm).
+        float t = TwoPi<float>() * (thread_index / static_cast<float>(nb_threads));
```
oktomus: We don't have much control over colors here. Have you considered creating a color palette and picking colors in it like this : 
`color_index = thread_index % nb_threads % color_palette_size` ? 
Like this we don't cast anything and we can choose the colors :)

oktomus: Moreover, you can simply drop the `nb_threads` value if you use a color palette.

src/appleseed/renderer/kernel/rendering/generic/tilejob.cpp
```diff
@@ -134,8 +134,9 @@ void TileJob::execute(const size_t thread_index)
             : nullptr;
 
     // Call the pre-render tile callback.
-    if (tile_callback)
-        tile_callback->on_tile_begin(&m_frame, m_tile_x, m_tile_y);
+    if (tile_callback) {
+        tile_callback->on_tile_begin(&m_frame, m_tile_x, m_tile_y, thread_index, m_tile_renderers.size());
```
oktomus: Be careful with naming, you used `nb_threads` everywhere but it's actually the number of tile renderer.

dictoon: That should be fine because there is always one tile renderer per thread. But yeah, it would be more explicit to pass the thread count.

src/appleseed.bench/mainwindow/rendering/renderwidget.cpp
dictoon: We use the "count" terminology throughout the code base, i.e. `thread_count` rather than `nb_threads`.

dictoon: `thread_index` can't be passed just to `RenderWidget` since by definition it varies per thread, and there's only one `RenderWidget`. It must be passed to `highlight_tile()`.

It's a different matter for `thread_count`, but if we're passing one, we can as well pass both.

src/appleseed.bench/mainwindow/rendering/renderwidget.cpp
```diff
@@ -205,7 +207,16 @@ void RenderWidget::highlight_tile(
 
     // Draw a bracket around the tile.
     const size_t BracketExtent = 5;
-    const std::uint8_t BracketColor[3] = { 255, 255, 255 };
+    std::uint8_t BracketColor[3] = { 255, 255, 255 }; // white
+    if (thread_index >= 0) {
```
dictoon: `thread_index` is a `std::size_t` so it's always >= 0.

src/appleseed.bench/mainwindow/rendering/renderwidget.cpp
```diff
@@ -205,7 +207,16 @@ void RenderWidget::highlight_tile(
 
     // Draw a bracket around the tile.
     const size_t BracketExtent = 5;
-    const std::uint8_t BracketColor[3] = { 255, 255, 255 };
+    std::uint8_t BracketColor[3] = { 255, 255, 255 }; // white
```
dictoon: Remove initialization if you're going to overwrite it immediately after.

src/appleseed.studio/mainwindow/rendering/renderwidget.cpp
```diff
@@ -231,7 +233,16 @@ void RenderWidget::highlight_tile(
 
     // Draw a bracket around the tile.
     const size_t BracketExtent = 5;
-    const std::uint8_t BracketColor[3] = { 255, 255, 255 };
+    std::uint8_t BracketColor[3] = { 255, 255, 255 }; // white
```
dictoon: Same remarks here.

In the meantime I'll try to refactor this code to share the `RenderWidget` class across appleseed.studio and appleseed.bench.

src/appleseed.bench/mainwindow/rendering/renderwidget.cpp
dictoon: As discussed in #dev, I think this is a reasonable solution for the time being.

src/appleseed/renderer/kernel/rendering/tilecallbackbase.h
```diff
@@ -64,7 +64,9 @@ class TileCallbackBase
     void on_tile_begin(
         const Frame*            frame,
         const size_t            tile_x,
-        const size_t            tile_y) override
+        const size_t            tile_y,
+        const size_t            thread_index = -1,
```
dictoon: Why assign default values? Moreover, `-1` will be converted to an unsigned value (`~0`) so if default values are really needed, I would rather use `~0`.

src/appleseed/renderer/kernel/rendering/generic/tilejob.cpp
laurelkeys: Should I add `thread_count` to the `TileJob` constructor and use it instead of `m_tile_renderers.size()`?

I'd then pass it to `TileJobFactory::create`, so we could use `m_thread_count` in here: 

https://github.com/appleseedhq/appleseed/blob/8053115547b9b804087586e3e8610eab42b1f6f4/src/appleseed/renderer/kernel/rendering/generic/genericframerenderer.cpp#L366-L374

src/appleseed/renderer/kernel/rendering/tilecallbackbase.h
dictoon: Remove default values as explained in #dev.

src/appleseed/renderer/kernel/rendering/generic/tilejob.cpp
dictoon: I think the way you did it (passing both thread index and thread count) is better.

dictoon: Ah sorry, I got confused:

> Should I add thread_count to the TileJob constructor and use it instead of m_tile_renderers.size()?
> I'd then pass it to TileJobFactory::create, so we could use m_thread_count in here

Yes, that sounds better indeed.

src/appleseed.bench/mainwindow/rendering/qttilecallback.h
```diff
@@ -64,7 +64,9 @@ class QtTileCallback
     void on_tile_begin(
         const renderer::Frame*  frame,
         const size_t            tile_x,
-        const size_t            tile_y) override;
+        const size_t            tile_y,
+        const size_t            thread_index = ~size_t(0),
+        const size_t            thread_count = ~size_t(0)) override;
```
oktomus: I think you can simply remove the default values.

src/appleseed/renderer/kernel/rendering/tilecallbackbase.h
```diff
@@ -64,7 +64,9 @@ class TileCallbackBase
     void on_tile_begin(
         const Frame*            frame,
         const size_t            tile_x,
-        const size_t            tile_y) override
+        const size_t            tile_y,
+        const size_t            thread_index = ~size_t(0),
+        const size_t            thread_count = ~size_t(0)) override
```
oktomus: Remove default values also here.

src/appleseed/renderer/kernel/rendering/generic/tilejob.cpp
laurelkeys: Fixed it 🙂

PR #2793, opened 2020-03-03
Move memory related classes from foundation/utility to memory

est77:


src/appleseed/foundation/array/arrayallocator.h
```diff
@@ -30,7 +30,7 @@
 
 // appleseed.foundation headers.
 #include "foundation/math/scalar.h"
-#include "foundation/utility/memory.h"
+#include "foundation/memory/memory.h"
```
dictoon: Maybe we could rename `memory.h`?

src/appleseed/foundation/platform/python.h
```diff
@@ -39,15 +39,15 @@
 #endif
 
 // Boost headers.
-#include "boost/python/detail/wrap_python.hpp"  // has to be first, to avoid redefinition warnings
 #include "boost/python.hpp"
+#include "boost/python/detail/wrap_python.hpp"  // has to be first, to avoid redefinition warnings
```
dictoon: This change should be reverted I think.

My guess is that you wrote a Python script to reorder the `#include` statements? ;) If so, could you share it?

src/appleseed/foundation/utility/minizip/ioapi.h
```diff
@@ -41,9 +41,9 @@
 
 #endif
 
+#include "zlib.h"
```
dictoon: Revert changes to third party files.

dictoon:
Thanks, looks great! You can merge once you've addressed the minor issues I brought up.

src/appleseed/foundation/platform/python.h
est77: I used the sort includes script in the repo. But it does not work very well in some cases.
For example it messes with _begin_xxx and _end_xxx headers.

src/appleseed/foundation/array/arrayallocator.h
est77: I prefer to leave it like this. having memory.h and memory/ with some other headers looks a bit arbitrary to me.

dictoon: I'm not sure to understand your point :) Mine was: now that we have a `memory` subfolder, `memory.h` isn't very descriptive anymore and maybe we could rename it. Or not. Just wondering.

PR #2796, opened 2020-03-07
Move foundation/utility/string.h to string/string.h

est77:


dictoon:
What else do you plan to move to `foundation/string/`?

est77:
> What else do you plan to move to `foundation/string/`?

interned strings for constant strings and maybe a compact string for use when storing object names and things like that. A string in gcc is 32 bytes and that sucks if you only want to store strings.

sandbox/examples/cpp/alembicassembly/alembicassembly.cpp
```diff
@@ -46,7 +46,7 @@
 #include "foundation/utility/api/specializedapiarrays.h"
 #include "foundation/utility/job/iabortswitch.h"
 #include "foundation/utility/otherwise.h"
-#include "foundation/utility/string.h"
+#include "foundation/string/string.h"
```
dictoon: It looks like you didn't run the sorting script?

Issue #1487, opened 2017-06-25
Emit warning when alpha map and volume are assigned to the same object instance


biart:
Users must be informed that the behaviour is undefined in the mentioned case.

Rajiv2605:
@Biart95 I would like to contribute to this. I am new to this repository as well as open source. It would be great if you could help me get started.

dictoon:
@Rajiv2605 Hello! If you're comfy with C++ (that's a prerequisite), we gladly accept contributions!

The very first step would be to build appleseed on your machine. Here is a detailed guide: https://github.com/appleseedhq/appleseed/wiki/Building-appleseed.

Don't hesitate to ask on our [appleseed-dev mailing list](https://groups.google.com/forum/#!forum/appleseed-dev) if you have questions or if you encounter issues.

Biart95:
@Rajiv2605 Hello! Welcome to Appleseed!

This issue is very good to start with. The problem is that alpha mapping in Appleseed just ignores the surface (as if it does not exist) with the given probability. It is not clear how we should treat participating media in aforementioned case. At the moment the behavior is not specified, and it is good to inform users about this.

We already have a similar warning about **EDF** + **alpha**, so I would recommend to find a place in the code where we already emit this warning and put new warning there.

Perhaps in the future we would like to treat surface alpha like it is done in some renderers: if there are material with **alpha**, **BSDF** and **Phase Function**, then with probability **alpha** we compute bounce according to **BSDF**, with probability **1 - alpha** we skip the bounce but proceed to raymarch the media inside. However, this approach needs to be discussed and we should understand if this will work correctly with BSSRDFs and transmissive BSDFs.

Rajiv2605:
@dictoon @Biart95 I am currently running Ubuntu 16.04 LTS in which I have g++ 5.4.0 and Qt 5.9. As it is mentioned in the docs that Appleseed can be built only with g++ 4.8 and Qt 4.x, is there anything I can do?

est77:
appleseed builds fine with gcc 5.4. Qt4 is required to build studio. But you can install qt4 from most package managers on linux.

Aaron19960821:
@dictoon When I investigate this issue, I found that we already have a warning message in
src/appleseed/renderer/modeling/material/genericmaterial.cpp:97.
I think we can close it now. :)

oktomus:
@Aaron19960821 The warning we already have is not for the same issue.

[...]

PR #2798, opened 2020-03-08
Emit warning when material is assigned both an alpha map and a volume

Aaron19960821:
To address #1487

### DIFF
--- a/src/appleseed/renderer/modeling/material/genericmaterial.cpp
+++ b/src/appleseed/renderer/modeling/material/genericmaterial.cpp
@@ -101,6 +101,7 @@ namespace
                     "this may lead to unexpected or unphysical results.",
                     context.get());
             }
+
             if  (m_render_data.m_volume && has_alpha_map())
             {
                 RENDERER_LOG_WARNING(
--- a/src/appleseed/renderer/modeling/material/genericmaterial.cpp
+++ b/src/appleseed/renderer/modeling/material/genericmaterial.cpp
@@ -101,6 +101,13 @@ namespace
                     "this may lead to unexpected or unphysical results.",
                     context.get());
             }
+            if  (m_render_data.m_volume && has_alpha_map())
+            {
+                RENDERER_LOG_WARNING(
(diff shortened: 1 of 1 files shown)


src/appleseed/renderer/modeling/material/genericmaterial.cpp
```diff
@@ -101,6 +101,13 @@ namespace
                     "this may lead to unexpected or unphysical results.",
                     context.get());
             }
+            if  (m_render_data.m_volume && has_alpha_map())
```
oktomus: Separate both `if` with a blank line.

Aaron19960821: Thanks, Done.

dictoon: Check `m_render_data.m_alpha_map` instead of using `has_alpha_map()`, as it's done in the EDF+alpha map case. Ultimately it should be the same thing, but there's no reason to be inconsistent.

src/appleseed/renderer/modeling/material/genericmaterial.cpp
```diff
@@ -102,6 +102,14 @@ namespace
                     context.get());
             }
 
+            if  (m_render_data.m_volume && has_alpha_map())
+            {
+                RENDERER_LOG_WARNING(
+                    "%smaterial is assigned with alpha map and volume at the same time: "
```
dictoon: *is assigned an*

Replace colon by semicolon, again in the name of consistency with the other similar message.

dictoon: Also, _a volume_.

src/appleseed/renderer/modeling/material/genericmaterial.cpp
Aaron19960821: done, thanks.

Aaron19960821:
Hi @dictoon I think this PR is ready for checking in, could you give it another review?

dictoon:
Looks good now, thanks.

PR #2799, opened 2020-03-09
Add interned string class to foundation

est77:


src/appleseed/CMakeLists.txt
```diff
@@ -765,6 +766,8 @@ source_group ("foundation\\resources\\logo" FILES
 set (foundation_string_sources
     foundation/string/string.cpp
     foundation/string/string.h
+    foundation/string/name.cpp
```

src/appleseed/foundation/string/name.h
```diff
@@ -0,0 +1,107 @@
...
+//
+
+#pragma once
+
+// appleseed.main headers.
+#include "main/dllsymbol.h"
+
+// Standard headers.
+#include <cstddef>
+#include <iostream>
```
dictoon: I think you only need `<ostream>`?

PR #2801, opened 2020-03-14
Add compute_tile_highlight to foundation/ColorMap

Mango-3:
- Adds the compute_tile_highlight color function to foundation/ColorMap and makes it dll exportable for usage with the plugins.

### DIFF
--- a/src/appleseed/foundation/image/colormap.cpp
+++ b/src/appleseed/foundation/image/colormap.cpp
@@ -36,6 +36,7 @@
 
 // Standard headers.
 #include <algorithm>
+#include <array>
 #include <cassert>
 #include <limits>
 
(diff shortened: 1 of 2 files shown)


src/appleseed/foundation/image/colormap.h
```diff
@@ -93,6 +96,10 @@ class ColorMap
 
     Color3f evaluate_palette(float x) const;
 
+    APPLESEED_DLLSYMBOL std::array<float,4> compute_tile_highlight_color(
```
dictoon: That's not exactly how I saw it:

First, I would suggest to rename the function since it's not really specific to tile highlights. Maybe `evaluate_cosine_palette()`? There are many more of those by the way: https://www.shadertoy.com/view/ll2GD3, in the future we could consider adding more as needed.

Also, I think this should be a free function instead of static method as it's not related to the `ColorMap` class.

Finally, it would be more logical to return a `Color3f`.

src/appleseed/foundation/image/colormap.h
```diff
@@ -93,6 +96,10 @@ class ColorMap
...
+        const size_t thread_index,
```
dictoon: If we rename the method, those should be renamed as well (`index`, `count` sound good enough).

src/appleseed/foundation/image/colormap.cpp
```diff
@@ -248,6 +249,23 @@ Color3f ColorMap::evaluate_palette(float x) const
     return lerp(m_palette[ix], m_palette[ix + 1], w);
 }
 
+std::array<float,4> ColorMap::compute_tile_highlight_color(const size_t thread_index, const size_t thread_count)
+{
+    // Choose one color per thread (see https://www.shadertoy.com/view/wlKXDm).
+    std::array<float,4> frame_color;
+
+    assert(thread_count != 0);
```
dictoon: Move first.

src/appleseed/foundation/image/colormap.cpp
```diff
@@ -248,6 +249,23 @@ Color3f ColorMap::evaluate_palette(float x) const
...
+
+    float t = 2 * 3.14159265359f * (thread_index / static_cast<float>(thread_count));
```
dictoon: Make all variables const.

PR #2803, opened 2020-03-15
Metal BRDF changes

est77:
- Implemented fresnel conductor variant from "Fresnel Equations Considered Harmful".
- Use the energy compensation variant from "Practical multiple scattering compensation for microfacet models"
- Update OSL shaders.

### DIFF
--- a/src/appleseed.shaders/src/appleseed/as_metal.osl
+++ b/src/appleseed.shaders/src/appleseed/as_metal.osl
@@ -62,6 +62,15 @@ shader as_metal
         string help = "Reflectance at grazing incidence.",
         int as_max_param_id = 2
     ]],
+    float in_edge_reflectance_weight = 0.0
+    [[
+        string as_maya_attribute_name = "edgeReflectanceWeight",
+        string as_maya_attribute_short_name = "f90w",
+        string label = "Edge Reflectance Weight",
+        string page = "Fresnel",
+        string help = "Edge reflectance weight.",
+        int as_max_param_id = 26
+    ]],
     int in_distribution = 0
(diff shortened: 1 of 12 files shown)


src/appleseed/renderer/modeling/bsdf/metalbrdf.cpp
```diff
@@ -299,25 +297,17 @@ namespace
         {
             if (values->m_energy_compensation != 0.0f)
             {
-                const float Ess =
-                    get_directional_albedo(
-                        std::abs(dot(outgoing, n)),
-                        values->m_roughness);
+                const float Ess = get_directional_albedo(
```
dictoon: `ess` (lower case)

What does it stand for?

src/appleseed.shaders/src/appleseed/as_metal.osl
```diff
@@ -62,6 +62,15 @@ shader as_metal
         string help = "Reflectance at grazing incidence.",
         int as_max_param_id = 2
     ]],
+    float in_edge_reflectance_weight = 0.0
```
dictoon: Shouldn't the default be 1?

est77: No. Should be 0 for compatibility with the metal roughness PBR workflow.

src/appleseed/renderer/modeling/bsdf/metalbrdf.cpp
est77: This comes from the original paper. E is for energy and ss for single scattering.

Issue #2754, opened 2019-12-28
Don't instanciate a TextureStore each time we pick something in the scene


oktomus:
We recently added printable settings on texture store (8d949848e376a8e93fb516dcaabdb7f7dfe88b2b). These settings are printed when the store is created. Thanks to this, we noticed a bug in **appleseed.studio**: each time the user wants to pick a material or an object, a new TextureStore is created.

The reason is that a `ScenePicker` is instantiated each time we call `pick()`. 

Because the picker is project based, we could just create it once.

Make sure to test in debug, and make sure to test picking during and after rendering, and after loading another project.

archisha-chandel:
![image](https://user-images.githubusercontent.com/47753303/75351503-46a5ce00-58ce-11ea-8517-dd73a6a9ff3a.png)
So, this should not be printed everytime an object is picked or selected?

oktomus:
@archisha-chandel The `store settings` block should not

achalpandeyy:
Hi!
Is anybody looking into this? I'd like to take a stab at it.

PR #2805, opened 2020-03-17
ScenePicker is no longer created everytime we pick something

achalpandeyy:
Addresses #2754 

Apart from the other regular instantiations of `TextureStore`, it is only created once when we pick an object the first time in the scene or drag and drop a material the first time in the scene.

I've tested the code with all the test cases mentioned in the issue.

If you could think of a more elegant solution than what I've done in `MainWindow::slot_open_cornellbox_builtin_project()`, please let me know! :)


src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -1450,6 +1450,11 @@ void MainWindow::slot_open_cornellbox_builtin_project()
     if (!can_close_project())
         return;
 
+    // Reset the ScenePickers for certain handlers, so that they don't hold a reference to
+    // Lazy<Tree> and consequently it can be destroyed without obstruction.
+    for (const_each<RenderTabCollection> i = m_render_tabs; i; ++i)
+        i->second->update_handlers();
```
oktomus: This snipped of code will make your code work only for the cornellbox project. There is a method `on_project_change` in this class, check it out. I also think all tabs are recreated when the project changed.

achalpandeyy: Actually, loading of the built-in Cornell Box project seems like a _special_ case, where, we _first_ load the new project on the [line](https://github.com/appleseedhq/appleseed/blob/1a0907c5dc3c491686cb2be4b3d4f9410965ba35/src/appleseed.studio/mainwindow/mainwindow.cpp#L1458), and _then_ call `on_project_change` on the [line](https://github.com/appleseedhq/appleseed/blob/1a0907c5dc3c491686cb2be4b3d4f9410965ba35/src/appleseed.studio/mainwindow/mainwindow.cpp#L1461) which, as you said recreates all the `RenderTab`s,

Whereas, in other projects (loaded via the Open Project menu item in appleseed.studio), we _first_ remove the `RenderTab`s on the [line](https://github.com/appleseedhq/appleseed/blob/1a0907c5dc3c491686cb2be4b3d4f9410965ba35/src/appleseed.studio/mainwindow/mainwindow.cpp#L238) and _then_ we load in the new project on the [line](https://github.com/appleseedhq/appleseed/blob/1a0907c5dc3c491686cb2be4b3d4f9410965ba35/src/appleseed.studio/mainwindow/mainwindow.cpp#L245).

I hope I'm making myself clear.

As it stands now, loading in other projects (not just the built-in Cornell Box) in any order works fine, I've tested it myself. Is there a way you can pull the changes and see for yourself, or maybe I can record a video?

The main purpose of the snippet above is to bypass [this](https://github.com/appleseedhq/appleseed/blob/master/src/appleseed/foundation/utility/lazy.h#L369) line (`Lazy<renderer::TriangleTree>`)

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -300,6 +300,7 @@ bool MainWindow::pack_project(QString filepath)
 
 void MainWindow::close_project()
 {
+    remove_render_tabs();
```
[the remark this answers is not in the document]
achalpandeyy: This is done to remove the reference to `Lazy<renderer::TriangleTree>` before it gets destroyed (as the project is getting destroyed by the next line `m_project_manager.close_project()`) so that [this](https://github.com/appleseedhq/appleseed/blob/master/src/appleseed/foundation/utility/lazy.h#L369) `assert` don't go off.

_Why is there a reference to `Lazy<renderer::TriangleTree>` ?_
`RenderTab` has a `ScenePicker` member.
`ScenePicker` has an `Intersector` member.
`Intersector` has a `TriangleTreeAccessCache` member.
`TriangleTreeAccessCache` is defined as,
```
typedef foundation::AccessCacheMap<
    TriangleTreeContainer,
    TriangleTreeAccessCacheLines,
    TriangleTreeAccessCacheWays,
    foundation::PoolAllocator<void, TriangleTreeAccessCacheLines * TriangleTreeAccessCacheWays>
> TriangleTreeAccessCache;
```
Note that it has a `TriangleTreeContainer`, which is defined as,
```
typedef std::map<
    foundation::UniqueID,
    foundation::Lazy<TriangleTree>*
> TriangleTreeContainer;
```
Note that it has `foundation::Lazy<TriangleTree>*` which doesn't let the `foundation::Lazy<TriangleTree>` to get destroyed, because the aforementioned `assert` goes off.

For more clarification, I'm including the call stack of what happens if I __remove__ the call to `remove_render_tabs`

callstack

I hope it makes things clearer. As usual, if I'm doing something wrong please let me know. :)

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -1450,6 +1447,8 @@ void MainWindow::slot_open_cornellbox_builtin_project()
     if (!can_close_project())
         return;
 
+    remove_render_tabs();
```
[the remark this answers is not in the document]
achalpandeyy: Please see the above similar conversation.

src/appleseed.studio/mainwindow/mainwindow.cpp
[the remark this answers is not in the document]
oktomus: Wow, great explanations !

src/appleseed.studio/mainwindow/rendering/materialdrophandler.cpp
```diff
@@ -43,24 +43,29 @@ namespace appleseed {
...
+    const renderer::ScenePicker&    scene_picker)
+  : m_project(project), m_rendering_manager(rendering_manager), m_enabled(false), m_scene_picker(scene_picker)
 {
 }
 
+void MaterialDropHandler::set_enabled(const bool enabled)
+{
+    m_enabled = enabled;
+
+    m_project.update_trace_context();
```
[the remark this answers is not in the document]
achalpandeyy: This will always work because this will only happen once the `AssemblyTree` is built.
`m_project.update_trace_context();` will synchronize the `TraceContext` so that `Intersector` can use it in `ScenePicker::pick`.

The warning which I removed checked if there exists a `TraceContext` at all.
Since we are constructing a `ScenePicker` with the `RenderTab`, a `TraceContext` will come into existence with the constructor of `ScenePicker` even if the `AssemblyTree` is _not_ yet built.

src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
```diff
@@ -523,6 +528,11 @@ void RenderingManager::slot_camera_change_end()
 {
 }
 
+void RenderingManager::slot_assembly_tree_built()
+{
+    m_render_tab->enable_scene_picking();
```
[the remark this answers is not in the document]
achalpandeyy: Actually, name of the method is a bit misleading.
There's nothing to enable in the `ScenePicker`.
This method enables the _handlers_ which use the `ScenePicker`, namely `ScenePickingHandler` and `MaterialDropHandler`.

Unless the new picker for light paths uses the `ScenePickingHandler` (or `MaterialDropHandler`) I think it should _not_ be a problem.

I agree though, the name is misleading, do you think `refresh_scene_picker` makes more sense in this context?

src/appleseed.studio/mainwindow/mainwindow.cpp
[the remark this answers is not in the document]
dictoon: Is this change still relevant?

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -899,10 +900,6 @@ void MainWindow::set_rendering_widgets_enabled(const bool is_enabled, const Rend
             // Set/clear rendering region.
             render_tab->set_render_region_buttons_enabled(
                 is_enabled && is_project_open && rendering_mode != RenderingMode::FinalRendering);
-
```
dictoon: Not sure why you removed this?

src/appleseed.studio/mainwindow/mainwindow.cpp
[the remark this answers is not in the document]
dictoon: Same question, is this still relevant?

src/appleseed.studio/mainwindow/rendering/materialdrophandler.cpp
```diff
@@ -43,20 +45,22 @@ namespace appleseed {
...
-    const QString&          material_name)
+    const foundation::Vector2d&     drop_pos,
+    const QString&                  material_name)
 {
     m_drop_pos = drop_pos;
     m_material_name = material_name.toStdString();
 
-    if (!m_project.has_trace_context())
+    // Tell the user to render once, if assembly tree hasn't been built yet.
+    if (m_project.get_trace_context().get_assembly_tree().get_nodes().size() == 0)
```
dictoon: Why this change? What's the problem with `has_trace_context()`?

src/appleseed.studio/mainwindow/rendering/scenepickinghandler.cpp
```diff
@@ -225,7 +229,8 @@ namespace
 
 ItemBase* ScenePickingHandler::pick(const QPoint& point)
 {
-    if (!m_project.has_trace_context())
+    // Tell the user to render once, if assembly tree hasn't been built yet.
+    if (m_project.get_trace_context().get_assembly_tree().get_nodes().size() == 0)
```
dictoon: Same question here.

src/appleseed.studio/mainwindow/mainwindow.cpp
oktomus: This line was here to prevent the user from selecting things or change materials **while it's rendering**

src/appleseed.studio/mainwindow/mainwindow.cpp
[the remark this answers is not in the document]
achalpandeyy: Yes, it'll be relevant as long as we have `ScenePicker` as a member of `RenderTab`.

src/appleseed.studio/mainwindow/mainwindow.cpp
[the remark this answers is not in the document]
achalpandeyy: Yes, see above.

src/appleseed.studio/mainwindow/rendering/materialdrophandler.cpp
achalpandeyy: `m_project.has_trace_context()` will always return true, since it gets created with the `ScenePicker` which in turn gets created with the `RenderTab`

src/appleseed.studio/mainwindow/rendering/scenepickinghandler.cpp
achalpandeyy: See above.

src/appleseed.studio/mainwindow/mainwindow.cpp
achalpandeyy: Removal of this snippet wasn't intended, it made sense with the previous iteration of the PR.
I'll add it back in.

> This line was here to prevent the user from selecting things or change materials while it's rendering
This isn't why I removed it. Also, it wasn't intended behaviour.
This wasn't affect material dropping at all.

achalpandeyy:
@oktomus Did you see the above conversation with @dictoon regarding this issue?

oktomus:
@achalpandeyy Yes I did. But by user should not be able to "pick" while its rendering. And if the user picks on an empty render and it doesn't find anything, that's ok.

If you leave the previous `has_trace_context`, do you ever get any errors ?

achalpandeyy:
> If you leave the previous `has_trace_context`, do you ever get any errors ?



> @achalpandeyy Yes I did. But the user should not be able to "pick" while its rendering. And if the user picks on an empty render and it doesn't find anything, that's ok.
> 
> If you leave the previous `has_trace_context`, do you ever get any errors ?

You would get a `Read Access Violation` if you put `has_trace_context` there because you try to pick when `AssemblyTree` isn't created.

achalpandeyy:
`TextureStore` is created only one time per `Project`, just before the rendering is started for the first time.
Rest of the system holds references to the `TextureStore`.

src/appleseed.studio/mainwindow/rendering/materialdrophandler.h
```diff
@@ -29,6 +29,7 @@
 #pragma once
 
 // appleseed.renderer headers.
+#include "renderer/api/scenepicker.h"
```
oktomus: You don't need that include here.

src/appleseed.studio/mainwindow/rendering/materialdrophandler.cpp
```diff
@@ -33,6 +33,8 @@
 #include "mainwindow/rendering/renderingmanager.h"
 
 // appleseed.renderer headers.
+#include "renderer/kernel/intersection/assemblytree.h"
+#include "renderer/kernel/intersection/tracecontext.h"
```
oktomus: Looks like these are unused

src/appleseed/foundation/math/bvh/bvh_tree.h
```diff
@@ -58,6 +58,9 @@ class Tree
     // Return the size (in bytes) of this object in memory.
     size_t get_memory_size() const;
 
+    // Return the nodes in the tree.
+    const NodeVector& get_nodes() const;
```
oktomus: Is suggest adding `is_empty` instead.

src/appleseed/renderer/modeling/project/project.cpp
```diff
@@ -126,6 +127,7 @@ struct Project::Impl
     LightPathRecorder                   m_light_path_recorder;
     std::unique_ptr<TraceContext>       m_trace_context;
     RenderingTimer                      m_rendering_timer;
+    std::unique_ptr<TextureStore>       m_texture_store = nullptr;
```
oktomus: This is `nullptr` by default. No need to do `= nullptr`.

src/appleseed.studio/mainwindow/rendering/rendertab.cpp
```diff
@@ -498,7 +498,6 @@ void RenderTab::recreate_handlers()
     // Set initial state.
     m_pixel_inspector_handler->set_enabled(false);
     m_camera_controller->set_enabled(false);
-    m_scene_picking_handler->set_enabled(true);
```
oktomus: Have you tested different cases to ensure that the scene picker is enabled when it should ? 
- Start/stop rendering
- Interactive rendering
- Abort rendering
- Open a new project, start a new project

src/appleseed.studio/mainwindow/rendering/scenepickinghandler.cpp
```diff
@@ -48,6 +48,8 @@
 #include "renderer/api/project.h"
 #include "renderer/api/scene.h"
 #include "renderer/api/surfaceshader.h"
+#include "renderer/kernel/intersection/assemblytree.h"
+#include "renderer/kernel/intersection/tracecontext.h"
```
oktomus: Not required

src/appleseed/foundation/math/bvh/bvh_tree.h
oktomus: And i think you don't use it anymore do you ? If not, remove this

src/appleseed/renderer/modeling/project/project.cpp
```diff
@@ -436,6 +435,17 @@ void Project::on_frame_end(
     impl->m_rendering_timer.measure();
 }
 
+TextureStore& Project::get_texture_store(const ParamArray& params) const
+{
+    if (!impl->m_texture_store)
+    {
+        assert(impl->m_scene.get());
+        impl->m_texture_store.reset(new TextureStore(*impl->m_scene, params));
```
oktomus: Something looks wrong here (tell me if I'm wrong). 

The texture store is cached in `m_texture_store`. But to create a texture_store, you need to give it some parameters. What if `get_texture_store` is called with different parameters ? 

@dictoon What do you think ?

src/appleseed/renderer/modeling/project/project.h
```diff
@@ -192,6 +194,9 @@ class APPLESEED_DLLSYMBOL Project
         const Project&                  project,
         const BaseGroup*                parent) override;
 
+    // Get the texture store.
+    TextureStore& get_texture_store(const ParamArray& params = ParamArray()) const;
```
oktomus: No need for a default value.

src/appleseed/renderer/meta/tests/test_tracer.cpp
```diff
@@ -228,8 +228,8 @@ TEST_SUITE(Renderer_Kernel_Lighting_Tracer)
         Tracer                                  m_tracer;
 
         Fixture()
-          : m_trace_context(FixtureParams::FixtureBaseClass::m_scene)
-          , m_texture_store(FixtureParams::FixtureBaseClass::m_scene)
+          : m_trace_context(FixtureParams::FixtureBaseClass::m_project)
+          , m_texture_store(FixtureParams::FixtureBaseClass::m_project.get_texture_store())
```
oktomus: Here for example, the texture store was initially created without any parameters. With the changed you made to project, we end up having the same texture_store everywhere. Isn't it a problem ?

src/appleseed/renderer/kernel/intersection/tracecontext.h
```diff
@@ -54,7 +57,7 @@ class APPLESEED_DLLSYMBOL TraceContext
 {
   public:
     // Constructor, initializes the trace context for a given scene.
-    explicit TraceContext(const Scene& scene);
+    explicit TraceContext(const Project& project);
```
oktomus: Why this change ? If I understand correctly, we only need the scene for `get_scene`.

src/appleseed/renderer/kernel/intersection/assemblytree.cpp
```diff
@@ -77,9 +77,9 @@ namespace renderer
 // AssemblyTree class implementation.
 //
 
-AssemblyTree::AssemblyTree(const Scene& scene)
+AssemblyTree::AssemblyTree(const Project& project)
   : TreeType(AlignedAllocator<void>(System::get_l1_data_cache_line_size()))
-  , m_scene(scene)
+  , m_project(project)
```
oktomus: We just need the scene here. Why this change ?

oktomus: I think it's to TriangleTree the project so that it can find the texture store. If so, no need to change the members of AssemblyTree, just do `m_scene(scene)` or `m_scene(project.get_scene())`.

achalpandeyy:
> Thanks @achalpandeyy ! Looks great :)
> 
> There is one thing that I'm concerned about. Previously, we used to have multiple texture store (I saw one with 0 parameters and 1 with some paramters). And now, we have the same texture store everywhere and we are not sure which parameters are used. I think that's problematic but I don't know how TextureStore are used.
> 
> @dictoon Can you have a look on this ?

@oktomus, this is a vaild point. I'd love to test it.

Do you know of any test scenes which use the "texture_store" param? I couldn't find any.

src/appleseed.studio/mainwindow/rendering/rendertab.cpp
achalpandeyy: Yes I've tested with all the cases.

If you look closely you would find that this line is redundant because just a couple of lines above when we create a `ScenePickingHandler`, the `m_enabled` member of it is set to true (in constructor).

src/appleseed/renderer/modeling/project/project.cpp
achalpandeyy: Seems like you're right, I'll remove it.

src/appleseed/foundation/math/bvh/bvh_tree.h
achalpandeyy: Nice catch, we don't want it anymore.

src/appleseed/renderer/kernel/intersection/assemblytree.cpp
achalpandeyy: Doesn't keeping both `Scene` and `Project` as members redundant? I mean `Scene` could be easily accessed from `m_project`, so why keep it?

src/appleseed/renderer/kernel/intersection/tracecontext.h
achalpandeyy: We need `Project` here to pass it down to `AssemblyTree` -- all this to, as you've guessed in one of the comments, to provide `TriangleTree` with the `TextureStore`.

> If I understand correctly, we only need the scene for get_scene.
I'm not sure what do you mean by this.

src/appleseed/renderer/meta/tests/test_tracer.cpp
achalpandeyy: What I could gather from my conversation with @dictoon on Discord is that,

Everywhere we used to create a `TextureStore` with default value of `ParamArray`, is for one-time use only and typically loads a single texture, so the params doesn't matter.

But still, I would like @dictoon to verify this.

oktomus:
> Do you know of any test scenes which use the "texture_store" param? I couldn't find any.
Have you tried `grep store` or something inside the test scenes folder ?

achalpandeyy:
> > Do you know of any test scenes which use the "texture_store" param? I couldn't find any.
> 
> Have you tried `grep store` or something inside the test scenes folder ?

I haven't tried running `grep` to find it.

Again, from my conversation with @dictoon on Discord, I've found that no scene is supposed to set these are params. They are characteristic of the machine and not the scene.

There is an option to force-set these params in appleseed.studio's "Rendering Settings". In the event someone actually does use this, a `TextureStore` would be properly initialized with these params just before rendering starts for the first time (in `MainWindow::start_rendering`), then the remaining system uses this `TextureStore`.

src/appleseed/renderer/modeling/project/project.cpp
achalpandeyy: This is no longer relevant.

src/appleseed/renderer/kernel/intersection/assemblytree.cpp
dictoon: Storing a reference to the project in `AssemblyTree` is the right solution.

src/appleseed.studio/mainwindow/mainwindow.cpp
```diff
@@ -1254,6 +1254,9 @@ void MainWindow::start_rendering(const RenderingMode rendering_mode)
         rendering_mode == RenderingMode::InteractiveRendering ? "interactive" : "final";
     const ParamArray params = get_project_params(configuration_name);
 
+    if (!project->has_texture_store())
```
dictoon, line 1257: Having to manually initialize the texture store before a render is a big problem. In fact, you haven't updated all the other places where we start a render, meaning that rendering there is now broken (appleseed.cli, appleseed.bench, Max plugin, Maya plugin, Blender plugin, Gaffer integration and possibly many others).

dictoon:
Thanks for your work @achalpandeyy! I've thoroughly checked this PR, and overall it looks very good. I replaced a few `#include` by forward declarations where this had become possible.

My main issue now has to do with the new requirement (if I'm not mistaken) that the texture store must be manually initialized before starting a render. See my comment on that. We need to find a solution that doesn't require doing any manual initialization.

dictoon:
It seems that the unit tests don't pass, possibly for the same reason (texture store not initialized).

src/appleseed.studio/mainwindow/mainwindow.cpp
achalpandeyy: One way to not manually initialize the TextureStore is to treat it analogous to TraceContext i.e. initialize the first time you need it and then use a reference everywhere else.

I'll try to implement this solution.

If you have something better in mind, please let me know.

Issue #2797, opened 2020-03-08
Explicitly give signals and slots when calling `connect`


oktomus:
Instead of doing 
```cpp
connect(
        m_viewport_widget, SIGNAL(customContextMenuRequested(const QPoint&)),
        SLOT(slot_viewport_widget_context_menu(const QPoint&)));
```
do
```cpp
connect(
        m_viewport_widget, &ViewportWidget::customContextMenuRequested,
        this, &FinalRenderViewportTab::slot_viewport_widget_context_menu);
```

This allow to detect `connect` errors at compile time instead of runtime. 
Also, navigating to signals or slots in Visual Studio is easier.

For signals or slots that have multiple arg possibilities, you can specify which one to chose with `qOverload`. Example, this 
```cpp
connect(
        m_display_transform_combo, SIGNAL(currentIndexChanged(QString)),
        m_viewport_widget, SLOT(slot_display_transform_changed(QString)));
```
become 
```cpp
connect(
        m_display_transform_combo, qOverload<const QString&>(&QComboBox::currentIndexChanged),
        m_viewport_widget, &ViewportWidget::slot_display_transform_changed);
```

LeaBroudo:
I'd like to work on this issue, as I'm new to appleseed. Are there any particular files I can find this in? grep is taking a bit too long.

oktomus:
@LeaBroudo run grep in a specific folder. For instance `src/appleseed.studio`.

LeaBroudo:
Got it! I'll start on it now.

PR #2809, opened 2020-03-27
Switch to modern connect syntax in appleseed.bench & appleseed.qtcommon

LeaBroudo:
My first pass on the "Explicitly give signals and slots when calling `connect` #2797" issue. I've made all of the signals and slots explicit in /src/appleseed.qtcommon/ and  /src/appleseed.bench/.


src/appleseed.qtcommon/widgets/doubleslider.cpp
```diff
@@ -41,15 +41,19 @@ namespace
 DoubleSlider::DoubleSlider(QWidget* parent)
   : QSlider(parent)
 {
-    connect(this, SIGNAL(valueChanged(int)), this, SLOT(setValue(int)));
+    connect(
+        this, &DoubleSlider::valueChanged, 
+        this, static_cast<void (DoubleSlider::*)(const int)>(&DoubleSlider::setValue));
```
oktomus: I'm not sure about this one. Why you cast it ?

LeaBroudo: For this one, when I tried to just use "qOverload" as you recommended, I kept getting an error saying "error: address of overloaded function 'setValue' does not match required type 'qOverload<const int>'". I found this error was because my version of Qt was before 5.6, (see https://stackoverflow.com/questions/16794695/connecting-overloaded-signals-and-slots-in-qt-5). When I built appleseed there was no mention of a minimum Qt version, so I assumed I should make the code work for more Qt versions than less.

LeaBroudo: Should I implement it instead for only Qt 5.6 and above?

oktomus: appleseed supports Qt 5.12 minimum, but I think we forgot to mention it on the wiki for mac devices :/

LeaBroudo: Ah ok, then if appleseed wants to keep supporting a minimum of Qt 5.12, I should leave the static cast. Should I leave a comment in the code that mentions this?

LeaBroudo: Also should I squash the commits?

oktomus: > Ah ok, then if appleseed wants to keep supporting a minimum of Qt 5.12, I should leave the static cast. Should I leave a comment in the code that mentions this?

I thought the static cast was here because qOverload wasn't working on your machine ? It should work with Qt 5.12 so we shouldn't need any static cast.

> Also should I squash the commits?

If you want yes, but we can also do it when merging so it's not required :)

LeaBroudo:
I updated `doubleslider.cpp` to use `QOverload`. I wasn't able to use `qOverload`, because that method only works for Qt 5.7 and later. `QOverload<>::of` does the same thing, and works for earlier Qt versions (plus it's not ugly like the static casts!)

dictoon:
Thanks @LeaBroudo for your contribution!

To be clear, Qt 5.12 is much newer than Qt 5.6 (12 > 6), not to be mixed up with Qt 5.1.2.

We indeed target Qt 5.12 as a minimum version, but as you noticed appleseed currently builds fine with older versions. The minimum version is stated in the Windows build instructions but not the Linux or macOS ones, our bad.

Issue #2284, opened 2018-11-01
Print more runtime details in Python utilities


dictoon:
Print:
- Date/Time
- Python version
- Path to script
- Script version

MahidharBandaru:
Hi! I would like to take up this issue. By saying Python utilities are you referring to the scripts in this directory?
[scripts](https://github.com/appleseedhq/appleseed/tree/master/scripts)

jsxs0:
I believe, **date/time** can be added by 
`import datetime
now = datetime.datetime.now()
print
print "Date/time:"
print str(now)
print
print "Current date and time:"
print "Current year: %d" % now.year
print "Current month: %d" % now.month
print "Current day: %d" % now.day
print "Current hour: %d" % now.hour
print "Current minute: %d" % now.minute
print "Current second: %d" % now.second
print "Current microsecond: %d" % now.microsecond
print
print "Current date and time using strftime:"
print now.strftime("%Y-%m-%d %H:%M")
print
print "Current date and time using isoformat:"
print now.isoformat()
`
If this piece of code is added in the code debugger, we can probably see the date/time
**Python version**
```
import sys
print sys.version #from Script
```
**Path to script**
`os.path.realpath(__file__)`
OR, I could also use
`import os
print os.__file__`
**Script version**
```
import platform;

print(platform.sys.version);
```
Or, a simpler version could be:
`import sys
print sys.version`

dictoon:
@jsxs0 Please don't post tutorials about basic Python functions, these information are available in abundance on the web.

dictoon:
@MahidharBandaru Yes. Two scripts that would benefit from these information are `appleseed.package.py` and `runtestsuite.py` (both in the console and the HTML report).

tyxchen:
I can work on this issue.

Out of curiosity, what is meant by "script version"? Would the Git hash of the commit where the script was last changed suffice?

PR #2810, opened 2020-03-28
Print additional runtime details in Python utilities

tyxchen:
This PR resolves issue #2284.

In addition to adding printing of the runtime details to the console, I've also added printing of the details to the HTML reports generated by runtestsuite.py and the log generated by appleseed.benchmark. I've also slightly refactored appleseed.benchmark to use argparse for parsing arguments, and added version numbers to all of the utility scripts.

### DIFF
--- a/scripts/appleseed.benchmark.py
+++ b/scripts/appleseed.benchmark.py
@@ -28,11 +28,15 @@
 # THE SOFTWARE.
 #
 
+from __future__ import print_function
+import argparse
 import datetime
 import os
 import re
 import subprocess
 import sys
+
+from utils import print_runtime_details  # local module
 
 
 # -------------------------------------------------------------------------------------------------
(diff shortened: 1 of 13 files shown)


scripts/oslextractmeta.py
```diff
@@ -337,16 +345,16 @@ def cli():
...
-        The JSON dictionary consists of a 'header' and a 'shader' part.
-        jsondict['shader'] will return a dictionary with all shaders. The
-        user interface of the shader is stored as a sub-dictionary, the
-        metadata can be retrieved using the 'ui' key on the elements, e.g.:
-
-        for x in jsondict['shaders'].values():
-            print x['ui']
-        ''')
+oslextractmetadata stores the user interface and metadata of a
+compiled OSL (openshadinglanguage) shader(s) into a JSON file.
```
oktomus, line 349: You will end up with `of acompiled OSL` here

tyxchen: Wouldn't the newline be preserved in a triple-quoted string?

oktomus: Good point. I honestly don't know ^^

tyxchen: Just confirmed, the newline is printed in the help text.

tyxchen:
@oktomus Is there anything more that needs to be fixed in this PR?

scripts/appleseed.benchmark.py
```diff
@@ -28,12 +28,16 @@
...
 
+from __future__ import print_function
+import argparse
 import datetime
 import os
 import re
 import subprocess
 import sys
 
+from utils import print_runtime_details  # local module
```
dictoon, line 39: Nice!

PR #2816, opened 2020-03-31
Fix compilation and regressions

dictoon:


src/appleseed.qtcommon/widgets/doubleslider.cpp
```diff
@@ -51,7 +54,7 @@ DoubleSlider::DoubleSlider(QWidget* parent)
 DoubleSlider::DoubleSlider(const Qt::Orientation orientation, QWidget* parent)
   : QSlider(orientation, parent)
 {
-   connect(
+    connect(
         this, &DoubleSlider::valueChanged, 
         this, QOverload<const int>::of(&DoubleSlider::setValue));
```
LZaw: The compiler complains about this `QOverload` and the one in line 49.

src/appleseed.studio/mainwindow/rendering/qttilecallback.cpp
```diff
@@ -64,8 +65,8 @@ namespace
           : m_render_widget(render_widget)
         {
             connect(
-                this, SIGNAL(signal_update()),
-                m_render_widget, SLOT(update()),
+                this, &QtTileCallback::signal_update,
+                m_render_widget, QOverload<void>::of(&QWidget::update),
```
LZaw: Given that `QOverload` causes problems in `doubleslider.cpp`, I'm sure this one will do so as well (the compiler didn't get here yet, so we can't know for sure).

src/appleseed.bench/mainwindow/rendering/qttilecallback.cpp
```diff
@@ -54,7 +54,7 @@ QtTileCallback::QtTileCallback(RenderWidget* render_widget)
 {
     connect(
         this, &QtTileCallback::signal_update,
-        m_render_widget, &QWidget::update,
+        m_render_widget, QOverload<void>::of(&QWidget::update),
```
LZaw: Same argument as below.

PR #2824, opened 2020-04-04
Optimize OSL shaders before on_render_begin is called

est77:
This will allow building the importance map for OSL environments in on_render_begin.


dictoon:
Thanks!

PR #2831, opened 2020-04-06
Various improvements in appleseed.bench

dictoon:


src/appleseed.bench/mainwindow/mainwindow.cpp
```diff
@@ -106,26 +107,24 @@ namespace bench {
 // MainWindow class implementation.
 //
 
-#ifndef NDEBUG
-#define FAST_DEBUG_MODE
+// Use a simpler scene and lower rendering quality in debug builds.
+#ifdef APPLESEED_DEBUG
+    #define FAST_DEBUG_MODE
```
oktomus, line 112: 🏆

oktomus: I didn't knew we had this kind of feature :o

dictoon: Without it running appleseeed.bench in debug is not an option...

PR #2832, opened 2020-04-06
Enable clang 8 for Linux on Travis

LZaw:


src/appleseed/renderer/kernel/rendering/progressive/samplegeneratorjob.cpp
```diff
@@ -152,6 +152,8 @@ void SampleGeneratorJob::execute(const size_t thread_index)
         acquired_sample_count,
         abortable ? "" : " (non-abortable)",
         pretty_time(t2 - t1).c_str());
+#else
+    (void)(m_job_index);
```
est77, line 156: Can't we use APPLESEED_UNUSED here?

LZaw: @FrancoisBeaune tried that in
https://github.com/appleseedhq/appleseed/commit/c7a02cba95b9c4009fb0c25bce221e22d8c8e751
But it caused some problems, which is why he went with this solution. The commit containing this change is part of #2834 which already got merged.
I just fetched it over here, to test it on the new setup.

PR #2838, opened 2020-04-11
Ray differentials fixes and improvements.

est77:


src/appleseed/renderer/kernel/shading/shadingpoint.cpp
```diff
@@ -579,6 +579,29 @@ void ShadingPoint::compute_world_space_partial_derivatives() const
     }
 }
 
+namespace
+{
+
+template <typename T>
```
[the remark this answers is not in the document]
dictoon: Also `intersect_ray_plane()` would be a little more consistent.

src/appleseed/renderer/modeling/bsdf/bsdfsample.cpp
```diff
@@ -40,13 +40,48 @@ using namespace foundation;
 
 namespace renderer
 {
+namespace
+{
+
+void compute_normal_derivatives(
```
dictoon, line 46: Indent content of anonymous namespaces.

src/appleseed/renderer/kernel/shading/shadingpoint.cpp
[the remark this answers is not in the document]
est77: done

src/appleseed/renderer/modeling/bsdf/bsdfsample.cpp
est77: done

src/appleseed/renderer/modeling/bsdf/diffusebtdf.cpp
```diff
@@ -145,6 +145,7 @@ namespace
                 sample.m_value.m_beauty = sample.m_value.m_diffuse;
 
                 sample.m_min_roughness = 1.0f;
+                sample.compute_diffuse_differentials(outgoing);
```
dictoon, line 148: Given that we're dealing with transmission here, are the differentials correct? Just wondering about the lack of reflected vs transmitted versions of the diffuse case.

est77: Diffuse differentials are always 1/25 of the sphere in the direction of the outgoing vector.
There is no special case for transmitted diffuse differentials.

PR #2840, opened 2020-04-13
Run test scenes when building master

oktomus:
The wiki explaining how it work is available [here](https://github.com/appleseedhq/appleseed/wiki/Custom-continuous-integration).

- Test report is available [here](http://107.170.222.178:8181/)
- Job log is available [here](http://107.170.222.178:8181/job.log)
- Previous test report is available [here](http://107.170.222.178:8181/last_report.html)
- If the job is running, [this page](http://107.170.222.178:8181/running.html) will be answering

Another Travis job in Ship mode was added to the jobs list. We should consider removing some jobs to use less computer power.

The process to run the test scenes is the following:
1. Wait for the Ship build job to be done
2. Send the build to a custom server
3. Run tests every 24h with a crontab

### DIFF
--- a/scripts/travis/deploy-linux-build.sh
+++ b/scripts/travis/deploy-linux-build.sh
@@ -40,15 +40,15 @@ THISDIR=`pwd`
 echo "travis_fold:start:report"
 echo "Creating a report..."
 
-touch report.txt
+touch build_report.txt
 
-echo "web url=$TRAVIS_BUILD_WEB_URL" >> report.txt
-echo "job url=$TRAVIS_JOB_WEB_URL" >> report.txt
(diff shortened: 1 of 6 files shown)


scripts/travis/deploy-linux-build.sh
```diff
@@ -0,0 +1,79 @@
...
+
+
+#--------------------------------------------------------------------------------------------------
+# Create a report.
+#--------------------------------------------------------------------------------------------------
+
+echo "travis_fold:start:report"
+echo "Creating a report..."
+
+touch build_report.txt
```
dictoon, line 43: I suppose the report name could be moved into a variable instead of repeating it on the lines below, but that's a minor improvement.

Issue #2808, opened 2020-03-26
Enable entities to build importance map for OSL shaders in on_render_begin()


lorentzo:
### Motivation:
This issue occurred while solving the issue #701 in PR #2784. Segmentation error occured if `build_importance_map()` is called from `on_render_begin()` in `OSLEnvironmentEDF`. The reason is that OSL shaders (that are required for building importance map) haven't been compiled.

@est77:
> Compiling shaders shouldn't be done in `on_render_begin()`. Shaders have different lifetime than other entities. It is desirable to keep them alive as much as possible.

**NB:** `on_render_begin()` should be used opposed to  `on_frame_begin()` when possible. See [this discussion on Discord](https://discordapp.com/channels/430063582777049089/430064067873472526/682538301927850010) for further explanation. But not in this case.

@est77:
> When creating a master renderer, one shading system is created in `CPURenderDevice`. Instead of creating a new one, the shading system should be reused.  The problem with creating a new shading system is seen for example when entities (light, environment, ...) build an importance map (e.g. one shading system per light is not acceptable).

### Solution

Proposed by @est77 

1. In masterernderer.cpp, shader compilation (line 359) should be moved before `on_render_begin()` is called so that shaders are compiled when `on_render_begin()` is called.

2. Then, building an importance map for entities should be moved in `on_render_begin()`. 

**PROBLEM**:
* Interactive mode in appleseed.studio is crashing when OSL textures are used and above change is done.

3. Finally, shading system should be passed as an argument in `on_render_begin()` to all entities.

PR #2841, opened 2020-04-14
Enable OSL importance map build in on_render_begin

lorentzo:
### Intro
This PR is for issue  #2808. Investigate there for more introduction information.

### Contribution
* Moving `build_importance_map()` from `on_frame_begin()` to `on_render_begin()` in `OSLEnvironmentEDF`. This enables less workload in `on_frame_begin()`.
* Moving `m_render_device->initialize()` before `on_render_begin()` in `initialize_and_render_frame()` in `masterrenderer.cpp`. This enables OSL shader compilation before work in `on_render_begin()` is performed in entity (this case in `OSLEnvironmentEDF` entity).

### Further work
Passing the shading system as an argument in `on_render_begin()`.

### DIFF
--- a/src/appleseed/renderer/kernel/rendering/masterrenderer.cpp
+++ b/src/appleseed/renderer/kernel/rendering/masterrenderer.cpp
@@ -343,6 +343,13 @@ struct MasterRenderer::Impl
         // Construct an abort switch that will allow to abort initialization or rendering.
         RendererControllerAbortSwitch abort_switch(renderer_controller);
 
+        // Initialize the render device. Here is OSL shader compilation performed.
+        const bool success =
+            m_render_device->initialize(
+                m_resource_search_paths,
+                m_tile_callback_factory,
+                abort_switch);
+
         // Let scene entities perform their pre-render actions. Don't proceed if that failed.
         // This is done before creating renderer components because renderer components need
         // to access the scene's render data such as the scene's bounding box.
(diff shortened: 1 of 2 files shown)


src/appleseed/renderer/kernel/rendering/masterrenderer.cpp
```diff
@@ -354,12 +361,6 @@ struct MasterRenderer::Impl
             return renderer_controller.get_status();
         }
 
-        // Initialize the render device.
-        const bool success =
-            m_render_device->initialize(
-                m_resource_search_paths,
-                m_tile_callback_factory,
-                abort_switch);
         if (!success || abort_switch.is_aborted())
```
dictoon, line 364: This piece of code goes together with the block you moved. It needs to be moved as well (and updated since `recorder` won't have been created yet).

lorentzo: Hi,
If I am not mistaken this PR can be ignored. 
This problem was further handled by @est77 so I have stopped working on this PR.

lorentzo: Also, I just remembered one more thing.
@est77 mentioned that, while solving this issue, he had a problem with Appleseed crushing when rendering would be stopped and restarted (and maybe also when the material was changed).
I am thinking that he forgot about moving this piece of code that you mention here. Maybe that would help.

PR #2849, opened 2020-04-15
Remove Disney Material

est77:
Depends on PR #2848


src/appleseed.studio/mainwindow/project/materialcollectionitem.cpp
```diff
@@ -257,17 +150,6 @@ void MaterialCollectionItem::do_create_material(const char* model)
         new EntityBrowser<Assembly>(Base::m_parent));
 
     std::unique_ptr<CustomEntityUI> custom_entity_ui;
```
dictoon, line 152: We can probably get rid of this variable.

src/appleseed.studio/mainwindow/project/materialitem.cpp
```diff
@@ -108,17 +96,6 @@ void MaterialItem::slot_edit(AttributeEditor* attribute_editor)
 
     std::unique_ptr<CustomEntityUI> custom_entity_ui;
```
dictoon, line 97: And this one.

PR #2850, opened 2020-04-16
Fix reading and writing embedded OSL code

### LINKED ISSUE #2830: Embedded with OSL shader with < character makes project file invalid
Repro case: [31 - multiple source shaders.zip](https://github.com/appleseedhq/appleseed/files/4440385/31.-.multiple.source.shaders.zip)

Properly encoding the XML entities should fix the problem, but it didn't...

LZaw:
This PR fixes #2830 
The `replace_special_xml_characters` function did only get called on values for XML element attributes, i.e. on `attribute_value` in this example:
```<someTag attribute_key="attribute_value"/>```
I changed it to be called on every part of an XML element (except the indentation) as well as on the OSL Code.
Moreover, when reading the OSL Code, the parser only read up to the next escaped character, which caused the `m_code` to only be the last part of the OSL Code after the last escaped character. I changed this as well as the resulting formatting issues (due to `trim_both`).

So far, I only tested the results with the scene in the Issue, but I'd be happy to test it with more scenes, if needed.

### CODE REVIEW

File: src/appleseed/foundation/string/string.h [hunk 1]
```diff
@@ -896,6 +896,11 @@ inline std::string replace_special_xml_characters(const std::string& s)
     return result;
 }
 
+inline std::string replace_special_xml_characters(const char* s){
```
dictoon, line 899: This function isn't needed as a `const char*` is implictly convertible to a `std::string`, hence the original function can be used without issues.

src/appleseed/foundation/utility/xmlelement.h
```diff
@@ -156,13 +156,14 @@ inline void XMLElement::write(const ContentType content_type)
     assert(!m_is_open);
 
     // Open the tag.
-    std::fprintf(m_file, "%s<%s", m_indenter.c_str(), m_name.c_str());
+    std::fprintf(m_file, "%s<%s", m_indenter.c_str(), replace_special_xml_characters(m_name).c_str());
```
dictoon, line 159: The name of the tag certainly shouldn't contain any special characters. I would suggest to revert this.

src/appleseed/foundation/utility/xmlelement.h
```diff
@@ -156,13 +156,14 @@ inline void XMLElement::write(const ContentType content_type)
...
 
     // Emit the attributes.
     for (const_each<AttributeVector> i = m_attributes; i; ++i)
     {
+        const std::string attribute_key = replace_special_xml_characters(i->first);
```
dictoon, line 164: Similarly, it's not possible for an attribute name to contain special characters. This should be reverted.

src/appleseed/renderer/modeling/project/projectfilewriter.cpp
```diff
@@ -863,7 +863,9 @@ namespace
             write_collection(shader.shader_params());
 
             if (shader.get_source_code())
+            {
```
dictoon, line 866: We prefer not to add braces for single line statements in order to increase vertical code density.

src/appleseed/renderer/modeling/project/projectfilereader.cpp
```diff
@@ -1817,7 +1817,10 @@ namespace
             const XMLCh* const  chars,
             const XMLSize_t     length) override
         {
-            m_code = trim_both(transcode(chars));
+            if(!m_code.empty()){
```
dictoon, line 1820: My theory is that with the fix you did in `write_osl_code()`, this change isn't needed. It shouldn't required really, and it looks a bit hacky (why would adding a space be always valid?). Could you test that theory?

src/appleseed/foundation/string/string.h
LZaw: Okay, I'll check.

src/appleseed/foundation/utility/xmlelement.h
LZaw: I thought about it too. In the end of the day, if there are any special characters in the tag, we got problems anyways, but I thought it would be more thorough to do it that way.

src/appleseed/renderer/modeling/project/projectfilereader.cpp
LZaw: I already did. Even in the new setup the read stops at every special character (i.e. this function gets called for multiple blocks belonging to the same code).
The space is indeed hacky. I had a more explicit version at hand, which seemed a bit huge. Since "trim" only removes spaces, I'd count the amount of removed spaces and added them again as soon as a new codesnippet gets added.
I'll test your theory tomorrow again in different setups and report back.

src/appleseed/foundation/utility/xmlelement.h
dictoon: Less code is always better (code is a liability, not an asset) in every dimension: speed, compactness, readability, etc. Since we know per the XML spec that tags cannot contain special characters, we definitely don't want to do any escaping which involves memory allocations, memory fragmentation, string copies, etc.

src/appleseed/renderer/modeling/project/projectfilereader.cpp
dictoon: I would be quite surprised if with proper escaping the whole text wasn't read at once.

LZaw:
I changed all the things you requested and put a print-statement in the `characters` function as well as the `get_code` function of the reader, to test your theory.

Reading the attached file will result in split reads (split on each escaped character), as described above. The scene is the one from the issue, with escaped characters.

Reading the original (unescaped) source file results in an error:
`2020-04-17T10:00:38.394202Z <002>   114 MB error   | while reading <some_path>/31 - multiple source shaders.appleseed, at line 140, column 15: element name expected.`
That's expected behaviour, since the original file still has `<` symbols in the shader source.

[escaped_multiple_shaders.zip](https://github.com/appleseedhq/appleseed/files/4492195/escaped_multiple_shaders.zip)

dictoon:
Thanks for looking into this problem in details, I really appreciate.

You are right: [according to Xerces-C++'s documentation](https://xerces.apache.org/xerces-c/apiDocs-3/classContentHandler.html#aed552fab959bfa7e9135691e1d21ab5a), `characters()` will be called for chunks of text, not necessarily the entire text at once:

> The Parser will call this method to report each chunk of character data. SAX parsers may return all contiguous character data in a single chunk, or they may split it into several chunks; however, all of the characters in any single event must come from the same external entity, so that the Locator provides useful information.

So we must take care of concatenation ourselves.

I've tried your code and it appears to work correctly in the limited tests I've done, however the logic is still fragile and error-prone. It turns out that we can do a lot simpler:

    class OSLCodeElementHandler
      : public ElementHandlerBaseType
    {
      public:
        explicit OSLCodeElementHandler(ParseContext& context)
        {
        }

        void characters(
            const XMLCh* const  chars,
            const XMLSize_t     length) override
        {
            m_code += transcode(chars);
        }

        std::string get_code() const
        {
            return trim_both(m_code);
        }

      private:
        std::string m_code;
    };

I'll push this to this PR and merge it.

Thanks again!

PR #2851, opened 2020-04-16
Fix bugs

dictoon:


oktomus:
@dictoon You haven't added the test scenes I provided using textures for emission. Is that intentional ?

src/appleseed/renderer/kernel/lighting/lighttypes.cpp
```diff
@@ -358,21 +359,35 @@ void EmittingShape::make_shading_point(
...
         break;
 
       case SphereShape:
         {
-            const double theta = static_cast<double>(param_coords[0]);
-            const double phi = static_cast<double>(param_coords[1]);
+            const double r = m_geom.m_sphere.m_radius;
+
+            const double phi = TwoPi<double>() * u;
+            const double theta = HalfPi<double>() - v * Pi<double>();
```
oktomus, line 372: In `mappings.h`, `sample_sphere_uniform()`, we do this:
```cpp
const T phi = TwoPi<T>() * s[0];
const T cos_theta = T(1.0) - T(2.0) * s[1];
const T sin_theta = std::sqrt(T(1.0) - cos_theta * cos_theta);
```

Maybe we could follow the exact same pattern here or the other ? Which would be:
```cpp
const double phi = TwoPi<double>() * u;
const double cos_phi = std::cos(phi);
const double sin_phi = std::sin(phi);

const double cos_theta = 1.0 - 2.0 * v;
const double sin_theta = std::sqrt(1.0 - cos_theta * cos_theta);
```

I would prefer changing `sample_sphere_uniform` and do as you do here because `theta` is computed on its own, which is clearer.

dictoon: Actually, `sample_sphere_uniform()` is the correct way to do it, and the way I do it is not equivalent and is in fact incorrect.

PR #2854, opened 2020-04-19
New project format

est77:
- Split XML project file reading and writing


dictoon:
I haven't looked at the code yet but conceptually I think this could (or even should) be its own PR.

est77:
> I haven't looked at the code yet but conceptually I think this could (or even should) be its own PR.

I can open a PR with only this. It is just basic refactoring.

PR #2856, opened 2020-04-22
Move XML project file reading and writing to their own classes

est77:


src/appleseed.qtcommon/project/projectmanager.cpp
```diff
@@ -96,8 +97,7 @@ void ProjectManager::load_project_async(const std::string& filepath)
 
 bool ProjectManager::load_builtin_project(const std::string& name)
 {
-    ProjectFileReader reader;
-    m_project = reader.load_builtin(name.c_str());
+    ProjectFileReader::load_builtin(name.c_str());
```
dictoon, line 100: You're dismissing the return value, this no longer works.

est77: Fixed

PR #2859, opened 2020-05-03
Exclude source code from the deployed build

oktomus:


scripts/travis/deploy-linux-build.sh
```diff
@@ -67,9 +68,10 @@ echo $DEPLOY_SSH_KEY >> $HOME/.ssh/known_hosts
 export SSHPASS=$DEPLOY_PASSWORD
 sshpass -e rsync \
     -raz --stats --no-perms --no-owner --no-group --delete \
-    --exclude 'src' \
-    --exclude 'docs' \
-    --exclude 'cmake' \
+    --include="build_report.txt" \
```
dictoon, line 71: Much nicer!

I would also suggest to expand all options (`-e`, `-raz`) to their long forms, and either having them all on one line, or all on a separate line, but not a mix of the two.

PR #2863, opened 2020-05-08
Remove unused build files on DO

oktomus:
References:
- https://linux.die.net/man/1/rsync
- https://stackoverflow.com/questions/15687755/how-to-use-rsync-to-copy-only-specific-subdirectories-same-names-in-several-dir
- https://stackoverflow.com/questions/9952000/using-rsync-include-and-exclude-options-to-include-directory-and-file-by-pattern

### DIFF
--- a/scripts/travis/deploy-linux-build.sh
+++ b/scripts/travis/deploy-linux-build.sh
@@ -75,11 +75,10 @@ sshpass -e rsync \
     --compress \
     --stats \
     --no-perms --no-owner --no-group \
-    --include="build_report.txt" \
-    --include="prebuilt-linux-deps/***" \
-    --include="sandbox/***" \
-    --include="scripts/***" \ 
-    --exclude="*"
+    build_report.txt \
+    prebuilt-linux-deps \
+    sandbox \
+    scripts \
     ./ \
     $DEPLOY_USER@$DEPLOY_URL:$DEPLOY_FOLDER
 
--- a/scripts/travis/deploy-linux-build.sh
+++ b/scripts/travis/deploy-linux-build.sh
@@ -73,10 +73,11 @@ sshpass -e rsync \
     --stats \
     --no-perms --no-owner --no-group \
     --include="build_report.txt" \
-    --include="sanbox/" \
-    --include="scripts/" \
-    --include="prebuilt-linux-deps/" \
-    --delete-after \
+    --include="prebuilt-linux-deps/***" \
+    --include="sanbox/***" \
+    --include="scripts/***" \
+    --exclude="*" \
+    --delete-after --delete-excluded \
     ./ \
     $DEPLOY_USER@$DEPLOY_URL:$DEPLOY_FOLDER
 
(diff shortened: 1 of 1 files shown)


scripts/travis/deploy-linux-build.sh
```diff
@@ -73,10 +73,11 @@ sshpass -e rsync \
     --stats \
     --no-perms --no-owner --no-group \
     --include="build_report.txt" \
-    --include="sanbox/" \
-    --include="scripts/" \
-    --include="prebuilt-linux-deps/" \
-    --delete-after \
+    --include="prebuilt-linux-deps/***" \
+    --include="sanbox/***" \
```
LZaw, line 77: Shouldn't this be `sandbox`?

oktomus: yes it should, thanks

scripts/travis/deploy-linux-build.sh
```diff
@@ -64,19 +64,21 @@ echo "Deploy travis build on the server..."
...
     --compress \
     --stats \
     --no-perms --no-owner --no-group \
     --include="build_report.txt" \
-    --include="sanbox/" \
-    --include="scripts/" \
     --include="prebuilt-linux-deps/" \
-    --delete-after \
+    --include="sandbox/" \
+    --include="scripts/" \
```
LZaw, line 81: You need the `--exclude` statement as well. Otherwise the whole appleseed directory will be copied over.
Also, in the `--include`s you need the `***`, as you had it a few commits ago.

scripts/travis/deploy-linux-build.sh
```diff
@@ -64,19 +64,21 @@ echo "Deploy travis build on the server..."
 # Add server public key to known hosts.
 echo $DEPLOY_SSH_KEY >> $HOME/.ssh/known_hosts
 
-# Send build to server.
+# Remove previous build from server.
```
dictoon, line 67: Don't you want to preserve the previous build?

scripts/travis/deploy-linux-build.sh
oktomus: I tested to run it only with include and it was working as expected

scripts/travis/deploy-linux-build.sh
oktomus: no. The test runner copy the build somewhere else so it won't affect the ruuning test job or something

scripts/travis/deploy-linux-build.sh
oktomus: When there is include rules, rsync will use the first include match to check if a file can be copied. It will be ignored otherwise.

oktomus: *** was required only when i was excluding and deleting. Before that we didn't have *** and it was ok

scripts/travis/deploy-linux-build.sh
```diff
@@ -64,19 +64,21 @@ echo "Deploy travis build on the server..."
...
-    --include="build_report.txt" \
-    --include="sanbox/" \
-    --include="scripts/" \
-    --include="prebuilt-linux-deps/" \
-    --delete-after \
+    build_report.txt \
+    prebuilt-linux-deps \
+    sandbox \
+    scripts \
     ./ \
```
LZaw, line 82: If you name every folder and file explicitely, you have to remove the base dir, otherwise it will be copied over as well.

PR #2865, opened 2020-05-18
Refactor the vignette post-processing stage

laurelkeys:
Refactored the vignette post-processing stage to split its computation to threads.

Main changes:  
- Added 5 new test scenes (`sandbox/tests/test scenes/vignette/`).
- Added `thread_count` as a parameter of `PostProcessingStage`'s `execute()`.
- Created a new `EffectJob` which implements `IJob`.
- Created a new `renderer/modeling/postprocessingeffect/` folder, to narrow the actual algorithm implementation for post-processing effects into instances of `IEffectApplier` (the stage's `execute()` now simply spawns jobs to execute its `apply()`).


src/appleseed/renderer/api/postprocessing.h
```diff
@@ -37,3 +37,6 @@
 #include "renderer/modeling/postprocessingstage/postprocessingstagetraits.h"
 #include "renderer/modeling/postprocessingstage/renderstamppostprocessingstage.h"
 #include "renderer/modeling/postprocessingstage/vignettepostprocessingstage.h"
+#include "renderer/modeling/postprocessingeffect/ipostprocessingeffect.h"
+#include "renderer/modeling/postprocessingeffect/postprocessingeffectjob.h"
+#include "renderer/modeling/postprocessingeffect/vignettepostprocessingeffect.h"
```
oktomus, line 42: What's the difference between `stage` and `effect` ?

src/appleseed/renderer/modeling/postprocessingeffect/ipostprocessingeffect.h
```diff
@@ -0,0 +1,71 @@
...
+namespace renderer      { class Frame; }
+
+namespace renderer
+{
+
+//
+// Interface of a post-processing effect algorithm applier.
+//
+
+class IEffectApplier
```
oktomus, line 47: I think `IImageEffectApplier` or `IPostProcessEffectApplier` is a better name

src/appleseed/renderer/modeling/postprocessingeffect/postprocessingeffectjob.cpp
```diff
@@ -0,0 +1,120 @@
...
+using namespace foundation;
+
+namespace renderer
+{
+
+//
+// EffectJob class implementation.
+//
+
+EffectJob::EffectJob(
```
oktomus, line 53: Same here with the naming, `EffectJob` is too generic

src/appleseed/renderer/modeling/postprocessingeffect/postprocessingeffectjob.cpp
```diff
@@ -0,0 +1,120 @@
...
+    const EffectApplierVector&  effect_appliers,
+    const Frame&                frame,
+    const size_t                tile_x,
+    const size_t                tile_y,
+    const size_t                thread_count)
+  : m_effect_appliers(effect_appliers)
```
oktomus, line 59: We should name this vector the same way we do in other job renderers. Something like `m_jobs`  ? It makes more sense then when you call `execute()` on a job.

src/appleseed/renderer/modeling/postprocessingeffect/postprocessingeffectjob.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+    // Generate tiles in linear order.
+    std::vector<size_t> tiles;
+    linear_ordering(tiles, props.m_tile_count);
+
+    // Make sure the right number of tiles was created.
+    assert(tiles.size() == props.m_tile_count);
+
+    // Create effect jobs, one per tile.
+    for (size_t i = 0; i < props.m_tile_count; ++i)
```
oktomus, line 100: `std::` anything you can 
```suggestion
    for (std::size_t i = 0; i < props.m_tile_count; ++i)
```

src/appleseed/renderer/modeling/postprocessingeffect/vignettepostprocessingeffect.h
```diff
@@ -0,0 +1,87 @@
...
+namespace renderer  { class Frame; }
+
+namespace renderer
+{
+
+//
+// Vignette-specific parameters.
+//
+
+struct VignetteParams {
```
oktomus, line 50: In other places, we generally don't create a struct for the params. We simply give params and we do `params.get_float(...)" or something where we need the actual values. Is there a specific reason you created this struct ?

src/appleseed/renderer/modeling/postprocessingeffect/postprocessingeffectjob.cpp
laurelkeys: I think this is correct, i.e., the job has a "list of appliers" and the job factory that has a "list of jobs".
Like the [TileJob](https://github.com/appleseedhq/appleseed/blob/master/src/appleseed/renderer/kernel/rendering/generic/tilejob.h#L64) which has a "list of renderers" but no job list. It's the [TileJobFactory](https://github.com/appleseedhq/appleseed/blob/master/src/appleseed/renderer/kernel/rendering/generic/tilejobfactory.h#L80) that has a `tile_jobs`.

src/appleseed/renderer/modeling/postprocessingeffect/vignettepostprocessingeffect.h
laurelkeys: Do you mean using a `ParamArray`? I had originally thought about it, but as Franz mentioned on Discord, they shouldn't be used to pass values "internally", for this we can use standard arguments.

Thus, I preferred to group them into a POD-struct as I think it's clean and easier to spot them this way, but I can rewrite `VignetteApplierFactory` to receive each one separately. What do you think?

oktomus: I misunderstood the code, my bad :) You did right

src/appleseed/renderer/modeling/postprocessingeffect/postprocessingeffectjob.cpp
```diff
@@ -0,0 +1,120 @@
...
+#include "renderer/modeling/frame/frame.h"
+
+// appleseed.foundation headers.
+#include "foundation/image/canvasproperties.h"
+#include "foundation/image/image.h"
+#include "foundation/image/tile.h"
+#include "foundation/math/ordering.h"
+
+// Standard headers.
+#include <cassert>
```
oktomus, line 42:
```suggestion
#include <cassert>
#include <cstddef> 
```

For size_t

src/appleseed/renderer/modeling/postprocessingeffect/vignettepostprocessingeffect.cpp
```diff
@@ -0,0 +1,163 @@
...
+#include "renderer/modeling/frame/frame.h"
+
+// appleseed.foundation headers.
+#include "foundation/image/canvasproperties.h"
+#include "foundation/image/color.h"
+#include "foundation/image/image.h"
+#include "foundation/math/scalar.h"
+#include "foundation/math/vector.h"
+
+#include <iostream>
```
oktomus, line 42: Was it for debug ?

src/appleseed/renderer/modeling/postprocessingeffect/vignettepostprocessingeffect.cpp
```diff
@@ -0,0 +1,163 @@
...
+                    const float inverse_biquadratic_radial_falloff = 1.0f / (quadratic_radial_falloff * quadratic_radial_falloff);
+
+                    Color4f pixel;
+                    tile.get_pixel(x, y, pixel);
+
+                    pixel.rgb() *= inverse_biquadratic_radial_falloff;
+                    tile.set_pixel(x, y, pixel);
+                }
+            }
+
```
oktomus, line 122:
```suggestion
```

src/appleseed/renderer/modeling/postprocessingeffect/vignettepostprocessingeffect.cpp
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+      private:
+        // Settings.
+        const float                     m_intensity;
```
oktomus, line 127: I think you can use/store VignetteParams reference directly here.

src/appleseed/renderer/modeling/postprocessingeffect/vignettepostprocessingeffect.cpp
```diff
@@ -0,0 +1,163 @@
...
+{
+    delete this;
+}
+
+IEffectApplier* VignetteApplierFactory::create()
+{
+    return new VignetteApplier(m_params);
+}
+
+IEffectApplier* VignetteApplierFactory::create(
```
oktomus, line 157: Do you use both `create` methods ?

src/appleseed/renderer/modeling/postprocessingeffect/vignettepostprocessingeffect.h
```diff
@@ -0,0 +1,87 @@
...
+    // Settings.
+    float       intensity;
+    float       anisotropy; // 0 = no anisotropy (i.e. perfectly rounded)
+                            // 1 = full anisotropy (i.e. the aspect ratio of the frame is respected)
+
+    // Context.
+    float       frame_width;
```
oktomus, line 57: I don't think these should be in VignetteParams. Maybe you can simply pass them as parameters to the  `apply()` or execute()` method ?

src/appleseed/renderer/modeling/postprocessingeffect/postprocessingeffectjob.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+//
+// EffectJobFactory class implementation.
+//
+
+void EffectJobFactory::create(
+    const Frame&                            frame,
+    const EffectJob::EffectApplierVector&   effect_appliers,
+    const std::size_t                       thread_count,
+    EffectJobVector&                        effect_jobs)
```
oktomus, line 87: Why passing jobs by reference ? Can we return the jobs instead ? When you call this, effect_jobs is empty so there is no need to fill it by reference

src/appleseed/renderer/modeling/postprocessingeffect/vignettepostprocessingeffect.cpp
laurelkeys: Yes, my bad 😅

src/appleseed/renderer/modeling/postprocessingeffect/postprocessingeffectjob.cpp
laurelkeys: `<cstddef>` is included in the header file, shouldn't I omit it from the `.cpp` then?

src/appleseed/renderer/modeling/postprocessingeffect/vignettepostprocessingeffect.cpp
laurelkeys: No, only the static one (`vignettepostprocessingstage.cpp:124`), the other is required by the interface. Iirc something similar is done in the `GenericFrameRendererFactory`.

Since there's no way to enforce the definition of a static `create()` with an interface, is it better to remove it and use the non-static one? This would imply the need to create an instance of `VignetteApplierFactory` (for each `VignetteParam`).

laurelkeys: Removed the static one.

src/appleseed/renderer/modeling/postprocessingeffect/postprocessingeffectjob.cpp
oktomus: Yes, you are right.

src/appleseed/renderer/api/postprocessing.h
```diff
@@ -37,3 +37,6 @@
 #include "renderer/modeling/postprocessingstage/postprocessingstagetraits.h"
 #include "renderer/modeling/postprocessingstage/renderstamppostprocessingstage.h"
 #include "renderer/modeling/postprocessingstage/vignettepostprocessingstage.h"
+#include "renderer/modeling/postprocessingstage/effect/ipostprocessingeffect.h"
```
oktomus, line 40: I don't think this needs to be exposed to the API. I think we expose only stages and that should be enough to apply post process.

src/appleseed/renderer/kernel/rendering/masterrenderer.cpp
```diff
@@ -529,11 +529,12 @@ struct MasterRenderer::Impl
         }
 
         // Execute post-processing stages.
+        const size_t thread_count = get_rendering_thread_count(m_params);
```
oktomus, line 532:
```suggestion
        const std::size_t thread_count = get_rendering_thread_count(m_params);
```

src/appleseed/renderer/modeling/postprocessingstage/colormappostprocessingstage.cpp
```diff
@@ -199,7 +199,7 @@ namespace
             return true;
         }
 
-        void execute(Frame& frame) const override
+        void execute(Frame& frame, const size_t thread_count) const override
```
oktomus, line 202:
```suggestion
        void execute(Frame& frame, const std::size_t thread_count) const override
```

src/appleseed/renderer/modeling/postprocessingstage/effect/postprocessingeffectjob.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+        // Create the tile job.
+        effect_jobs.push_back(
+            new ImageEffectJob(
+                effect_appliers,
+                frame,
+                tile_x,
+                tile_y));
+    }
+    return effect_jobs;
```
oktomus, line 115:
```suggestion

    return effect_jobs;
```

src/appleseed/renderer/modeling/postprocessingstage/vignettepostprocessingstage.cpp
```diff
@@ -94,45 +100,49 @@ namespace
...
+            VignetteApplierFactory effect_applier_factory(effect_params);
+            ImageEffectJob::EffectApplierVector effect_appliers;
+
+            effect_appliers.reserve(thread_count);
+            for (std::size_t i = 0; i < thread_count; ++i)
+                effect_appliers.push_back(effect_applier_factory.create());
+
+            // Create effect applier jobs.
+            ImageEffectJobFactory effect_job_factory;
+            ImageEffectJobFactory::EffectJobVector effect_jobs = effect_job_factory.create(
```
oktomus, line 129: should be formatted like this:
```cpp
             ImageEffectJobFactory::EffectJobVector effect_jobs = 
                 effect_job_factory.create(
                     frame,
                     effect_appliers);
```

src/appleseed/renderer/modeling/postprocessingstage/vignettepostprocessingstage.cpp
```diff
@@ -94,45 +100,49 @@ namespace
...
+                                                                    frame,
+                                                                    effect_appliers);
+
+            // Schedule effect applier jobs.
+            JobQueue job_queue;
+            for (const auto effect_job : effect_jobs)
```
oktomus, line 135: Is it a const reference ? (i.e.  `const auto&`).

also, we use const when the type can be easily determined like ` const auto something = static_cast<TYPE>(...)`. Maybe you should use explicit type here

src/appleseed/renderer/api/postprocessing.h
```diff
@@ -31,6 +31,7 @@
 // API headers.
 #include "renderer/modeling/postprocessingstage/colormappostprocessingstage.h"
 #include "renderer/modeling/postprocessingstage/ipostprocessingstagefactory.h"
+#include "renderer/modeling/postprocessingstage/multithreadpostprocessingstage.h"
```
oktomus, line 34: I don't think you need to expose that

src/appleseed/renderer/modeling/postprocessingstage/effect/postprocessingeffectjob.cpp
```diff
@@ -0,0 +1,117 @@
...
+{
+    // Retrieve frame properties.
+    const CanvasProperties& props = frame.image().properties();
+
+    // Generate tiles in linear order.
+    std::vector<std::size_t> tiles;
+    linear_ordering(tiles, props.m_tile_count);
+
+    // Make sure the right number of tiles was created.
+    assert(tiles.size() == props.m_tile_count);
```
dictoon, line 91: This assertion feels redundant. It made sense in the context of rendering tile jobs where the number of jobs was indirectly controlled via `m_tile_count_x` and `m_tile_count_y`.

src/appleseed/renderer/modeling/postprocessingstage/effect/postprocessingeffectjob.cpp
```diff
@@ -0,0 +1,117 @@
...
+ImageEffectJobFactory::EffectJobVector ImageEffectJobFactory::create(
+    const Frame&                    frame,
+    const IImageEffectApplier&      effect_applier) const
+{
+    // Retrieve frame properties.
+    const CanvasProperties& props = frame.image().properties();
+
+    // Generate tiles in linear order.
+    std::vector<std::size_t> tiles;
+    linear_ordering(tiles, props.m_tile_count);
```
dictoon, line 88: A more cache-friendly ordering is the Hilbert one, I would suggest that this would be a better default.

src/appleseed/renderer/modeling/postprocessingstage/multithreadpostprocessingstage.h
```diff
@@ -0,0 +1,59 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+#pragma once
+
+// appleseed.renderer headers.
+#include "renderer/modeling/postprocessingstage/postprocessingstage.h"
+#include "renderer/modeling/postprocessingstage/effect/ipostprocessingeffect.h"
```
dictoon, line 33: You can simply forward-declare `IImageEffectApplier`.

src/appleseed/renderer/modeling/postprocessingstage/multithreadpostprocessingstage.cpp
```diff
@@ -0,0 +1,87 @@
...
+        effect_job_factory.create(
+            frame,
+            effect_applier);
+
+    // Schedule effect applier jobs.
+    JobQueue job_queue;
+    for (ImageEffectJob* const effect_job : effect_jobs)
+        job_queue.schedule(effect_job);
+
+    // Create a job manager to wait until jobs have effectively stopped.
```
dictoon, line 77: Technically this comment is incorrect as the main purpose of the job manager isn't to wait until jobs have stopped :)

src/appleseed/renderer/modeling/postprocessingstage/multithreadpostprocessingstage.h
```diff
@@ -0,0 +1,59 @@
...
+    MultithreadPostProcessingStage(
+        const char*             name,
+        const ParamArray&       params);
+
+    // Execute this post-processing stage on a given frame
+    // by scheduling a job for each tile to apply the image effect.
+    virtual void execute_on_tiles(
+        Frame&                          frame,
+        const IImageEffectApplier&      effect_applier,
+        const std::size_t               thread_count = 1) const;
```
dictoon, line 56: Include `<cstddef>`.

src/appleseed/renderer/modeling/postprocessingstage/multithreadpostprocessingstage.h
```diff
@@ -0,0 +1,59 @@
...
+{
+  public:
+    // Constructor.
+    MultithreadPostProcessingStage(
+        const char*             name,
+        const ParamArray&       params);
+
+    // Execute this post-processing stage on a given frame
+    // by scheduling a job for each tile to apply the image effect.
+    virtual void execute_on_tiles(
```
dictoon, line 53: Why aren't you overriding `PostProcessingStage::execute()` here? That was the idea: a multithreaded post-processing stage is just another "standard" stage, it just happens to execute a post-processing effect in a multithreaded fashion. The post-processing effect should be passed to the constructor.

src/appleseed/renderer/modeling/postprocessingstage/vignettepostprocessingstage.cpp
```diff
@@ -94,45 +98,31 @@ namespace
             return true;
         }
 
-        void execute(Frame& frame) const override
+        void execute(Frame& frame, const std::size_t thread_count) const override
         {
+            // Skip vignetting if the intensity is zero.
+            if (feq(m_intensity, 0.0f))
```
dictoon, line 104: I would suggest to only skip if `m_intensity` is exactly equal to 0.0f.

src/appleseed/renderer/modeling/postprocessingstage/vignettepostprocessingstage.cpp
```diff
@@ -94,45 +98,31 @@ namespace
...
-                    image.get_pixel(x, y, pixel);
-
-                    pixel.rgb() *= inverse_biquadratic_radial_falloff;
-
-                    image.set_pixel(x, y, pixel);
-                }
-            }
+
+            // Initialize effect-specific settings and context.
+            const VignetteParams effect_params {
```
dictoon, line 110: Move brace to its own line.

src/appleseed/renderer/modeling/postprocessingstage/vignettepostprocessingstage.cpp
```diff
@@ -94,45 +98,31 @@ namespace
...
+                static_cast<float>(props.m_canvas_width),
+                static_cast<float>(props.m_canvas_height)
+            };
+
+            // Instantiate an effect applier.
+            const VignetteApplierFactory effect_applier_factory(effect_params);
+            const std::unique_ptr<const IImageEffectApplier> effect_applier(effect_applier_factory.create());
+
+            // Apply the effect onto each image tile.
+            MultithreadPostProcessingStage::execute_on_tiles(
```
dictoon, line 122: I see now why you introduced `execute_on_tiles()` but IMHO you can simplify further: any post-processing stage based on `MultithreadPostProcessingStage` should mostly just pass the right effect (here the vignetting effect) to `MultithreadPostProcessingStage` and let that class handle everything.

src/appleseed/renderer/modeling/postprocessingstage/effect/postprocessingeffectjob.cpp
laurelkeys: Should I keep it if I change `linear_ordering` to `hilbert_ordering`?

dictoon: Yes, good point.

PR #2868, opened 2020-05-24
WIP: light path expression

joeyskeys:
I used to work with Luis and this's the feature he suggested me to implement, and here it is.

LPE AOV uses OSL's accumulator to do the calculation which is kinda different from appleseed's current AOV system.
The relationship between LPE AOV accumulator and LPE AOV is not one to one like the other applessed AOVs do, so there's a totally new entry in the project for LPE AOVs and the code structure is not quite the same.
I think the integration of LPE into appleseed's AOV system is mostly done, what comes next is the generate correct light path events and I believe there're still some problems need to solved.

As documented in [OSL's wiki](https://github.com/imageworks/OpenShadingLanguage/wiki/OSL-Light-Path-Expressions), currently supported events are:

- C: Camera
- R: Surface Reflection
- T: Surface Transmission
- V: Volume
- L: Light
- O: Emitting object
- B: Background

Plus a customed one:
  * X: Termination by null material, recursion bound etc...
This event is added coz I found a lot light path termiated due to different terminations, need some suggestions on this part.

Scattering events support info:
- D: Diffuse (supported)
- G: Glossy (supported)
- S: Singular, some people like it to be called reflection and refraction (this is confusing and is generated for ScatteringMode::Specular)
- s: Straight (confusing one, not supported)

Custom label is not supported yet.

There's a test scene sandbox/tests/test scenes/aovs/06 - LPE aov.appleseed, you can play with it first. DCC software supported is not implemented yet so you can only add LPE AOV in the text editor, sorry for the inconvenience. Save the AOVs via the "save main image and aovs" button in appleseed.studio.
In the above scene only mesh light is tested.
Volume support is added but not tested, need to setup another test scene.

And for the scattering mode part, I really need you guys' options.
Previously hit event is handled before bsdf is evaluated which caused the light path event generated to be incorrect, in the current implementation I postponed the handling of hit event(after bsdf is evaluated), which caused emitter event could be generated at index 0 and is conflicting with the assertion in LightPathStream::create_path_from_sampled_emitter.
I don't know if my current implementation will cause other problems.

A known issue is that the "<" symbol in the expression will cause problem when parsing project file, haven't try to solve this yet.

And in the last, here are some aov pictures rendered from the test scene:
direct_diffuse
image
greenwallonsphere
image
indirect_diffuse
image
specular
image
terminate
image
transparent
image


dictoon:
Hi @joeyskeys,

This is an awesome feature: Would you be interested in turning it into a mergeable PR?

joeyskeys:
Hi @dictoon,

Yeah sure! I'll be working on this ASAP. Been away from the discord channel since the covid-19 outbreak.
Anyway I'll be working on this soon.

Issue #2704, opened 2019-09-21
Remove all occurrences of foundation::each<> and foundation::const_each<>


dictoon:
Replace them by range-based `for` loops.

oktomus:
You mentioned why somewhere but I forgot where 🤦‍♂ Then my question is why ?

dictoon:
Since C++11 introduced range-based `for` loops, we no longer need our custom mechanism anymore: it doesn't bring anything over range-based `for` loops. The only case where a similar custom mechanism could be useful is for reverse iteration.

hbina:
If no one is working on this I would like to take it

oktomus:
Go ahead @hbina  :)

hbina:
@oktomus I have been trying to extract the tar file here https://github.com/appleseedhq/linux-deps/releases/download/v2.1/appleseed-deps-shared-2.1.tgz
but for whatever reason I have insufficient permission. Do I really need to do this?

oktomus:
@hbina For any problem you encounter with compilation, you should ask for help in #compiling on our discord https://discord.gg/dNCE5J8

darthgera123:
hi is this issue still open? I would like to work on it

dictoon:
Yes it is still open.

prat31:
I would like to work on this. Is it still open?

dictoon:
It is.

AyushRaj8271:
hey @dictoon! Is this issue still open?

oktomus:
> 
> 
> hey @dictoon! Is this issue still open?

It is.

oktomus:
@AyushRaj8271 It's possible that the last PR #2869 doesn't replace all occurences. You can check in the code

ganesh1828:
Hello, is this issue still open ? If yes I would like to work on it. Can you please assign me this issue.

oktomus:
Hi @ganesh1828 ,you can work on it directly and submit a PR.

ganesh1828:
Hi @oktomus / @dictoon I think this issue is solved in #2869 PR. You can close this issue.

PR #2869, opened 2020-05-28
Replace some usages of each<> and const_each<>

h44rd:
Hi,

This PR aims to solve issue #2704.

Two instances of `foundation::each` in the file `src/appleseed/renderer/kernel/intersection/treerepository.h` and one instance of `foundation::each_const` in `src/appleseed/renderer/utility/paramarray.h` has been replaced with  a range based for loop.

I have run tests under 'Debug' and all of them have passed.

Please review and let me know if there are any further changes required.

Thanks,
Haard

### DIFF
--- a/src/appleseed/renderer/kernel/intersection/treerepository.h
+++ b/src/appleseed/renderer/kernel/intersection/treerepository.h
@@ -86,7 +86,7 @@ class TreeRepository
 template <typename TreeType>
 TreeRepository<TreeType>::~TreeRepository()
 {
-    for (std::pair<const std::uint64_t, TreeInfo>& tree : m_trees)
+    for (std::pair<std::uint64_t, TreeInfo>& tree : m_trees)
         delete tree.second.m_tree;
 }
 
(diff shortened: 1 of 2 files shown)


src/appleseed/renderer/kernel/intersection/treerepository.h
```diff
@@ -86,8 +86,8 @@ class TreeRepository
 template <typename TreeType>
 TreeRepository<TreeType>::~TreeRepository()
 {
-    for (foundation::each<TreeContainer> i = m_trees; i; ++i)
-        delete i->second.m_tree;
+    for (auto &i : m_trees)
```
dictoon, line 89: Whenever possible, when the type is definitely known, I'd prefer that we don't use `auto`.

src/appleseed/renderer/utility/paramarray.h
```diff
@@ -578,9 +578,9 @@ bool ParamArray::contains(
     const StringVec&            allowed_values,
     const T&                    value)
 {
-    for (foundation::const_each<StringVec> i = allowed_values; i; ++i)
+    for (auto const &i : allowed_values)
```
dictoon, line 581: `const auto& i` (ampersand goes on the type's side).

But as I said above, `auto` should only be used when the type isn't known, or is particularly cumbersome.

src/appleseed/renderer/kernel/intersection/treerepository.h
dictoon: Another thing: `i` means "iterator": it was the correct name when using `each<>` and `const_each<>` but it's no longer the case with range-based for loops. Here, `tree` would be a better name.

dictoon:
Turns out the `const` were necessary, interestingly... Woops!

PR #2875, opened 2020-06-24
Add Bloom post-processing stage

laurelkeys:
Added Bloom as a new post-processing stage. Its 5 parameters are:
- `iterations`: number of downsampling iterations used for blurring, which indirectly controls the spread distance / radius
- `intensity`: strength of the Bloom effect (i.e. blending factor with the original image)
- `threshold`: filters out pixels under this level of brightness
- `soft_threshold`: makes the trasition between under/over-threshold gradual (0 = hard, 1 = soft)
- `debug_blur`: toggle to show only the Bloom target image, which is composited with the final render

Main changes:
- Included 11 test scenes (which render in ~30sec on a Core i5-4670K @ 3.40GHz)
- Created 4 new `ImageEffectApplier`s used in the Bloom stage, for:
	- (Weighted) additive color blending
	- Applying a bright-pass filter
	- Resampling images with a slight modification of Marius Bjørge's "Dual filtering"
	- Resampling images by a factor of 2 (faster than the method above)

The implementation is pretty similar to Keijiro Takahashi's [KinoBloom](https://github.com/keijiro/KinoBloom) (used in [EEVEE](https://github.com/blender/blender/blob/master/source/blender/draw/engines/eevee/shaders/effect_bloom_frag.glsl#L6)) and to Unity's Post-Processing [Bloom](https://github.com/Unity-Technologies/PostProcessing/blob/v2/Documentation~/Bloom.md).

However, instead of using the sampling filters from Jorge Jimenez's [SIGGRAPH2014 presentation](http://advances.realtimerendering.com/s2014/#_NEXT_GENERATION_POST), it is based on Marius Bjørge's ["Bandwidth-Efficient Rendering"](https://community.arm.com/cfs-file/__key/communityserver-blogs-components-weblogfiles/00-00-00-20-66/siggraph2015_2D00_mmg_2D00_marius_2D00_notes.pdf) presentation from SIGGRAPH2015 (and also, it is not implemented with shaders).


oktomus:
> iterations: number of downsampling iterations used for blurring, which indirectly controls the spread distance / radius

For the user, it makes more sense to expose the distance or radius parameter.

src/appleseed/renderer/modeling/postprocessingstage/effect/resamplex2applier.cpp
```diff
@@ -0,0 +1,142 @@
...
+    const std::size_t dst_width = image.properties().m_canvas_width;
+    const std::size_t dst_height = image.properties().m_canvas_height;
+
+    const Vector2f scaling_factor(
+        static_cast<float>(m_src_width - 1) / (dst_width - 1),
+        static_cast<float>(m_src_height - 1) / (dst_height - 1));
+
+    assert(
+        m_mode == SamplingMode::DOUBLE
+        ? dst_width / 2 == m_src_width && dst_height / 2 == m_src_height    // scale x2 with bilinear filtering
```
oktomus, line 89: How can you ensure this test is robust regarding the fact you're using integers ? A more robust way to test it is `m_src_width * 2 == dst_width`.

src/appleseed/renderer/modeling/postprocessingstage/bloompostprocessingstage.cpp
```diff
@@ -0,0 +1,425 @@
...
+                3,
+                props.m_pixel_format);
+
+            // Compute how many downsampling iterations we can do before a buffer has a side smaller than 2 pixels.
+            const float max_iterations = std::log2(min_pyramid_dimension / 2.0f);
+            const std::size_t iterations = clamp<std::size_t>(m_iterations, 1, static_cast<std::size_t>(max_iterations));
+
+            if (iterations < m_iterations)
+            {
+                RENDERER_LOG_INFO(
```
oktomus, line 190: Warning instead of info ?

src/appleseed/renderer/modeling/postprocessingstage/bloompostprocessingstage.cpp
```diff
@@ -0,0 +1,425 @@
...
+
+using namespace foundation;
+
+namespace renderer
+{
+
+namespace
+{
+    //
+    // Bloom post-processing stage.
```
oktomus, line 66: You should explain the general algorithm somewhere (passes process and stuff). Be sure to also link your references

src/appleseed/renderer/modeling/postprocessingstage/bloompostprocessingstage.cpp
```diff
@@ -0,0 +1,425 @@
...
+            // Copy the original image but ignore its alpha channel.
+            const std::size_t shuffle_table[4] = { 0, 1, 2, Pixel::SkipChannel };
+            Image prefiltered_image(image, props.m_pixel_format, shuffle_table);
+
+            // Filter out dark pixels.
+            const BrightPassApplier bright_pass(m_threshold, m_soft_threshold);
+            bright_pass.apply_on_tiles(prefiltered_image, thread_count);
+
+            //
+            // Create and initialize the blur buffer pyramids.
```
oktomus, line 212: Can you explain why we need such buffer ?

src/appleseed/renderer/modeling/postprocessingstage/effect/resampleapplier.cpp
```diff
@@ -0,0 +1,208 @@
...
+            //
+            // Reference:
+            //
+            //   "Bandwidth-Efficient Rendering" SIGGRAPH2015 Presentation
+            //   https://community.arm.com/cfs-file/__key/communityserver-blogs-components-weblogfiles/00-00-00-20-66/siggraph2015_2D00_mmg_2D00_marius_2D00_notes.pdf
+            //
+            //
+
+            const Color3f result = (
+                [&]() -> const Color3f
```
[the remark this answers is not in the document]
laurelkeys: I added it to make `result` a `const`. I don't think this showed any performance hit, but I'll double check it and see if using a lambda resulted in something weird in the assembly 🙂

src/appleseed/renderer/modeling/postprocessingstage/effect/resampleapplier.cpp
```diff
@@ -0,0 +1,208 @@
...
+                {
+                    if (m_mode == SamplingMode::UP)
+                    {
+                        return (
+                            blerp(m_src_image_with_border, fx - 1.0f, fy) +
```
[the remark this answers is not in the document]
laurelkeys: I tried fetching all of them beforehand for downsampling (that's the "jump from 6 lines to ~100" I mentioned some time ago 😅), but it didn't make any noticeable difference so I didn't try it for upsampling.

I haven't looked into fetching a row of pixels at once though.

src/appleseed/renderer/modeling/postprocessingstage/effect/resamplex2applier.cpp
laurelkeys: I'm using it this way to account for integer truncation, e.g.:
If the original image is 101x101, the first downscaling will result in a 50x50 image. Thus, we'll later have to upsample from 50x50 to 101x101, so `101 / 2 == 50` works but `50 * 2 == 101` wouldn't.

src/appleseed/renderer/modeling/postprocessingstage/bloompostprocessingstage.cpp
```diff
@@ -0,0 +1,425 @@
...
+            std::vector<Image>&         blur_pyramid_up,
+            const std::size_t           level_count,
+            const CanvasProperties&     max_level_props,
+            const std::size_t           max_tile_size)
+        {
+            blur_pyramid_down.reserve(level_count);
+            blur_pyramid_up.reserve(level_count);
+
+            // Note that lower levels have larger images.
+            std::size_t level_width = max_level_props.m_canvas_width;
```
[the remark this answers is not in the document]
laurelkeys: These represent the "current level" dimensions, so at the end of each for-loop iteration they are divided by two.
I can move them to the init statement to make it clearer:
```cpp
for (std::size_t
        level = 0, 
        level_width = max_level_props.m_canvas_width, 
        level_height = max_level_props.m_canvas_height
    ; level < level_count
    ; ++level)
{
    // ...
}
```
(I'm not sure what would be the standard way to format it in appleseed)

laurelkeys:
> Maybe you can find a way to expose only one parameter for the threshold instead of 2 ? Or call one of them differently ? I think Unity calls it "soft knee"

I used `threshold` instead of `knee` as I thought that'd make its purpose clearer (i.e. to soften the threshold, so 0 is a harsh boundary and 1 is a "soft" one), since using `knee` refers to shape of the function that is used for thresholding:
WZlkyidpry

However, it may be better to use `soft knee` so it's more familiar, Unity uses that (as you mentioned) and Blender calls it `knee` (GIMP calls if `softness`, btw). What do you think?

src/appleseed/renderer/modeling/postprocessingstage/bloompostprocessingstage.cpp
```diff
@@ -0,0 +1,425 @@
...
+            std::size_t level_height = max_level_props.m_canvas_height;
+
+            for (std::size_t level = 0; level < level_count; ++level)
+            {
+                const CanvasProperties level_props(
```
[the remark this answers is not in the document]
laurelkeys: `level_width` and `level_height` aren't constant: https://github.com/appleseedhq/appleseed/pull/2875#discussion_r452376103.

src/appleseed/renderer/modeling/postprocessingstage/bloompostprocessingstage.cpp
laurelkeys: Better, ty 🙂

src/appleseed/renderer/modeling/postprocessingstage/bloompostprocessingstage.cpp
laurelkeys: When we blur an image its intensity fades. So, by adding the intermediate blurring steps we can avoid this, e.g.:
Jfw0rZ5bno

src/appleseed/renderer/modeling/postprocessingstage/effect/resamplex2applier.cpp
oktomus: Maybe you can explain this directly in the code ?

src/appleseed/renderer/modeling/postprocessingstage/effect/resampleapplier.cpp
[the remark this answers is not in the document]
oktomus: The result is const but the code is less easy to read. I think it's better to not use lambda and keep the code easy to read. You can make an inline method if you want to make a const with all this.

oktomus:
> However, it may be better to use soft knee so it's more familiar, Unity uses that (as you mentioned) and Blender calls it knee (GIMP calls if softness, btw). What do you think?

I think the less parameter we have, the better it is. And if we can use something that is already established by the community, it's also better.

oktomus:
I have tested the bloom build you provided.

First of all, thanks for sharing a build, it really makes reviewing/testing easier.

-----

I have encountered python issues with your build. But it's not related to your work and not a problem for testing. Although it's worth saying, as it may be related to your build. There is module import errors in the python console where opening appleseed.studio.

---

The bloom post process works well, good job ! 

In order to be able to use that post process, I think we need some kind of changes to see the post process effect interactively. A first step could be to add a button somewhere that render the post process without rendering the whole scene. Maybe in `Rendering > Render post process`. A better way to do this would be to run post process when releasing a slider in the post process editor window. That would be perfect. What do you think @dictoon ? This could be done in a separate PR.

Regarding performance, it takes around 6seconds on a 4K frame on my 6-core 4Ghz AMD Processor, which I is too slow. But improvements can be made in a future work.

sandbox/tests/test scenes/bloom/_shaders/as_blender_areaLight.oso
```diff
@@ -0,0 +1,37 @@
+OpenShadingLanguage 1.00
```
oktomus, line 1: These shaders are already in the sandbox I think. Can you try to reference them instead of duplicating them ? You can try `grep -rn as_blender` in the `test scenes` folder to see how other scenes are using it.

PR #2880, opened 2020-07-05
Fix false colors not being applied to all tiles on a final render

laurelkeys:
This PR fixes the following bug (which happens when we have post-processing stages, and `False Colors` is enabled):

**Edit:** I undid the changes I mention below because they don't fit well with https://github.com/appleseedhq/appleseed/pull/2885 (though, I intend to add them in a later PR, once we discuss what'd lead to the best "user experience").

~Also, these changes make post-processing effects always run when a rendering is stopped (e.g. `Shift+F5`).~
~If this behavior is desirable, https://github.com/appleseedhq/appleseed/pull/2878 can be closed. Otherwise, we can keeps effects only being applied to final renderings with:~

```diff
@@ -226,7 +226,7 @@ struct MasterRenderer::Impl
             render_info.insert("render_time", m_project.get_rendering_timer().get_seconds());

             // Don't proceed further if rendering failed.
-            if (result.m_status == RenderingResult::Failed)
+            if (result.m_status != RenderingResult::Succeeded)
             {
                 controller.on_rendering_abort();
                 return result;
@@ -241,12 +241,7 @@ struct MasterRenderer::Impl
             // Insert post-processing time into frame's render info.
             render_info.insert("post_processing_time", stopwatch.get_seconds());

-            switch (result.m_status)
-            {
-              case RenderingResult::Succeeded: controller.on_rendering_success(); break;
-              case RenderingResult::Aborted: controller.on_rendering_abort(); break;
-              assert_otherwise;
-            }
+            controller.on_rendering_success();
         }
```

### DIFF
--- a/src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
+++ b/src/appleseed.studio/mainwindow/rendering/renderingmanager.cpp
@@ -133,7 +133,7 @@ RenderingManager::RenderingManager(StatusBar& status_bar)
     //
     // They are using a queued connection because the emitting thread is different from
     // the receiving thread (the emitting thread is the master renderer thread, and the
-    // receiving thread is the UI thread of the main window (presumably).
+    // receiving thread is the UI thread of the main window (presumably)).
     //
     // They are using a blocking queue connection because we need the receiving slot to
     // have returned in the receiving thread before the emitting thread can continue.
--- a/src/appleseed/renderer/kernel/rendering/masterrenderer.cpp
+++ b/src/appleseed/renderer/kernel/rendering/masterrenderer.cpp
@@ -225,8 +225,8 @@ struct MasterRenderer::Impl
             // Insert rendering time into frame's render info.
             render_info.insert("render_time", m_project.get_rendering_timer().get_seconds());
 
-            // Don't proceed further if rendering failed.
-            if (result.m_status == RenderingResult::Failed)
+            // Don't proceed further if rendering failed or aborted.
+            if (result.m_status != RenderingResult::Succeeded)
             {
                 controller.on_rendering_abort();
                 return result;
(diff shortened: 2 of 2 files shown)


laurelkeys:
> I'm just concerned about the fact that we trigger `on_rendering_abort` for both the `Aborted` and `Failed` states. Have you heavily tested appleseed.studio ? Also, be sure to open and try different projects when testing to find any corner case.

`on_rendering_abort` is used for both:

https://github.com/appleseedhq/appleseed/blob/964c5a91d27760370f20b15f07088dc6a0064e36/src/appleseed/renderer/kernel/rendering/irenderercontroller.h#L58-L59

However, since the value of `m_status` is set by `do_render()` it'll either be `Aborted` or `Suceedded` (it does not return `Failed`). Thus, the `if (result.m_status == RenderingResult::Failed)` could simply be removed.

But, if `do_render()` is later changed to return a `Failed` state we won't be treating this case (or we could add it to the `switch` statement, though it probably wouldn't make sense to call `postprocess()` if the rendering failed). 

I'm not sure, this `if` statement was trying to be "future proof", so it may be better to remove it. What do you think?

laurelkeys:
> Have you heavily tested appleseed.studio ? Also, be sure to open and try different projects when testing to find any corner case.

I have tested it with a couple of different combinations of post-processing stages (`Color Map`, `Render Stamp`, `Vignette`) on quite a few images (small, large, non-square).

Also tried setting different color maps for `False Colors` and everything appeared to be working 😅

oktomus:
> 
> 
> > Have you heavily tested appleseed.studio ? Also, be sure to open and try different projects when testing to find any corner case.
> 
> I have tested it with a couple of different combinations of post-processing stages (`Color Map`, `Render Stamp`, `Vignette`) on quite a few images (small, large, non-square).
> 
> Also tried setting different color maps for `False Colors` and everything appeared to be working 😅

It's really important to test different scenes on top of the default "Cornell Box" scene.

oktomus:
The bug needs to be confirmed to continue to exist once #2877 is merged.

laurelkeys:
> The bug needs to be confirmed to continue to exist once #2877 is merged.

It's still present after commit https://github.com/appleseedhq/appleseed/commit/0acfc054e9167ecfb25e3b5815ed0066e06b43f5.

PR https://github.com/appleseedhq/appleseed/pull/2896 adds the remaining changes to fix the issue, so I'm closing this one.

PR #2884, opened 2020-07-28
Add Tone Map post-processing stage

laurelkeys:
Added 7 new tone map operators:
- Two ACES approximations:
  - `ACES (Narkowicz)`: by Krzysztof Narkowicz (ref. [[1]](https://knarkowicz.wordpress.com/2016/01/06/aces-filmic-tone-mapping-curve/))
  - `ACES (Unreal)`: by Romain Guy (ref. [[2]](https://www.shadertoy.com/view/llXyWr))
- Reinhard's operators from ["Photographic Tone Reproduction for Digital Images"](http://www.cmap.polytechnique.fr/~peyre/cours/x2005signal/hdr_photographic.pdf):
  - `Reinhard`: Equation 3 in the paper (the simple 1/(1+x) tone mapping)
  - `Reinhard (Extended)`: Equation 4 in the paper
- Three filmic curves:
  - `Filmic (Hejl)`: by Jim Hejl (ref. [[3]](http://filmicworlds.com/blog/filmic-tonemapping-operators/), [[4]](https://github.com/tizian/tonemapper))
  - `Filmic (Uncharted)`: by John Hable, used in Uncharted 2 (ref. [[5]](http://filmicworlds.com/blog/filmic-tonemapping-operators/))
  - `Filmic (Piecewise)`: also by John Hable, developed with the aim of being simpler / more intuitive (ref. [[6]](http://filmicworlds.com/blog/filmic-tonemapping-with-piecewise-power-curves/))

The first filmic curve is almost "plug-n-play", while the other two are pretty customizable.

I think these seven operators provide a good starting point for *appleseed*.
While there are many other interesting ones that could definitely have been included (look at [`tonemapper`](https://github.com/tizian/tonemapper) for instance), IMHO the most important thing after adding these is to improve the "user experience" when using them on *appleseed.studio*.

This mainly involves: **1)** applying the effect as the parameters are changed (in the same way False Colors does it currently); and **2)** having a graph of the tone curve in the stage settings, specially for understanding the parameters of the two filmic curves from John Hable.

---

### Example
tmo_comparison

### Original
original


src/appleseed/renderer/modeling/postprocessingstage/tonemappostprocessingstage.cpp
```diff
@@ -0,0 +1,544 @@
...
+            if (m_clamp_colors)
+            {
+                const ClampColorsApplier clamp_colors;
+                clamp_colors.apply_on_tiles(image, thread_count);
+            }
+
+        }
+
+      private:
+        ToneMapApplier*     m_tone_map;
```
dictoon, line 250: Would make sense to use `std::unique_ptr<>` here.

dictoon (commented):
Only had a cursory look so far but this is top quality code and it looks nearly ready to merge (if not ready). We'll have a closer look with Kevin shortly and hopefully merge it in the next few days.

In the meantime, made a small comment.

PR #2885, opened 2020-08-06
Preview post-processing stage changes in appleseed.studio

laurelkeys:
This adds a rendered preview when parameter values are changed for post-processing stages (similar to the current behavior of `False Colors`).

Note that only the effect that is being changed is applied for preview (i.e. if there are multiple stages, this won't show all of them).

Other "nice-to-have"s are:
  - Render effects preview on a downscaled version of the frame, so it's more responsive (and make this a customizable option)
  - Some way to "opt-out" of this (e.g. a checkbox), either for all stages or for a specific one, untill all effects are "real-time"

gS09PO8IEs


oktomus:
Franz and I just tested your PR. Amazing work ! 

We are concerned about the processing time taken by the bloom effect or other future resource-intensive effect. I think we already discussed about this, but maybe you can add a checkbox to control the preview ? Like the following : 

image

That checkbox won't be visible when first adding the effect, as the preview doesn't work there.

PR #2887, opened 2020-08-15
Add Chromatic Aberration post-processing stage

laurelkeys:
Added ([lateral](https://en.wikipedia.org/wiki/Chromatic_aberration#Types)) Chromatic Aberration as a new post-processing stage.

This stage has two parameters, for artistic control:
- `strength`: controls the effect intensity by increasing distortion towards the edges of the image (thus, radially increasing the "color fringes")
- `fringing`: controls the amount of colors we use for the effect, i.e.:
  - the smallest value will result in the default "three-color split", found in most implementations of CA
  - higher values lead to a more "rainbow-like" effect, as shown in [this presentation](https://www.youtube.com/watch?v=RdN06E6Xn9E&t=428s) (see image comparisons below)

_Note: "fringing" might be a bad/misleading name.. renaming it to something like "smoothness" would probably be better._

---

| no aberration | default values |
|---|---|
| image | image |

_obs.: using a higher `strength` value than the default for the images below:_
| lowest `fringing` | highest `fringing` |
|---|---|
| image | image |


dictoon (approved):
Absolutely perfect! Thanks for the amazing work Tiago.

Issue #1243, opened 2017-02-07
Double wrapping of child BSDFs in BSDFBlend, BSDFMix and OSLBSDF?


dictoon:
These BSDFs have child BSDFs. The problem is that both these BSDFs and their children are wrapped with `renderer::BSDFWrapper<>`. That's likely not correct.

PR #2889, opened 2020-08-28
Fixing double wrapping for OSLBRDF, BSDFBlend and BSDFMix children

lorentzo:
## Description

This PR related to issue Double wrapping of child BSDFs in BSDFBlend, BSDFMix and OSLBSDF #1243.
It fixes double wrapping when OSLBRDF, BSDFBlend or BSDFMix are used.
Idea is that BSDFWrapper is not applied if the current wrapped BRDF is OSLBRDF, BSDFBlend or BSDFMix.

NOTE: this PR is also required for #2886.

### DIFF
--- a/src/appleseed/renderer/modeling/bsdf/bsdfblend.cpp
+++ b/src/appleseed/renderer/modeling/bsdf/bsdfblend.cpp
@@ -169,7 +169,7 @@ namespace
                 sampling_context,
                 values->m_child_inputs[bsdf_index],
                 adjoint,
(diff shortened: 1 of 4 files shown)


LZaw:
Did you verify, if the problem also occurs with `bsdf_mix` and `bsdf_blend`? If they do, could you please include a solution for that too, to close #1243 fully?

dictoon:
Thanks Lovro for your work!

As we discussed previously, we are going with the implementation from #2901.

PR #2890, opened 2020-09-01
Fix bold formatting in test suite runner's output

LZaw:
When running the testsuit, as soon as an entry was skipped, all lines are drawn bolt (see first screenshot). This PR resets the style to be normal after each print (see second screenshot).
Before:
Bolt
After:
After


dictoon (approved):
Thanks!

PR #2893, opened 2020-09-25
Fix assert in plasticbrdf.cpp

andreasdeclercq:
This assert is being triggered because probability can be zero. Since in the specular/glossy case the assert allows zero probability, I assume we can do the same thing for the diffuse case.


dictoon (approved):
Thanks for the fix, much appreciated!

PR #2898, opened 2020-10-26
Fix race condition in localsampleaccumulationbuffer.cpp

jnijs:
The m_active_level atomic is modified in store_samples outside the critical section protected by m_lock. Ensure the same level is used as source and target level.

### DIFF
--- a/src/appleseed/renderer/kernel/rendering/localsampleaccumulationbuffer.cpp
+++ b/src/appleseed/renderer/kernel/rendering/localsampleaccumulationbuffer.cpp
@@ -257,7 +257,7 @@ void LocalSampleAccumulationBuffer::develop_to_frame(
 #endif
 
     // Copy tile first so the critical section can end before doing the expensive develop_to_tile.
-    const std::uint32_t active_level = m_active_level.value(); // Ensure the same active_level is used for both arrays.
+    const std::uint32_t active_level = m_active_level; // Ensure the same active_level is used for both arrays.
     const AccumulatorTile& level = *m_levels[active_level];
     AccumulatorTile& read_level = *m_read_levels[active_level];    
     read_level.copy_from(level);
--- a/src/appleseed/renderer/kernel/rendering/localsampleaccumulationbuffer.cpp
+++ b/src/appleseed/renderer/kernel/rendering/localsampleaccumulationbuffer.cpp
@@ -257,11 +257,9 @@ void LocalSampleAccumulationBuffer::develop_to_frame(
 #endif
 
     // Copy tile first so the critical section can end before doing the expensive develop_to_tile.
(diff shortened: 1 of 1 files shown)


oktomus:
Thanks for your contribution @jnijs ! Can you tell me more about the bug, what happened ? A crash, assert failure, something else ?

jnijs:
The asserts in Tile::copy_from were triggered because the level and read_level tile are of different dimensions when the m_active_level atomic gets modified in between getting the current level tile from the 2 arrays.
There are never crashes since the target tile will always >= the source tile.

src/appleseed/renderer/kernel/rendering/localsampleaccumulationbuffer.cpp
```diff
@@ -259,6 +259,9 @@ void LocalSampleAccumulationBuffer::develop_to_frame(
     // Copy tile first so the critical section can end before doing the expensive develop_to_tile.
     const AccumulatorTile& level = *m_levels[m_active_level];
     AccumulatorTile& read_level = *m_read_levels[m_active_level];
+    const std::uint32_t active_level = m_active_level.value(); // Ensure the same active_level is used for both arrays.
+    const AccumulatorTile& level = *m_levels[active_level];
+    AccumulatorTile& read_level = *m_read_levels[active_level];
```
oktomus, line 264: You declare level and read_level two times. That won't build, see the last build log: https://travis-ci.org/github/appleseedhq/appleseed/jobs/738901308#L1880

Isn't this better ? 

```suggestion
    const std::uint32_t active_level = m_active_level.value(); // Ensure the same active_level is used for both arrays.
    const AccumulatorTile& level = *m_levels[active_level];
    AccumulatorTile& read_level = *m_read_levels[active_level];    
```

oktomus (changes requested):

src/appleseed/renderer/kernel/rendering/localsampleaccumulationbuffer.cpp
jnijs: Sorry about that. This is what I intended to do. Something must have gone wrong when merging the fix from somewhere else. Thanks for the fix.

oktomus:
Hi @jnijs, there is still an error at compile time.
https://travis-ci.org/github/appleseedhq/appleseed/jobs/739207799#L1879

```cpp
/home/travis/build/appleseedhq/appleseed/src/appleseed/renderer/kernel/rendering/localsampleaccumulationbuffer.cpp:260:55: error: ‘class boost::atomics::atomic<unsigned int>’ has no member named ‘value’
     const std::uint32_t active_level = m_active_level.value(); // Ensure the same active_level is used for both arrays.
                                                       ^~~~~
```

dictoon (approved):
Looks good to me, thanks!

PR #2899, opened 2020-10-30
Fix warnings with gcc 10.2.0

theartful:
Compiling appleseed in Release mode using gcc 10.2.0 causes a couple of warnings that due to "Werror" were failures. These were my changes to address them.
I'm using boost 1.72.0 and OpenShadingLanguage 1.11.8

### DIFF
--- a/src/appleseed/foundation/core/exceptions/stringexception.h
+++ b/src/appleseed/foundation/core/exceptions/stringexception.h
@@ -76,7 +76,7 @@ inline StringException::StringException(const char* what)
 
 inline StringException::StringException(const char* what, const char* s)
 {
-    char buf[sizeof(m_what) + sizeof(m_string) + 2];
+    char buf[sizeof(m_what) - 1];
     portable_snprintf(buf, sizeof(buf), "%s: %s", what, s);
     set_what(buf);
 
--- a/src/appleseed/foundation/platform/thread.h
+++ b/src/appleseed/foundation/platform/thread.h
@@ -295,7 +295,7 @@ class APPLESEED_DLLSYMBOL ThreadFlag
 // Spinlock class implementation.
 //
 
-inline Spinlock::Spinlock() : m_sp{BOOST_DETAIL_SPINLOCK_INIT}
+inline Spinlock::Spinlock() : m_sp BOOST_DETAIL_SPINLOCK_INIT
 {
 }
 
--- a/src/appleseed/foundation/platform/thread.h
+++ b/src/appleseed/foundation/platform/thread.h
@@ -295,10 +295,8 @@ class APPLESEED_DLLSYMBOL ThreadFlag
 // Spinlock class implementation.
 //
 
-inline Spinlock::Spinlock()
+inline Spinlock::Spinlock() : m_sp{BOOST_DETAIL_SPINLOCK_INIT}
 {
-    boost::detail::spinlock initialized_sp = BOOST_DETAIL_SPINLOCK_INIT;
-    std::memcpy(&m_sp, &initialized_sp, sizeof(initialized_sp));
 }
 
 inline bool Spinlock::try_lock()
--- a/src/appleseed/renderer/kernel/shading/shadingpoint.cpp
+++ b/src/appleseed/renderer/kernel/shading/shadingpoint.cpp
@@ -1114,7 +1114,7 @@ void ShadingPoint::initialize_osl_shader_globals(
 
         // Opaque state pointers.
         m_shader_globals.renderstate = const_cast<ShadingPoint*>(this);
-        memset(&m_osl_trace_data, 0, sizeof(OSLTraceData));
+        m_osl_trace_data = {};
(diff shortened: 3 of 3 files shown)


src/appleseed/foundation/core/exceptions/stringexception.h
```diff
@@ -76,7 +76,7 @@ inline StringException::StringException(const char* what)
 
 inline StringException::StringException(const char* what, const char* s)
 {
-    char buf[sizeof(m_what) + sizeof(m_string) + 2];
+    char buf[sizeof(m_what) - 1];
```
theartful, line 79: This causes a "truncation" when using strncpy here:
https://github.com/appleseedhq/appleseed/blob/0acfc054e9167ecfb25e3b5815ed0066e06b43f5/src/appleseed/foundation/core/exceptions/exception.h#L92-L103
`m_what` is going to be assigned to `buf`, and the size of `buf` is unnecessarily larger than `m_what`. Newer versions of gcc require that the size of the source be strictly less than the size of destination, and hence the `-1`

src/appleseed/foundation/platform/thread.h
```diff
@@ -295,10 +295,8 @@ class APPLESEED_DLLSYMBOL ThreadFlag
 // Spinlock class implementation.
 //
 
-inline Spinlock::Spinlock()
+inline Spinlock::Spinlock() : m_sp{BOOST_DETAIL_SPINLOCK_INIT}
 {
-    boost::detail::spinlock initialized_sp = BOOST_DETAIL_SPINLOCK_INIT;
-    std::memcpy(&m_sp, &initialized_sp, sizeof(initialized_sp));
```
theartful, line 301: Causes warning of using `memcpy` with non trivial data type.

src/appleseed/renderer/kernel/shading/shadingpoint.cpp
```diff
@@ -1114,7 +1114,7 @@ void ShadingPoint::initialize_osl_shader_globals(
 
         // Opaque state pointers.
         m_shader_globals.renderstate = const_cast<ShadingPoint*>(this);
-        memset(&m_osl_trace_data, 0, sizeof(OSLTraceData));
+        m_osl_trace_data = {};
```
theartful, line 1117: Warning of using `memset` with non trivial datatype. Value initialization should act as memset to 0 with primitive datatypes, so this should be equivalent as far as I can see.

src/appleseed/foundation/core/exceptions/stringexception.h
oktomus: This won't work if `m_what` is empty or if `m_what` is smaller that `m_string`. I'm not sure if it's a problem or not, but this need to be confirmed.

theartful: I'm not sure I get what you mean. "buf" is only assigned to "m_what" so it has to be the same size. If it's any longer, then the rest will be truncated by `strncpy`. Assuming `portable_snprinttf` 
```C++
     portable_snprintf(buf, sizeof(buf), "%s: %s", what, s);
```
acts sensibly assuming and does not crash if size of the formatted string"%s: %s" is larger than `sizeof(buf)` (which it should otherwise why the "n" in s'n'printf) , this should be alright, no?

src/appleseed/renderer/kernel/shading/shadingpoint.cpp
dictoon: Agreed, that's better.

src/appleseed/foundation/core/exceptions/stringexception.h
dictoon: `buf` needs to be big enough to accommodate:

1. The content of `what`
2. The 2-character string `: `
3. The content of `s` (which we accept to truncate to the size of `m_string` if necessary).

This is the reasoning behind making `buf` large enough to accommodate `sizeof(m_what) + sizeof(m_string) + 2` characters (including the terminating `\0`).

Is there a flaw in this reasoning?

dictoon (commented):
Thanks for your contribution!

Two of the fixes are unquestionably valid, but I'm not sure about the one in `StringException`.

src/appleseed/foundation/core/exceptions/stringexception.h
theartful: The reasoning is correct, but in both cases the outcome is the same. Whether `buf` can accommodate "%s : %s" or not, it's going to be truncated anyways to be of size `m_what` as of here
https://github.com/appleseedhq/appleseed/blob/e0bf003c584327e4b4d063e5d2f479f5316a479d/src/appleseed/foundation/core/exceptions/stringexception.h#L79-L81
where we set "m_what = buf". This causes gcc to raise the truncation warning. So the change is to truncate its size from the beginning.

dictoon:
Thanks for the fixes @theartful, and sorry for the delay :) We just merged it.

PR #2900, opened 2020-11-01
Fix compiling with OSL 1.11.8

theartful:
Fixes #2892

PR #2902, opened 2020-11-16
Change OSL to Version 1.11.9 and LLVM to Version 9

LZaw:
Before merging, I still have to update OSL in the Linux and Windows dependency package, to have every OS work on the same (major) version.

### DIFF
--- a/scripts/travis/build-macos.sh
+++ b/scripts/travis/build-macos.sh
@@ -71,7 +71,7 @@ brew info python@2
 
 brew upgrade boost
 brew upgrade qt
-brew install boost-python embree llvm@7 lz4 openimageio@2.0.4 openvdb pugixml xerces-c zlib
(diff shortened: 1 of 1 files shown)


oktomus (approved):
Thank you ! Can't wait to see that "build fail" on each PR disappear 🤣

theartful:
I think the change in #2900 (in the cmake file) should fix the `function was not declared` errors in osl.

Mango-3 (commented):
line 168    `-DWITH_DISNEY_MATERIAL=ON \` should be removed. We don't use the Disney material anymore. It has been removed from the sources with commit 1e8e30c72d782b478467a1ec5cf9e13d710c49ac

LZaw:
Thank you for the hint.
I removed it, as well as the SeExpr.

PR #2904, opened 2020-11-30
Fix clamp usage in frensel.h

dgmzc:
clamp is to be used with a return value.

This caused negative values to be calculated for some samples.


LZaw (approved):
Well spotted.

PR #2906, opened 2020-12-26
Fix CMake paths Linux/static build

hikiko:
CMake installs the python files in `<installation_dir>/lib/python2.7`.

But in file: `src/appleseed.studio/python/pythoninterpreter.cpp` they are expected to be in `<installation_dir>/lib/python` (see "Compute path" code). 

As a result when starting appleseed.studio they can't be found and the following warnings appear:
`Python's site-packages directory does not exist.`
`appleseed Python module's directory does not exist.`
as well as errors for not properly imported modules in Python console.

Changed the paths in CMake to match the paths in `pythoninterpreter.cpp`.

### DIFF
--- a/src/appleseed.python/CMakeLists.txt
+++ b/src/appleseed.python/CMakeLists.txt
@@ -308,7 +308,7 @@ endif ()
 # Installation.
 #--------------------------------------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


LZaw (approved):

dictoon:
Thanks for the PR @hikiko! Much appreciated.

PR #2910, opened 2021-03-18
Fix orennayarbrdf.cpp: no longer try to normalizes zero vectors.

andreasdeclercq:
Calculate the norm of each vector first and do an early return if zero.
This avoids asserts in debug mode and divisions by null in release mode.


oktomus:
Thanks for your contribution @andreasdeclercq !

> This avoids asserts in debug mode and divisions by null in release mode.

Can you please explain in which case you get divisions by null ? That would be great if you could provide a scene so that we can test.

PR #2915, opened 2021-08-26
Fix Apple Silicon build

Willem871:
Only include x86 specific headers and code when APPLESEED_X86 is defined.
This fixes the build of system.cpp on Apple Silicon.

### DIFF
--- a/src/appleseed/foundation/platform/system.cpp
+++ b/src/appleseed/foundation/platform/system.cpp
@@ -71,7 +71,9 @@
     #include <sys/param.h>
     #include <sys/sysctl.h>
     #include <sys/types.h>
+#ifdef APPLESEED_X86
     #include <cpuid.h>
+#endif
 
 // Linux.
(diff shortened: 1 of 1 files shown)


pjessesco:
Hi @Willem871 , is it possible to build on macOS in current master branch?? 

It would be nice if you can share how you build on recent version of macOS.

Willem871:
Hi @pjessesco. We use the master branch of commit [a6de6fb5c9618b2523d9b5eb4f8f4d9311234c53]( https://github.com/DLTcollab/sse2neon).
We have working macOS (Big Sur) builds (both Intel and ARM).
Next to this fix we also needed the [sse2neon](https://github.com/DLTcollab/sse2neon) library to convert the SSE instructions in src/appleseed/foundation/platform/sse.h to NEON on ARM architectures. We didn't submit a pull request for this one, since we had to do some custom tweaks.

PR #2925, opened 2023-08-02
Fix various typos

luzpaz:
Fixes various user-facing and non-user-facing typos. Found via `codespell -q 3 -S "./src/thirdparty,./src/appleseed/foundation/utility/minizip" -L fo,localy,mis,ot`

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -441,7 +441,7 @@ epub_identifier = 'https://github.com/appleseedhq/appleseed'
 # The format is a list of tuples containing the path and title.
 #epub_pre_files = []
 
-# HTML files shat should be inserted after the pages created by sphinx.
+# HTML files that should be inserted after the pages created by sphinx.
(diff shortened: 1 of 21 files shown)


luzpaz:
Soft bunp

luzpaz:
bump

luzpaz:
no response?

virtualzavie (approved):
For what it's worth, the proposed change does exactly what it says in the title, and looks good to me.

gospodnetic:
Please push an update with resolved conflicts

luzpaz:
@gospodnetic done

virtualzavie (approved):
I did another pass after the update, and the proposed change still looks good.

There are no functional modifications, as it strictly consists in fixing typos in comments and strings. The only caveat would be if there is some localisation or some other string matching based system that could be impacted.

Otherwise I would suggest to merge the PR since it is a harmless improvement (even if a minor one).

